PR #2, opened 2016-08-12
Update Italian localisation 

darthpelo:


roundhill:
We actually use GlotPress for the translations, so you can fix it here: https://translate.wordpress.com/projects/simplenote/ios/it/default/

I will update the README with translation contribution info. Thanks for contributing!

Issue #17, opened 2016-08-14
'Done' button can delete email address when adding collaborator


kwonye:
https://twitter.com/DanielGenser/status/567779387810906112/photo/1

Opened by @roundhill

jleandroperez:
Fixed via #71

Props to @khoiln!!

PR #71, opened 2016-08-18
#17 Done button could delete collaborator

khoiln:
#17 
Done button could delete email address while user adding new collaborator


jleandroperez:
Fix confirmed as well!!! Again, thanks for helping us out @khoiln!

:shipit:

Issue #21, opened 2016-08-14
Alphabetise tag list


kwonye:
It would be great to be able to sort the list of tags alphabetically

another request #2685197-t

Another request #2726533-t

Opened by @rachelsquirrel

rantoncuadrado:
#43423423-zen

gracie:
Also requested in #1211447-zen

cademis:
+1 for this change. I understand this is by design cross-platform, however, I find myself manually ordering my tags alphabetically to allow me to find tags quickly (150+ tags)

Shtalin:
Why it is open? I can sort tags

jleandroperez:
Closing this one since there's an option to Sort Tags alphabetically.

PR #74, opened 2016-08-20
Alphabetize note tags

khoiln:
#21

### DIFF
--- a/Simplenote/Classes/Note.m
+++ b/Simplenote/Classes/Note.m
@@ -395,7 +395,7 @@ SEL notifySelector;
 }
 
 - (void)updateTagsArray {
-    tagsArray = tags.length > 0 ? [[tags objectFromJSONString] mutableCopy] : [NSMutableArray arrayWithCapacity:2];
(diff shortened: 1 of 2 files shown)


jleandroperez:
Hello there @khoiln!!

Thanks for the PR, but this is actually by design. Tags can be arranged by hand, and the order is expected to be reflected in all of the devices.

khoiln:
Oops sorry I didn't know that :).

jleandroperez:
No worries sir!! thanks for helping us out!!

supernovia:
@jleandroperez it occurs to me people may be asking about the ability to alphabetize by title, actually, since the word "tags" appears above the list of titles with an arrow. Maybe people think the arrow means "These things below are tags" when they are in fact note titles, and the arrow is there to indicate options for tags.

jleandroperez:
@supernovia i'm sorry i've missed that ping!! ^^

Not sure to which app you refer to, can't see any arrows in the iOS app. Maybe the macOS one?. Still, Simplenote already has an option that automatically **Sorts Notes Alphabetically** (Open the Tags area, bottom button `Settings`, it should be at the top).

In the macOS app, such option is available via **View >> Sort Notes Alphabetically**.

Thanks for the feedback!!

PR #75, opened 2016-08-20
Change back button title to "Search" if the user was searching

khoiln:
# 28

When I changed the title of the back button at -viewWillAppear, the button seems to blink for couple ms, not sure if I did something wrong.

Can some one help me review this?


jleandroperez:
Looks good @khoiln!

Thanks for helping us out!!

:shipit:

Issue #80, opened 2016-08-26
iPad: Share and Collaborate Buttons Stopped Working (iOS 9.3.4)


roundhill:
A user reported and I reproduced on my iPad:

1. Open a note, tap info button.
2. Tap on `Send`. Nothing happens.
3. Tap on `Collaborate`. Nothing happens.

The other buttons do appear to work fine, like history and trash. Also, the blue background behind 'markdown' is a lot lighter on the iPad for some reason.

roundhill:
Screenie of the light blue background:
<img width="187" alt="screen shot 2016-08-26 at 9 21 04 am" src="https://cloud.githubusercontent.com/assets/789137/18012485/7dfa2730-6b6e-11e6-87bb-5222d419f0fa.png">

eduardozulian:
Reported also on #2794030-t

PR #81, opened 2016-08-26
Issues/80: Action sheet buttons should now work correctly on iPad.

### LINKED ISSUE #79: Increase contrast for toggle highlighting
Requested via [Twitter](https://twitter.com/user/status/768880679933194241).

User reports that toggle highlighting in the app on their iPad Pro is too light and not very visible.

Screen shot: 
cqxy50qxgaene2q jpg_small

frosty:
Fixes #79 and #80. 

The action button popover had been changed from using UIPopoverController (deprecated) to presenting with a modal presentation style. Because it now presents as a view controller, this was blocking the subsequent presentation of other view controllers. We now dismiss the action popover before displaying share / collaborate.

Also blocked the popover from adapting its style whilst presented to fix a layout issue when:

* Viewing the app on iPad with another app in a narrow multitasking split on the right.
* Show the popover
* Rotate to portrait

Normally the popover would adapt to show as a modal, but it doesn't lay itself back out correctly. The simplest fix seems to be to retain the popover presentation in this layout.

I've also updated the background colour of the popover to make the contrast match the action sheet (#79).

### To test:

* On iPad, try using the Send and Collaborate buttons in the note action popover!

Needs review: @jleandroperez

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController.h
+++ b/Simplenote/Classes/SPNoteEditorViewController.h
@@ -50,8 +50,6 @@
     // sheets
     SPActivityView *noteActivityView;
     SPActionSheet *noteActionSheet;
-    UIPopoverController *noteActionPopover;
-    UIPopoverArrowDirection *sharePopover;
     SPActionSheet *versionActionSheet;
     UIActionSheet *deleteActionSheet;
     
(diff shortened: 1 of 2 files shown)


jleandroperez:
Working great, thanks for fixing this @frosty!!

:shipit:

PR #82, opened 2016-08-26
Updates Markdown localization strings with English placeholders.

frosty:
See title :)

Some users reported an issue where these strings weren't localized to their language and they were seeing the placeholder strings:

screen shot 2016-08-26 at 21 26 54

So I've changed the placeholder strings to be in English, so it should at least be more readable.

Needs review: @jleandroperez


jleandroperez:
Looking great!!

:shipit:

PR #92, opened 2016-10-14
Convert Swift Code to V3

roundhill:
Most of this was automated, but I needed to update one usage of `NSMutableURLRequest` to `URLRequest`.

cc @jleandroperez


jleandroperez:
Looks good @roundhill!

:shipit:

PR #97, opened 2016-10-28
Update Tag Menu to New Design

roundhill:
Updates the tag drawer/menu to the new design as outlined by @drw158. It ended up changing quite a few things about how the code works in `SPTagsListViewController` mainly:
- The UITableView `edit` process no longer expands the tableview to full screen, instead you edit the tags list in place (see screenshots)
- The TableView now only uses one section (yay!), but it requires a lot of views for the headers and footers to be created in the code (boo!). I think it's actually a bit less complex now which is nice.

Also, a view was needed to be added to `SPNoteListViewController` to show a border at the left side when the drawer is opened. (See `SPBorderedTableView`)

I really wanted a storyboard/swift conversion for this, but didn't have the time so stuck with good ol' Obj-c. 😬

Screenshots:
simulator screen shot oct 28 2016 2 28 11 pm
simulator screen shot oct 28 2016 2 28 15 pm
simulator screen shot oct 28 2016 2 30 07 pm

@jleandroperez needs review! Thanks!

PR #105, opened 2016-11-10
Update Icons

roundhill:
Updated icons to newer versions used in Electron app. Also removed animation when drawer would open, as it didn’t make sense any longer.

**Examples**
screen shot 2016-11-10 at 3 42 23 pm

screen shot 2016-11-10 at 3 42 35 pm

screen shot 2016-11-10 at 3 42 41 pm

I didn't update some of the icons, but they won't be visible any longer with the merge of #97. This also doesn't change any of the icons in the 'info' panel, that should come in a separate PR.

cc @jleandroperez and @drw158


jleandroperez:
Looks beautiful @roundhill 

:shipit: !!!

Issue #39, opened 2016-08-14
Enable Apple's spotlight search to search Simplenote notes


kwonye:
User requested that we enable searching of Simplenote notes with Spotlight

63442-t

https://twitter.com/carosi/status/656518502778576896

https://twitter.com/user/status/688927948196462596

Opened by @rachelsquirrel

kevmarsden:
Another request in 2942273-t

jleandroperez:
Closed via #107 by @michalkosmowski 

Thank you Michal!!

PR #107, opened 2016-11-25
Spotlight search support

michalkosmowski:
All notes are now indexed and can be found via Spotlight. #39 
I'm not sure if `- (void)bucketDidFinishIndexing:(SPBucket *)bucket` is the best place to hook up indexing. I'm open to any suggestions.

That's how it looks:
simulator screen shot 25 nov 2016 22 18 15


roundhill:
Thanks for contributing. Looks neat! Is there a way to disable this? A lot of users like to keep their notes private and might not want them to be searchable outside of the app.

michalkosmowski:
My pleasure @roundhill. 
Yup, one could disable Spotlight on a 'per app' basis in Settings > General > Spotlight Search. We could definitely expose it as well in app's settings, but I don't recall any app doing it like that.

roundhill:
> Yup, one could disable Spotlight on a 'per app' basis in Settings > General > Spotlight Search. We could definitely expose it as well in app's settings, but I don't recall any app doing it like that.

Nice! Hey @jleandroperez this seems like something worth adding to the app, would you mind reviewing this PR?

Simplenote/SPAppDelegate.m
```diff
@@ -295,6 +295,25 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     return YES;
 }
 
+- (BOOL)application:(UIApplication *)application continueUserActivity:(NSUserActivity *)userActivity restorationHandler:(void (^)(NSArray * _Nullable))restorationHandler {
+    
+    if (userActivity.userInfo != nil) {
+        if (userActivity.userInfo[CSSearchableItemActivityIdentifier] != nil) {
+            NSString *uniqueIdentifier = userActivity.userInfo[CSSearchableItemActivityIdentifier];
```
jleandroperez: Suggestion: Since in ObjC messages to `nil` are allowed, perhaps something like this would be easier to read:

```
NSString *uniqueIdentifier = userActivity.userInfo[CSSearchableItemActivityIdentifier];
if (uniqueIdentifier == nil) {
   return false;
}

SPBucket *noteBucket = [_simperium bucketForName:@"Note"];
Note *note = [noteBucket objectForKey:uniqueIdentifier];
if (note == nil) {
   return false;
}

(...)

return true;
```

(Reason: less indentation is way easier to understand, plus, this way we'd tell the OS if a Note is no longer available).

Simplenote/SPAppDelegate.m
```diff
@@ -295,6 +295,25 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+            SPBucket *noteBucket = [_simperium bucketForName:@"Note"];
+            Note *note = [noteBucket objectForKey:uniqueIdentifier];
+            double delayInSeconds = 0.05;
+            dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC));
+            dispatch_after(popTime, dispatch_get_main_queue(), ^(void){
```
jleandroperez: Perhaps `dispatch_async` instead of `dispatch_after` would do the trick? (that way you could avoid the delayInSeconds constant)

Simplenote/SPAppDelegate.m
```diff
@@ -295,6 +295,25 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+                [_noteListViewController openNote:note fromIndexPath:nil animated:NO];
+                [self showPasscodeLockIfNecessary];
```
jleandroperez: Is the `showPasscodeLockIfNecessary` 1000% needed? (already being called in `didFinishLaunching` and `applicationWillResignActive`

Simplenote/SPAppDelegate.m
```diff
@@ -683,6 +704,17 @@ - (void)bucketDidFinishIndexing:(SPBucket *)bucket
...
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
+            NSArray *deleted = [bucket objectsForPredicate:[NSPredicate predicateWithFormat:@"deleted == YES"]];
+            if (deleted.count > 0) {
+                NSArray *deletedIDs = [deleted valueForKey:@"simperiumKey"];
+                [[CSSearchableIndex defaultSearchableIndex] deleteSearchableItemsWithIdentifiers:deletedIDs completionHandler:nil];
+            }
+            
+            NSArray *notes = [bucket objectsForPredicate:[NSPredicate predicateWithFormat:@"deleted == NO"]];
+            [[CSSearchableIndex defaultSearchableIndex] indexSearchableNotes:notes];
+        });
```
jleandroperez: Although `didFinishIndexing` is a suitable place to do this, there are few problems with this snippet:

- `SPBucket` is not thread safe. `objectsForPredicate` runs agains the Main Thread MOC, which would potentially cause issues when / if called from a BG thread
- The `deleted` flag is just a temporary indicator, which (AFAIK) will be gone after the context effectively saves. This could cause erratic behavior, due to the Core Data architecture (BG writer saving asynchronously). I _suspect_ inspecting the `deleted` property of an object that's gone, but still registered in the MOC, may break the app.

How about if, instead, we move the `indexSearchableNote` / `deleteSearchableItems` calls to this SPBucket delegate method?

```
bucket:(SPBucket *)bucket didChangeObjectForKey:(NSString *)key forChangeType:(SPBucketChangeType)change memberNames:(NSArray *)memberNames
```

Simplenote/Classes/CSSearchable+Helpers.swift
```diff
@@ -0,0 +1,66 @@
...
+        let index = note.preview.index(note.preview.startIndex, offsetBy: note.titlePreview.utf16.count)
+        contentDescription = note.preview.substring(from: index)
+    }
+    
+}
+
+extension CSSearchableItem {
+
+    convenience init(note: Note) {
+        let set = CSSearchableItemAttributeSet(note: note)
```
jleandroperez: Nipticky comment: perhaps rename `set` > `attributeSet` ? (for clarity)

Simplenote/Classes/CSSearchable+Helpers.swift
```diff
@@ -0,0 +1,66 @@
...
+        self.init(uniqueIdentifier: note.simperiumKey, domainIdentifier:"notes", attributeSet: set)
+    }
+
+}
+
+extension CSSearchableIndex {
+
+    func indexSearchableNote(_ note: Note) {
+        let item = CSSearchableItem(note: note)
+        CSSearchableIndex.default().indexSearchableItems([item]) { error in
```
jleandroperez: Since this method is always called as `[CSSearchableIndex defaultSearchableIndex]`, perhaps we could nuke  `CSSearchableIndex.default()`, and call `indexSearchableItems` directly?

Simplenote/Classes/CSSearchable+Helpers.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+    
+    func indexSearchableNotes(_ notes: [Note]) {
+        var items: [CSSearchableItem] = []
+        
+        for note in notes {
+            items.append(CSSearchableItem(note: note))
+        }
+        
+        CSSearchableIndex.default().indexSearchableItems(items) { error in
```
jleandroperez: Same comment as above, perhaps we could nuke the `CSSearchableIndex.default()` snippet

Simplenote/Classes/CSSearchable+Helpers.swift
```diff
@@ -0,0 +1,66 @@
...
+            if let error = error {
+                NSLog("Couldn't index notes in spotlight: \(error.localizedDescription)");
+            }
+        }
+    }
+    
+    func deleteSearchableNote(_ note: Note) {
+        CSSearchableIndex.default().deleteSearchableItems(withIdentifiers: [note.simperiumKey]) { error in
```
jleandroperez: (Same as above, nuke `CSSearchableIndex.default()` if possible)

Simplenote/SPAppDelegate.m
michalkosmowski: Makes sense, I'll change that.

Simplenote/Classes/CSSearchable+Helpers.swift
michalkosmowski: Oh that's right, I missed the point of execution. Nice catch!

Simplenote/SPAppDelegate.m
michalkosmowski: I took this chunk from `- (BOOL)application:(UIApplication *)application handleOpenURL:(NSURL *)url`. I'll extract it to a separate method, and reuse in those 2 scenarios. Seems to me that it's necessary in a case when the user leaves the app on Settings, then comes back to the app via Spotlight or URL. At this point the view stack needs to be cleaned up first, then we can show a new note (via URL) or a note from Spotlight Results.

Simplenote/SPAppDelegate.m
michalkosmowski: I was considering `didChangeObjectForKey` too, but it would fire on every object separately. I'll do that on a private context instead.

The `deleted` flag is actually a field of Note, used to differentiate notes in the thrash. The one you're thinking about is `isDeleted` and I completely agree with the rest of the comment.

michalkosmowski:
@jleandroperez please take a look now. I ended up using a regular private context, but there may be a better way using Simperium built-in stuff. We could handle `case SPBucketChangeInsert:` in `didChangeObjectForKey` as well, and ditch the `indexSpotlightItems` method. As I commented earlier that would be slower, but definitely easier to read. Let me know what do you think.

Not to be nitpicking but you've misspelled nipticking 😉

jleandroperez:
One extremely minor caveat: users who upgrade from the previous binary won't get their notes indexed, until Simperium (eventually) goes thru a reindex. I'll get that addressed in another PR, and will cc you.

Beautiful PR @michalkosmowski!!!. Thanks so much for helping us improve Simplenote!!

:shipit:

PR #118, opened 2016-12-23
Nukes dead code

jleandroperez:
### Details:

In this PR we're just removing a couple snippets that haven't been in use, for a while.
Plus: adding a missing *super* call i've stumbled upon!

cc @roundhill


roundhill:
LGTM :shipit:

jleandroperez:
Thanks Dan!

PR #125, opened 2017-01-17
Don’t dismiss presented views in the Editor if it is showing the pin lock.

### LINKED ISSUE #120: Fingerprints not required to access the app again
A user reported the following behavior:

> Love the app on my iphone. I've just noticed that things are not as secure as they once were. I use fingerprints, ios10.2, on an iPhone 6. When I switch away from Simplenote to another app I am not required to re-enter my passcode or fingerprint. In the past every time I came back to the app I had to log in. 

As I don't have an iPhone to test, @roundhill tested it and could also reproduce it. His thoughts about it:

> It seems to happen when you are in the note editor, but not in the list when you leave the app

roundhill:
#120 was caused by the editor view controller attempting to dismiss views that it presented. The problem was that this could include the PIN lock controller too! I've added a check in the app delegate for when it is showing the PIN lock, which seems to have resolved the issue.

**To test**
* Enable the pin lock and go to the editor view
* Press the home button, then return to the app
* You should see the PIN lock entry

Also check that the app works as expected when you leave it with modals open, on both iPhone and iPad.

@frosty you touched this code before so I'm hoping you can review!

Fixes #120

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -130,10 +130,6 @@ CGFloat const SPMultitaskingCompactOneThirdWidth = 320.0f;
                                                  selector:@selector(didReceiveVoiceOverNotification:)
                                                      name:UIAccessibilityVoiceOverStatusChanged
                                                    object:nil];
-        [[NSNotificationCenter defaultCenter] addObserver:self
-                                                 selector:@selector(applicationWillResignActive:)
-                                                     name:UIApplicationWillResignActiveNotification
-                                                   object:nil];
     }
     
     return self;
@@ -284,10 +280,6 @@ CGFloat const SPMultitaskingCompactOneThirdWidth = 320.0f;
 {
     [super willTransitionToTraitCollection:newCollection withTransitionCoordinator:coordinator];
     [self refreshNavBarSizeWithCoordinator:coordinator];
-}
-
-- (void)applicationWillResignActive:(NSNotification *)notification {
-    [self dismissActivityView];
 }
 
 - (void)refreshNavBarSizeWithCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
@@ -1443,7 +1435,7 @@ CGFloat const SPMultitaskingCompactOneThirdWidth = 320.0f;
 }
 
 - (void)dismissActivityView {
-    if (self.presentedViewController && ![[SPAppDelegate sharedDelegate] isPresentingPinLock]) {
+    if (self.presentedViewController) {
         [self dismissViewControllerAnimated:YES completion:nil];
     }
 
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -21,7 +21,6 @@
 #import "SPObjectManager.h"
 #import "SPAddCollaboratorsViewController.h"
 #import "JSONKit+Simplenote.h"
-#import "DTPinLockController.h"
 #import "SPHorizontalPickerView.h"
 #import "SPVersionPickerViewCell.h"
 #import "SPPopoverContainerViewController.h"
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -21,6 +21,7 @@
 #import "SPObjectManager.h"
 #import "SPAddCollaboratorsViewController.h"
 #import "JSONKit+Simplenote.h"
+#import "DTPinLockController.h"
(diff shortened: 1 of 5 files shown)


Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -21,6 +21,7 @@
 #import "SPObjectManager.h"
 #import "SPAddCollaboratorsViewController.h"
 #import "JSONKit+Simplenote.h"
+#import "DTPinLockController.h"
```
frosty: I don't think this is necessary?

roundhill: So true! I accidentally left it in there from some earlier work. Removed in fc572dd

frosty:
@roundhill I'm getting a crash (iPhone) when following these steps:

* Fresh install of the app, log in
* Turn on pin lock
* Open a note
* Tap Collaborate

```
2017-01-18 10:06:39.816626 Simplenote[36656:686016] [Assert] Trying to dismiss the presentation controller while transitioning already. (<_UIFullscreenPresentationController: 0x7fd437c5f6c0>)
2017-01-18 10:06:39.835112 Simplenote[36656:686016] [Assert] transitionViewForCurrentTransition is not set, presentation controller was dismissed during the presentation? (<_UIFullscreenPresentationController: 0x7fd437c5f6c0>)
```

It looks like the contacts view makes the app resign active, which triggers the call to `dismissViewController` whilst the view is still setting itself up.

roundhill:
> I'm getting a crash (iPhone) when following these steps:

That's strange, seems to be something with how the app prompts for access to the contacts. It works ok after you grant access and repeat the steps.

frosty:
Yep, I think it's because the first time the permissions alert is displayed on top.

I'm also seeing the history view staying visible when the pin lock screen is active:

simulator screen shot 18 jan 2017 12 25 46

I'm not sure if it's _caused_ by this PR, but seems like something we can probably address here.

roundhill:
Oh my, the crash happens in the production version too. I'll fix it here.

- [ ] Contacts prompt crash
- [ ] History view overlay

roundhill:
A lot of bugs were coming up with regards to how the pin lock is currently presented as I was trying to fix this, so I've actually refactored it to present itself in a UIWindow. This works a lot better because we are guaranteed that the PIN lock will cover all UI, including action views.

Changing it to a UIWindow revealed a few more bugs related to the tags view controller listening to keyboard show/hide events when it shouldn't. I've also fixed that.

These changes touch a lot of core areas of the app, so a thorough test is needed. Enable the PIN lock and ensure that the app works as expected when you leave and return to it from whatever state you left it.

I also added a specific check to not show the PIN lock when the contacts view is first displayed and the user is viewing the permissions prompt. The code is [not pretty though](https://github.com/Automattic/simplenote-ios/pull/125/files#diff-3d7ffb11fe2a17fede94feba3e9c5aa0R932), so I'm open to any other ideas!

Thanks for your review @frosty and @jleandroperez

Simplenote/SPAppDelegate.m
```diff
@@ -7,6 +7,7 @@
 //
 
 #import "SPAppDelegate.h"
+#import <Contacts/Contacts.h>
```
frosty: Kind of a nitpick, but Apple recommend using the module-based `@import` for system frameworks these days.

Simplenote/SPAppDelegate.m
```diff
@@ -325,6 +326,7 @@ - (void)onboardingDidFinish:(NSNotification *)notification
 
 - (void)applicationWillResignActive:(UIApplication *)application
 {
+    [self.tagListViewController removeKeyboardObservers];
```
frosty: Instead of exposing this method in the header of the tag list view controller and calling it from here, the tag list view controller could register for the system `UIApplicationWillResignActiveNotification` notification and call the method itself. Then the app delegate doesn't need any knowledge of the tag list view controller.

roundhill: I believe I attempted using that, but the notification wouldn't fire until after the code was run in the app delegate. We need the listener to go away before the app adds the new UIWindow for the Pin lock.

Simplenote/SPAppDelegate.m
roundhill: Ah, thanks. I've been out of the iOS game for a while :) 

Updated in 7969f67

Simplenote/SPAppDelegate.m
frosty: Okay! Sounds reasonable 👍

jleandroperez:
@roundhill @frosty  thank you both for fixing this one!!

PR #136, opened 2017-02-26
Fix: Localized string for the Trash option not being translated

### LINKED ISSUE #131: Localized string for the Trash option not being translated
Looks like the strings for "Trash" and "Edit" are not being translated into the system language when you switch it from English to a different one.

Steps to reproduce:
- On iPhone go to Settings > Language & Region > iPhone Language.
- Switch the language to a non-English one.
- Open Simplenote.
- Click on the top left corner (the tag symbol) to open the tag and options menu.
- "All notes" & "Settings" will be translated to the language, but not "Trash".

Attached screenshot: https://cloudup.com/cOuTXWWLnIB

Regarding "TAGS", it might be translations in other languages (for example in Spanish is `etiquetas`, but I'm unsure if that's intended)

diejmon:
Fixes #131


jleandroperez:
Thanks for helping us out @diejmon!!

:shipit:

PR #148, opened 2017-06-20
Update build instructions to use dev app and key

roundhill:
This PR adds a hard coded dev app key to the app so that developers will not need to create a Simperium account and app in order to build and test the app.


jleandroperez:
Thanks Dan!!

PR #149, opened 2017-06-23
WIP: Add todo lists?

dmsnell:
Okay so obviously I have no idea what I'm doing and I don't even know
Objective-C. However, I wanted to try and start digging into it and
created this to mimick Automattic/simplenote-electron#add/todo-lists

I'm lost and would love some feedback. In Simplenote-electron I kept
track of how many times each match was found from the RegExp and then
used that index as a reference back into the app to update the note to
reflect clicks.

Here in iOS I just have no idea. The main idea is to reference back out
of the WebView so we can update the note but I'm not sure how to get to
the note in the `SPMarkdownPreviewViewController` let alone update it.

This will probably die a painful death but thanks for looking!


kwonye:
Thanks for this, @dmsnell! Love that you did it vs just asking.

@jleandroperez any ideas? I have no clue how simplenote works. And @frosty do you have any clue? If I recall correctly you did markdown support on iOS?

dmsnell:
Android is interesting because it will let me inject Java objects into the Javascript where I can call from within the WebView into my application. 

Some of the problem I think may also stem from the fact that it's just awkward using a WebView instead of native rendering of Markdown.

@frosty I'm pretty sure you did the awesome job of implementing Markdown. Did you look into parsing and generating native widgets?

Also, kudos to all you iOS devs out there. This is such a huge barrier to entry for me, knowing nothing about it.

dmsnell:
Android is interesting because it will let me inject Java objects into the Javascript where I can call from within the WebView into my application. 

Some of the problem I think may also stem from the fact that it's just awkward using a WebView instead of native rendering of Markdown.

@frosty I'm pretty sure you did the awesome job of implementing Markdown. Did you look into parsing and generating native widgets?

Also, kudos to all you iOS devs out there. This is such a huge barrier to entry for me, knowing nothing about it.

Simplenote/Classes/SPMarkdownParser.m
```diff
@@ -29,22 +29,40 @@ + (NSString *)renderHTMLFromMarkdownString:(NSString *)markdown
...
+                              "  this.sendRawMessageToiOS = function(message){"
+                              "    console.log(\"Message string to iOS: [\" + message+ \"]\");"
+                              "    var iframe = document.createElement(\"IFRAME\");"
+                              "    iframe.setAttribute(\"src\", \"jscall://\" + message);"
+                              "    document.documentElement.appendChild(iframe);"
+                              "    iframe.parentNode.removeChild(iframe);"
+                              "    iframe = null;"
+                              "  };"
+                              "}"
+                              "</script>";
```
jleandroperez: Hey there @dmsnell !!! can we nuke this JS snippet?

Simplenote/Classes/SPMarkdownPreviewViewController.m
```diff
@@ -108,6 +108,17 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
...
+    NSString *jsCallPrefix = @"jscall://";
+    NSString *fullURLRequested = request.URL.absoluteString;
+    
+    BOOL isJavaScriptCall = ([fullURLRequested hasPrefix:jsCallPrefix]);
+    
+    if (isJavaScriptCall) {
+        NSString *messageConveyed = [fullURLRequested substringFromIndex:jsCallPrefix.length];
+        NSLog(@"Your message was: %@", messageConveyed);
+        return NO;
+    }
```
jleandroperez: (And can we nuke this one as well?)

Simplenote/Classes/SPMarkdownParser.m
```diff
@@ -29,22 +29,40 @@ + (NSString *)renderHTMLFromMarkdownString:(NSString *)markdown
     hoedown_html_renderer_free(renderer);
     
     NSString *htmlString = [[NSString alloc] initWithData:htmlData encoding:NSUTF8StringEncoding];
+    
+    NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:
+                                  @"\n\\s+- \\[(x|X|o|o| )\\]\\s+([^\n]+)" options:0 error:nil];
```
jleandroperez: Would it be possible to add a brief comment, documenting what does this one do?

Thanks!!!

Simplenote/Classes/SPMarkdownPreviewViewController.m
dmsnell: just a scratchpad during my exploration

Simplenote/Classes/SPMarkdownParser.m
dmsnell: I have no idea! my main challenge appears to be getting events from inside the WebView out into the Objective-C code. this was shamelessly copied from StackExchange. again, since I don't know what I'm doing much of this is "doodling" as I learn

Simplenote/Classes/SPMarkdownParser.m
dmsnell: Definitely. Unfortunately it doesn't even do what I need it to do, which is link the rendered text to the plaintext in the note. In JavaScript I was able to do this because the RegExp replace takes a callback where I could track the match index in the full string. Here however I think I need to actually break this apart into a loop and build the string incrementally.

jleandroperez:
Ahhhhhhh @dmsnell thought the PR was ready for review, wasn't really following!!. We've got direct access to the original HTML (and prior to that, we can tamper with the Note's contents, directly).

What do we need to do sir?. It's possible we don't really need the JS bridge at all.

dmsnell:
> It's possible we don't really need the JS bridge at all.

Yes on two accounts. First, I was asking @frosty why we used a WebView instead of parsing the Markdown into native UI elements. My guess is that when he (or whoever it was, I forget) added the Markdown support that finding a good native library was difficult.

That being said, we could improve the performance and consistency as side-effects of using a more natural event mechanism by replacing the WebView.

On the other hand, if something here exists as it does in Android we could get away with fewer hacks. The ultimate goal, of course, is responding to the mouse-click on the todo items in such a way that the corresponding spot in the plaintext note itself toggles.

frosty:
Hey @dmsnell! Cool that you're taking a look at this. I did indeed do Markdown – essentially we use a web view because it's way easier to simply convert the Markdown to HTML and put that in a web view, than to try and produce a native view from it. There were lots of existing libraries for rendering Markdown to HTML, maybe one (unmaintained) for converting to some kind of native view.

We do now have a view in WPiOS that displays HTML in a native `UITextView`, so I guess it's possible we could adapt that to render the Markdown output with some special cases for todo lists... @aerych might have some insight on whether that'd be feasible.

dmsnell:
Closing this because I don't think it's a good approach even if it _could_ turn into shippable code. We probably need to have swipe-able todo items or something and I think that that would be best done with some native UI widgets.

Anyone disagree? Is tapping on an HTML checkbox a legitimate experience?

If we do HTML we can probably get the feature out across the mobile apps and the Electron app okay. I don't know about in app.simplenote.com or the macOS native app. If we want swiping and natural-feeling todos then I'm lost again.

Thanks for the feedback here!

jleandroperez:
I think swipeable is the way to go, but would need some thought (and coordination with the android fellows!!).

Thanks for checking into this @dmsnell!!

Issue #159, opened 2017-09-28
Content Loss on iOS 11 when dashes are added


roundhill:
From a user report:

> I’ve encountered what appears to be a major bug in the Simplenote app on my iPad since upgrading to iOS11. If I type a bunch of hyphens (the one over the “x”) in the middle of a body of text (—-), which I often do to mark off something I want to get back to, the text after the hyphens disappear after I close the file. This happened a couple of times before I figured out what was going on. In one instance, I was able to recover what I lost from the history, but in another case, I was too late in noticing it and lost a half day’s work.
The hyphen has changed with the new iOS11 keyboard. You used to be able to type single hyphens ad infinitum, but now if you type two in a row in merges them into a long dash. I don’t know how this is messing with your app, but it is.

To reproduce, just add some dashes anywhere in a note, looks like it needs to be 4 of them that get merged into two long dashes, and then go back to the notes list. All content after the dashes will be deleted :(

Arc676:
I can confirm that this happened to me too. At first I thought it was a generic sync error, but if another user has experienced this I guess it must be the dashes. 

I’m running Simplenote on iPhone.

gracie:
Also reported in #708974-zd

> I‘m using iOS 11.0.1 and I used several short dashes (The character next to the forward slash on the iOS keyboard.) as an underline.

serabi:
Also reported in #725635-zd

> [truncation] happens exactly when using the iOS app and have inserted multiple dashes on an empty line.

localastronaut:
Same error happening. Loosing all text between line breaks “—”. Can’t recover through history. Lost very important notes

jleandroperez:
Already fixed via #161. New binary out soon, thanks everyone for the reports!

PR #161, opened 2017-10-02
Disabling Smart Dashes + Quotes

jleandroperez:
### Details:
I've stumbled upon this exact same issue in the Aztec Editor. In this PR we're disabling Smart Quotes + Smart Dashes, since both of them trigger a replacement that's not being picked up by any of the callbacks.

This effectively fixes the Out of Sync bug reported in #159

Needs Review: @roundhill 
Thanks in advance sir!


roundhill:
Confirmed that this fixes the issue 👍 

I'm seeing a weird bug though where the note will slide up underneath the navigation bar when going to the detail view:
weird animation

Something with compiling with iOS 11 SDK now? Do you see the same @jleandroperez?

roundhill:
> Something with compiling with iOS 11 SDK now? Do you see the same @jleandroperez?

This is indeed the case, I've started #162 to address these UI issues.

@jleandroperez I think this one is good to merge 👍

jleandroperez:
Thanks for the review @roundhill!!

Issue #157, opened 2017-09-26
Layout issues on iPhone X in Landscape


roundhill:
Tried out the app in an iPhone X simulator, looks like we need to make some adjustments to the landscape layout:

<img width="864" alt="screen shot 2017-09-26 at 12 51 24 pm copy" src="https://user-images.githubusercontent.com/789137/30881618-40db487c-a2bb-11e7-8e79-ccb67cd16163.png">

davipontesblog:
Another report here - 953732-Zen

wildmaples:
This also seems to be fixed on iOS 11 and 12.

PR #162, opened 2017-10-03
iOS 11 / iPhone X Fixes

roundhill:
iOS 11 changed the way that a few layout properties work, including `contentInset` which messed up the rendering of a few areas of the app. This PR attempts to fix those things, including what I reported here: https://github.com/Automattic/simplenote-ios/pull/161#issuecomment-333606701

There was also some rendering issues on iPhone X in landscape which I addressed here as well. I still think we may need another PR for more iPhone X fixes as we discover them :)

Refs #157

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -582,7 +582,6 @@ CGFloat const SPMultitaskingCompactOneThirdWidth = 320.0f;
     // push off updating note text in order to speed up animated transition
     dispatch_async(dispatch_get_main_queue(), ^{
         _noteEditorTextView.attributedText = [note.content attributedString];
-        _noteEditorTextView.contentOffset = CGPointMake(0, -_noteEditorTextView.contentInset.top);
     });
     
     [self resetNavigationBarToIdentityWithAnimation:NO completion:nil];
--- a/Simplenote/Classes/SPTitleView.swift
+++ b/Simplenote/Classes/SPTitleView.swift
@@ -1,0 +1,17 @@
+//
+//  SPTitleView.swift
+//  Simplenote
+//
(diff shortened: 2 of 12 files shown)


Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -582,7 +582,6 @@ - (void)updateNote:(Note *)note {
     // push off updating note text in order to speed up animated transition
     dispatch_async(dispatch_get_main_queue(), ^{
         _noteEditorTextView.attributedText = [note.content attributedString];
-        _noteEditorTextView.contentOffset = CGPointMake(0, -_noteEditorTextView.contentInset.top);
```
jleandroperez: We should probably keep this one for iOS < 11 devices. Maybe wrap this one around an `if #available` check?

Simplenote/Classes/SPTitleView.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+//  Created by Dan Roundhill on 10/2/17.
+//  Copyright © 2017 Automattic. All rights reserved.
+//
+
+import UIKit
+
+class SPTitleView: UIView {
+
+    override var intrinsicContentSize: CGSize {
```
jleandroperez: This would be SO MUCH EASIER with Autolayout!! One of these days we should drop all of the manual layout code here. Badly

roundhill:
> I'm unable to see any difference in the iOS 10 simulator. Perhaps this behavior was also changed in iOS 10?

Yes, it seemed to fix an issue on iOS 11 by removing that line, and appeared to have no effect on earlier versions... but it is a bit scary ;)

jleandroperez:
Merging this PR in, we'll address anything else that might show up in other PR's.

Awesome work @roundhill !!

PR #165, opened 2017-10-23
Support Face ID for iPhone X

frosty:
This PR updates some UI and code to better support both Touch ID and Face ID (coming with iPhone X). 

While Face ID uses the same `LAContext` methods as Touch ID, I've updated any UI that references Touch ID to instead reference Face ID when appropriate. I've also tweaked the names of some methods and variables to use the term 'biometry' instead of 'touchID', so they more appropriately cover both types of authentication.

Finally, I've added a definition for `NSFaceIDUsageDescription` to the Info.plist, as I expect the app will  either crash or be able to utilize Face ID at all if this isn't present.

To test:

* Ensure Touch ID authentication still works as it did.
* Ensure the Touch ID setting in Settings is still correct on a Touch ID capable device.
* Try modifying these lines to set Face ID available to YES to test the alternative code path before we have access to an iPhone X: https://github.com/Automattic/simplenote-ios/pull/165/files#diff-ede834af434c1ec45f56abdc16ea99e9R152

Needs review: @jleandroperez

### DIFF
--- a/Simplenote/Classes/SPOptionsViewController.m
+++ b/Simplenote/Classes/SPOptionsViewController.m
@@ -278,7 +278,7 @@ typedef NS_ENUM(NSInteger, SPOptionsDebugRow) {
                     cell.textLabel.text = NSLocalizedString(@"Touch ID", @"Offer to enable Touch ID support if available and passcode is on.");
                     cell.selectionStyle = UITableViewCellSelectionStyleNone;
 
-                    BOOL isTouchIDTurnedOn = [[SPAppDelegate sharedDelegate] allowTouchIDInsteadOfPin];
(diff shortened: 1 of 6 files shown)


roundhill:
Do you think we could get this in before the iPhone X release, or must we wait until we have one in our hands?

frosty:
@roundhill I'd love us to do that, but I'm not sure if we need to fix our team issues first? @jleandroperez might know.

roundhill:
> not sure if we need to fix our team issues first

Yeah for sure. We're still waiting on one more fix from Apple... 🕓

jleandroperez:
Yup... unfortunately, we need to wait on Apple's fix. Which is way overdue by now!

frosty:
I'm not 100% sure what the flow is, but I think so yes.

roundhill:
I've tested this for a while, it works great. Thanks @frosty!

PR #175, opened 2017-11-13
Revert smart dash and quotes removal

roundhill:
Apple fixed the issue with smart dashes in iOS 11.1. I think we should add back the smart dash support since it is no longer an issue to honor the user's typing settings.

This reverts commit 67f0b676c3c9b835d9c3b6a80800a562d3b646da.


jleandroperez:
@roundhill how about if we lock up the fix so that it only runs in iOS 11.0? (pushed a small commit to do so). That way we bring the fix to anyone who is still in 11.0.

Mind taking a peek?

Thanks Dan, you're awesome!!

roundhill:
> how about if we lock up the fix so that it only runs in iOS 11.0

Yeah great idea 👍

roundhill:
:shipit:

jleandroperez:
Thanks Dan!

PR #177, opened 2017-11-14
Update Fabric Build Config

roundhill:
We'll now pull the keys from the `config.plist` for fabric. We also don't need to worry about the fabric key that used to be in `Simplenote-Info.plist` because we only use Crashlytics.

**To Test**
* Update your `config.plist` with the proper fabric keys, and run the app for release (or comment out the `USE_CRASHLYTICS` flags in `SPAppDelegate.m`.
* Cause a crash somewhere :)
* Next time you run the app, you should see crashlytics sending the crash data in the log.

I think we'll need to confirm that the dsym files are being properly uploaded to crashliytics as well after this change.

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -1284,7 +1284,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if [[ \"Distribution AppStore\" != \"${CONFIGURATION}\" ]]; then\n   echo \"Skipping Fabric\";\n   exit 0;\nfi\n\n\"${PODS_ROOT}/Fabric/run\"\n";
+			shellScript = "FABRIC_API_KEY=$(defaults read ${SRCROOT}/Simplenote/config.plist FabricApiKey);\nFABRIC_BUILD_KEY=$(defaults read ${SRCROOT}/Simplenote/config.plist FabricBuildKey);\n\nif [ ${FABRIC_API_KEY} == \"not-required\" ] || [ ${FABRIC_BUILD_KEY} == \"not-required\" ]; then\necho \"Could not find fabric keys. Check your config.plist.\";\nexit 0;\nfi\n\n\"${PODS_ROOT}/Fabric/run\" ${FABRIC_API_KEY} ${FABRIC_BUILD_KEY}\n";
 		};
 		B596467F1C46BD83001537E5 /* ShellScript */ = {
 			isa = PBXShellScriptBuildPhase;
--- a/Simplenote/SPAppDelegate.m
+++ b/Simplenote/SPAppDelegate.m
@@ -230,8 +230,6 @@
     NSString *email = self.simperium.user.email;
     NSString *key = [SPCredentials simplenoteCrashlyticsKey];
 
-    [Fabric with:@[CrashlyticsKit]];
-
     [Crashlytics startWithAPIKey:key];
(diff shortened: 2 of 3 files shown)


Simplenote/SPAppDelegate.m
```diff
@@ -230,8 +230,6 @@ - (void)setupCrashlytics
     NSString *email = self.simperium.user.email;
     NSString *key = [SPCredentials simplenoteCrashlyticsKey];
```
kwonye: This stays nil for me even after I set `FabricApiKey` and `FabricBuildKey`. Looks like we're still pulling from `simplenoteCrashlyticsKey`. Shouldn't we pull for the fabric key in `SPCredentials.simplenoteCrashlyticsKey()`?

roundhill: I _think_ we still need to have all three keys. The two for fabric are used in the project script strictly to upload the dsym.

Simplenote/SPAppDelegate.m
```diff
@@ -230,8 +230,6 @@ - (void)setupCrashlytics
...
 
-    [Fabric with:@[CrashlyticsKit]];
```
kwonye: I think we still need this no? Looking at the install guide here (albiet with possibly a small modification): https://fabric.io/kits/ios/crashlytics/install

`[Fabric with:@[[Crashlytics class]]];`

roundhill: I think you're right! I thought I read somewhere that it wasn't needed. But I added it back and I'm still seeing the crashes being sent:

```[Crashlytics:Crash:Reports] Submitting async /var/mobile/Containers/Data/Application/D48F905D-C8A2-45A1-846E-92B1517C6D1F/Library/Caches/com.crashlytics.data/com.codality.NotationalFlow/v3/prepared/BE27225B-B740-4745-85D5-A4FF00B3C853.multipartmime```

Comment (kwonye) [reply Simplenote/SPAppDelegate.m]: Ok thanks. I'd prefer it stay in just in case since the documentation says to do it.

**Comment by kwonye** [comment]
Now that it's on device it builds and sends crashes for me! :shipit:

File: Simplenote.xcodeproj/project.pbxproj [hunk 3]
```diff
@@ -1284,7 +1284,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if [[ \"Distribution AppStore\" != \"${CONFIGURATION}\" ]]; then\n   echo \"Skipping Fabric\";\n   exit 0;\nfi\n\n\"${PODS_ROOT}/Fabric/run\"\n";
+			shellScript = "FABRIC_API_KEY=$(defaults read ${SRCROOT}/Simplenote/config.plist FabricApiKey);\nFABRIC_BUILD_KEY=$(defaults read ${SRCROOT}/Simplenote/config.plist FabricBuildKey);\n\nif [ ${FABRIC_API_KEY} == \"not-required\" ] || [ ${FABRIC_BUILD_KEY} == \"not-required\" ]; then\necho \"Could not find fabric keys. Check your config.plist.\";\nexit 0;\nfi\n\n\"${PODS_ROOT}/Fabric/run\" ${FABRIC_API_KEY} ${FABRIC_BUILD_KEY}\n";
```
jleandroperez: Incoming late comment! only detail with this one is: this will run all the time, even for simulator builds.

(This line of code uploads the DSYM behind the scenes). Can we restore the `if` that skips fabric?

PR #180, opened 2017-11-21
Keychain Migration

roundhill:
Here's an attempt at migrating the token that is stored in the keychain to the new keychain.

This is pretty hard to test, but I was able to see it working by:

* Installing Simplenote from the app store and signing in on my test device.
* Delete the share target :)
* Changing the bundle identifier to `com.codality.NotationalFlow.MigrationDev`
* Set the UserDefault to the username at the beginning of the `migrateIfNecessary()` method: `UserDefaults.standard.set("{your-username", forKey:usernameKey)`
* Run the app to your device and you should be signed in (after dismissing the welcome view because the app thinks it is the first run)

This all hinges on the `SPUsername` user default persisting across the app updates, so we need to test that!


roundhill:
Was thinking that we should probably delete the token from the old keystore after we've migrated, or else sign out might not work?

jleandroperez:
@roundhill (IMHO) i'd probably do the first release, add some tooling (to verify if the migration is working out), and then ship a dot update with cleanup.

**JUST** in case!!.

roundhill:
Yup that sounds like a good plan to me

jleandroperez:
**Update:**

I've tweaked the targets as follows:

- Distribution AppStore
- Distribution Internal (Hockey App)
- Release (Matches all of the AppStore Identifiers)
- Debug

(Unless i'm missing something!!), in order to test the upgrade flow, as close as possible, we'd need to:

1. Install the AppStore build
2. Set the build target to **release** (for both, main app and Share Extension)
3. Hit run
4. Verify that the app remains logged in

We can definitely ship a TestFlight build to test the migration flow!!. One thing i'd love to note:

Removing the "Extra" new Entitlements effectively cause the flow, outlined above, to fail (the app just won't install!). So... things are looking good!!

jleandroperez (approved):
**Amazing** job sir!! thanks so much for pulling this together!!

:shipit:

roundhill:
🤞

mokagio:
@jleandroperez this is a long shot, but do you remember why you decided to have an AppStore and a Release target? 

When you and @roundhill worked on this PR, was Simplenote already under the Automattic App Store Connect account?  The acquisition happened [in 2013](https://blog.simperium.com/2013/01/24/simperium-joins-automattic/) but maybe the app stayed in the Simperium account for a while? 🤔 

I'm asking because I'm curious about the `previous-application-identifiers` entry added in 85a45111. That's what that's for, right? To account for app group id changes.

We're debating whether `previous-application-identifiers` is required or not in new targets and their entitlemens in #1408. My guess is that it is not and the few tests I've done validating an archived build made without it with the App Store seem to point in that direction. Maybe you have something to add?

😄

jleandroperez:
Hey there @mokagio !!! sorry I'm late, just got back from AFK.

At the time, it was common practise to have a `Release` target (mirroring the AppStore one), but with **device codesign wired**. This allowed us to test the app with the `DEBUG` macro disabled, while testing on the device.

As per the AppStore connect Q, plase refer to this PR post for the details `p77Llu-9yY-p2`. I believe it's not required for new targets (since the main binary should perform the keychain migration, and the old identifier should be gone).

Thank you for checking sir!!

jleandroperez:
OH!! I forgot to mention: the KeychainMigrator was implemented, specifically, to handle the TeamID migration over to the A8C developers portal.

mokagio:
Thanks @jleandroperez 🙌

roundhill:
@jleandroperez those were dark days. Miss you, pal! 😄

jleandroperez:
Miss you too @roundhill !!! Good ol' times, hope you're doing great over there!!!

PR #188, opened 2018-01-25
Add Alert Dialog Warning Upon Signout with Unsynced Notes

roundhill:
Some users are losing notes when they sign out because for one reason or another they have unsynced notes. This is a quick solution to that which searches for unsynced notes upon signout and warns the user if so. An option is presented to view the web app to reconcile their unsynced notes:

simulator screen shot - iphone x - 2018-01-25 at 09 51 42

The most reliable way I found to check if a note was unsynced was if its `ghostData` object had a `version` of `0`.

**To Test**
* Take the app offline, and create a note.
* Visit settings tap signout. You should see the alert dialog.
* Verify all of the buttons work as expected.
* Also verify that you don't see the prompt if all of your notes have synced.


jleandroperez:
@roundhill added the same mechanism we've got in macOS, so that we detect both, notes created while offline, and pre-existant notes that have been modified. Scenarios detailed below!


### Testing: Note created offline

1. Go offline (airplane mode)
2. Add a new note
3. Try to signout

Verify that the Unsent Changes alert shows up.

### Testing: Existing Note
1. Go offline (airplane mode)
2. Edit an existing note
3. Try to signout

Verify that the Unsent Changes alert shows up.

Please, feel free to merge if all looks good on your end!! thanks for doing this Dan!!

jleandroperez:
cc @dmsnell

roundhill:
Thanks @jleandroperez 👍

PR #202, opened 2018-05-17
Add WP.Com Sign In

roundhill:
This PR adds support for signing to Simplenote via your WordPress.com account. It supports creating a new Simplenote account if one didn't exist yet, or alternatively it will sign the user to an existing account.

wpcom-signin-ios

**To Test**
* From the login screen tap 'Sign in with WordPress.com'
* Authorize a WordPress.com account
* 3 things should happen:
 ** If a Simplenote account already exists for the account, it should be signed in successfully.
 ** If a Simplenote account didn't exist for the account, a new one should be created and you'll see the welcome note.
 ** If the WordPress.com account hasn't been verified by email, you should get an error message.

We're storing the WP.com token in the keychain for future integrations. If a user signed out, it will be deleted.

_Note: The reviewer will need some new secrets to add to their config file. Ping me on slack to get those._


Simplenote/SPAppDelegate.m
```diff
@@ -812,11 +815,72 @@ - (BOOL)application:(UIApplication *)application handleOpenURL:(NSURL *)url
...
+            [SPKeychain setPassword:token forService:kSimplenoteWPServiceName account:user error:&error];
+        }
+        
+        self.simperium.user = [[SPUser alloc] initWithEmail:user token:token];
+        [self.navigationController dismissViewControllerAnimated:YES completion:nil];
+        [self.simperium authenticationDidSucceedForUsername:user token:token];
+        
+        if ([isNew isEqualToString:@"true"]) {
+            [self createWelcomeNoteAfterDelay];
+        }
```
jleandroperez: Suggestion!!

Do you think it would be much of a trouble to move this whole new block of code to a new class, for encapsulation purposes?. Something like...

```
@interface WordPressAuthenticationHandler
+ (BOOL)isWordPressAuthenticationUrl:(NSURL*)url;
+ (SPUser *)extractSimplenoteUserFromUrl:(NSURL*)url;
@end
```

jleandroperez: (It'd be neat if we could, probably, have another class, App-Level, to encapsulate all of the authentication as well).

```
- (BOOL)application:(UIApplication *)application handleOpenURL:(NSURL *)url
{    
     //  ..
     else if ([WordPressAuthenticationHandler supportsUrl:url]) {
         SPUser *user = [WordPressAuthenticationHandler extractSimplenoteUserFromUrl:url];
         if (user == nil) {
             // error handling
            return;
         }

         [[SimplenoteAuthenticator shared] updateSimperiumUser:user];
     }
}
```

(ROUGH, very ROUGH idea), but the goal would be to move a little bit of the logic elsewhere, so that we begin cleaning up a little bit the AppDelegate. Lemme know what you think!!

roundhill: Nice idea, I added a new `WPAuthHandler` class in 3495c30. I also added support for a better error message if the user hasn't activated their wp.com account yet.

jleandroperez:
Code looks good @roundhill!! (thanks for those changes!!). Few notes:

## iPad Air UI Issues:

Note that the `Sign In With WordPress.com` button is misaligned.

simulator screen shot - ipad air 2 - 2018-05-21 at 20 42 21

## Error 500:
I'm triggering a constant Error 500 scenario:

1. Logged into WordPress.com with my default (a8c) account
2. Approved Simplenote
3. Constantly getting an Error 500 (screenshot below).

simulator screen shot - iphone se - 2018-05-21 at 20 45 12

The error is probably some sort of a8c-accounts-only issue (maybe?).

## Cannot Sign In:
1. Logged into WordPress.com (with a random, testing account, non a8c)
2. Approved Simplenote's permissions request
3. I got a brand new Simplenote account
4. Logged out
5. Tried to log in again using wpcom
6. Getting the `Couldn't Sign In` error consistently (screenshot below)

I've debugged this one. The callback we're getting looks like this (backend issue?):

```
(lldb) po url
simplenote://auth?error=true&code=0
```

simulator screen shot - iphone se - 2018-05-21 at 20 49 34

## Modal Style:

I've noticed SFSafariViewController gets pushed into the Simperium's Navigation Controller. Is that on purpose?. If we wanted "popup" behavior, we'd need to do this:

```
    SFSafariViewController *sfvc = [[SFSafariViewController alloc] initWithURL:[NSURL URLWithString:encodedUrl]];
    sfvc.modalPresentationStyle = UIModalPresentationFormSheet;
    [self presentViewController:sfvc animated:YES completion:nil];
```

I mean this one! =D  `modalPresentationStyle = UIModalPresentationFormSheet`

Lemme know if i can help debugging anything Dan!!!

roundhill:
> I'm triggering a constant Error 500 scenario:

Sorry about that, I think I broke the site when I pushed something yesterday. This should go away now.

jleandroperez:
@roundhill Dan!! still having few issues:

- **Error 500:** Still persistant for a8c accounts
-  **Cannot Sign In:** Regular WP account (still the callback seems to be `simplenote://auth?error=true&code=0`)
- **Broken iPhone SE Layout:** See the screenshot below!

simulator screen shot - iphone se - 2018-05-24 at 09 22 37

jleandroperez:
@roundhill Issue with (my personal) a8c account seems fixed. *But* still seeing the Cannot Sign In error. Scenario:

1. Sign in with WordPress.com
2. Enter credentials for a WPCOM account that **does not** have a Simplenote account
3. Verify Simplenote Logs in correctly (and a Simplenote account is created)
4. Add a note
5. Logout
6. Attempt to sign into the account you've just created (same credentials as in step 1)

Result: `Cannot Sign In` alert. Same error as before: the callback is `simplenote://auth?error=true&code=0`

jleandroperez:
@roundhill FWIW the wpcom username that triggered this is `jleandropereztest35` (but i can repro with any other testing account)

roundhill:
> Result: Cannot Sign In alert. Same error as before: the callback is simplenote://auth?error=true&code=0

Ugh, I was not actually saving the newly created user to the datastore! Fixed up server side. It should work fine now if you follow those steps again.

roundhill:
@jleandroperez added some tracks events in dc8bb40. Hopefully close to being ready?

jleandroperez:
@roundhill i'm getting this one now!. Is it possible a URL was changed... and maybe i don't have the latest `config.plist`?

simulator screen shot - iphone se - 2018-06-04 at 11 56 47

roundhill:
Thanks again @jleandroperez!

PR #218, opened 2018-08-22
Add Pin Lock Timeout Option

### LINKED ISSUE #215: Add Timeout Setting for PIN Lock
Let's add a timeout setting to the pin lock that will give the user some time to return to the app without the lock screen:

We can match the native display timeout settings, but I think adding 15 seconds would be good:

* Off
* 15 seconds
* 30 seconds
* 1 minute
* 2 minutes
* 3 minutes
* 4 minutes
* 5 minutes

roundhill:
Adding a new setting for the pin lock timeout that will dismiss the lock screen if the chosen amount of time hasn't passed yet. I'm not particularly fond of how much code it took to add this feature, but it does appear to work nicely :)

Screenshot:
simulator screen shot - iphone 7 - 2018-08-22 at 13 57 35

**To Test**
* Enable the passcode lock in Settings.
* Leave the app and come back, you should immediately see the passcode lock.
* Change the `Lock Timeout` setting to something other than `Off`.
* Leave the app and come back within the timeframe that you set. You should return to the app where you left off with no pin lock presented.
* Setting it back to `Off` should imeddiately show the pin lock again when returning to the app.

I've been 🤔 about this related to security. I suppose you could change the clock of the device to try and match when the user last left the app to bypass the lock screen, but you'd have to know when they last left the app so I'm not too worried about that. Any other security considerations the reviewer may have is most welcome!

Fixes #215

### DIFF
--- a/Simplenote/Classes/SPOptionsViewController.m
+++ b/Simplenote/Classes/SPOptionsViewController.m
@@ -342,7 +342,6 @@ typedef NS_ENUM(NSInteger, SPOptionsDebugRow) {
                     break;
                 }
                 case SPOptionsSecurityRowRowBiometry: {
-                    // Let the timeout row render the cell (next switch statement) if biometry is unavailable
                     if ([self biometryIsAvailable]) {
                         cell.textLabel.text = self.biometryTitle;
                         cell.selectionStyle = UITableViewCellSelectionStyleNone;
@@ -355,6 +354,8 @@ typedef NS_ENUM(NSInteger, SPOptionsDebugRow) {
                         
                         break;
                     }
+                    
+                    // No break here so we intentionally render the Timeout cell if biometry is disabled
                 }
                 case SPOptionsSecurityRowTimeout: {
                     cell.textLabel.text = NSLocalizedString(@"Lock Timeout", @"Setting for when the passcode lock should enable");
--- a/Simplenote/Classes/SPOptionsViewController.m
+++ b/Simplenote/Classes/SPOptionsViewController.m
@@ -153,7 +153,6 @@ typedef NS_ENUM(NSInteger, SPOptionsDebugRow) {
     [self.doneToolbar sizeToFit];
     
     UIBarButtonItem *doneButtonItem = [[UIBarButtonItem alloc] initWithTitle: NSLocalizedString(@"Done", @"Done toolbar button")                                                                                    style:UIBarButtonItemStylePlain target:self                                                                 action:@selector(pinTimeoutDoneAction:)];
-    
     UIBarButtonItem *fixedSpace = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemFixedSpace                                                                                    target:nil action:nil];
     
     fixedSpace.width = self.doneToolbar.frame.size.width;
--- a/Simplenote/Classes/SPOptionsViewController.m
+++ b/Simplenote/Classes/SPOptionsViewController.m
@@ -235,8 +235,9 @@ typedef NS_ENUM(NSInteger, SPOptionsDebugRow) {
         }
             
         case SPOptionsViewSectionsSecurity: {
-            int rowsToRemove = [self pinLockIsEnabled] ? 1 : 2;
-            return self.biometryIsAvailable ? SPOptionsSecurityRowRowCount : SPOptionsSecurityRowRowCount - rowsToRemove;
+            int rowsToRemove = self.biometryIsAvailable ? 0 : 1;
+            int disabledPinLockRows = [self biometryIsAvailable] ? 2 : 1;
+            return [self pinLockIsEnabled] ? SPOptionsSecurityRowRowCount - rowsToRemove : disabledPinLockRows;
         }
             
         case SPOptionsViewSectionsAbout: {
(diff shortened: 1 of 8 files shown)


jklausa:
How about using a monotonic clock like http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_gettime.html? It'll guard against someone changing the date/time manually _and_ more importantly, would get around bunch of weird behaviors users might see when traveling across timezones.

jklausa:
(Should be available on iOS since 10+.)

jklausa:
I'll admit I am not 100 % sure what security promises we're trying to make with the pin screen in the first place, but for full-paranoia-mode I'd think about maybe also storing the timeout preference in Keychain.

roundhill:
> I'd think about maybe also storing the timeout preference in Keychain

Great idea, I'll do that 👍

Simplenote/Classes/SPOptionsViewController.m
```diff
@@ -313,6 +350,15 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
                 }
+                case SPOptionsSecurityRowTimeout: {
+                    cell.textLabel.text = NSLocalizedString(@"Lock Timeout", @"Setting for when the passcode lock should enable");
+                    cell.selectionStyle = UITableViewCellSelectionStyleDefault;
+                    cell.accessoryView = nil;
+                    cell.tag = kTagTimeout;
+                    
+                    NSInteger timeoutPref = [[NSUserDefaults standardUserDefaults] integerForKey:kPinTimeoutPreferencesKey];
+                    [cell.detailTextLabel setText:timeoutPickerOptions[timeoutPref]];
+                }
```
jleandroperez: Same as above, missing `break;` on this one. (Nothing breaks though)

Simplenote/Classes/SPOptionsViewController.m
```diff
@@ -313,6 +350,15 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
                     cell.accessoryView = self.biometrySwitch;
                     cell.tag = kTagTouchID;
```
jleandroperez: Missing break on this one `break;`.

(ObjC's behavior is to fall thru unless there's a break)

Simplenote/Classes/SPOptionsViewController.m
```diff
@@ -649,4 +708,33 @@ - (BOOL)themePref
...
+
+- (void)pinTimeoutDoneAction:(id)sender
+{
+    [self.pinTimeoutTextField resignFirstResponder];
+}
+
+- (NSAttributedString *)pickerView:(UIPickerView *)pickerView attributedTitleForRow:(NSInteger)row forComponent:(NSInteger)component {
+    VSTheme *theme = [[VSThemeManager sharedManager] theme];
+    NSAttributedString *attributedTitle =
+    [[NSAttributedString alloc] initWithString:timeoutPickerOptions[row] attributes:@{NSForegroundColorAttributeName:[theme colorForKey:@"textColor"]}];
```
jleandroperez: Maybe put the allocation and the assignment in the same line? (and move the `attributes` to a variable to shorten the line length?)

Simplenote/SPAppDelegate.m
```diff
@@ -308,6 +308,48 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+    NSInteger timeoutPref = [[NSUserDefaults standardUserDefaults] integerForKey:kPinTimeoutPreferencesKey];
+    NSArray *timeoutValues = @[@0, @15, @30, @60, @120, @180, @240, @300];
+    
+    if (timeoutPref > timeoutValues.count) {
+        return 0;
+    }
+    
+    return [timeoutValues[timeoutPref] integerValue];
+}
+
```
jleandroperez: Nipticky Suggestion! how about moving `shouldBypassPinLock` + `getPinLockTimeoutSeconds` (and as much as possible, all related to the Pinlock!) to a class to contain all that logic?

`PinLockManager` || `PinLockController` || `LockController` || `LockManager` || `Anything better!`

Goal is to help reduce a bit the AppDelegate and contain all of the related code in a single spot.

roundhill: Haha I thought "I bet @jleandroperez will want this code to be moved to a static class" at one point ;)

Simplenote/Classes/SPOptionsViewController.m
roundhill: Because of how we hide the `biometry` row, I actually needed to leave this break in here. I added a comment in the code. I'm open to alternative ideas :)

Simplenote/SPAppDelegate.m
roundhill: Added a new swift class for this 👍

roundhill:
@jleandroperez Thank you! Ready for another look 👀

roundhill:
> As a result, the FaceID lock will be onscreen (even though it should have been bypassed!)

Yikes! Fixing this, good catch.

roundhill:
@jleandroperez Ready once again :) Thanks.

roundhill:
ok @jleandroperez I dug deep and fixed up a few things. Mainly I should have used `applicationDidBecomeActive` vs. `applicationWillEnterForeground`. I also fixed a pre-existing bug where the touch/face Id prompt would not show during certain situations. Here's a good test list:

- Swiping up to reveal control center should show face Id when returning to the app. Same for entering the multitask view. The timeout should work here as well, dismissing the pin lock if within the timeout period.
- Starting the app cold should show the pin lock or bypass it where appropriate.
- Going back to the home screen and back to the app multiple times should work as expected.

roundhill:
Thank you once again @jleandroperez!

Issue #170, opened 2017-11-02
iPad: User can get stuck on trash view


roundhill:
You can get stuck in the trash view if you:

* Tap on the search field
* Dismiss the keyboard using the soft key

It appears that the `Cancel` button is missing to the right of the navigation bar, but only on an iPad.

Issue #95, opened 2016-10-22
Collaborate Issues


charliescheer:
A user reported being unable to use collaborate over iOS.  

I tested with @jerrysarcastic and was in some cases able to get the notes to sync and in others was unable to make it work.  We could not nail down exactly what was different when it would sync and when it wouldn't.  

2883659-t

charliescheer:
Not the most details bug report that I wrote many years ago...

I recall doing a bunch of testing around this at the time with @jerrysarcastic  but can't recall too many specifics about what we tested and how.

Did some collaborating tests just now with iOS and it is working fine.  Closing this issue

Issue #65, opened 2016-08-14
Sync'ing Settings


kwonye:
#### Web app settings:

Sorting: Sort By: Modified/Created/Alphabetically Order: Descending, Ascending
Preview lines: 0 - 4
Tab Key Behavior (web only): Tab => Indent? yes/no
#### iOS app settings:
- Sort notes alphabetically: yes/no 
- Condensed note list: yes/no 
- Dark theme: yes/no

**Proposed Bucket:** prefs

There may be multiple objects in the bucket corresponding to different device class
#### Suggested object names/device classes:
- default 
  This is the one we will create for everyone, the web app might use this as well.
- ios-iphone 
  For iPhones, iOS will create this and maintain iOS-specific settings here (the ones prefixed with ios-)
- ios-ipad 
  For iPads, iOS will create this and maintain settings here (so users can have a separate ipad/iphone profile)
- android-phone 
  For android phone devices
- android-tablet 
  For android tablet devices
#### Details:

When modifying a setting, if modifying a global setting, modify the default object, if modifying a device-class specific setting, create or modify the specific object, like ios-iphonefirst

When reading a setting, read the default object and apply whichever settings are applicable for your device, then if it exists, read your device-class specific object, (iphone will look for ios-iphone in that bucket, ipad will look for ios-ipad), and apply the device specific settings from that object (the ones prefixed with ios- if you are an iOS device)

cc @fredrocious (YES this is a rip off from your shared note, minus the TouchID setting)
## 
#### Sample JSON Payload:

```
{
"sort_by" :  possible_values( "modified", "created", "alpha" )
"sort_order" : possible_values( -1, 1 ) (descending, ascending)
"markdown" : possible_values( false, true )

"web_tab" : possible_values( false, true )
"web_preview_lines" : possible_values( 0 - 4 )
"web_theme" : possible_values( "default" )

"ios_preview_lines" : possible_values( 0 - 4 )
"ios_theme" : possible_values( "default", "dark" )

"android_preview_lines" : possible_values( 0 - 4 )
"android_theme" : possible_values( "default", "dark" )
"android_font_size" : ?
}
```

Opened by @jleandroperez

bummytime:
Ref: #204

jleandroperez:
Added to the list of Feature Requests. Thanks eeeveryone!

Issue #35, opened 2016-08-14
Feature request: Add an option to disable data detectors on iOS


kwonye:
A user has commented that data detectors can be quite annoying while writing notes and other text. Such as writing a date that automatically highlights blue.

An ideal feature in this scenario, would be to offer a preference of disabling data detectors throughout Simplenote.

Replicate;
1. Write a date, ex. 1/30
2. Hide keyboard.
3. Date becomes highlighted in blue.

Opened by @kurzee

serabi:
A user is finding the data detectors in iOS make the application unusable for them. They would like to be able to disable data detectors throughout Simplenote. 

To replicate the problem, write an album or movie title as a note, hit return, and iOS turns the title into a clickable link that pops up a preview of an IMDB page.

jleandroperez:
I've added this to our feature requests. Thanks everyone for the suggestion!

PR #223, opened 2018-09-10
Typos on #35, #65, #95, #107, #170, #200, #206

kant:
Plus some more.


roundhill:
Hi! Thanks for contributing. We actually use GlotPress to update the translation files. Can you check over they are and make any suggestions needed to the translations?

https://translate.wordpress.com/projects/simplenote/ios/es/default/

Issue #191, opened 2018-02-09
Character count for Chinese/Japanese/Korean characters is incorrect


Piyopiyo-Kitsune:
Currently Simplenote does not count words correctly for Chinese, Japanese, and Korean when I tested on my iPad. This was originally 
Ideally, for these languages a character count would suffice as it's not commonplace to do word counts like we do in English. 

iPad(ver 4.4.5)
![image uploaded from ios](https://user-images.githubusercontent.com/17694715/36003642-c4f78f2a-0ce3-11e8-8caf-01b47a0ab581.png)
It says 15 words, when there are actually 7 words. 

Total characters(without spaces): 18

bummytime:
Fixed by #224

PR #224, opened 2018-09-14
fix-191-character-count

LispLY:
[issue 191](https://github.com/Automattic/simplenote-ios/issues/191)
Fixing character count in Chinese and Japanese.Sorry but Korean not included.

In China and Japan we usually count words by character so I made this fix.
In Korea things maybe like English or other Latin languages.

When  word count needed, determine the language first. If it's "zh"(Chinese) or "ja"(Japanese), we use the new characterCount, else we just use the old wordCount method.
I was tried to code in Objective-C but it was a little hard because NSCharacterSet is  an NSObject rather than a real Set. 

To be fixed:
1. Korean.
2. When there is not a dominant language, like 40% English plus 60% Japanese, the dominant language would be "undefined", and we just use the old wordCount method. Now we test the language by whole string. Testing the language by sentence may help but it is a little expensive.
3. Dominant language testing is only available since iOS 11, so we use the old way in iOS 10 and earlier versions.


roundhill:
How about we just show both word and character count like we recently did for the Android app?

https://user-images.githubusercontent.com/3122461/43176802-f2163faa-8ff7-11e8-8f61-e17800b4f15f.png

LispLY:
Showing both word and character count would be good and I am going to try.

BTW the word count in Android version doesn't work properly either when there's Chinese character.
Usually we count every single Chinese character as a "word" when we do word count. (Both Pages and MS Word do it this way. )

LispLY:
I've pushed the new version with character count and word count.
Now we need the localized string @"%d Characters" (Number of Characters in a note).

roundhill:
@LispLY sorry for the delay in getting back to you. I think this is looking good. Is there any chance you can add the locale-specific thousands separator to the counts?

screen shot 2018-11-19 at 2 00 09 pm

So that would be `2,175` and `17,052` for English.

LispLY:
@roundhill OK, I'll try.

LispLY:
@roundhill Locale-specific thousands separator is added.
Now￼￼ it's like:

US￼￼:
image

Italy:
image

roundhill:
Wonderful! Thanks so much for your hard work @LispLY. :shipit:

PR #235, opened 2018-10-11
Fix PIN lock not showing after reboot

### LINKED ISSUE #234: Pin Lock Not Activated After Reboot
When you have pin lock active on Simplenote 4.6.5 running on iOS 12 and reboot your phone, you can then access Simplenote without inputting the pin code.

Steps to reproduce.

1.) Turn on pin lock.

2.) Restart your phone.

3.) Open Simplenote. It will not prompt for the pin. The pin screen flashes for a quick second and then it opens the app without it. 

Initially reported in #1474205-zen and I was able to reproduce.

roundhill:
Fixes #234 with some additional logic to account for when the device was rebooted.

We were missing some logic:

 * The `shouldBypassPinLock` method should bail early if the user hasn't set a timeout.
 * When the device restarts, it starts the monotonic clock over. We should show the pin lock if the last stored time is greater than the current clock count.

**To Test**
* Enable passcode lock timeout, and reboot your device. You should see the passcode lock after starting the app.
* Repeat but with the passcode lock timeout turned to `Off`. The passcode lock should also appear.
* The passcode lock should still operate normally in other conditions.

### DIFF
--- a/Simplenote/SPPinLockManager.swift
+++ b/Simplenote/SPPinLockManager.swift
@@ -19,12 +19,22 @@ class SPPinLockManager: NSObject {
             return false
         }
         
+        let maxTimeoutSeconds = getPinLockTimeoutSeconds()
+        // User has timeout set to 'Off' setting (0)
+        if (maxTimeoutSeconds == 0) {
+            return false
+        }
+        
         var ts = timespec.init()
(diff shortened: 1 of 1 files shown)


roundhill:
Thanks @jleandroperez!

PR #237, opened 2018-10-17
Move .strings files to UTF-8

jtreanor:
`.strings` files are currently a mixture of UTF-8 and UTF-16LE. This converts all of them to UTF-8 and makes the appropriate Xcode project file updates.

This prepares Simplenote for the upcoming changes to translate.wordpress.com to use UTF-8.


roundhill:
:shipit:

Issue #255, opened 2018-12-05
Note editor: tags move down into the home indicator area


LispLY:
On iPhone X (or newer devices) in the note editor scene, the tags move down into the home indicator area after keyboard shows and hides.

before toggling the keyboard: (normal)
![img_802113bfc5e0-1](https://user-images.githubusercontent.com/12289707/49487648-6478b480-f87e-11e8-9702-3d412df37316.jpeg)

after toggling the keyboard: (tags moved down)
![img_4052420bca3f-1](https://user-images.githubusercontent.com/12289707/49487657-6c385900-f87e-11e8-815e-13d9e3e6a273.jpeg)

Simplenote version 4.6.6

PR #256, opened 2018-12-05
fix the tags position issue in iPhone X.

LispLY:
[issue 255](https://github.com/Automattic/simplenote-ios/issues/255)

In note editor we didn't count in the safeAreaInsets when recalculating the editor frame after the keyboard toggling.
Tried to fix that.

BTW, do you guys think if the note editor should expand to the bottom of screen or not  in iPhone X?


roundhill:
:shipit: thanks @LispLY!

> BTW, do you guys think if the note editor should expand to the bottom of screen or not in iPhone X?
IMO it's better if it stays within the safe areas. You still have plenty of editing space on the X/XR

Issue #55, opened 2016-08-14
Problems when typing Korean characters as a tag


kwonye:
A Korean user reports that when they type in the tag - 한글 - the app changes it to ㅎㅏㄴㄱㅡㄹ 

I've verified this with @gracejiyoung 

ticketid=2543059

"Similar problem happens with Japanese characters which may be related. The user inputs a Japanese letter and a number shows up on the screen instead. #2681786-t"

Opened by @rachelsquirrel

kristarella:
Also reported in #2755474

kristarella:
And #3164734-t

rachelsquirrel:
Also reported as an issue in the Electron app #831214.

Issue #190, opened 2018-02-03
Renaming tags bug


wenchoheelio:
You create a tag titled 'personal' (for instance) in a note and then you realise that you want it to be capitalised so you go to the tag, press edit, force touch the keyboard to move the cursor to just before the first letter, press backspace.. 

Bug number 1 - you'll see that the cursor jumps to the end of the word after you delete the first letter.

You then forcepress on the keyboard to go to the beginning of the word to put the capital letter in... the cursor jumps back to the end of the word again, but all looks good. You press 'Done' and sometimes (it appears not always)... 

Bug number 2 - the letter stays as a lowercase letter.

Issue #174, opened 2017-11-09
Unexpected characters appearing when typing double byte characters


jamiepalatnik:
This was reported here: 782536-zen

When I will be entering double byte character エントリー as a tag, unexpected characters will be entered like えええんんんエ. Would you please fix the issue?

I replied to the user to ask for more information about the app version and if this has happened at other times.

rachelsquirrel:
This is probably related to these issues:

https://github.com/Automattic/simplenote-ios/issues/58
https://github.com/Automattic/simplenote-ios/issues/55

rachelsquirrel:
Similar report in 871812-zen

eduardozulian:
Another report in 877182-zen

PR #259, opened 2018-12-25
fix #139, #55 Chinese input.

### LINKED ISSUE #139: Chinese inputs are not correctly defined
A note from one of our user in #3138309-T:

> Use the Chinese stroke input method, enter the label can not be used normally, each press a key will appear a long string of compliance, as shown.
>
> img_4538

LispLY:
Fixed [#139](https://github.com/Automattic/simplenote-ios/issues/139), [#55](https://github.com/Automattic/simplenote-ios/issues/55).

1. UITextfield conforms to UITextInput protocol. When inputing text in ideographic language, undetermined text is store as markedText. I guess explicitly setting the text of a UITextField may cause information lost about the markedText. I guess the glitch came from here. 
2. When text changed, `[SPTagEntryField onTextChanged:textField]` is called after `[delegate  textField:textField shouldChangeCharactersInRange:range replacementString:string]` . The latter already handled input with spaces, so I guess it’s not necessary to handle spaces in SPTagEntryField onTextChanged:textField].

### DIFF
--- a/Simplenote/Classes/SPTagEntryField.m
+++ b/Simplenote/Classes/SPTagEntryField.m
@@ -85,27 +85,6 @@
 }
 
 - (void)onTextChanged:(UITextField *)textField {
-    BOOL endEditing = NO;
-
-    NSString *text = textField.text;
-    if ([text hasPrefix:@" "]) {
-        text = nil;
-        endEditing = YES;
-    } else if ([text rangeOfString:@" "].location != NSNotFound) {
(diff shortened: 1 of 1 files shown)


LispLY:
Seems it also fixs issue [#190](https://github.com/Automattic/simplenote-ios/issues/190), [#174](https://github.com/Automattic/simplenote-ios/issues/174) and maybe some others.

roundhill:
Tests well for me, seems to still work fine. Thanks @LispLY! :shipit:

PR #258, opened 2018-12-13
Add Checklists

roundhill:
This PR adds supports for 'checkable' checklists in the note editor. There's a new button in the toolbar that will insert a new list:

screen shot 2018-12-12 at 9 50 57 am

Behind the scenes, the note saves with markdown checklist syntax (e.g. `- [ ], - [x]`). The note editor now uses a regex to parse which lines start with checklist markdown, and replaces them with tappable `NSTextAttachment` objects in the attributedText of the editor UITextView. See the `attributedStringWithChecklistAttachments` method.

When a note is saved, the new method `getPlainTextContent` searches for the `NSTextAttachments`, and replaces them with their plain text counterpart.

The checklists are also rendered in the note list.

Known issue:
- [x] The checkboxes aren't tapple on iOS 10. (making this require >= iOS 11)

**To Test**
* Open a note, and create a checklist! Pressing `return` at the end of a checklist item should automatically add a new one on the next line. The keyboard should also be capitalized after this action (I had to add some iffy workarounds for this 😬).
* Open the note on app.simplenote.com, it should have the correct markdown syntax displayed.
* Update the note on the web browser, the app should update and display any modified checklists appropriately.
* Test things that modify the note content, like the history browser and ensure they work as expected.
* Run the new tests!

PR #262, opened 2019-01-05
Render Checklists in the Markdown Preview

roundhill:
During testing of #258, users noticed that the markdown preview was not converting the checklist markdown syntax into a checkbox.

This PR fixes that by updating the markdown parser to `Hoextdown` (we used `Hoedown` before), which can parse the checklist syntax into HTML:

simulator screen shot - iphone x - 2019-01-04 at 15 14 24

Unfortunately there's no pod for `Hoextdown`, so I included it in the source of the project (which is the majority of the new code in this PR).

**To Test**
* Create a checklist in the editor.
* Enable markdown for the note from the info panel.
* Swipe to view the markdown preview. The checkboxes should display!
* The checkboxes should _not_ be tappable in the preview.


frosty:
Just noticed this does introduce a bunch of compiler warnings in the new files. Do we want to do anything about that?

roundhill:
> Just noticed this does introduce a bunch of compiler warnings in the new files. Do we want to do anything about that?

Hmm good catch. Do you think it'd be bad to ignore all warnings from the `Hoextdown` folder code? I can try and fix them but haven't used C in... 20 years :)

frosty:
I think it would be a reasonable compromise for now, particularly as they're part of another project. If we were still using the pod, we'd just be suppressing them with `inhibit_all_warnings!` anyway :)

roundhill:
Thanks @frosty!

PR #264, opened 2019-01-07
Fix linking addresses, phone numbers, etc in the note editor

### LINKED ISSUE #246: Link Clicking/Editing can be sporadic
Occasionally links don't seem to `Linkify` in the editor, and it can be hard to discover that you need to long-press to open a link in the browser.

Perhaps we can do something like Notes.app, it links urls when you open a note, but as soon as the editor receives focus all links are disabled in edit mode.

roundhill:
To fix #246 I removed the custom linkifier for the text editor in favor of the default iOS one. I missed that we needed to update the `DetectorType` back to `All` in order to auto-link things like addresses and phone numbers.

**To Test**
* Add a url, link and phone number to a note.
* Dismiss the keyboard.
* Links should be added to all the things :)

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -96,7 +96,7 @@ CGFloat const SPMultitaskingCompactOneThirdWidth = 320.0f;
         _noteEditorTextView = [[SPEditorTextView alloc] init];
         
         // Add Link Data Detector:
-        _noteEditorTextView.dataDetectorTypes = UIDataDetectorTypeLink;
+        _noteEditorTextView.dataDetectorTypes = UIDataDetectorTypeAll;
 
         // Note:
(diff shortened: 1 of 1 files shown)


Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -96,7 +96,7 @@ - (instancetype)init {
         _noteEditorTextView = [[SPEditorTextView alloc] init];
         
         // Add Link Data Detector:
```
frosty: Nitpick: can we remove this comment now?

roundhill:
> Also, is it an issue that phone numbers and addresses have a different style to web links?

I'm not sure what's going on there. I reverted to an earlier version with the older linkifier and saw the same behavior. Filed https://github.com/Automattic/simplenote-ios/issues/265

PR #269, opened 2019-01-09
Fix editor going off screen after switching themes

### LINKED ISSUE #268: Editor can get in a state where it is scrolled off of the screen
1. create a note in light mode.
2. switch to dark mode
3. open the note again
4. try to add text
5. watch the note screen disappear!

roundhill:
Fixes #268 

Changing themes after you have loaded the soft keyboard would cause the editor to scroll off the screen:

editor-bug

We have a pretty awful hack here that reattaches views to the windows of the app in order to reapply the new theme coloring. I poked around for alternative solutions but didn't come up with much, so this may be a necessary evil:

https://github.com/Automattic/simplenote-ios/blob/c589afcbb6419644d5f4cedebfc2637c7dffe15e/Simplenote/Classes/SPOptionsViewController.m#L785-L796

The issue was caused by that code messing with the view in a window called `UITextEffectsWindow`, which appears to be where the soft keyboard lives. I've fixed it by checking for that class and not doing anything to its views at all.

Alternative solutions are most welcome here!

### DIFF
--- a/Simplenote/Classes/SPOptionsViewController.m
+++ b/Simplenote/Classes/SPOptionsViewController.m
@@ -787,17 +787,12 @@ typedef NS_ENUM(NSInteger, SPOptionsDebugRow) {
     // HACK:
     // Yes, another one. Reference: http://stackoverflow.com/questions/17070582/using-uiappearance-and-switching-themes
     // This is required, so UINavigationBar picks up the new style
-    for (UIWindow *window in [[UIApplication sharedApplication] windows]) {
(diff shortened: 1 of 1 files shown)


roundhill:
> So, this does fix the issue! I was just wondering... I think we only ever have one window at a time? Can we just run this code on `[[UIApplication sharedApplication] keyWindow]` instead of looping through all the windows?

Good call! That seems to work fine from my testing. There was one other window it was touching previously besides the keyboard, something to do with transition animations.

PR #274, opened 2019-01-18
Add Support for Nested Checklists

roundhill:
I missed in the original checklists PR that GitHub checklists can be nested, so I'm adding support for that in the app:

screen shot 2019-01-18 at 8 34 05 am

**To Test**
* Create a checklist at any position on a line. It should insert at the proper position.
* Pressing enter should 'autobullet' a new checklist item on next line at the same position.
* Selecting multiple lines and/or one line of text and then inserting a checklist should still work as expected.

### DIFF
--- a/Simplenote/NSMutableAttributedString+Styling.m
+++ b/Simplenote/NSMutableAttributedString+Styling.m
@@ -12,6 +12,10 @@
 
 @implementation NSMutableAttributedString (Styling)
 
+const int RegexExpectedMatchGroups  = 3;
+const int RegexGroupIndexPrefix     = 1;
+const int RegexGroupIndexContent    = 2;
+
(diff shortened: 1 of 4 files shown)


Simplenote/NSMutableAttributedString+Styling.m
```diff
@@ -37,19 +37,25 @@ - (void)addChecklistAttachmentsForColor: (UIColor *)color  {
     int positionAdjustment = 0;
     CGFloat fontSize = [UIFont preferredFontForTextStyle:UIFontTextStyleHeadline].pointSize + 4;
     for (NSTextCheckingResult *match in matches) {
+        if ([match numberOfRanges] < 3) {
```
frosty: Should we use constants for these numbers here?

Issue #281, opened 2019-02-14
Text Selection Options Require Long Press Instead of Tap


mdrockwell:
#### Steps to reproduce
1. Open a note.
2. Place the cursor where you'd like to select text.
3. Tap to display the Select/Select All options.

#### What I expected
When tapping within a text field, the Select/Select All options should be displayed as they do throughout the rest of the system.

#### What happened instead
Nothing is displayed. You must tap and hold or double tap in order to display the Select/Select All options.

#### Simplenote version
4.6.8

#### iOS version
12.1.4

#### Screenshot / Video

https://d.pr/free/v/5gHGIK

Note: I'm single tapping throughout all parts of the video, aside from when the text selection options display — the first time I double tap, the second time I tap and hold.

This was initially reported in 1792352-zen and I was able to reproduce it.

darnelldibbles:
This appears to be a same issue reported by a user on Twitter:

https://twitter.com/ridogi/status/1097869854332325888/video/1

roundhill:
Fixed in #282

PR #282, opened 2019-02-15
Fixes for note editor text selection

### LINKED ISSUE #280: Cannot Highlight/Copy First Word of a Line
#### Steps to reproduce
1. Open a note in Simplenote.
2. Try to double tap to highlight the first word of the note.
3. See that it sort of jumps and won't highlight. 

#### What I expected
Double tapping the first word of the note would highlight the word and bring up options to copy/cut

#### What happened instead
The screen flashes the copy/cut options and the cursor jumps. The word is not highlighted

#### Simplenote version
4.6.8

#### iOS version
12.1.2

#### Screenshot / Video
https://cld.wthms.co/JuRxey

Also reported in #1784941-zen

roundhill:
Fixes #280, #281

For checklists we added a `UITapGestureRecognizer` on the editor `UITextView`. We already had been doing some tricks using `touchesBegan`, `touchesEnded` etc so I believe there were conflicts in using two 'touch listeners'. 

The `touchesBegan` workaround appeared to be something for iOS 7 (we target 10 now), so I removed them and moved some of its logic to the `UITapGestureRecongnizer`.

**To Test**
* In general make sure text editing works as expected in the editor. Scrolling and checklists should still work properly. Tapping to change cursor location should work everywhere in a note.
* Double-tapping a word should select it. Note: You can't do this unless the editor is active, which I think is expected.
* Tapping on the same cursor location twice should show the pop-up menu for text selection. #281 
* I've added a fix for iOS 10 users, they shouldn't see the checklist button in the toolbar because we can't render checkboxes on iOS 10.

### DIFF
--- a/Simplenote/Classes/SPEditorTextView.h
+++ b/Simplenote/Classes/SPEditorTextView.h
@@ -12,10 +12,7 @@
 
 extern NSString *const CheckListRegExPattern;
 
-@interface SPEditorTextView : SPTextView {
-    BOOL touchBegan;
-	CGPoint tappedPoint;
-}
+@interface SPEditorTextView : SPTextView
 
 @property (nonatomic) BOOL editing;
 @property (nonatomic) BOOL lockContentOffset;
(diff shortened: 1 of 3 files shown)


roundhill:
Thanks @frosty!

Issue #305, opened 2019-05-15
Markdown live preview


bummytime:
Currently when markdown is enabled in the editor, the user has to flip to preview mode in order to see how their marked-up text renders:

![markdown_preview](https://user-images.githubusercontent.com/154014/57785003-8102fd80-76f6-11e9-8e55-bf686bff5363.gif)

This dual-mode experience presents a few issues:
* Swiping left from the editor screen is not intuitive/obvious
* Many users continually flip back and forth between modes while editing
* (Subjective) Markdown markup doesn't look as nice as already-formatted text
* Users coming from more traditional editors (Notes.app, Word, etc) are not used to this
* The markup causes the user to lose focus on their content

So, my suggestion here is that we implement a live markdown preview where markup is rendered in real time. At the lowest levels, notes are still markdown...stored exactly as they are today. There are a few editors that already do this — a great example is [Typora](https://typora.io/).

Additionally, this new mode could be a global (and/or note-level) setting (on by default) that users can disable if they prefer the "classic" Simplenote editor experience.

Ref: https://github.com/Automattic/simplenote-macos/issues/326

sarahfg:
Just a note of appreciation on this report. I would not have known how to get to the preview on iOS if not for your screen recording here @bummytime

jos22:
+1 11326375-zd-a8c

PR #314, opened 2019-05-27
Markdown Preview: Fixing Anchor Links Crash

### LINKED ISSUE #247: Crash when tapping in-note links
#### Steps to reproduce

Using below to make a "table of contents":

```
# Table of Contents
1. [Example](#example)
2. [Example2](#example2)
3. [Third Example](#third-example)

## Example
## Example2
## Third Example
```

When viewing ios "preview" mode, tapping on the blue links to locations in the same note causes app to crash and close. 

#### What I expected
Tap on link, note scrolls to expected header.   

#### What happened instead
Crash and App closed

#### Simplenote version
Ios 4.6.5

#### iOS version
12.0.1

jleandroperez:
### Fix
This PR fixes a crash triggered after pressing Markdown Anchor Links.

Closes #247

### Test
1. Create a new note with the following contents:

  ```
# Table of Contents
1. [Example](#example)
2. [Example2](#example2)
3. [Third Example](#third-example)

## Example
## Example2
## Third Example
```
2. Press over any of the three links
3. Verify nothing breaks!

### Known Issues:
As far as I can tell, **Hoedown** seems incapable of dealing with "Anchor" links. The generated HTML strips out any `<a name="">` tags.

There seems not to be an easy fix, but in the light of #305, it's probably a good idea not to invest much time hunting that down.

@bummytime Any feedback would be more than welcome!!
Thanks in advance!!

WOOOOOO!!!

### DIFF
--- a/Simplenote/Classes/SPMarkdownPreviewViewController.m
+++ b/Simplenote/Classes/SPMarkdownPreviewViewController.m
@@ -101,19 +101,25 @@
 #pragma mark - WKNavigationDelegate
 
 - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler {
-    
-    if (navigationAction.navigationType == WKNavigationTypeLinkActivated) {
-        if ([SFSafariViewController class]) {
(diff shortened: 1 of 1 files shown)


jleandroperez:
Thank you sir!

PR #316, opened 2019-05-29
Fix CircleCI Config

jtreanor:
### Fix

This makes a few changes to fix CircleCI for the project:

- Updated the Orb so that `ios/test` defaults to latest iOS version (see https://github.com/wordpress-mobile/circleci-orbs/pull/23) and to allow it to run as a command (see https://github.com/wordpress-mobile/circleci-orbs/pull/24).
- Copy the demo `config.plist` to allow the build to pass.
- Don't ignore `Simplenote.xcworkspace` from git. Ignoring it means it is not present in cached CI builds. We could implement custom caching logic if there is an objection to this.

~**Note:** This is a draft because https://github.com/wordpress-mobile/circleci-orbs/pull/24 is awaiting review.~ Now merged, so this can be reviewed.

### Test

- The 'Test' CircleCI check is green and it ran the tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   # This uses the Orbs located at https://github.com/wordpress-mobile/circleci-orbs
-  ios: wordpress-mobile/ios@0.0.25
+  ios: wordpress-mobile/ios@dev:ios-updates
 
 workflows:
   simplenote_ios:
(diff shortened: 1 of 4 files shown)


jleandroperez:
(And apologies I crashed this PR!!)

jtreanor:
Thanks @jleandroperez! No apology needed 😄

PR #325, opened 2019-06-07
Share Extension: TextBundle Support

jleandroperez:
### Details:
- In this PR we're bringing the Share Extension's Swift to current style
- We've also moved few private types to their own files
- **ShareExtension**'s Activation Rule has been updated, based on [Matt's masterful work for WPiOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/9543)
- We now support **TextBundle**, **TextPack**, plain Text Files and Markdown files shared from 3rd party apps

Thank you @nheagy for helping me out all the way with this one!!
(And of course, thanks @bummytime for helping me out as well!!!)

cc @bummytime Ready for a first pass sir!

Ref. #323

---

### Scenario: Apple Notes
1. Launch Apple Notes
2. Open any of your notes from the main list
3. Press over the top right button (Share Sheet Action)
4. Pick Simplenote
5. Verify the Note's Contents are onScreen
6. Press over the top right **Post** button
7. Switch over to Simplenote
8. Verify the new note has been properly added

### Scenario: Bear + Share Sheet
1. Launch Bear
2. Open any of your notes
3. Press over the top right button (Share Sheet Action)
4. Pick Simplenote
5. Verify the Note's Contents are onScreen
6. Press over the top right **Post** button
7. Switch over to Simplenote
8. Verify the new note has been properly added

### Scenario: Bear + Export Note
1. Launch Bear
2. Open the Notes List
3. Swipe over any of the cells from the right (Right to Left!)
4. Press over the **More** button
5. Press over **Export note**
6. Pick **Text Bundle** and select Simplenote
7. Verify the Note's Contents are onScreen
8. Press over the top right **Post** button
9. Switch over to Simplenote
10. Verify the new note has been properly added
11. Swipe from the right and verify **Markdown Preview** is presented

Repeat for TXT + Markdown + Bear Note please!

PR #328, opened 2019-06-12
Adding TextBundleWrapper

jleandroperez:
### Details:

In this PR we're adding [TextBundleWrapper](https://github.com/shinyfrog/TextBundle/tree/master/TextBundle), which is a tool that will allow us to extract contents from  TextBundle-encoded-file.

We're importing the code as is. The goal of this PR is to actually shrink the size of its sibling #325.

### Testing:
Just verify the project builds fine!

cc @bummytime 
Thanks in advance!


jleandroperez:
Thank *you* sir!!

Issue #327, opened 2019-06-11
Update typography, icons, and colors to match new design guidlines 


bummytime:
Let's update the icons, typography, and colors throughout the app to match the latest design guidelines put forth by @SylvesterWilmott. Most of the updates should be contained within our [DB5](https://github.com/brentsimmons/DB5) [plist](https://github.com/Automattic/simplenote-ios/blob/9a9d3a73329fa2cc6de97240df46df9ac8ad0793/Simplenote/Simplenote-DB5.plist) and [markdown css](https://github.com/Automattic/simplenote-ios/tree/develop/Simplenote/Resources) files.

## Todos
- [x] Update everywhere UIFont is instantiated to now use the system font + [dynamic type styles](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/#dynamic-type-sizes) → #331
- [x] Adjust the line spacing in the editor to 1.4 em (24pt) → #331
- [ ] Normalize all color usage in light/dark themes (aka use the design guideline color specs)

## Current state of the app:
| Area | Light Theme | Dark Theme |
| ------------- | ------------- | ------------- |
| Note List | ![note_list_light](https://user-images.githubusercontent.com/154014/59308336-70e32d00-8c66-11e9-8cdb-1c0383a31b16.png) | ![note_list_dark](https://user-images.githubusercontent.com/154014/59308358-7b9dc200-8c66-11e9-9477-81c86485a6e1.png) |
| Tag List | ![tag_list_light](https://user-images.githubusercontent.com/154014/59308429-a4be5280-8c66-11e9-9c86-688afb1366a5.png) | ![tag_list_dark](https://user-images.githubusercontent.com/154014/59308431-a720ac80-8c66-11e9-9c75-446fa7b9ef56.png) |
| Settings | ![settings_light](https://user-images.githubusercontent.com/154014/59308476-bd2e6d00-8c66-11e9-8d21-24f35ae21c3f.png) | ![settings_dark](https://user-images.githubusercontent.com/154014/59308480-bf90c700-8c66-11e9-8f15-3397402c3970.png) |
| Note detail | ![note_detail_light](https://user-images.githubusercontent.com/154014/59308524-df27ef80-8c66-11e9-903c-e0d3251bb993.png) | ![note_detail_dark](https://user-images.githubusercontent.com/154014/59308530-e0f1b300-8c66-11e9-9515-d4bdcf0150bf.png) |
[...]

PR #331, opened 2019-06-17
Use dynamic type styles + increase editor line spacing

bummytime:
As part of brand refresh project, this PR updates typography across the app. 

Previously we leveraged hard-coded font names and sizes in DB5. In this PR we taking the very first step of removing DB5 and swapping out everywhere we call `[self.theme fontForKey:..` with `[UIFont preferredFontForTextStyle:...`. 

Addresses part of #327.

## Dynamic Type
Type usage across the app now leverages [dynamic type styles](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/#dynamic-type-sizes) instead of setting specific fonts and font sizes:

1_bf
2_bf
3_bf
4_bf

## Editor updates

Within the editor we now set the title to `UIFontTextStyleHeadline` and the body text to `UIFontTextStyleBody`. Additionally we set adjust the line spacing to approximately 1.4em.

### iPhone Xs iOS 12.2:
iphonexs_12 2

### iPhone 7, iOS 10.3:
iphone7_10 3

### iPad Pro 11" 12.2:
ipad_12 2

# Testing

To test this properly, we need to double check each screen/view app on a variety of devices and iOS levels. As a baseline I would try to look at the following devices:
* iPhone SE
* iPhone 7
* iPhone Xs
* iPad Pro

and verify everything looks good from iOS 10-12.

For extra credit, try changing the default font sizes in Settings.app and see if the app responds well (warning there are probably other bugs to file when checking this 😄 )

@jleandroperez Would you be so kind to take a peek at this?


bummytime:
Thank you @jleandroperez 🙇 !

PR #338, opened 2019-06-21
Display Published Icon for individual Notes

### LINKED ISSUE #312: Show share/publish icon in note list
The (old) web app had the published icon for any individual note:

sharing_icon

iOS does not. Let's fix that. 

Ref: https://github.com/Automattic/simplenote-macos/issues/4

jleandroperez:
### Details:
- We're now displaying an Accessory ImageView whenever a note is published
- **SPTransitionController** has been heavily tuned so that animations don't flicker when showing / hiding the Shared indicator
- New tool in charge of rendering a Note into a UIImage (wooo!)
- Thank you @astralbodies !!!! This PR was... days in the making

Closes #312

### Testing:
1. Launch Simplenote
2. Open any note
3. Press the ( i ) indicator and **publish** the note
4. Go back to the Notes List
5. Verify that the (newly) shared note displays a new asset on the right

**Please:** Repeat this test in:

- [x] iPhone SE
- [x] iPhone 8
- [x] iPhone 8 Plus
- [x] iPhone Xs
- [x] iPhone Xs Max
- [x] iPad Air Fullscreen
- [x] iPad Air Sideover
- [x] iPad Air Multitasking

**Plus** make sure that this works with:
- [x] Notes that contain a single preview line (IE. note with just one line of text)
- [x] Notes with multiple preview lines


Simplenote/Classes/SPTableViewCell.m
```diff
@@ -52,7 +56,14 @@ - (instancetype)initWithStyle:(UITableViewCellStyle)style reuseIdentifier:(NSStr
         _previewView.isAccessibilityElement = NO;
         _previewView.clipsToBounds = NO;
         [self.contentView addSubview:_previewView];
-        
+
+        _accessoryImageView = [UIImageView new];
+        _accessoryImageView.contentMode = UIViewContentModeCenter;
+
+        // Believe me: Using the cell's accessoryView is a nightmare. iPhone Xs Max and iPad has different metrics.
```
bummytime: +1 😄

SylvesterWilmott:
LGTM! Thanks @jleandroperez 

Just to confirm, is the left and right margins of that icon 16pt?

jleandroperez:
@SylvesterWilmott The share icon is, actually, positioned inside the frame we normally use to display the Note's preview.

This means that Readability metrics (handled by the system) affect the right margin. This is visible in iPad Devices (see screenshots below).

As per the left margin, it's set so that the text would never overlap the icon itself (and iOS itself appears to be adding an extra bit).

Please let me know if you'd like to fine tune anything. Thank you!!

Simulator Screen Shot - iPhone SE - 2019-07-04 at 11 53 04

Simulator Screen Shot - iPad Air (3rd generation) - 2019-07-04 at 11 51 30
Simulator Screen Shot - iPad Air (3rd generation) - 2019-07-04 at 11 51 31

SylvesterWilmott:
@jleandroperez can we explicitly set the left margin of the icon to 16pt so the text doesn't get so close to the icon as in the phone screenshot above?

jleandroperez:
### For documentation purposes:

- Padding between Preview Text <> Accessory Icon is now set to 16pt.

Thank you @bummytime and @SylvesterWilmott for reviewing this one!!


Simulator Screen Shot - iPhone SE - 2019-07-05 at 13 03 50

Issue #294, opened 2019-04-11
Cursor Jumps with Long Checklist


mgalbritton:
#### Steps to reproduce
1. Create a note with a checklist
2. Make the checklist have many items, say over 30
3. Close note; then re-open checklist
4. Check or uncheck an item
5. Screen will scroll somewhere but it's unclear because the cursor is not visible

#### What I expected
For the cursor to not jump to a spot in the note after checking or unchecking an item in a list

#### What happened instead
The cursor jumps to a random spot in the note and it's not easy to see the cursor to then move it back to where you were.

Looks very similar to what was happening here https://github.com/Automattic/simplenote-ios/issues/271

#### Simplenote version
4.7.0

#### iOS version
12.2 on an iPad air 2

#### Screenshot / Video
See #1915895-zen

bummytime:
Looks like we scroll to the bottom of the note — here is a video of this issue:

![checklist_issue](https://user-images.githubusercontent.com/154014/57651570-21401180-7593-11e9-9bc2-cbe7f8d95a7c.gif)

This issue seems to occur on 1st load of the note screen before the keyboard is raised. After the keyboard appears (and the note is in edit mode), the note does not scroll to the bottom when a list item is checked. Additionally when lowering the keyboard again, checking an item does not scroll the screen to the bottom:

![checklist_issue2](https://user-images.githubusercontent.com/154014/57651924-18037480-7594-11e9-860b-aace55f0bd1c.gif)

jleandroperez:
Closed via #341 by @mbaltaks.

Thank you sir!!

bummytime:
Thanks @mbaltaks!!

mbaltaks:
My pleasure! This one had bugged me for a while, and when I finally looked for a way to report the problem, I find that the source is available! So an hour later it's fixed, and I wished I'd known about this months ago. Live and learn. :)

PR #341, opened 2019-06-29
SPEditorTextView: fix 294 by setting the insertion point

mbaltaks:
Changing the state of a checkbox is a change event that leaves the
insertion point at the end of the note, which triggers a scrolling bug.

### Fix
https://github.com/Automattic/simplenote-ios/issues/294

### Test
Open note with more than one list, where the length of the note exceeds the size of the screen.
Without bringing up the keyboard, simply press a checkbox to toggle the state, and watch the note scroll to the end.

### Review
Very small change.


bummytime:
Thanks for your PR submission @mbaltaks. @jleandroperez would you mind reviewing this?

PR #349, opened 2019-07-05
Remove the Simperium link from About

bummytime:
This PR reorganizes the About VC's code a tad (not perfect, but better 😄 ) and removes the Simperium link:

@jleandroperez care for a quick review?


bummytime:
@jleandroperez, some things never change 🤝 😃 (thanks for the quick review!)

PR #355, opened 2019-07-10
Add lanes for automatic build and upload

loremattei:
This PR adds a few lanes to automate the build and the upload of the binaries to TestFlight and HockeyApp. 

# To Test:
1. Make sure to add the new required entries to your `~/.simplenoteios-env.default`.
2. Checkout a new branch.
3. change `VERSION_LONG` in `Version.Public.xcconfig` to a new version that has been uploaded yet. The best thing during testing is to change only the last number because it's the one that's not seen by the users.
4. Commit the changes.  
5. Run `bundle install`.
6. Run `bundle exec fastlane build_and_upload_release`.
7. Verify that the system reads the correct version and confirm.
8. Verify that the builds run without errors.
9. When prompted, login into TestFlight.
10. Verify that the process finishes without errors.
11. Verify that the build has been uploaded to HockeyApp.
12. Verify that the build has been uploaded to TestFlight.


loremattei:
> (which I realized after uploading probably should've been 4.8.1.1 🤦‍♂)

You're right here. 

> Not in the steps, but I would've expected the confirmation message would mention Hockey if we're uploading there. Instead, it just said:

Good point: I updated it.
 
> I have the FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD environment variable set, so I wasn't prompted to log in to Testflight

I guess this is not a change request, even if you marked it with a red cross :-) :-) :-) 

> Verify that the build has been uploaded to HockeyApp. [ I logged into Hockey, but I don't see Simplenote there. Am I looking in the wrong place? I saw the upload process succeed, so I'm thinking it went somewhere]

I never dug into the Hockey API, but I guess that like many others, it has a stage area for the uploads and then binaries are moved to the right application folder after a sort of "commit" command. I think your upload is only in the stage area. 
I now wonder how you did set up your env file if you don't see the app in Hockey :-) 
I pinged you in a Slack convo about the setup... you may want to check it before testing again. 
I also sent an invite for Hockey.

jkmassel:
One other thing @loremattei – I left a note in P9P2 about possibly moving the `.simplenoteios-env.default` file into the secret store. Could be in a future PR; doesn't have to block this one at all!!

But figured I'd note it here in case you want to take it into consideration for this PR? :)

loremattei:
>  Doing this will change version scheme in Hockey + App Store Connect. If this isn't a thing we want, we shouldn't merge this PR.

Yep, we are fine with this 👍

PR #361, opened 2019-07-15
[WIP] Nü Onboarding: Master PR

jleandroperez:
### Details:
[TBD]

### Testing:
- [ ] Verify rotations are disabled in every screen
- [ ] Verify the Input Text Views's outer border is slightly different when they become the first responder
- [ ] Verify the Keyboard always matches the kind of input text we're expecting
- [ ] Verify the Reveal Password button effectively displays the (hidden) text. Upon release, the password should get hidden again
- [ ] Verify the 1P button is hidden whenever 1P is not available
- [ ] Verify that the Primary Action Button (SignIn / SignUp) is grayed out whenever the input is actually invalid.
- [ ] Verify the Authentication UI **is not dismissable** in iOS 13.
- [ ] Validate that pressing **Next** / **Done** triggers inline warnings, whenever the input is invalid.
- [ ] Validate that Inline Warnings are dismissed the second the input becomes valid.
- [ ] Validate that both SignUp / SignIn display a spinner
- [ ] Verify Fallback to Login Mechanism
- [ ] Verify that pressing the main action renders both warnings, whenever applicable

[TBD]


jleandroperez:
Closing this one in favor of #384 and #385

PR #366, opened 2019-07-23
Updates Cocoapods to Mk 1.7.5

jleandroperez:
### Details:
Simplenote, as is, won't build in Xcode 11. [Details here](https://github.com/CocoaPods/CocoaPods/issues/9013).

In this PR we're bumping up our Cocoapods locked version to 1.7.5, so that the universe is in order, once again.

@bummytime may i bug you with this one?

Thanks in advance!!

### Testing:
- [x] Verify things build correctly in Xcode 10
- [x] (If possible) Verify the project builds in Xcode 11.


Coeur:
Be careful that you're moving from bundler 1.17.3 to bundler 2.0.3.
But I believe it's all good.

Gemfile.lock
```diff
@@ -259,4 +259,4 @@ DEPENDENCIES
   xcpretty-travis-formatter!
 
 BUNDLED WITH
-   1.17.3
+   2.0.1
```
bummytime: @jleandroperez I see we are moving the Bundler 2.0.1 in this PR. Is that required to fix the builds? Also, should we be using the [latest Bundler release v2.0.2](https://github.com/bundler/bundler/releases/tag/v2.0.2)?

/cc @loremattei in case there are some implications that we should consider....

jleandroperez:
@bummytime Reverted to Bundler Mark 1! Thank you sir!!

Thank you @jkmassel as well, and thank you @Coeur for the heads up!

jleandroperez:
Thanks everyone!!

PR #369, opened 2019-07-23
Notes List: iOS 13 clipsToBounds Bugfix

jleandroperez:
### Details:
We've got a simple bug in iOS 13. Yet, quite annoying:

1. Launch Simplenote on an iOS 13 device
2. Log into your account
3. Swipe over any of the notes in the list

As a result, you should get something like the following screenshot.

In this PR we're introducing a **one liner** fix. (Yay!)

Screen Shot 2019-07-23 at 5 16 26 PM

---

### Testing:
1. Launch Simplenote on an iOS 13 device
2. Log into your account
3. Swipe over any of the notes in the list
4. Verify the UI looks as expected


jleandroperez:
Thank you Matt!!

PR #373, opened 2019-07-26
Share Extension: First Responder Handling

jleandroperez:
### Details:
In this PR we're doing a super minor enhancement:

1. The keyboard is now expected to animate along with the Share Extension animation
2. And whenever the Share Extension is dismissed, the keyboard should animate along!

### Testing:
1. Launch Simplenote (**easy steps**
2. Switch over to Safari
3. Trigger Simplenote's Share Extension

- [x] Verify that the keyboard animates along, and the TextView is the first responder
- [x] Verify that dismissing the Share Extension causes the keyboard to animate along


jleandroperez:
Thank you sir!!

PR #378, opened 2019-07-30
Share extension: Absolute URL fallback for URLExtractor.buildExternalLinkNote

### LINKED ISSUE #377: Share extension: Unable to share URL's from certain apps
When sharing a URL from certain apps, the share extension UI is empty:

simple_share_news360

The issue is [this line here](https://github.com/Automattic/simplenote-ios/blob/develop/SimplenoteShare/Extractors/URLExtractor.swift#L119). Specifically `context.attributedContentText?.string` is sometimes `nil`. So the `guard` fails and the note ends up being blank.

To address this, if the above `guard` fails, lets create a new note just with the `url.absoluteString` (assuming it's not a file link).

### Reproduced
1. Open an article in the News360 app
2. Share the article to Simplenote (v4.9 and above)
3. Notice the UI is empty
4. Close the Simplenote share extension
5. share the article again, but this time to Notes.app
6. Notice the link is shared.

bummytime:
**Hotfix for v4.9**

This PR fixes `URLExtractor.buildExternalLinkNote` such that if content cannot be extracted from `context.attributedContentText` we at least insert the absolute URL into a new note.

fix_for_url_share

Fixes #377 

## Testing

1. Install and run the [News360 app](https://news360.com/)
2. Attempt to share an article to Simplenote
3. Verify the article link appears in the note.

@jleandroperez Would you mind giving this PR a 👀 ?

/cc @jtreanor

### DIFF
--- a/SimplenoteShare/Extractors/URLExtractor.swift
+++ b/SimplenoteShare/Extractors/URLExtractor.swift
@@ -116,11 +116,16 @@ private extension URLExtractor {
     /// Builds a Note for an external link
     ///
     func buildExternalLinkNote(with url: URL, context: NSExtensionContext) -> Note? {
-        guard url.isFileURL == false, let payload = context.attributedContentText?.string else {
+        guard url.isFileURL == false else {
             return nil
         }
 
-        let content = payload + "\n\n[" + url.absoluteString + "]"
+        var content = ""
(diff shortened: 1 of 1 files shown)


bummytime:
Thanks @jleandroperez

Issue #359, opened 2019-07-12
Nü Onboarding


jleandroperez:
### Mockup:

![login_and_signup_ios-1](https://user-images.githubusercontent.com/1195260/61151709-f30c7e80-a4bc-11e9-8ac8-468bee7bcefd.png)

jleandroperez:
Closing this one as done!! Wooo!!!

PR #383, opened 2019-08-05
Nü Onboarding: Mark I

jleandroperez:
### Details:
In the aims of keeping the main PR as small as possible, we're offloading few key components:

- **SPTextInputView:** Renders a UITextField with custom UI, which reacts to first responder events, and updates its look.
- **SPSquaredButton:** Convenience UIButton with rounded corners.

We're also:
- Adding new **UIColor(s)** for onboarding purposes, to both, DB5.plist and xcassets.
- Updating SPTextField to support insets.

Ref. #359

### Testing:
These two classes haven't been wired in this PR. Please:

- [x] Verify the code makes sense
- [x] Verify the app still runs!

This will be havily tested in the master Onboarding PR. Thank you!
cc @bummytime 

Thanks in advance!!


jleandroperez:
Thank you Bummy!!

PR #393, opened 2019-08-13
Bugfix: Onboarding Constraints

jleandroperez:
### Details:
There's a couple missing constraints in **SPAuthViewController**, which cause the UI to overflow in small screens (such as the iPhone SE).

In this small PR we're bringing order to the universe.

### Testing:
**Please:** Repeat the following steps for these devices:
- [ ] iPhone SE
- [ ] iPhone Xs
- [ ] iPhone Xs Max
- [ ] iPad Multitasking

1. Fresh install the app
2. Launch Simplenote and press the `Log In` button
3. Verify the Authentication UI looks okay

cc @frosty or @aerych 
Thanks in advance!!


jleandroperez:
Tracked down the warning in console. Boils down to a component with **preferredWidth** set to **Auto** (which shouldn't break!). Cool thing is that: doesn't happen in Xcode 11 / iOS 13.

I'll keep an eye on it, thank you for the heads up!!!

Thank you Eric!!!

PR #395, opened 2019-08-13
Notes List: Peek and Pop

jleandroperez:
### Details:
This PR adds Peek and Pop support to the notes list.

**Baby Swift steps, TINY refactors. I'm literally restraining myself.**

Closes #394

cc @frosty 
Thanks in advance James!!

### Testing: Note with Text
1. Log into your account
2. Force press over any of the notes that contains text
3. Verify the Preview shows up
4. Verify the keyboard does not show up
5. Press further
6. Verify the Note Editor is Onscreen
7. Verify the editor still doesn't show up

### Testing: Empty Note
1. Log into your account
2. Force press over any empty note
3. Verify the Preview shows up
4. Verify the keyboard does not show up
5. Press further
6. Verify the Note Editor is Onscreen
7. Verify the editor becomes the first responder

PR #401, opened 2019-08-19
Restoring First Responder Behavior

jleandroperez:
### Details:
As a side effect of #395, i'm afraid, the First Reponder behavior has been altered. Simplenote is expected to automatically display the keyboard **only** when the note being edited is effectively empty.

Thanks in advance gentlemen!
cc @bummytime or @aerych 

### Testing
- [x] Verify that opening an **Empty Note** immeditely causes the Keyboard to show up
- [x] Verify that opening a **NON Empty Note** doesn't cause the keyboard to show up


jleandroperez:
Thank you Eric!!

PR #405, opened 2019-08-20
Nü Credentials Mechanism

jleandroperez:
### Details:
This PR brings a **VERY** simplistic mechanism to deal with Secrets.

Gentlemen, would you be game for a *REALLY COOL* PR?
Thanks in advance!!

cc @jkmassel @jtreanor @bummytime 

P.s.: @jkmassel @jtreanor **SERIOUSLY**... thank you for being so awesome, and always helping us out!!

### Testing:
- [ ] `git pull` in your local copy of the `.mobile-secrets`
- [ ] Verify that you can effectively sign into your (production) account


jtreanor:
Thanks for making this change @jleandroperez!

Put of curiosity, what made you decide to take this approach (generating the code) vs keeping the `config.plist` but storing it in the secrets repo and copying it using the credentials tool?

jleandroperez:
Hey @jtreanor!! Few notes on the motivation behind this PR:

- The secrets are still stored in the *Secrets Repository* (JSON format!)
- I had a million problems with the `config.plist` file getting lost when switching branches
- A fresh checkout of the repository doesn't mean **anymore** that you need to manually copy the `config.plist` file.
- This approach (same one we use in WooCommerce) allows us to be explicit. `config.plist` is a bag of unknown things, which may or may not contain a given key. After this PR, you can just build Simplenote and open `SPCredentials.swift`. You'll see *right there* if you've got the right keys.

Thank you sir!!

🔥 🔥 🔥

Scripts/build-phases/replace_secrets.rb
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env ruby
+##
```
jkmassel: @jleandroperez – WDYT about skipping this and storing the file in mobile secrets as a `.swift` file rather than as JSON? It makes this script unnecessary as we can just copy the file directly without having the extra translation step.

Simplenote/SPLegacyCredentials.swift
```diff
@@ -101,13 +105,13 @@ class SPCredentials: NSObject {
...
     }
     
-    @objc static func WPCCClientSecret() -> String {
-        if let value = configDictionary().value(forKey: "WPCCClientSecret") {
-            return (value as! String)
-        } else {
-            return ""
-        }
-    }
+//    @objc static func WPCCClientSecret() -> String {
```
jkmassel: Do we want to keep this around if it's commented out? Would it be better to just remove it entirely?

Simplenote/simplenote_app_credentials_demo.json
```diff
@@ -0,0 +1,14 @@
+{
+  "appbot_key": "not-required",
+  "bithockey_id": "not-required",
+  "google_analytics_id": "not-required",
+  "itunes_app_id": "not-required",
+  "itunes_review_url": "not-required",
+  "sentry_dsn": "not-required",
```
jkmassel: We'll need something that looks like a valid DSN here. 

If we wind up merging the templates, we could get rid of this file entirely – the `not-required` bits could just become comments in the templates.

SimplenoteTests/SPCredentialsTests.swift
```diff
@@ -0,0 +1,25 @@
...
+@testable import Simplenote
+
+
+// MARK: - SPCredentials Unit Tests
+//
+class SPCredentialsTests: XCTestCase {
+
+    /// Verifies that the new Credentials mechanism match the legacy values
+    ///
+    func testCredentialsMatchTheirLegacyCredentialsCounterpart() {
```
jkmassel: The legacy credentials don't exist in CI, so this test can't pass – it only works on a local machine that already has the credentials present. IMHO we probably don't need it if we hand-verify both methods of credential import (from mobile secrets and from the template). WDYT?

README.md
```diff
@@ -60,10 +60,10 @@ Simplenote is powered by the [Simperium Sync'ing protocol](https://www.simperium
 After you've created your own Simperium application, copy the sample config. For example:
 
 ```
-cp Simplenote/config-demo.plist Simplenote/config.plist
+cp Simplenote/simplenote_app_credentials_demo.json ~/.mobile-secrets/iOS/simplenote/simplenote_app_credentials.json
```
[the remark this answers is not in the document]
jleandroperez: *Actuallly* in this PR we're adding the `simplenote_app_credentials_demo.json` file (which is in this repository!), so this **shouldn't** be an issue.

Nevertheless, I'll be **DRAMATICALLY SIMPLIFYING** things up, thanks to you!!! (THANKS for all the feedback!!)

Scripts/build-phases/replace_secrets.rb
jleandroperez: I mostly replicated the WooCommerce mechanism... which was implemented (**i think**) before the mobile-secrets repository was implemented.

And honestly, didn't even realize things can be a million times simpler. **Thank you** for mentioning this!!

Simplenote/SPLegacyCredentials.swift
jleandroperez: Update: This file is gone! thank you!

jleandroperez:
@jkmassel **Thank you** for reviewing this one!. Ready for another look sir... only pending is `what to do with SentrySN's placeholder`. Shoul it be a blank string?

.circleci/config.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Copy demo config.plist
-          command: cp Simplenote/config-demo.plist Simplenote/config.plist
+          name: Copy demo SPCredentials
```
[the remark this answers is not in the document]
jtreanor: I don't have a particularly strong opinion here and I see both sides. However, I would have a slight preference for the comfort of @jleandroperez's approach. It's not a lot of overhead to copy the demo file, but there is a risk of that happening accidentally if somehow the credentials aren't copied.

jleandroperez:
Thank you for reviewing this one sir!!

PR #412, opened 2019-08-29
SignUp >> Login String

jleandroperez:
### Details:
In this PR we're shipping a tiny minor string-fix. Zero behavior is altered.

cc @frosty May i trouble you with a really quick review?
Thank you!

### Test
1. Fresh install the app
2. Press over **Create an Account**
3. Enter a username + password that already exists

Verify that an alert shows up, and Simplenote is in uppercase:

```
The email you've entered is already associated with a simplenote account.
```


jleandroperez:
Thank you James!!

PR #419, opened 2019-09-09
Nukes IS_XCODE_11 guards

jleandroperez:
### Details:
In this PR we're removing every IS_XCODE_11 guard we've got, both, in ObjC and Swift.

### Testing:
- Verify the project builds correctly
- Verify the code makes sense!
- Run a smoke test over the app


jleandroperez:
Thank you James!!!

PR #421, opened 2019-09-15
Update CircleCI config for XCode 11

loremattei:
This PR updates the CircleCI configuration to use XCode 11 containers as we're going to release 4.10 built on it. 

### Test
1. Check code changes.
2. Make sure CircleCI is green.


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org' do
   gem 'rake'
-  gem 'cocoapods', '1.7.5'
+  gem 'cocoapods', '1.6.1'
```
jtreanor: Was this downgrade intentional?

loremattei: Yep. We are on 1.6 on other apps, so I thought it would be nice to sync.

jtreanor: I think 1.7.x is required to work correctly with Xcode 11. Is that right @jkmassel?

PR #425, opened 2019-09-20
Fixing Search Race Condition

jleandroperez:
### Fix
In this PR we're fixing a race condition, which is preventing the Search feature from highlighting the matches.

@aerych May I trouble you with a quick review?

Thanks in advance!!

### Test
1. Log into your account
2. Search for any note's contents
3. Press over any of the matches
4. Verify you get the bottom bar to iterate over the matches
5. Repeat 1-2, but instead of pressing, perform the Force Press gesture
6. Verify the bottom bar allows you to iterate over matches

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -767,6 +767,7 @@ CGFloat const SPMultitaskingCompactOneThirdWidth = 320.0f;
     if (string.length > 0) {
         bSearching = YES;
         _searchString = string;
+        searchResultRanges = nil;
         [self.navigationController setToolbarHidden:NO animated:YES];
     }
 }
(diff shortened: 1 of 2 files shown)


Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -616,12 +616,14 @@ - (void)updateNote:(Note *)note {
     }
     
     _currentNote = note;
-    [_noteEditorTextView scrollToTop];
-    
-    // push off updating note text in order to speed up animated transition
+    [self.noteEditorTextView scrollToTop];
+
+    // Synchronously set the TextView's contents. Otherwise we risk race conditions with `highlightSearchResults`
+    self.noteEditorTextView.attributedText = [note.content attributedString];
```
frosty: Is there a potential for this to cause any UI lagginess with large notes?

jleandroperez: Measured with a really seriously long document, it's in the order of milliseconds. And since it should always run on the main thread, **should not** be an issue.

Thanks for bringing that up!

jleandroperez:
@frosty Peek + Pop issue fixed. I'll be... refactoring everything all over. Gradually!!

Thank you so much for reviewing this one!!!

PR #436, opened 2019-10-01
Updates Deployment Target

jleandroperez:
### Details:
In this PR we're bumping up the deployment target to iOS 11, and fixing several warnings here and there.

cc @danielebogo sir, may I bug you with YET another PR?
Thanks in advance!!

### Test:
Please: run a smoke test on the app, and just check if anything out of the ordinary shows up. We've fixed a few minor warnings, but zero behavior was altered.

- [x] Login flow
- [x] Notes List
- [x] Tags List
- [x] Editor

### Release
`RELEASE-NOTES.txt` was updated in a8260b6 with:

> iOS 10 Is now the minimum supported OS.


RELEASE-NOTES.txt
```diff
@@ -1,6 +1,7 @@
 4.11.0
 ======
 -   Notes List refreshed to display Pinned state on the right hand side.
+-   iOS 10 Is now the minimum supported OS.
```
dmsnell: might be helpful to explain why we're making this change

```
- iOS 10 is now the minimum supported OS: deprecating XYZ, using new ABC interfaces, etc…
```

Simplenote/Classes/SPAboutViewController.swift
```diff
@@ -36,17 +36,17 @@ private extension SPAboutViewController {
 
     func setupDoneButton() {
         doneButton.translatesAutoresizingMaskIntoConstraints = false
-        doneButton.setTitle(NSLocalizedString("Done", comment: "Verb: Close current view"), for: UIControl.State.normal)
-        doneButton.setTitleColor(UIColor.white, for: UIControl.State.normal)
-        doneButton.setTitleColor(lightBlue, for: UIControl.State.highlighted)
-        doneButton.addTarget(self, action: #selector(onDoneTap(_:)), for: UIControl.Event.touchUpInside)
-        doneButton.contentHorizontalAlignment = UIControl.ContentHorizontalAlignment.right
+        doneButton.setTitle(NSLocalizedString("Done", comment: "Verb: Close current view"), for: .normal)
```
dmsnell: just curious as one new to Swift: what's the resolution of `.normal`? Is there a kind of expanding context resolution? does it default to `this`/`self` and walk up the inheritance chain until it finds the nearest `.normal` property?

Simplenote/Classes/SPAuthViewController.swift
```diff
@@ -243,23 +242,6 @@ private extension SPAuthViewController {
...
-        stackView.removeConstraint(stackViewTopConstraint)
-
-        /// Establish a new top constraint, against the topLayoutGuide.
-        stackViewTopConstraint = stackView.topAnchor.constraint(equalTo: topLayoutGuide.bottomAnchor,
-                                                                constant: stackViewTopConstraint.constant)
-
-        /// And, please, forgive me for pushing this awful method.
-        stackViewTopConstraint.isActive = true
-    }
-
```
dmsnell: curios, and the answer might belong above in the release notes, but I'm assuming that something in iOS 11 makes these constraints obsolete? are these constraints saved in a visual model somewhere? do they disappear because the underlying `UIViewController` changed its interface?

jleandroperez: Layout is set at the Interface Builder level (which is, then, translated into actual instances by NSCoder, when the `xib` is unarchived).

iOS 11 introduced a "Layout Guide" named [safeLayoutGuide](https://developer.apple.com/documentation/uikit/uiview/2891102-safearealayoutguide), which lets you define layout constraints for elements that must not be covered by navigation bars (or few ther components).

Since that layoutGuide was non-existant in iOS 10, we took a different approach there. Now, because we're bumping up the deployment target (minimum OS required), it's okay to remove that workaround-method.

Simplenote/Classes/SPAboutViewController.swift
jleandroperez: Swift infers / knows the namespace, by checking against the method's signature API. Suppose you've got this codebase:

```
enum Test {
   case first
   case second
}

func doSomething(with test: Test) { /*...*/ }
```

You can do this:

```
doSomething(with: Test.first)
```

But since Swift already knows the parameter should be of `Test` type, it's okay for you to do...

```
doSomething(with: .first)
```

And that's one of the tiny reasons why we love Swift. Don't we?

dmsnell: is that specifically because of `with:`?

I've seen similar systems in other languages that do things this way and even _with_ that sam keyword

Simplenote/Classes/SPAuthViewController.swift
dmsnell: awesome. that would be really nice to read in the RELEASE-NOTES too…

```
 - iOS11 is new minimal to rely on `safeLayoutGuide` and remove prior custom constraints
```

I'm not trying to leverage my weight here; just coming in as an outsider sharing my thoughts. thanks for helping me understand!

jleandroperez: Huh I did reply on the RELEASE-NOTES comment, but somehow it didn't get actually posted!!

Thing is: the RELEASE-NOTES are picked by a script, localized, and posted to the AppStore. It should be as human friendly as possible!

Thanks for bringing it up though!!!

Simplenote/Classes/SPAboutViewController.swift
jleandroperez: Oh nope, the "namespace resolution" capability is not tied up to the method's signature. You can literally use whatever in there, and it'll always work, regardless.

Relatively recent addition!

RELEASE-NOTES.txt
jleandroperez: (Just for the record!!) re-replying here. Catch is: the Release Notes are picked up by one of our deployment scripts, and goes directly to the AppStore (prior to being translated).

For that reason, we should keep things as human friendly as possible. Thanks for bringing this up D!

jleandroperez:
Thank you Daniele!

PR #438, opened 2019-10-02
Updating WordPress-AppBot dependency

jleandroperez:
### Details:
This PR contains literally zero changes. We're not locking against a Appbot commit hash anymore.
Instead... now we've got a beautiful tag, which should silence further Peril warnings.

cc @frosty 
James, may I bug you with a VERY small PR?
Thank you!!

### Testing:
- [ ] Please verify the app builds correctly!


dmsnell:
Where did we get "further Peril warnings"?

Were they appearing from rewritten history? A branch tip that moved past the commit?

jleandroperez:
@dmsnell the peril warning was referring to an external dependency that was being locked by means of a commit hash.

I've tagged those changes (in GIT), and patched up the **Podspec** (which is the way we're bringing in that external dependency).

No history was ever rewritten, promise you!!

Thank you both!!

PR #440, opened 2019-10-02
Splitting Notes List Initializer

jleandroperez:
### Description:
This PR introduces zero new features. We're splitting large methods into smallers, and moving code around. 

Goal: reduce #432's footprint!

@SergioEstevao sir, can I bug you with a... hopefully-easy-to-test PR?
**Thank you** for being so awesome!!

### Details:
- Notes List
  - We've split (part of the) initializer into a smaller method
  - Converted several ivars into proper ObjC properties
  - New **configureNavigationButtons** method, to be called by the initializer
- New **UISearchBar** extension
- SidebarContainer: New hook
- Wired proper constants here and there!

### Test:
**Please** perform a smoke test on the Notes List:

- [x] Verify: SearchBar's UI looks properly (position to be adjusted via #432!)
- [x] Verify: Notes List gets properly updated when Dark Mode is enabled
- [x] Verify: Condensed List mode refreshes proprely the Notes List
- [x] Verify: Navigation Bar shows the `Show Sidebar` button (top left)
- [x] Verify: Navigation Bar shows the `New Note` button (top right)
- [x] Verify: Navigation Bar shows the `Empty Trash` button (top right + viewing Trashed notes)

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -133,6 +133,7 @@
 		B50789FE1C1F5517009F097A /* SPInteractivePushPopAnimationController.m in Sources */ = {isa = PBXBuildFile; fileRef = 17C421CE1BE0BAB100752B56 /* SPInteractivePushPopAnimationController.m */; };
 		B50789FF1C1F5592009F097A /* markdown-default.css in Resources */ = {isa = PBXBuildFile; fileRef = 1748382E1BC1CC2D00E834AF /* markdown-default.css */; };
 		B5078A001C1F5595009F097A /* markdown-dark.css in Resources */ = {isa = PBXBuildFile; fileRef = 17BC3D461BC46BA800535DF3 /* markdown-dark.css */; };
+		B50E99F4234242130092F274 /* UISearchBar+Simplenote.swift in Sources */ = {isa = PBXBuildFile; fileRef = B50E99F3234242130092F274 /* UISearchBar+Simplenote.swift */; };
 		B50F479F1D1D77FC00822748 /* UIAlertController+Helpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = B50F479E1D1D77FC00822748 /* UIAlertController+Helpers.swift */; };
 		B50F47A41D1D78EA00822748 /* SPRatingsHelper.m in Sources */ = {isa = PBXBuildFile; fileRef = B50F47A31D1D78EA00822748 /* SPRatingsHelper.m */; };
 		B50F47A61D1D791B00822748 /* NSURL+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = B50F47A51D1D791B00822748 /* NSURL+Extensions.swift */; };
@@ -398,6 +399,7 @@
 		A774E3E4F5041D36FBE400BD /* Pods-Automattic-SimplenoteShare.distribution adhoc.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Automattic-SimplenoteShare.distribution adhoc.xcconfig"; path = "Pods/Target Support Files/Pods-Automattic-SimplenoteShare/Pods-Automattic-SimplenoteShare.distribution adhoc.xcconfig"; sourceTree = "<group>"; };
 		AE066B9C23C9166D41F1A732 /* Pods-SimplenoteTests.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-SimplenoteTests.debug.xcconfig"; path = "Pods/Target Support Files/Pods-SimplenoteTests/Pods-SimplenoteTests.debug.xcconfig"; sourceTree = "<group>"; };
 		B303123B163A5C7805D10DC7 /* Pods-Automattic-Simplenote.distribution adhoc.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-Automattic-Simplenote.distribution adhoc.xcconfig"; path = "Pods/Target Support Files/Pods-Automattic-Simplenote/Pods-Automattic-Simplenote.distribution adhoc.xcconfig"; sourceTree = "<group>"; };
+		B50E99F3234242130092F274 /* UISearchBar+Simplenote.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = "UISearchBar+Simplenote.swift"; path = "Classes/UISearchBar+Simplenote.swift"; sourceTree = "<group>"; };
 		B50F479E1D1D77FC00822748 /* UIAlertController+Helpers.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = "UIAlertController+Helpers.swift"; path = "Classes/UIAlertController+Helpers.swift"; sourceTree = "<group>"; };
 		B50F47A21D1D78EA00822748 /* SPRatingsHelper.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = SPRatingsHelper.h; path = Classes/SPRatingsHelper.h; sourceTree = "<group>"; };
 		B50F47A31D1D78EA00822748 /* SPRatingsHelper.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; name = SPRatingsHelper.m; path = Classes/SPRatingsHelper.m; sourceTree = "<group>"; };
@@ -1003,6 +1005,7 @@
 				B513F2B72319A9A40021CFA4 /* UITableViewCell+Simplenote.swift */,
 				B52AC185233587F8007B2E75 /* UITraitCollection+Simplenote.swift */,
 				B58039852322D4F90083C916 /* UIView+ImageRepresentation.swift */,
+				B50E99F3234242130092F274 /* UISearchBar+Simplenote.swift */,
 			);
 			name = Extensions;
 			sourceTree = "<group>";
@@ -1967,6 +1970,7 @@
 				46A3C9A617DFA81A002865AE /* SPActionSheet.m in Sources */,
 				E2922AC7180CAC62003AF914 /* SPSidebarViewController.m in Sources */,
 				375D24B221E01131007AB25A /* buffer.c in Sources */,
+				B50E99F4234242130092F274 /* UISearchBar+Simplenote.swift in Sources */,
 				46A3C9A717DFA81A002865AE /* DTPinLockController.m in Sources */,
 				B56A696522F9D54600B90398 /* SPLabel.swift in Sources */,
 				46A3C9A817DFA81A002865AE /* SPBorderedView.m in Sources */,
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -16,11 +16,15 @@ extension SPNoteListViewController {
     /// Refreshes the SearchBar Style.
     ///
     @objc
-    func styleSearchBar() {
+    func styleSearchBar(_ searchBar: UISearchBar) {
+        let backgroundImage = UIImage()
+        let backgroundColor = UIColor.color(name: .backgroundColor)?.withAlphaComponent(Constants.searchBarBackgroundAlpha)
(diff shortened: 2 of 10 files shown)


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -133,6 +133,7 @@
 		B50789FE1C1F5517009F097A /* SPInteractivePushPopAnimationController.m in Sources */ = {isa = PBXBuildFile; fileRef = 17C421CE1BE0BAB100752B56 /* SPInteractivePushPopAnimationController.m */; };
 		B50789FF1C1F5592009F097A /* markdown-default.css in Resources */ = {isa = PBXBuildFile; fileRef = 1748382E1BC1CC2D00E834AF /* markdown-default.css */; };
 		B5078A001C1F5595009F097A /* markdown-dark.css in Resources */ = {isa = PBXBuildFile; fileRef = 17BC3D461BC46BA800535DF3 /* markdown-dark.css */; };
+		B50E99F4234242130092F274 /* UISearchBar+Simplenote.swift in Sources */ = {isa = PBXBuildFile; fileRef = B50E99F3234242130092F274 /* UISearchBar+Simplenote.swift */; };
```
dmsnell: these are autogenerated? I feel like I see them in every PR.

jleandroperez: Yup, handled by Xcode!

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -50,6 +42,13 @@ @interface SPNoteListViewController () <ABXPromptViewDelegate,
                                         UITextFieldDelegate,
                                         SPTransitionControllerDelegate>
 
+@property (nonatomic, strong) UIBarButtonItem           *addButton;
```
SergioEstevao: I don't like this alignment... :)

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -276,36 +226,72 @@ - (void)updateNavigationBar {
...
 
+- (void)configureNavigationButtons {
+
+    /// Button: New Note
+    ///
+    self.addButton = [UIBarButtonItem barButtonWithImage:[UIImage imageNamed:@"icon_new_note"]
+                                          imageAlignment:UIBarButtonImageAlignmentRight
+                                                  target:self
+                                                selector:@selector(addButtonAction:)];
+    _addButton.isAccessibilityElement = YES;
```
SergioEstevao: Why use the iVar instead of the property?

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -756,7 +742,7 @@ - (void)updateFetchPredicate
 {
     SPAppDelegate *appDelegate = [SPAppDelegate sharedDelegate];
     if (appDelegate.selectedTag != nil &&
-        [appDelegate.selectedTag compare:@"trash"] == NSOrderedSame) {
+        [appDelegate.selectedTag compare:kSimplenoteTagTrashKey] == NSOrderedSame) {
```
SergioEstevao: if you are using a constant now why not jut do an isEqual?

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -998,9 +983,9 @@ - (void)sidebarDidHide {
     self.tableView.allowsSelection = !(tagFilterType == SPTagFilterTypeDeleted);
     [self.tableView setBorderVisibile:NO];
     
-    addButton.customView.tintAdjustmentMode = UIViewTintAdjustmentModeAutomatic;
-    addButton.enabled = YES;
-    emptyTrashButton.enabled = (tagFilterType == SPTagFilterTypeDeleted && [self numNotes] > 0) || tagFilterType != SPTagFilterTypeDeleted ? YES : NO;
+    self.addButton.customView.tintAdjustmentMode = UIViewTintAdjustmentModeAutomatic;
+    self.addButton.enabled = YES;
+    self.emptyTrashButton.enabled = (tagFilterType == SPTagFilterTypeDeleted && [self numNotes] > 0) || tagFilterType != SPTagFilterTypeDeleted ? YES : NO;
```
SergioEstevao: why do we need the ? here: ```tagFilterType != SPTagFilterTypeDeleted ? YES : NO;```

why not use just this: ```tagFilterType != SPTagFilterTypeDeleted ```

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -1018,14 +1004,14 @@ - (void)setWaitingForIndex:(BOOL)waiting {
     if (tagFilterType == SPTagFilterTypeDeleted && waiting)
         return;
     
-    if (waiting && self.navigationItem.titleView != activityIndicator && (self.fetchedResultsController.fetchedObjects.count == 0 || _firstLaunch)){
+    if (waiting && self.navigationItem.titleView != _activityIndicator && (self.fetchedResultsController.fetchedObjects.count == 0 || _firstLaunch)){
         
-        if (!activityIndicator)
-            activityIndicator = [[UIActivityIndicatorView alloc] initWithActivityIndicatorStyle:(SPUserInterface.isDark ? UIActivityIndicatorViewStyleWhite : UIActivityIndicatorViewStyleGray)];
+        if (!_activityIndicator)
```
SergioEstevao: You probably don't want to change much on this PR, but why not move this to the getter of the property in order to create it automatically when it's nil?

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -1074,7 +1060,7 @@ - (void)resetTitleView {
     [self animateTitleViewSwapWithNewView:_searchBarContainer
                                completion:^{
                                    self->bResetTitleView = NO;
-                                   [self->activityIndicator stopAnimating];
+                                   [self->_activityIndicator stopAnimating];
```
SergioEstevao: `->` ??? is this still a thing? Any reason for doing this?

Simplenote.xcodeproj/project.pbxproj
dmsnell: I'm not sure if we can do this easily or not, but in our JavaScript PRs we were having ugly diff noise from things like the `package-lock.json` file which is very similar to this - auto-generated and un-reviewable practically speaking.

I wonder if we can set some switch to hide these diffs from the review screen? As in, start collapsed with the `See changes` link to expand it…

dmsnell: lol: https://github.com/marcelofabri/pbxproj-diff-hider

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -13,22 +13,17 @@ extension SPNoteListViewController {
...
-        searchBarContainer.backgroundColor = .clear
-
-        // Apply font to search field by traversing subviews
-        for textField in searchBar.subviewsOfType(UITextField.self) {
-            textField.font = .preferredFont(forTextStyle: .body)
-            textField.textColor = .color(name: .textColor)
-            textField.keyboardAppearance = SPUserInterface.isDark ? .dark : .default
+        switch selectedTag {
+        case kSimplenoteTagTrashKey:
+            title = NSLocalizedString("Trash-noun", comment: "Title: Trash Tag is selected")
```
dmsnell: would I assume correctly if I were to say that this is `Trash-noun` but above we have `All Notes` because the translations have already been built based on those things, and that's the reason to leave them instead of update `All Notes-noun`? Or is it that `All notes` seems less ambiguous than `Trash` and that's why they are different?

Simplenote/Classes/SPNoteListViewController.m
dmsnell: as an outsider coming in, what don't you like about it?

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -196,6 +141,31 @@ - (void)updateRowHeight {
     [self.tableView reloadData];
 }
 
+
+#pragma mark - Notifications
```
dmsnell: what does "notifications" mean to Simplenote?

jleandroperez: In this context, it refers to [this](https://developer.apple.com/documentation/foundation/nsnotificationcenter).

**NSNotificationCenter** is a publisher / subscriber implementation, meant to channel events in-app. There are Notifications posted by the OS itself / UIKit components, and notes posted by ourselves.

The pragma mark in here reads as `NSNotification handling methods below`, so to speak!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
jleandroperez: I've kept the string this way, because the translations are already in place. You would never see `Trash-noun` in the actual UI (as long as there's an entry [here](https://github.com/Automattic/simplenote-ios/blob/develop/Simplenote/en.lproj/Localizable.strings)... or in the appropriate language).

**NSLocalizedString** takes the first parameter, and picks up the appropriate `Localizable.strings` file, based on the system language. Then, the target string is picked up.

In this case, `Trash-noun` [maps here](https://github.com/Automattic/simplenote-ios/blob/develop/Simplenote/en.lproj/Localizable.strings#L218), and you get `Trash` onscreen.

Simplenote/Classes/SPNoteListViewController.m
jleandroperez: LOL

This kind of "vertical alignment" has been... the source of conversations thru the years. Looks super cool, but has a maintainability fee (that's all, pretty much!).

😄

SergioEstevao: Sorry about this comment @dmsnell. This is like an ongoing conversation that I have with @jleandroperez about his preference of aligning vertically variable names.

I don't like it because if you this, every time you change a variable name or type you need to realign the variables introducing extra noises on diffs and unnecessary changes.

Simplenote/Classes/SPNoteListViewController.m
jleandroperez: Ahhhhhh I copied and pasted, and didn't notice that!! Patching up, thank you!!

Simplenote/Classes/SPNoteListViewController.m
jleandroperez: Fixed, thank you!

Simplenote/Classes/SPNoteListViewController.m
jleandroperez: Mostly because... it's being set in the instruction immediately before, and this way you avoid going thru the getter.

**BUT** that being said, for the sake of consistency!! Updating all of the ivar indirections to access directly the properties.

Thank you for bringing this up!

Simplenote/Classes/SPNoteListViewController.m
jleandroperez: Done, thanks!

Simplenote/Classes/SPNoteListViewController.m
jleandroperez: Sold!! Thank you!

jleandroperez:
@SergioEstevao **THANK YOU** for reviewing this one!!!

All changes addressed, ready for another look!

Simplenote.xcodeproj/project.pbxproj
jleandroperez: Personally I would never try to hide these diffs @dmsnell !!! Anything that breaks in there (bad merge, whatever) might render the entire project unusable.

We do actually check on those changes, and make sure files are not added twice (for instance). And it's especially useful when merging branches!

Simplenote/Classes/SPNoteListViewController.m
dmsnell: are there any standard deterministic auto-formatters for Objective-C/Swift?

have folks discussed using something like that to eliminate the conversations in code review about style/indentation?

Simplenote/Classes/SPNoteListViewController+Extensions.swift
dmsnell: thanks. I had assumed it was because they already existed.

Simplenote/Classes/SPNoteListViewController.m
dmsnell: interesting. a kind of app message-bus?

jleandroperez: Yes!! sent you details about this over slack!

Simplenote/Classes/SPNoteListViewController.m
jleandroperez: I've seen tools that auto format code... but... implementing / integrating those, in this particular case, takes effort we can't invest there.

Mostly because we *probably should* invest that time in migrating code over to Swift, and adding unit test!

jleandroperez:
Thank you Sergio!!!

PR #441, opened 2019-10-03
Fixing Darkness

### LINKED ISSUE #40: When OS and Simplenote are both set to Dark theme, unreadable text/dark overlay shows
**New info: This seems to happen if Simplenote and iOS are both set to Dark theme, AND if iOS is set to Dark and Simplenote is set to System Default.**

A user reports that then they edit a note with the dark theme enabled the text in the note switches from blue to black and becomes difficult to read. I can reproduce this with some of my own notes but not all of them. Here's an example:
img_1967

Another report here:
https://twitter.com/ChanceArthur/status/632102403572113408

ticketid=2370648

jleandroperez:
### Fix
This PR fixes an issue in which the TextView's Attributes wouldn't be properly set, the first time the editor is open.

cc @aerych Eric!! SORRY!!! yet another quick fix!!
Thank you!!!

Closes #40

### Test
1. Fresh install the app
2. Log into your account
3. Enable Dark Mode `(Settings > Theme)`
4. Kill the app + relaunch
5. Open any of the notes

- [ ] Verify the Foreground and Background colors are properly set.

### Release
`RELEASE-NOTES.txt` was updated in 91b74c3 with: 
> Added markdown support

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@
 -   Search Bar style refresh!
 -   Notes List refreshed to display Pinned state on the right hand side.
 -   iOS 11 Is now the minimum supported OS.
+-   Fixed a bug in which the Editor Colors might not be properly nitialized
 
 4.10.0
 ======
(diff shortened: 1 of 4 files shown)


jleandroperez:
Thank you sir!!

Issue #408, opened 2019-08-22
Spring Cleaning: Master Issue


jleandroperez:
### Task: Notes List
- [x] Refresh Note Table Cells
  - [x] Implement Autolayout!
  - [x] Pinned / Unpinned indicator should now be in the right hand side
  - [x] Split title and subhead into 2 labels
  - [x] Add 2pt space between the labels
  - [x] Fonts: Headline (17pt semibold) and Subhead (15pt). Check with Sly!
  - [x] Add cell separators
  - [x] Status Icons should match the Line Height, and be vertically centered against the 1st line
- [x] Refresh NavigationBar Style
  - [x] Update searchBar background
  - [x] Remove bottom separator
  - [x] Implement SearchController: SearchBar should be rendered right below the NavigationBar

### Task: Tags List
- [x] Implement Autolayout
- [x] Dynamic Fonts
- [x] Relocate the Settings button

### Task: Wire New Assets
- [x] Notes List
- [x] Tags List
- [x] Editor

### Task: SidebarContainer
- [x] Update the SidebarContainer's behavior: We should slide the navigationBar as well
- [x] We should slide the Tag List 80% of the way
- [x] Fix TableView Separator Line
- [x] Fix Slide from the Left Gesture (Back)

---

### Mockups

<img width="200" alt="Spring Cleaning: Notes List" src="https://user-images.githubusercontent.com/1195260/65787078-86edfe00-e12e-11e9-9b50-46843383a5b6.png"><img width="200" src="https://user-images.githubusercontent.com/1195260/63538162-b1c0b300-c4ed-11e9-87ad-fad4c5c30905.png"> | <img width="200" src="https://user-images.githubusercontent.com/1195260/63538164-b1c0b300-c4ed-11e9-97b0-33276a413201.png"><img width="200" alt="Screenshot 2019-09-11 at 16 57 41" src="https://user-images.githubusercontent.com/1195260/64734865-79d9d980-d4b5-11e9-8e36-7ccc0e836850.png">

<img width="200" alt="Screen Shot 2019-08-23 at 10 13 09 AM" src="https://user-images.githubusercontent.com/1195260/63595241-d0c35180-c58e-11e9-985b-27a09fa3c34e.png"><img width="200" alt="Screen Shot 2019-08-23 at 10 13 32 AM" src="https://user-images.githubusercontent.com/1195260/63595242-d0c35180-c58e-11e9-8f74-8da034f76f6e.png">

[...]

PR #449, opened 2019-10-08
Simplifying UIImageName Innerworkings

jleandroperez:
### Details
In this PR we're removing VSTheme usage for UIImage asset resolution. From now on:

1. We'll only need to add UIImage(s) to Asset Catalogues. No need to deal with db5 mappings.
2. We'll also need to have special considerations for iOS <13 and iOS >=13
  1. iOS 13 supports asset-dynamic-Interface-Style resolution. Assets can have, right in the catalogue, their Dark counterpart.
 2. iOS <13 does not support asset-dynamic-Interface-Style resolution. We need to manually add assets with different names.

Ref. #408 

@SergioEstevao May I bother you with yet another review?
Thank you sir!!

### Test
- [ ] Verify the Notes List looks correctly in Light + Dark Mode
- [ ] Verify the Tags List looks correctly in Light + Dark Mode
- [ ] Verify the Editor looks correctly in Light + Dark Mode

**Please** repeat for iOS 11 and iOS 13, since both OSs now end up hitting a different resolution API.


jleandroperez:
Thank you sir!!

PR #451, opened 2019-10-08
Adding Missing UIImageName Cases

jleandroperez:
### Details:
This is a tech debt PR: 

1. `UIImageName` enum cases will not have `Image` as a prefix, anymore. It was... terribly redundant.
1. We're adding missing UIImageName definitions
2. And, of course, we're wiring them. All over

**Note:** Template Images are now marked as such via Assets Catalog. Ain't that epic?

cc @aerych (SORRY to bug you so much, lately!!!)
Ref. #408

### Test
- [x] Verify the Notes List looks correctly in Light + Dark Mode
- [x] Verify the Tags List looks correctly in Light + Dark Mode
- [x] Verify the Editor looks correctly in Light + Dark Mode

Please repeat for iOS 11 and iOS 13, since both OSs now end up hitting a different resolution API.


jleandroperez:
Dark Mode issue filed here #465.

Thank you Eric!!!!

Issue #452, opened 2019-10-08
Search bar overlaps note list


rachelmcr:
### Expected

I expect the note list to appear below the search bar at all times.

### Observed

When the app is first opened, the search bar overlaps the first note in the list.

The text from the notes is also visible behind the text in the search bar. This occurs even if you are actively doing a search and then scrolling the note list.

<img src="https://user-images.githubusercontent.com/8658164/66407865-d7dedb80-e9e5-11e9-8043-ee1f6c5085cd.PNG" width="50%">

Make|Model|iOS Version|App Version
-|-|-|-
iPhone|XS|13.1.2|4.11.0.0 beta

jleandroperez:
@rachelmcr Does this happen, on your end, 10 out of 10? is this at launch, without scrolling a tiny bit?

rachelmcr:
This always happens after completely closing (force-closing) the app and then launching the app, without any scrolling (without touching the device at all).

If I pull down on the note list the first note will come out from underneath the search bar and look normal, but if I then force-close and launch the app again it'll return to that state.

jleandroperez:
Closed via #454. Let's reopen if needed!

Thank you @rachelmcr !!

PR #454, opened 2019-10-08
Ensuring First Row is always visible at launch

jleandroperez:
### Fix:
I've been unable to reproduce Issue #452 on my end. This PR is meant to act as a workaround, if the worst case scenario happens, for unknown reasons.

We're force-scrolling the NoteList's TableView to the very first row, as soon as the ViewController is moved to its parent.

cc @aerych Sir, may I bug you with this one?
Thank you!!

### Test
1. Launch the app and log into your account
2. Verify the first Note Row is visible
3. Kill the app
4. Relaunch and verify the first row is visible

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -45,7 +45,7 @@ extension SPNoteListViewController {
     ///
     @objc
     func ensureFirstRowIsVisible() {
-        guard tableView.isHidden == false else {
+        guard !tableView.isHidden else {
             return
         }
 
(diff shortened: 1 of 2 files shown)


Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -37,6 +37,20 @@ extension SPNoteListViewController {
...
     }
+
+    /// Workaround: Scroll to the very first row. Expected to be called *just* once, right after the view has been laid out, and has been moved
+    /// to its parent ViewController.
+    ///
+    /// Ref. Issue #452
+    ///
+    @objc
+    func ensureFirstRowIsVisible() {
+        guard tableView.isHidden == false else {
```
aerych: You could `guard !tableView.isHidden else {` if you want to be a tad more terse.  Up to you. :)

jleandroperez:
Thank you Eric!!

jleandroperez:
@loremattei May I but you with a new 4.11 TF build?

(Thank you Lore!!!)

PR #460, opened 2019-10-09
Tags List Refresh: Mark I

jleandroperez:
### Details:
In this PR we're introducing several changes to TagListViewController, which don't affect its behavior anyhow:

- Added a NIB to deal with TagListViewController's main view initialization
- Legacy ObjC variables converted into properties
- Use of braces normalized
- Nuked headers copyright
- Simplified snippets here and there
- Removed unused / unneeded snippets

@etoledom May I trouble you with (yet another) PR?

**Thank you Eduardo!!!**

Ref. #408

### Testing:
- [x] Verify the Tag List looks as expected
- [x] Verify Tags can be deleted
- [x] Verify Tags can be reordered
- [x] Verify Tags can be renamed


jleandroperez:
Tracking all your feedback in #408, tech debt section (revaming all the things all over). 

As per the network call glitch... I'll spin over it. Only valid workaround **may** result in integrity issues (say, disabling the FRC refresh while the edit mode is active!).

Thanks **a lot** @etoledom !!!

PR #466, opened 2019-10-11
Fixing Dark Mode Glitch

jleandroperez:
### Fix
In this PR we're fixing a Dark Mode issue that affects iOS 11 devices.

cc @aerych (Thank you!!!)

### Test
1. Launch Simplenote on a device running iOS 11
2. Log into your account
3. Tap over the top left icon and press over the **Settings** row
4. Switch to Dark Mode
5. Open any notes from your list

Verify that the Editor's background is actually dark!

### Release
These changes do not require release notes, since affect only 4.12 (unreleased!).


jleandroperez:
Thank you Eric!!

PR #470, opened 2019-10-15
Fixing iOS13 Aggressive Links Behavior

jleandroperez:
### Fix:
In this PR we're patching up an iOS 13 behavior that's causing Links to be open, accidentally, during scroll gestures.

@aerych May I bug you with a quick review?
**Thank you!!!**

[WPiOS Sibling Issue Here](https://github.com/wordpress-mobile/WordPress-iOS/pull/12612)

### Test:
1. Log into your account
2. Add a note with multiple links
3. Close the note and reopen

- [ ] Verify that placing your finger over any of the links doesn't immediately open SFSafariViewController
- [ ] Verify that tapping over any link effectively opens a WebView

### Release
`RELEASE-NOTES.txt` was updated in a5de8d5 with:

> Fixed an iOS 13 bug that caused links to be accidentally opened

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -7,6 +7,7 @@
 -   Notes List refreshed to display Pinned state on the right hand side.
 -   iOS 11 Is now the minimum supported OS.
 -   Fixed a bug in which the Editor Colors might not be properly initialized
+-   Fixed an iOS 13 bug that caused links to be accidentally opened
 
 4.10.0
 ======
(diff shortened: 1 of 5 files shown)


aerych:
Before I dive too far into this...  this seemed like an odd thing so I had myself a google and found:
- https://discussions.apple.com/thread/250671997 ( iOS 13: Touchscreen opens links on web pages at the slightest touch)
- https://help.foxnews.com/hc/en-us/articles/360037165893-Link-Sensitivity-Issue-with-iOS-App
(Apple is expected to release a fix for this issue in the next iOS update (13.2).)

Have we checked the 13.2 beta to see if the behavior is fixed?  Maybe we don't need the patches.  iirc @frosty has the beta installed if we need a test device.

PR #490, opened 2019-10-23
Bugfix: UI Freeze!

### LINKED ISSUE #487: Freeze after Transitions
```
After/While on search mode, the app became completely frozen.
My first guess would be that there's an invisible view covering the whole app that blocks all user interactions. If that's even possible, I would start searching there.

Sadly I haven't been able to reproduce it consistently, but these steps might help:

Swipe to show and hide the side bar. (Not sure if needed)
Having enough notes to cover at least 2 times the screen, scroll the note lists to the end (not sure if needed)
Start searching.
Try to swipe to show the side bar (nothing should happen)
Type something to start searching
Try to swipe again
Repeat adding and removing letters and swiping trying to show the side bar a few times
Remove all search characters
Press cancel to cancel search.
Check that the app became unresponsive.
It also might help triggering pick&pop or opening a note while on search mode. But I was able to reproduce it without this too.

I wasn't able to reproduce this all the times, but it did happen many times. So if it doesn't work at first, try a few more times 
```

Thank you @etoledom !!!

jleandroperez:
### Fix
In this PR we're fixing a critical bug that caused complete freeze of the UI. Plus we're also fixing a (related) bug that might have also led to an invalid state.

@etoledom May I trouble you, YET AGAIN, to verify if this is okay on your end?

Thanks in advance!!!

Closes #487

### Scenario: Frozen UI
0. Launch Simplenote and log into your account
1. Swipe right to reveal tags
2. Swipe left to hide tags
3. Press over the Search Bar
4. Open any note (no need to type any keyword!)
5. Hit back
6. Swipe from the left edge of the screen, towards the center
7. Press the **Cance** button

- [x] Verify the UI isn't completely frozen

**Note:** With these repro steps, I've managed to reproduce the original issue 10/10

### Scenario: Invalid Tag Menu OP
0. Launch Simplenote and log into your account
1. Tap the top left navigation bar button to reveal the tags
2. Swipe from the center of the screen towards the left edge, to hide the Tags UI
3. Press over the search bar to activate search
4. Swipe from the left edge towards the center

- [x] Verify the Tags List doesn't show up while there's a Search Active

### Release

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController.h
+++ b/Simplenote/Classes/SPNoteListViewController.h
@@ -22,7 +22,6 @@ typedef NS_ENUM(NSInteger, SPTagFilterType) {
     BOOL bSearching;
     BOOL bListViewIsEmpty;
     BOOL bIndexingNotes;
-    BOOL bShouldShowSidePanel;
 
     SPTagFilterType tagFilterType;
 }
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -375,7 +375,6 @@
     bSearching = NO;
 
     self.searchText = nil;
-    [self.searchController dismiss];
 
     [self update];
 }
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -305,7 +305,6 @@
 - (void)sidebarButtonAction:(id)sender {
     
     [self.tableView setEditing:NO];
-    bShouldShowSidePanel = YES;
 
     [SPTracker trackSidebarButtonPresed];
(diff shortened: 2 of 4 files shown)


dmsnell:
What was the cause of the freeze? As in, why did it freeze when there was only one view controller?

Also is the fix to include the gesture check? What are the other rearrangements in this PR doing?

jleandroperez:
@etoledom Thank you sir!! I'm still trying to repro on my end (the new scenario!). No luck so far, I'm afraid.

@dmsnell Freeze was caused by a gesture recognizer kicking in a custom Transition, when there was a single ViewController in the navigation stack (breaking the UI in a bad, bad way). 

As per the other changes, there's a logic fix for an issue that left the sidebar in an inconsistent state, allowing you to pan while Search was active.

jleandroperez:
**Thank you** @etoledom, you're the best!!!

PR #497, opened 2019-10-23
Fixing Swipe to Dismiss

### LINKED ISSUE #491: Swipe to Dismiss doesnt work
### Expected
Swiping from the left edge of the screen should always dismiss the editor.

### Observed
Whenever a Note has Markdown enabled, the standard Swipe to dismiss gesture doesn't work

### Reproduced
1. Log into your account
2. Add a new note
3. Enable Markdown in it
4. Attempt to dismiss from the left edge

As a result, the Editor won't be dismissed. Note that disabling Markdown fixes this issue.

---

### Patch

`SPInteractivePushPopAnimationController` L76:

We need to verify the translation of the PanGesutreRecognizer. If it's Left to Right, and the topViewController isn't what we would expect (in this case, MarkdownViewController), the InteractivePushPop OP should not be initialized.

jleandroperez:
### Fix
In this PR we're fixing a broken logic that prevented the Editor from being dismissed via Swipe gestures, when Markdown enabled notes were onScreen.

@frosty Apologies to bug you, yet again. Sending you this tiny PR over, for historic reasons.
The header in InteractivePushPopAnimationController reads `08/10/2015`. Time goes by quickly, doesn't it?

Thanks in advance!!

Closes #491

### Test
1. Log into your account
2. Add a new note
3. Enable Markdown in it

- [ ] Verify that Swipe from Right to Left displays the Markdown Preview UI
- [ ] Verify that Swipe from Left to Right dismisses the Markdown Preview UI
- [ ] Verify that you can swipe from the left edge of the screen, **anywhere**, and the Editor gets dismissed

### Release
`RELEASE-NOTES.txt` was updated in 62bbfed with:

> Fixed a bug that caused Swipe to Dismiss gesture not to properly work in the Editor


jleandroperez:
Thank you James!!!

PR #500, opened 2019-10-24
Enables Preserve Vector Data flag in assets

jleandroperez:
### Fix
We've got reports that some images are a bit blurry... when they shouldn't. Bottom line: Xcode compiles vectorial images into Static Images, and runtime resizing may result in blurry images.

In this PR we're enabling the Assets Catalogue `Preserve Vector Data` flag. (Wow!).

@aerych Eric, May i bug you with an (inoccuous) review?. 
Zero code changes on this one, and... well, I wasn't aware of this flag!

**Thank you!!!**

### Test
Please verify the app builds fine!

### Release

### DIFF
--- a/Simplenote/Images.xcassets/Icons/icon_add.imageset/Contents.json
+++ b/Simplenote/Images.xcassets/Icons/icon_add.imageset/Contents.json
@@ -10,6 +10,7 @@
     "author" : "xcode"
   },
   "properties" : {
-    "template-rendering-intent" : "template"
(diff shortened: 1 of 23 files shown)


jleandroperez:
Thank you Eric!!!

jleandroperez:
@loremattei sir, would it be possible to have a new Beta Build? (two PR's with bugfixes merged in!).

**Thank you!!!**

PR #502, opened 2019-10-25
Fixing Rotation Bug

jleandroperez:
### Fix
In this PR we're fixing an issue triggered whenever the user performs the Sidebar Interactive Swipe (to either reveal or hide the Tags List), along with a device rotation.

@aerych May I bug you (YET!!!) again?
**Thank you!!!**

### Test
0. Launch Simplenote on an iPad device (I've tested this both on the device + simulator)
1. With the device in portrait, open the sidebar.
2. Start to drag the sidebar closed, and then rotate the device while the sidebar is partially closed (keeping your finger on the screen).

- [x] Verify the UI is laid out as expected!

### Release

### DIFF
--- a/Simplenote/Classes/SPSidebarContainerViewController.m
+++ b/Simplenote/Classes/SPSidebarContainerViewController.m
@@ -296,16 +296,7 @@ static const CGFloat SPSidebarAnimationCompletionFactorZero = 0.0;
 
 - (UIViewPropertyAnimator *)animatorForSidebarVisibility:(BOOL)visible
 {
-    CGRect mainFrame = self.mainView.frame;
-    CGRect sideFrame = self.sidebarView.frame;
-
-    if (self.isSidebarVisible) {
-        mainFrame.origin.x = 0;
-        sideFrame.origin.x = -sideFrame.size.width;
-    } else {
-        mainFrame.origin.x = SPSidebarWidth;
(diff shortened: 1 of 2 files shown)


dmsnell:
Is this resolved due to the use of the `CGAffineTransform` or due to the identity?

I guess before the swipe would result in a hidden sidebar when we wanted it or a visible one when we hid it?

Apart from the transform the logic looks the same: hinging on the visibility, unless `visible` is updated  while `isSidebarVisible` isn’t during the rotation...?

jleandroperez:
@dmsnell the gist of this is:

`UIViewPropertyAnimator`, upon rotation... *as far as I can tell* was messing up with frames. Potentially trying to compensate for the delta between width and height.

By setting an Affine Transform, `UIViewPropertyAnimation` (which, allegedly, the one responsible for the broken position), doesn't tamper with anything, and simply performs the translation.

In open state, we've got a Transform that moves the `dx = 300 points`, whereas in closed state we use "Identity Transform" (AKA No delta x).

I've **Also** tried switching over to Autolayout, rather than doing manual frame repositioning. What's super odd is: the `Broken Layout` issue also occurs if we do use autolayout in the container view. Using Transforms seems to fix the glitch, for good.

:fire:

jleandroperez:
Thanks a lot @aerych!!!

PR #503, opened 2019-10-25
Preventing Interactive Animation Race Condition

jleandroperez:
### Fix
In this PR we're fixing a race condition, triggered whenever a Swipe gesture is performed, before the **Sidebar Animation** concludes.

We're also fine tunning the Sidebar animation: new FadeIn, and properties adjusted.

@danielebogo May I bug you with (YET) another fun PR?
Thank you sir!!

### Test
1. Swipe Left to Right / Right to left, repeatedly, quickly

- [x] Verify that everytime the Notes List is onscreen, vertical scroll is not possible!
- [x] Verify the Tags List now has a Fade In (supersmooth) animation
- [x] Verify the Notes List is slightly faded out, whenever the Tags List is revealed

### DIFF
--- a/Simplenote/Classes/SPSidebarContainerViewController.m
+++ b/Simplenote/Classes/SPSidebarContainerViewController.m
@@ -349,8 +349,10 @@ static const CGFloat SPSidebarAnimationCompletionFactorZero = 0.0;
         BOOL didBecomeVisible = self.isSidebarVisible;
 
         [self.animator addCompletion:^(UIViewAnimatingPosition finalPosition) {
-            weakSelf.isPanningActive = NO;
-            [weakSelf endSidebarTransition:didBecomeVisible];
+            __strong __typeof(weakSelf) strongSelf = weakSelf;
+
+            strongSelf.isPanningActive = NO;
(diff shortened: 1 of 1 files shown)


Simplenote/Classes/SPSidebarContainerViewController.m
```diff
@@ -340,12 +349,12 @@ - (void)panGestureWasRecognized:(UIPanGestureRecognizer *)gesture
         BOOL didBecomeVisible = self.isSidebarVisible;
 
         [self.animator addCompletion:^(UIViewAnimatingPosition finalPosition) {
+            weakSelf.isPanningActive = NO;
             [weakSelf endSidebarTransition:didBecomeVisible];
```
danielebogo: Because this is still obj-c I would probably use `__strong __typeof(weakSelf) strongSelf = weakSelf;` to make sure it will never become _nil_ when `endSidebarTransition:` is called. What do you think?

jleandroperez: Updating, **thank you**!!!!

jleandroperez:
Thanks a lot Daniele!!

PR #508, opened 2019-11-01
NavigationBar Blur Adjustments

jleandroperez:
### Fix
In this PR we're adjusting the NavigationBar Blur Effect so that it's inactive until the underlying ScrollView is scrolled.

We're only applying this over to:

- Notes List
- Note Editor

@aerych Me again!!! promise this one is fun!!!
Thanks in advance!!!

### Test
1. Log into your account
2. Verify the Notes List's navigation bar's background color matches with the notes list BG
3. Scroll upwards
4. Verify the Blur gradually fades in, as you scroll
5. Press over any note
6. Verify the navigation bar's background color matches the editor
7. Scroll upwards
8. Verify the blur slooowly fades in!

### Release

---

### Demo!

RPReplay_Final1573174324
RPReplay_Final1573174493


Simplenote/Classes/SPBlurEffectView.swift
```diff
@@ -0,0 +1,111 @@
...
+        super.init(effect: UIBlurEffect.simplenoteBlurEffect)
+        setupTintView()
+        startListeningToNotifications()
+    }
+
+    required init?(coder: NSCoder) {
+        super.init(effect: UIBlurEffect.simplenoteBlurEffect)
+        setupTintView()
+        startListeningToNotifications()
+    }
```
aerych: A little nitpicky of me, but maybe factor out the setup and notification calls into a common method each init can call?

Simplenote/Classes/SPMarkdownPreviewViewController.m
```diff
@@ -15,13 +15,19 @@
...
 @end
 
 @implementation SPMarkdownPreviewViewController
 
+- (void)dealloc
+{
+    self.webView.scrollView.delegate = nil;
+    self.webView.navigationDelegate = nil;
+}
+
```
aerych: These are already weak references. Maybe we don't need to set them to nil in dealloc?

Simplenote/Classes/SPBlurEffectView.swift
dmsnell: @aerych can you help me understand if you are recommending this instead of…

```swift
init() {
	…
}

required init?(code: NSCode) {
	init();
}
```

also, is what I'm asking even possible?

Simplenote/Classes/SPBlurEffectView.swift
```diff
@@ -0,0 +1,111 @@
...
+        return output
+    }()
+
+    /// Closure that's expected to return the TintColor we want. Runs on `iOS <13` everytime the active Theme is updated.
+    ///
+    var tintColorClosure: (() -> UIColor)? {
+        didSet {
+            refreshTintColor()
+        }
+    }
```
dmsnell: `Closure` is already encapsulated in the type here - the name is redundant.

Screen Shot 2019-11-08 at 3 39 39 PM

Is it our practice to do this or would it be normal to also choose to name this `tintColor` or `getTintColor`?

Simplenote/Classes/SPBlurEffectView.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    /// Returns a BlurEffectView meant to be used as a NavigationBar Background
+    ///
+    @objc
+    static func navigationBarBlurView() -> SPBlurEffectView {
+        let effectView = SPBlurEffectView()
+        effectView.isUserInteractionEnabled = false
+        effectView.tintColorClosure = {
+            return .simplenoteNavigationBarBackgroundColor
+        }
```
dmsnell: updating a remote property seems normal, but this is replacing a remote method.

is this suspicious?

Simplenote/Classes/SPBlurEffectView.swift
aerych: As currently written, `init` and `init(coder:)` both call `setupTintView` and `startListeningToNotifications`. Basically, the work is identical apart from the call to `super`. Its a trivial amount of duplication, so not really anything to fuss over, but if we wanted to be a smidge nicer we could have a single method to do the common work that each init method could call.  Something like...

```
init() {
    super.init(effect: UIBlurEffect.simplenoteBlurEffect)
    performInitialSetup()
}

required init?(coder: NSCoder) {
    super.init(effect: UIBlurEffect.simplenoteBlurEffect)
    performInitialSetup()
}

func performInitialSetup() {
    setupTintView()
    startListeningToNotifications()
}
```

Alternatively, and maybe nicer still, but I'm skeptical if it would compile...

```
init() {
    init(effect: UIBlurEffect.simplenoteBlurEffect)
}

required init?(coder: NSCoder) {
    init(effect: UIBlurEffect.simplenoteBlurEffect)
}

override init(effect: UIBlurEffect) {
    super.init(effect: effect)
    setupTintView()
    startListeningToNotifications()
}
```

dmsnell: yeah I was asking more about leaving `init()` the same as it is and calling `init()` from `init(coder: NSCoder)` instead of creating a third method that both call

why go through the extra work? is it necessary because of Framework? or Swift?

Simplenote/Classes/SPBlurEffectView.swift
jleandroperez: Thing is: the property contains a `closure` that gives you a `tintColor`. It's not a `tintColor` directly, so I wanted to be extra clear with the name!

Simplenote/Classes/SPBlurEffectView.swift
jleandroperez: Not suspicios at all, just providing an anonymous function that will return the `bg color`. Reason why we do this is: iOS <13 don't have native support for dark mode, and we **will need to run** such closure, when the theme changes.

Simplenote/Classes/SPMarkdownPreviewViewController.m
jleandroperez: You're right!! thank you!

Simplenote/Classes/SPBlurEffectView.swift
jleandroperez: Actually it did compile! Just had to add the `convenience` keyword.

```
    convenience init() {
        self.init(effect: .simplenoteBlurEffect)
    }

    convenience required init?(coder: NSCoder) {
        self.init(effect: .simplenoteBlurEffect)
    }

    init(effect: UIBlurEffect) {
        super.init(effect: effect)
        setupTintView()
        startListeningToNotifications()
    }
```

Thank you @aerych!!!

jleandroperez: @dmsnell oh forgot to answer! `init(coder:` sometimes is required by the superclass, and enforced by the compiler. Not adding it just throws a build error!

jleandroperez:
Thank you all!!

Simplenote/Classes/SPBlurEffectView.swift
aerych: TIL you can mark a `required` initializer as `convenience`.  That makes zero sense to me but 🤷‍♂

jleandroperez: I didn't know it either... but the compiler is happy about it (wooo!!)

Simplenote/Classes/SPBlurEffectView.swift
dmsnell: Screen Shot 2019-11-11 at 10 58 46 AM

just curious

jleandroperez: in that case, you'd need to do this:

```
tintView.backgroundColor = tintColorClosure?()
```

(Since it's a pointer to an anonymous function "closure"), assigning it plainly to a property that expects a UIColor is expected to break.

jleandroperez: That assignment reads as:

```
1. If there's a closure set in this property (that's the ? part)
2. Then execute (that's the () part)
3. Otherwise, nil is assigned
```

dmsnell: right, I just meant that the IDE already makes it obvious if I use a closure under the assumption that it's a property. it won't compile, but before I even try it tells m "you doofus, you can't do that"

PR #512, opened 2019-11-04
Fixing Markdown Gestures

### LINKED ISSUE #435: Selecting text in Markdown mode causes swipe gesture instead of selecting text
If you are looking at the text of a note in markdown and tap on a word to select one word, then use the selection controls and slide them to select more text it is causing a swipe gesture instead of selecting more text, which causes the view to switch back to edit instead of markdown.

Check out this video to see this in action:
<a href="https://charlietestsiteblog.files.wordpress.com/2019/09/rpreplay_final1569618440.mp4" />video</a>

Notice that it starts to select more text and then changes to a swipe gesture

iOS 13.1
Simplenote 4.1

jleandroperez:
### Fix
In this PR we're fixing a bug that caused text selection in the Markdown UI to accidentally dismiss the topViewController.

@aerych May you do the honors? **Thanks in advance!!!**

Closes #435

### Details:
We're letting `UINavigationController.interactivePopGestureRecognizer` deal with **pop** gestures performed over the MarkdownPreview UI.

- `UINavigationController.interactivePopGestureRecognizer.delegate` must be implemented, since having a custom TitleBar in the Notes Editor disables its standard behavior
- We're keeping a close eye on #490 since that's why such delegate was removed, in the first place
- Ref. this [SO Issue](https://stackoverflow.com/questions/34942571/how-to-enable-back-left-swipe-gesture-in-uinavigationcontroller-after-setting-le?noredirect=1&lq=1)

### Test
1. Launch Simplenote and log into your account
2. Open (any) Markdown Enabled note
3. Swipe from he center of the screen towards the left edge, in order to reveal Markdown Preview
4. Verify you can select text, and extend the selection from both ends
5. Verify you can swipe to dismiss the Markdown Preview UI from the left edge
6. Verify you can swipe to dismiss the Editor's UI

**Sample video below!!**

### Release
`RELEASE-NOTES.txt` was updated in 95ebede with:

> Fixed a bug that caused the Markdown Preview UI to get dismissed accidentally

---

### Demo:

Demo


jleandroperez:
@aerych **Thank you Eric**, absolutely missed iPad. Should be good now!!

aerych:
Hola @jleandroperez!  I retested in the iPad simulator and on a device and it looks like the swipe gesture is working for me now. :) Good to :shipit: if you're ready.

One other observation I'll share.  I did a side by side test with a previous version, 4.10.1, and the swipe gesture behaves a bit differently. In the older version, the swipe-to-go-back gesture is easier to do on the markdown preview--it seems like its not necessary to swipe right from the edge.  In this branch, a swipe has to start precisely from the edge or its ignored.  Not sure if its an important difference but wanted to mention it just in case. 
Cheers!

jleandroperez:
Yeahhh I'm afraid that change (Gesture Recognizer) is on purpose. The short version is: Markdown Preview uses a WebView, and checking if there's selected text (or its bounds) implies using Javascript.

Given this... situation... I've opted for simply falling back to the system's default behavior.

Thank you @aerych!!!

PR #515, opened 2019-11-05
Update Fastlane to upload Simplenote Internal to AppCenter

loremattei:
This PR updates the Fastlane configuration to switch binary upload from HockeyApp to MSAppCenter because of HockeyApp sunsetting. 

### Test
1. Create a new [API Token in AppCenter](https://appcenter.ms/settings/apitokens).
2. Add a `APPCENTER_API_TOKEN` entry to your `~/.simplenoteios-env.default` with the token above. 
3. Bump the version of the app.
4. `bundle install`.
5. Run `bundle exec fastlane build_and_upload_internal` and verify that the app is built and uploaded to App Center.

### Review

### Release

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -191,13 +191,13 @@ ENV["PROJECT_ROOT_FOLDER"]="./"
 
     sh("mv ../build/Simplenote.ipa \"../build/Simplenote Internal.ipa\"")
 
-    hockey(
-        api_token: ENV["HOCKEY_API_TOKEN"],        
-        public_identifier: ENV["HOCKEY_PUBLIC_ID"],
-        notify: "0",
-        status: "1",
-        ipa: "./build/Simplenote Internal.ipa",
-        dsym: "./build/Simplenote.app.dSYM.zip",
+    appcenter_upload(
+      api_token: ENV["APPCENTER_API_TOKEN"],
+      owner_name: "Hogwarts-Organization",
+      owner_type: "organization", 
+      app_name: "Simplenote",
+      file: "./build/Simplenote\ Internal.ipa",
+      notify_testers: false 
     )
 
     sentry_upload_dsym(
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -191,12 +191,14 @@ ENV["PROJECT_ROOT_FOLDER"]="./"
 
     sh("mv ../build/Simplenote.ipa \"../build/Simplenote Internal.ipa\"")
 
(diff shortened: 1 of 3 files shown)


jkmassel:
In order to fully test this PR, I think I need to be added to the `Hogwarts-Organization/Simplenote` project in App Center. 

But so far the rest works!

loremattei:
@jkmassel thanks for looking!
You should be added to the project now, though you may need to login on the web the first time you try to access App Center.
Let me know if you have other issues with this!

fastlane/Fastfile
```diff
@@ -191,13 +191,13 @@ ENV["PROJECT_ROOT_FOLDER"]="./"
...
-        notify: "0",
-        status: "1",
-        ipa: "./build/Simplenote Internal.ipa",
-        dsym: "./build/Simplenote.app.dSYM.zip",
+    appcenter_upload(
+      api_token: ENV["APPCENTER_API_TOKEN"],
+      owner_name: "Hogwarts-Organization",
+      owner_type: "organization", 
+      app_name: "Simplenote",
+      file: "./build/Simplenote\ Internal.ipa",
```
jkmassel: Me again! 🤦‍♂ 

When running this, I got an error saying "Couldn't find build file at path '/Users/jeremymassel/Projects/simplenote-ios/build/Simplenote.ipa'".

It looks like by default, `appcenter_upload` might use the value of `IPA_OUTPUT_PATH`, which in this case doesn't resolve to the new path after the `mv` command above.

In the [example Fastfile](https://github.com/microsoft/fastlane-plugin-appcenter/blob/master/fastlane/Fastfile) for appcenter, I found that they use `file` in just one case, but `ipa` everywhere else  – changing this key to `ipa` made the upload work for me, so I'm wondering if that is "more" correct?

WDYT?

loremattei: mmm... looking at the code `ipa` seems to be deprecated[ here](https://github.com/microsoft/fastlane-plugin-appcenter/blob/0d66522efafad3849e177b475477d30cc5c923e3/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb#L383), while `file` seems to be the [current way to go ](https://github.com/microsoft/fastlane-plugin-appcenter/blob/0d66522efafad3849e177b475477d30cc5c923e3/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb#L399).

It seems it performs more checks with the `file` option... let me check if some of them may fail for us for some reason.

loremattei:
Thanks for the review!

PR #516, opened 2019-11-05
Switch to appcenter

loremattei:
### Fix
This PR updates the app to use the new MSAppCenter SDK instead of the HockeyApp one. AppCenter should be able to interact with the old SDK for a while, but since we're migrating, it  makes sense to jump on the new one now.

### Test
1. Pull the latest version of the `.mobile-secrets` repository.
2. Build the internal version of the app and make sure the build completes without errors. 

### Release


jleandroperez:
@loremattei  Apologies!!! I didn't notice before, but this branch is bringing a bunch of warnings. I'll downgrade to Xcode 11.1, and double check there. But it would look like there's an issue with the AppCenter binary?

```
note: Linking a static library that was built with -gmodules, but the module cache was not found.  Redistributable static libraries should never be built with module debugging enabled.
```

Screen Shot 2019-11-05 at 10 45 26 AM

loremattei:
Oh! Good catch @jleandroperez! Thanks for spotting this!

So, it looks like this is a known problem with the latest release of the MSAppCenter pod :-(. 
Luckily there's already a fix in their `develop` branch: https://github.com/microsoft/appcenter-sdk-apple/pull/1836, so it should be fixed with the next release. 

In the meantime, I checked the [changelogs](https://github.com/microsoft/appcenter-sdk-apple/releases) for the last two releases (which are the ones with this bug) and it looks like there's nothing that affects us, so I moved back this branch to version `2.3.0` that doesn't generate the warnings. 

Let me know if this is ok for you!

jleandroperez:
Definitely super!!! thank you @loremattei !!! Squirrel stands!!

:shipit:

PR #518, opened 2019-11-05
Disabling Markdown Swipe during Text Selection

### LINKED ISSUE #472: Copy Text Not Working iOS 13
When trying to copy text while in iOS 13, it will allow you to double tap a single word but if you try to swipe to expand the selection, it disallows copying overall. So you can either select a single word or the whole text of the note. This happens when Markdown is on. 

### Expected
When I select a word to copy, I should be able to swipe and highlight more text. 

### Observed
You cannot select more than one word. You can either select the single word or the whole text.

### Reproduced
1. Open a note.
2. Double tap on a word to copy it. Then try to swipe to expand the text selection to copy.
3. Notice how the highlighting disappears and does not allow you to select anymore. 

Video

https://cld.wthms.co/11WDxO

-|-|-|-
iPhone|11Pro|13.1.2|4.10.1

jleandroperez:
### Fix
In this PR we're disabling the Markdown Preview gesture, whenever there's text selected in the editor.

Fixes #472

@danielebogo Sir, may I bug you **yet again** with an easy one?

Thanks in advance!!

### Test
1. Open a note.
2. Enable Markdown
3. Double tap on a word to copy it

- [x] Verify that expanding the selected range (in any direction) won't trigger the Markdown Preview UI.
- [x] Verify that dragging from the right edge towards the left edge, anywhere outside the selected range, triggers the Markdown Preview interaction.

**Sample video below!!**

### Release
`RELEASE-NOTES.txt` was updated in f1f9c26 with:

> Fixes a bug that caused Markdown Preview to be pushed when extending Text Selection

### Test Case Demo!

Demo

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,6 +6,7 @@
 -   Fixed a bug that caused the StatusBar to be hidden when Simplenote's Settings were displayed
 -   Fixed a bug that caused Swipe to Dismiss gesture not to properly work in the Editor
 -   Fixed a bug that allowed Peek and Pop to open notes in the Trash.
+-   Fixed a bug tht caused Text Selection in the editor to push the Markdown Preview UI.
 
 4.12
 ====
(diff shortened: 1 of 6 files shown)


dmsnell:
Does it make sense to limit this to when the touch event occurs inside a selection range? Would it be weird to flat-out disable the swipe when _any_ text is selected regardless of the touch point?

jleandroperez:
@dmsnell that's actually the case!

```
In this PR we're disabling the Markdown Preview gesture, whenever there's text selected in the editor.
```

We're checking the bounds of the Selected Text, and disabling the "Slide in Markdown" feature, whenever the touch event is in there.

Thanks for the suggestion!!

jleandroperez:
Thank you Daniele!!!

jleandroperez:
@loremattei Sir, would it be possible to get another TF build? (Incoming hotfix!).

**Thanks in advance!!!**

PR #525, opened 2019-11-14
Fix: Surrogate Pairs Crash!

### LINKED ISSUE #524: Crash: Emojis Again!
### Details:
1. Add a new note
2. Enter the following emojis: `☺️🖖🏿`
3. Insert the following emoji in between: `😃`

As a result the app will crash. We'll end up with a malformed unicode sequence, which effectively breaks Apple's JSON Parser.

[macOS Sibling Here](https://github.com/Automattic/simplenote-macos/issues/397)

jleandroperez:
### Fix
In this PR we're mapping a (patched) version of Simperium, which makes sure `Surrogate Pairs` aren't split in the middle, and thus, causing invalid unicode strings.

@aerych May I bug you with a very very cool fix?
Thank you sir!!

Closes #524

### Test
1. Add a new note
2. Enter the following emojis: ☺️🖖🏿
3. Insert the following emoji in between: 😃

- [x] Verify the app doesn't break!

### Release
`RELEASE-NOTES.txt` was updated in acd509e with:
 
> Fixed an issue that might crash the app with a specific emojis combination

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -29,7 +29,7 @@ abstract_target 'Automattic' do
 		#
 		pod 'Automattic-Tracks-iOS', '~> 0.4'
 #		pod 'Automattic-Tracks-iOS', :git => 'https://github.com/Automattic/Automattic-Tracks-iOS.git', :branch => 'add/support-for-tracking-crashes'
-		pod 'Simperium', '~> 0.8.23'
+		pod 'Simperium', '~> 0.8.24'
 		pod 'WordPress-AppbotX', :git => 'https://github.com/wordpress-mobile/appbotx.git', :tag => '1.0.7'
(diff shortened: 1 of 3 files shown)


dmsnell:
> I'm only copying the single emoji glyph

@aerych I'm not sure what you are copying or seeing but copy/paste can be quite tricky here. it's best to stick to the emoji selector panel for inserting because there can be similar bugs when copying emoji - that is, the display which renders the string may be changing the actual Unicode characters that get copied.

aerych:
> the display which renders the string may be changing the actual Unicode characters that get copied.

Good to know! Retried just using the emoji keyboard and there was no extra white space inserted.

jleandroperez:
Yay!! thank you both gentlemen!!

PR #534, opened 2019-11-20
Disabling iOS 13 Aggressive Links Workaround

### LINKED ISSUE #533: Links: Tap Detection might fail
### Details:
Side effect of #470: not performing a quick press might cause the link detection to fail, and thus, links might not result in SafariViewController being pushed onScreen.

jleandroperez:
### Fix:
We've received reports of users having trouble opening links.As far as I've been able to see, anything that's not a quick tap might not be properly handled.

Rather than reversing UITextView's default inner behavior (which probably involves long press and tap!), we're disabling the Aggressive Links workaround that was shipped via #470 for iOS 13.0 - 13.1.

@aerych you were right... as always. We should probably have never shipped that workaround =D

Closes #533

### Test
1. Log into your account
2. Open a note that contains links

- [x] Verify pressing a link ends up with SafariViewController onScreen

**Please:** Repeat for iOS 13.0 (or 13.1) and iOS 13.2.

### Release
`RELEASE-NOTES.txt` was updated in e020d48 with:
 
> Fixed a bug that affects tap detections over links.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 4.15
 -----
- 
+-   Fixed a bug that affects tap detections over links.
+
 4.14
 ====
 -   Fixed a bug that prevented users from repositioning the caret in the Email field (Login / SignUp)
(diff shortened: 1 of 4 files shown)


jleandroperez:
The mindset behind this quick patch is:

- Restore the behavior we've always had (prior to the iOS 13.0/13.1 workaround)
- Keep the workaround **just** for the OS releases that had the aggressive link behavior

**Eventually** I'd love to remove all of the conditionals, and deal with this the standard way!!

Thank you @aerych!!!

Issue #507, opened 2019-10-29
Advanced Search


jleandroperez:
### Task: Data Model
- [x] Simperium bucket for recent search results (last 5 searches)

### Task: ResultsController
- [x] Encapsulate NSFetchedResultsController
- [x] Implement ResultsController <> UIKit bridge

### Task: NoteListController: State.results
- [x] Interact only with Note entities, single section

### Task: NoteListController: State.search
- [x] Interact with Notes and Tags entities
- [x] Limit the number of Tags to 5 entries
- [x] Keywords containing `tag:xxx` should filter notes containing such tag

### Task: List Interface
- [x] Replace FRC with NoteListController
- [x] NoteTableViewCell: Update Pinned and Shared Indicators location
- [x] New UITableViewCell for Tag Entities
- [x] Pressing over any `Tag` entity prefills the SearchBar with `tag:xxx`
- [x] When Search is active: dismiss the keyboard as soon as the user starts scrolling
- [x] Row Separators
- [x] Dismiss keyboard whenever **Search** is pressed
- [x] Deleting the Search Keyword should display the notes that were onscreen pre-search
- [x] Updating Search Keywords should always scroll to the top of the table
- [x] SearchBar: Restore Autocapitalization
- [x] Search should **Respect Condensed Mode Settings**
- [x] Swipeable Actions + 3d Touch: Only available for Notes (in Results or Search Mode)
- [x] SearchBar should be hidden when visualizing the Trash
- [x] Remove custom transition from List <> Editor
- [x] Highlighted Search Match's BG
- [x] Empty States: 
  - [x] **No Results**
  - [x] **No Notes**

### Task: SortBar
- [x] Note List should display the date relevant to the selected Sort Mode, when appropriate.
- [x] Location always fixed at the bottom. Shouldn't react to keyboard events.
- [x] Pressing the Sort Bar should display an Alert Sheet with the available options.
- [x] Switching Sort Order should automatically set the default order for the given "sort type".
   ```
   Arrow pointing up is A-Z
   Arrow point down is Z-A
   Default order should be A-Z
   Arrow pointing up is Oldest to newest
   Arrow pointing down is Newest to oldest
   Default order should be Newest to oldest
   ```

### Bugs:
- [x] Search Mode's Navigation Bar doesn't remain hidden, whenever you dismiss the Editor #610
- [x] Note List TableViewCell Layout Glitch (see comments in #586)
- [x] Keyboard First Responder #600 
[...]

PR #536, opened 2019-11-22
Search: Wiring Results

jleandroperez:
### Details:
This PR is part of a multi-step implementation of the new Advanced Search mechanism. 

- **SPResultsController:** Encapsulates the Search Result's FRC.
- **SPNoteTableViewCell:** New a `cellHeight` static API
- **SPSearchResultsViewController:** now renders actual live results

I just... hope the names aren't confusing. Suggestions always welcome!!!

@aerych Eric!!! May I bug you with this one?
Thanks in advance!!

Ref. #507

## Known Limitations:
- SearchResults bottom padding isn't accounting for the keyboard. Addressed via #538 

### Test: Note Cells UX
0. Set the `FeatureManager.advancedSearchEnabled` property to `true`
1. Log into your account
2. Tap over the **Search** field

- [x] Verify **all of your notes** show up immediately
- [x] Verify the Search Results cell look great!

### Test: Search Results
0. Set the `FeatureManager.advancedSearchEnabled` property to `true`
1. Log into your account
2. Tap over the **Search** field
3. Enter any keyword

- [x] Verify the related notes show up
- [x] Verify the Sort Options match the Notes List sort options
- [x] Verify that pressing over any results cell will push the Editor UI

### Test: NavigationBar's Blur
0. Set the `FeatureManager.advancedSearchEnabled` property to `true`
1. Log into your account
2. Scroll down so that the NavigationBar Blur is activated
3. Tap over the **Search** field

- [x] Verify that the NavigationBar's Blur remains active, and there is no flicker / anything weird!

### Test: Swipe Back
0. Set the `FeatureManager.advancedSearchEnabled` property to `true`
1. Log into your account
2. Tap over the **Search** field
3. Open any of the Search Results

- [x] Verify that you can swipe back from the left edge of the screen

### Release

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -104,6 +104,7 @@
 - (void)didMoveToParentViewController:(UIViewController *)parent {
     [super didMoveToParentViewController:parent];
     [self ensureFirstRowIsVisible];
+    [self ensureTransitionControllerIsInitialized];
 }
 
 - (void)viewDidAppear:(BOOL)animated {
@@ -131,7 +132,22 @@
     }
 }
 
-- (void)updateRowHeight {
+
+- (void)ensureTransitionControllerIsInitialized
+{
+    if (_transitionController != nil) {
+        return;
+    }
+
+    NSAssert(self.tableView != nil, @"tableView should be initialized before this method runs");
+    NSAssert(self.navigationController != nil, @"we should be already living within a navigationController before this method can be called");
+
+    self.transitionController = [[SPTransitionController alloc] initWithTableView:self.tableView navigationController:self.navigationController];
+    self.transitionController.delegate = self;
+}
+
+- (void)updateRowHeight
+{
     self.tableView.rowHeight = SPNoteTableViewCell.cellHeight;
     [self.tableView reloadData];
 }
@@ -630,11 +646,7 @@
     [SPTracker trackListNoteOpened];
 
     SPNoteEditorViewController *editor = [[SPAppDelegate sharedDelegate] noteEditorViewController];
-    if (!_transitionController) {
-        self.transitionController = [[SPTransitionController alloc] initWithTableView:self.tableView navigationController:self.navigationController];
-        self.transitionController.delegate = self;
-    }
-        
+
     BOOL isVoiceOverRunning = UIAccessibilityIsVoiceOverRunning();
     self.navigationController.delegate = isVoiceOverRunning ? nil : self.transitionController;
     editor.transitioningDelegate = isVoiceOverRunning ? nil : self.transitionController;
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -426,6 +426,7 @@
 
     // TODO: Nuke the following as soon as we can!
     if (FeatureManager.advancedSearchEnabled) {
+        [self.resultsViewController reset];
         return;
     }
 
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -486,10 +486,8 @@
 - (void)configureCell:(SPNoteTableViewCell *)cell atIndexPath:(NSIndexPath *)indexPath {
     
     Note *note = [self.fetchedResultsController objectAtIndexPath:indexPath];
-    
-    if (!note.preview) {
-        [note createPreview];
-    }
+
+    [note ensurePreviewStringsAreAvailable];
 
     UIColor *accessoryColor = [UIColor simplenoteNoteStatusImageColor];
 
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -132,14 +132,7 @@
 }
 
 - (void)updateRowHeight {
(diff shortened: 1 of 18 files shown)


Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -340,13 +340,23 @@ - (BOOL)searchControllerShouldBeginSearch:(SPSearchController *)controller
     }
 
     bSearching = YES;
-    [self.tableView reloadData];
+
+    // TODO: Nuke the following as soon as we can!
```
dmsnell: I'm not opposed to `TODO` comments but I recently removed one that said "remove this when it's no longer necessary" and I just had to guess that since it was about six years old that it wasn't necessary.

can we include the specific conditions in the `TODO` that communicate when it will be time to adjust the code?

```objective-c
// TODO: Nuke as soon as advanced search is deployed and CoreData tables are getting recent searches
```

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -340,13 +340,23 @@ - (BOOL)searchControllerShouldBeginSearch:(SPSearchController *)controller
...
     
     return bSearching;
 }
 
 - (void)searchController:(SPSearchController *)controller updateSearchResults:(NSString *)keyword
 {
+    if (FeatureManager.advancedSearchEnabled) {
+        [self.resultsViewController updateSearchResultsWithKeyword:keyword];
+        return;
+    }
```
dmsnell: how does this approach compare to the notifications/listeners I've seen elsewhere in the codebase?

would it not be very convenient to add hooks for the listeners here and have one class that initializes behind the feature gate which then hooks into all those listeners?

Simplenote/Classes/SPNoteListViewController.m
jleandroperez: In this particular case, definitely no need to extend the comment further. Reason is... the second Avanced Search is ready, all of the `FeatureManager` references will be nuked (and these comments alongside!!). That is, this isn't an isolated patch, of part of a quite big set of changes.

That's hopefully next friday! Yay!!

Simplenote/Classes/SPNoteListViewController.m
jleandroperez: This pattern is [old school delegation pattern](https://developer.apple.com/documentation/swift/cocoa_design_patterns/using_delegates_to_customize_object_behavior). Standard all across Cocoa!

Notifications are usually used to **further** decouple a relationship between two entities. In this particular case, one `SearchController` will **always** have just one delegate, so it's not really necessary to switch over to Notifications.

Notifications are used for scenarios such as: Application became Active / Inactive, and "global" events. **Except** in macOS, where you've got Notifications for scroll events, for instance.

Now, that being said, Notifications are definitely less performant than a delegate call. Wouldn't make any significant mark here, but that's yet another thing to consider!

jleandroperez:
**Update:** Few bugs fixed, (officially) ready for a look. Thank you Eric!!

Simplenote/Classes/SPResultsController.swift
```diff
@@ -0,0 +1,150 @@
...
+            refreshPredicate(keyword: keyword)
+        }
+    }
+
+    /// Designated Initializer
+    ///
+    ///  - mainContext: Main Thread's MOC
+    ///
+    init(mainContext: NSManagedObjectContext) {
+        self.mainContext = mainContext
```
aerych: Maybe consider an assert to confirm the passed context has main queue concurrency?

Simplenote/Classes/SPResultsController.swift
```diff
@@ -0,0 +1,150 @@
...
+}
+
+
+// MARK: - Private Methods
+//
+private extension SPResultsController {
+
+    /// Refreshes the ResultsController's Predicate
+    ///
+    func refreshPredicate(keyword: String) {
```
aerych: Would `updatePredicateAndFetch` be slightly more descriptive?

Simplenote/Classes/SPResultsController.swift
```diff
@@ -0,0 +1,150 @@
+import Foundation
+import CoreData
+
+
+// MARK: - SPResultsController
+//
+class SPResultsController {
```
aerych: Maybe tweak the naming to be specific to search? `SPResultsController` sounds like it might be a generic wrapper around an FRC.  `SPSearchResultsController` could work but its similar to `SPSearchResultsViewController`. Maybe that's not so bad?

Simplenote/Classes/SPResultsController.swift
aerych: Incidentally, tho I'm a fan of the dependency injection here I'm wondering if its not more straight forward to just hard code `SPAppDelegate.shared().managedObjectContext` since that's what's ultimately used anyway. Is there a scenario where there are different main queue contexts in play, or a test scenario were a different context would be used?

jleandroperez: Thank you Eric!! I ended up doing as follows, so that it's flexible enough for DI, but you can't really break things up.

Eventually I'd love to write a million unit tests.

```
init(mainContext: NSManagedObjectContext = SPAppDelegate.shared().managedObjectContext) {
        assert(mainContext.concurrencyType == .mainQueueConcurrencyType)
```

jleandroperez:
All feedback addressed, thanks a lot @aerych!!!

Issue #481, opened 2019-10-21
Tech Debt: Master Issue


jleandroperez:
### Tasks
- [ ] Implement (new) Height Calculation mechanism. Top Spacing: 12 + 22 + 1 + 20 + 20 + 13
- [ ] NoteTableViewCell: Replace UITextView(s) with UILabels. Ref. #427
- [ ] Optimize **SPTransitionController**: Generate animationOptions just once. Improve code readability **please**
- [ ] When editing the tag name, the edition was terminated when space is pressed. This seems to be by design, but I found it a bit counter-intuitive. Maybe starting/creating a new tag, or just doing nothing on space key press would be more appropriate.

PR #539, opened 2019-11-27
Search: Dismissing Keyboard upon scroll

jleandroperez:
### Details:
In this PR we're making the SearchBar to resign its First Responder state, whenever you scroll over the Search Results.

Ref. #507 

@aerych Sorry about spamming you!!!

### Test: Dismiss upon Scroll
0. Set the `FeatureManager.advancedSearchEnabled` property to `true`
1. Log into your account
2. Tap over the Search Bar
3. Scroll over the Results

- [x] Verify the Keyboard is dismissed as soon as you begin dragging
- [x] Verify that **even** if the keyboard is hidden, the `Cancel` button remains enabled
- [x] Verify that pressing over the SearchBar triggers the keyboard again
- [x] Verify that pressing the `Cancel` button, while the Keyboard is hidden, effectively lets you exit Search Mode
- [x] Verify that you can get the SearchBar to become the first responder while the TableView remains inertia-scrolling (perform a quick swipe down / up!)

### Release


Simplenote/Classes/SPSearchController.swift
```diff
@@ -113,6 +138,18 @@ private extension SPSearchController {
 
         displayResultsViewController()
     }
+
+    func notifyStatusChanged(active: Bool) {
+        guard let delegate = delegate else {
+            return
```
dmsnell: when is this? when the view has been dismissed but before the notification event propagates?

jleandroperez: This would be... 

- Before it becomes the First Responder (that's how the "element with the focus" / "attached to the keyboard" is called)
- Or after it's been dismissed

Simplenote/Classes/SPSearchController.swift
```diff
@@ -190,6 +223,28 @@ extension SPSearchController: UISearchBarDelegate {
...
+class SPSearchBar: UISearchBar {
+
+    /// **Custom** Behavior:
+    /// Normally resigning FirstResponder status implies all of the button subviews (ie. cancel button) to become disabled. This implies that
+    /// hiding the keyboard makes it impossible to simply tap `Cancel` to exit **Search Mode**.
+    ///
+    /// With this (relatively safe) workaround, we're keeping any UIButton subview(s)  enabled, so that you can just exit Search Mode anytime.
+    ///
+    @discardableResult
+    override func resignFirstResponder() -> Bool {
```
aerych: I wish there was another way but I can't think of anything better than this. 
This would be a good candidate for a UI test to catch if the behavior changes in a future iOS release.

jleandroperez: Added a note in the main [Tech Debt issue](https://github.com/Automattic/simplenote-ios/issues/481). Thank you sir!!

jleandroperez:
@aerych **Thanks a lot** for reviewing this one!!!. Apologies It took me a while to finally merge... Support + AFK caught me!.

All feedback addressed. Thank you Eric!!

PR #543, opened 2019-11-28
Add configure to rake dependencies

loremattei:
This PR adds a step to `rake dependencies` in order to call `configure apply` that makes sure that the `SPCredentials.swift` used during the build is the one specified in `.configure`.

### Test
1. Delete your local `Simplenote/Credentials/SPCredentials.swif`.
2. Run `rake dependencies`.
3. Verify that `Simplenote/Credentials/SPCredentials.swif` has been created again with the right values. 
4. Build the app and verify that it works as expected.
 
### DIFF
--- a/.configure
+++ b/.configure
@@ -3,8 +3,8 @@
   "pinned_hash": "eed3c3f412a87e2804f4c0291dde7523e24877fa",
   "files_to_copy": [
     {
-      "file": "iOS/simplenote/config.plist",
-      "destination": "Simplenote/config.plist"
+      "file": "iOS/simplenote/SPCredentials.swift",
+      "destination": "Simplenote/Credentials/SPCredentials.swift"
     }
   ],
   "file_dependencies": [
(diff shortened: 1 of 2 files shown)


jleandroperez:
@loremattei question for you!!!

We've switched off from the `Simplenote/config.plist` mechanism (and replaced it with `Simplenote/Credentials/SPEnvironment.swift`).

**SPEnvironment.swift** is also in the mobile secrets (I didn't delete the former `config.plist` just for us to be able to... pull any source and rebuild it without further issues).

Is this `config.plist` required by our CI?

loremattei:
Ah! I see! No, it's not required by the CI. I think we missed to update the  `.configure` file when the credential mechanism was switched. 
I'm pushing an update to consider `SPEnvironment.swift` instead of `config.plist`.

loremattei:
@jleandroperez, sir, this should be ready for another look!
Thank you!

PR #544, opened 2019-11-29
Merge Release/4.14 into develop

loremattei:
Merge Release/4.14 into develop

These changes do not require release notes.

PR #545, opened 2019-11-29
Fixing Update-Translations Script

jleandroperez:
### Fix
In this PR we're replicating WPiOS's [Update Translations](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/Scripts/update-translations.rb) script.

Instead of downloading the JSON + Strings translation, we snap everything in a single shot (which is what we do in WordPress).

The current mechanism has a couple issues, and it's producing malformed strings / comments. [More here!](https://github.com/Automattic/simplenote-ios/pull/544).

Gentlemen, may I bug you with this PR?
cc @loremattei or @jkmassel

Thanks in advance!!

 
### Test
1. Run `./Scripts/update-translations.rb`
2. Verify Simplenote builds correctly
3. (OR) simply verify that the CI is happy

**Note:** Since I was in the neighborhood, I've also squeezed in the fixed `.strings` files!!

### Review

### DIFF
--- a/Scripts/update-translations.rb
+++ b/Scripts/update-translations.rb
@@ -92,28 +92,12 @@ langs.each do |code,local|
   # Backup the current file
   system "if [ -e #{lang_dir}/Localizable.strings ]; then cp #{lang_dir}/Localizable.strings #{lang_dir}/Localizable.strings.bak; fi"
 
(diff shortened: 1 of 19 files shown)


jleandroperez:
Woooo!! Thank you for checking this out @loremattei !!

PR #554, opened 2019-12-06
Feature/add match signing

loremattei:
This PR adds automatic profiles and certificate handling by Fastlane Match, as a first step to add installable builds on CI. 

### Test
1. Verify that the CI is green.
2. Run `bundle install`.
3. Run `bundle exec fastlane build_and_upload_internal` and verify that the build succeeds and the binary is uploaded.
4. Build on XCode and verify that the build succeeds.

### Release


loremattei:
Damn 🤦‍♂️! You're right!
Pushed it! I'm sorry for the mistake.

PR #560, opened 2019-12-18
Nukes Integrity Failsafe

jleandroperez:
### Details:
**Years, years ago**, in a distant galaxy...

We had (multiple) Sync'ing bugs, in which an entity's properties would not be properly persisted, but the Ghost would reflect the "canonical" / "right" version of the note's payload.

A workaround was designed, which would run **just once**, when several conditions were met (no local changes, discrepancy between Ghost and Content detected).

In this PR we're removing this failsafe, since it's been +5 years already, and the underlying bug has been fixed.

### Test
- [x] Verify the unit tests look green
- [x] Verify the app launches correctly, and your notes show up after logging in!

### Release


bummytime:
Ref: https://github.com/Simperium/simperium-ios/issues/359

jleandroperez:
Thank you Bummy!!

Issue #447, opened 2019-10-07
Dark mode printing "prints" white text


mchowning:
### Expected
Printing a note from a device with dark mode active should print dark text on a white background.

### Observed
Printing a note from a device with dark mode active "prints" white text, so there is no visible text. Confirmed that printing from the web app properly displayed a preview with black text.

![ios_print_invisible mov](https://user-images.githubusercontent.com/4656348/66328602-f025fc00-e8fa-11e9-8dd8-1d06fd88632e.gif)

### Reproduced

1. Set phone to dark mode
2. Open a non-empty note in the Simplenote app
3. Tap the circled i icon
4. Tap send on the bottom sheet
5. Tap print
6. Observe that a blank page is shown in the preview
7. Pinch to zoom on the blank page to open full screen
8. Observe that a blank page is shown but that you can select and copy (invisible) text from the page

Device | iOS Version | App Version
--- | --- | ---
| iPhone Xr |13.1 | 4.10.1
|iPad Pro 11 | 13.1.2 | 4.10.1

A user also reported experiencing this issue on an iPad in 2395781-zen. That user included videos of the issue, but I am not posting them publicly here in order to respect the user's privacy as the video includes portions of one of the user's notes.

PR #571, opened 2020-01-10
Fixes Print Share Printing White in Darkmode

charliescheer:
### Fix
This PR fixes [issue 447](https://github.com/Automattic/simplenote-ios/issues/447)
If the device running Simplenote is set to Dark Mode, and the user goes to the Note Settings > Share > Print, the text would be formatted in white, so it would appear that what was printed was blank.

Added a check when the SimpleTextPrintFormatter is initialized to see if the device is in Dark mode, and switch the text in the print formatter to black before printing.

### Test
1. Set device to Dark Mode
2. Open a note and go to the sharing options
3. Choose Print.  The preview should display a white background with black text.

### Review
- [ ] Check and make sure the preview text is appearing in black
- [ ] Make sure the text is still formatted correctly when device is in light mode
- [ ] Check sharing to other apps to confirm only the print text is being reformatted

### DIFF
--- a/Simplenote/SimplenoteSimpleTextPrintFormatter.swift
+++ b/Simplenote/SimplenoteSimpleTextPrintFormatter.swift
@@ -1,0 +1,21 @@
+//
+//  SimplenoteSimpleTextPrintFormatter.swift
+//  Simplenote
+//
+//  Created by Charlie Scheer on 1/10/20.
+//  Copyright © 2020 Automattic. All rights reserved.
+//
+
+import UIKit
+
+class SimplenoteSimpleTextPrintFormatter: UISimpleTextPrintFormatter {
+    override init(text: String) {
+        super.init(text: text)
+        
+        if #available(iOS 12.0, *) {
+            if UIScreen.main.traitCollection.userInterfaceStyle == .dark {
+                self.color = UIColor.black
+            }
+        }
+    }
+}
--- a/Simplenote/SimplenoteSimpleTextPrintFormatter.swift
+++ b/Simplenote/SimplenoteSimpleTextPrintFormatter.swift
@@ -13,7 +13,7 @@ class SimplenoteSimpleTextPrintFormatter: UISimpleTextPrintFormatter {
         super.init(text: text)
         
         if #available(iOS 12.0, *) {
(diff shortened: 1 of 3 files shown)


Simplenote/SimplenoteSimpleTextPrintFormatter.swift
```diff
@@ -0,0 +1,21 @@
...
+//  SimplenoteSimpleTextPrintFormatter.swift
+//  Simplenote
+//
+//  Created by Charlie Scheer on 1/10/20.
+//  Copyright © 2020 Automattic. All rights reserved.
+//
+
+import UIKit
+
+class SimplenoteSimpleTextPrintFormatter: UISimpleTextPrintFormatter {
```
jleandroperez: Can we rename **SimplenoteSimpleTextPrintFormatter** > **SPSimpleTextPrintFormatter**  (OR) **SimpleTextPrintFormatter**?

ObjC convention is usually the app's initials + name, for namespaces purposes. Suppose a dependency used **SimpleTextPrintFormatter**, if you had a class with the exact same name, you'd get a compiler error.

In Swift that's not really needed anymore, each framework is namespaced by default (UIKit.UISimpleTextPrintFormatter)

That being said, since Simplenote is in a mixed state (ObjC / Swift), we're keeping around the SP prefix for a little longer (OR droppping it entirely, depending on how cool one looks, or the other, to be honest!).

Simplenote/SimplenoteSimpleTextPrintFormatter.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  SimplenoteSimpleTextPrintFormatter.swift
+//  Simplenote
+//
+//  Created by Charlie Scheer on 1/10/20.
+//  Copyright © 2020 Automattic. All rights reserved.
+//
```
jleandroperez: Would it be possible to drop the copyright? we still need to remove that from several classes, but trying to keep new code without it.

(We do the same in WPiOS and Woo)

charliescheer:
Changes made!  Thanks for taking a look!

jleandroperez:
Thanks for fixing this Charlie!!

PR #579, opened 2020-01-15
NSPredicate+Simplenote: Updates API(s)

jleandroperez:
### Details:
Since now we'll filter Tags by name, in this PR we're updating NSPredicate extension(s) API, so that it's crystal clear when a method returns a NSPredicate for Notes (or else).

Plus we're adding a new extension method to filter Tags by name.

Ref. #507

### Test
- [ ] Verify the unit tests are green!

### Release


jleandroperez:
Thaaank you!!!

PR #580, opened 2020-01-15
Implements NotesListController: Part I

jleandroperez:
### Details:
In this PR we're implementing **NotesListController**, in charge of encapsulating the NotesListViewController's datasource FSM.

With an API similar to **ResultsController**, it'll be in charge of wrapping access to the following entities:

- **Tags**
- **Notes**
- **Search History**

Initial FSM implementation consists of two states:

- **.results**: We're expected to render _just_ Notes
- **.searching(keyword: String)**: Expected to render Notes + Tags matching such keyword

In a follow up PR we'll add the **.history** state, which will allow us to deal with History entries.

Ref. #507

### Test:
Nothing has been wired yet. **But** we're making a really strong case for Unit Testing.... soooo please:

- [x] Verify the Unit Tests are green

Thank you!!

### Release


Simplenote/Classes/NotesListController.swift
```diff
@@ -0,0 +1,216 @@
...
+
+    /// Main Context
+    ///
+    private let viewContext: NSManagedObjectContext
+
+    /// Notes Controller
+    ///
+    private lazy var notesController: ResultsController<Note> = {
+        ResultsController<Note>(viewContext: viewContext, matching: state.predicateForNotes(filter: filter), sortedBy: sortMode.descriptorsForNotes)
+    }()
```
dmsnell: what's going on with the function wrapper? this is lazy by nature of the keyword, right, not by nature of the `{}()`? in fact, does this do anything?

jleandroperez: It's a lazily initialized property. Nil until the first time it's accessed, _then_ it'll contain a ResultsController for the Note entity.

It reads as `we want an ivar named notesController, of the ResultsController kind. Please initialize it with the result of this closure`.

Another detail is: in a closure / method, if there's a single line, the language assumes it's the return value. No need to write `return` anymore. 

🔥

dmsnell: sure, but wouldn't this be the same as the following code?

```swift
private lazy var notesController: ResultsController<Note> = \
    ResultsController<Note>(viewContext: viewContext, matching: state.predicateForNotes(filter: filter), sortedBy: sortMode.descriptorsForNotes)
```

jleandroperez: Let's settle on the following, since it's more swift-y!

```
    private lazy var tagsController = ResultsController<Tag>(viewContext: viewContext,
                                                             matching: state.predicateForTags(),
                                                             sortedBy: sortMode.descriptorsForTags)
```

Adding semicolons in a new line feels .. way too JS!

Thank you!

jleandroperez:
@bummytime @dmsnell thanks a lot for the super detailed review!!

🔥

PR #584, opened 2020-01-17
Implements NotesListController: Part II

jleandroperez:
### Details:
In this PR we're enhancing **NotesListController** by adding a callback, to be known as **onBatchChanges**.

Such closure is to be executed whenever the entities onScreen (either Tags or Notes) get updated. The trick here is: we're mixing results from **two** *NSFetchedResultsController(s)*, and rendering them in a single UITableView.

Because of this, we're adjusting IndexPaths, accordingly:

- Results Mode: *single* section, **only** Notes onscreen
- Search Mode: **two** sections. Section zero displays Tags, section One displays Notes

Ref. #507 

### Test
The changes in this PR haven't been wired just yet. Please:

- [x] Verify all of the new Unit Tests are green
- [x] Verify that the code makes sense!

### Release


jleandroperez:
Thank you Bummy!!

PR #589, opened 2020-01-22
Notes List: Tags UI

jleandroperez:
### Details:
This PR is all about Tags UI:

- New UITableViewCell subclass for Tag Results!
- And we're also getting a new SectionHeader, to be displayed when we're in Search Mode

cc @aerych (Thanks in advance Eric!!!)
Ref. #507

### Test: Regular Results!
1. Log into your account
2. Go thru your tags!

- [x] Verify no Section Header is displayed
- [x] Verify all the things look as expected!

### Test: Search Mode!
1. Log into your account
2. Press over the Search Bar
3. Enter a keyword that yields Tags + Notes

- [x] Verify that Tags look correctly
- [x] Verify that Tag rows are smaller than Note rows!
- [x] Verify that the Tag Section has a cool header that reads `Search by Tag:`
- [x] Verify that the Note Section has a header that reads `Notes`

### Test: Zero Tag Results
1. Press over the Search Bar
2. Enter a result that yields zero tags (but some notes!)

- [x] Verify the Tags Section is completely hidden
- [x] Verify that pressing over any of the Notes pushes the Editor

### Test: Empty State
1. Press over the Search Bar
2. Enter a result that yields zero results (no Tags nor Notes)

- [x] Verify that the legend **No results** shows up onScreen

### Release
`RELEASE-NOTES.txt` was updated in 0ecea89 with:

> Search results now display matching tags!

---


Simplenote/Classes/UITableView+Simplenote.swift
```diff
@@ -39,4 +39,14 @@ extension UITableView {
 
         return cell
     }
+
+    /// Returns a Header instance of the specified kind
+    ///
+    func dequeueReusableHeaderFooterView<T: UITableViewHeaderFooterView>(ofType type: T.Type) -> T {
+        guard let header = dequeueReusableHeaderFooterView(withIdentifier: T.reuseIdentifier) as? T else {
+            fatalError()
```
aerych: 😬 Understanding that this shouldn't be reached in production, does it need to be a fatalError?  An Assert would serve the same purpose in development and be production safe. Thoughts?

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -292,9 +328,8 @@ private extension SPNoteListViewController {
     /// Returns a UITableViewCell configured to display the specified Tag
     ///
     func dequeueAndConfigureCell(for tag: Tag, in tableView: UITableView, at indexPath: IndexPath) -> UITableViewCell {
-        // TODO: Wire a proper UITableViewCell for Tags. To be followed up in another PR!
-        let cell = UITableViewCell()
-        cell.textLabel?.text = "tag:" + tag.name
+        let cell = tableView.dequeueReusableCell(ofType: SPTagTableViewCell.self, for: indexPath)
+        cell.titleText = "tag:" + tag.name
```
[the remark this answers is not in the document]
jleandroperez: I've localized that in a follow up PR [here](https://github.com/Automattic/simplenote-ios/pull/591/files#diff-dd812dd2f4747acc9565cbb712c04671R341) (where the tag: operator is actually parsed).

Thanks for the heads up Eric!!!

Simplenote/Classes/UITableView+Simplenote.swift
jleandroperez: The huge thing there is: what do we return, if the target HeaderFooter cannot be instantiated, right?

Do we instantiate a new `T`? (what if such subclass is expecting `awakeFromNib` to fully initialize?). Do we return a base class instance? (that would probably crash down the road).

Or do we convert the return type into an Optional, and guard let elsewhere?(and thus, eventually rendering an empty view).

Of all the alternatives, crashing if that condition is met (cannot dequeue) kiiinda felt like the most reasonable one, since we'd catch it 100% on Sentry, and it's something that will either never happen, or happen 10/10.

WDYT? (Thanks for opening the conversation, LOVE these topics!!!)

Simplenote/Classes/SPTagTableViewCell.swift
```diff
@@ -0,0 +1,90 @@
...
+// MARK: - SPTagTableViewCell
+//
+extension SPTagTableViewCell {
+
+    /// Returns the Height that the receiver would require to be rendered
+    ///
+    /// Note: Why these calculations? why not Autosizing cells?. Well... Performance.
+    ///
+    static var cellHeight: CGFloat {
+        let verticalPadding: CGFloat = 12
```
aerych: I suppose there isn't much use in declaring a constant for this value, eh?

Simplenote/Classes/UIFont+Simplenote.swift
```diff
@@ -0,0 +1,17 @@
...
+import UIKit
+
+
+// MARK: - UIFont Simplenote Helpers
+//
+extension UIFont {
+
+    /// Returns the System Font for a given Style and Weight
+    ///
+    static func preferredFont(for style: TextStyle, weight: Weight) -> UIFont {
```
aerych: So this will be called once for every tag cell, and once for every header view, anytime style is refreshed, or when searching. Does that sound right? 
Not sure how expensive a call it is to redo this work over and over.  Only 5 tag cells so maybe not very. Would it be worth caching the response for a particular style and weight?

Simplenote/Classes/SPTagTableViewCell.swift
jleandroperez: Good call! Moving over to the `Style` enum!

Simplenote/Classes/UIFont+Simplenote.swift
jleandroperez: I **love** this idea. Working on it!! Thank you!

jleandroperez:
- **UIFont.preferredFont** cache is in place (NICE!!)
-  Properly defined the cellHeight constant
- Patched up the issue description (`Search by Tags:` > `Search by Tag:`): **You're the best Eric!!!**
- As per the empty state (you're right sir, `No Results` is the correct copy), we'll deal with those in yet another PR, just to keep things a bit atomic.
- The Search Results will also get a gray overlay , as soon as you enter Search Mode. Non implemented yet (that'll cover the current results)
- I'll patch up the scenario in which "keyword has been deleted" elsewhere as well. Taking notes in the master issue, **thank you!!!**

Aaaaand last but not least, I've checked closely the documentation for `dequeueReusableHeaderFooterView`, and I you're most definitely right.  It's not just the downcasting that might fail, the method itself might also return nil.

That is... unlike [this other API](https://developer.apple.com/documentation/uikit/uitableview/1614878-dequeuereusablecell), which is what I had in mind. So you're definitely right on spot, I've removed the `fatalError`.

Turns out, the caller allowed for nil SectionHeader, so worst case scenario, it'll be empty.

Thank you for your time and help @aerych !!!!!!

Simplenote/Classes/UIFont+Simplenote.swift
```diff
@@ -9,9 +9,46 @@ extension UIFont {
...
+
+    /// Internal Cache
+    ///
+    private static var cache = [UIFont.TextStyle: [UIFont.Weight: UIFont]]()
+
+    /// Returns the stored entry for the specified Style + Weight combination (If Any!)
+    /// - Note: This method is, definitely, non threadsafe!
+    ///
+    static func cachedFont(for style: UIFont.TextStyle, weight: UIFont.Weight) -> UIFont? {
+        assert(Thread.isMainThread)
```
aerych: ❤️

jleandroperez:
Thank you Eric!!

PR #591, opened 2020-01-23
Search: Tag Operator

### LINKED ISSUE #57: Search does not search tags
Search searches note body but not tags.

Steps to repro:
1.) Create a note called Note One, add the tag "Test Tag"
2.) Create a note called Note Two, add the text "Test Tag" to the note
3.) Type "Test Tag" into the search bar - only Note Two is returned on iOS and the Web App

Is this by design?

@jleandroperez 
"Tom / Mike designed this, ages ago, but i'd go ahead and say that yes, this is by design. And makes sense, it would be very weird if you were filtering by tag, and notes from another tag were to show up.

Except for OS X, in which case the tag list is always visible (unless you've collapsed it).

CC'ing @drw158 on this one. Sorry about the delay @Essayoh!!"

@beaucollins 
"In Android I have a "hidden" feature that would allow you to search tags by using a gmail-like syntax:

> tag:work blah

Would only search posts within the tag work.

If you had already selected a tag (e.g. "blog") and you did that same search, then it would only return posts tagged both "work" and "blog" matching the rest of the query."

@drw158 
"I believe the web app search functions similarly to the Android app also. You can type `tag:` and filter notes that way.

I can't speak for Mike and Fred, but yes I think this is by design. @Essayoh, in your original steps, you would just select the tag in the tag sidebar to see those notes, you wouldn't type the tag in the search bar, which is pretty slow on mobile devices.

I could see the desire to type a tag in the search bar to quickly see those notes on Desktop. We can keep that in mind for the web, Mac, Linux, and Windows apps."

Opened by @Essayoh

jleandroperez:
### Details:
This PR is all about supporting the **tag:** operator:

- Whenever the Search Text contains the `tag:keyword`, we'll filter out notes with tags whose names contain `keyword`
- Pressing Tag Results now updates the Search Text. We're always replacing the last word, with whatever **tag** the user entered
- And we definitely added **a ton** of new Unit Tests added

@bummytime incoming FUN pr!!!

Ref. #507 
Closes #57

### Scenario: Tag Filtering
1. Log into your account
2. Make sure you've got at least 3 tags
3. Press over the Search Bar
4. Enter `something else [VALID_TAG_SUBSTRING]` (Where only the last word is a partial match of a valid tag name)

- [x] Verify that Tags whose names match the last word do show up onScreen
- [x] Repeat the same test, but in (4) enter `one two tag:[VALID_TAG_SUBSTRING]`

### Scenario: Excluding Exact Matches
1. Log into your account
2. Make sure you've got at least one tag
3. Press over the Search Bar
4. Enter `tag:VALID_TAG_NAME` (with **VALID_TAG_NAME** being a full tag name)

- [x] Verify that such tag (VALID_TAG_NAME_ is excluded from the results

### Scenario: Pressing Tag Results (No Spaces)
1. Log into your account
2. Make sure you've got at least one tag
3. Press over the Search Bar
4. Enter a keyword that yields multiple Tag Results
5. Make sure there is no space after the keyword

- [x] Verify that pressing over any of the Tag Results replaces the `keyword` with the `tag:xxx` you've pressed in (5)
- [x] Verify that the Search Results are updated accordingly!

### Scenario: Pressing Tag Results (Space!)
1. Log into your account
2. Make sure you've got at least one tag
3. Press over the Search Bar
4. Enter a keyword that yields multiple Tag Results
5. Make sure to leave a space right after the keyword

- [x] Verify that pressing over any of the Tag Results simply gets the `tag:xxx` appended to the search text
- [x] Verify that the Search Results are updated accordingly!

### Release
`RELEASE-NOTES.txt` was updated in 73f9924 with:

> We're now supporting the `tag:keyword` search operator. Simplenote will yield Notes whose tag names contain such *keyword*.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 4.16
 -----
 -   Search results now display matching tags!
+-   We're now supporting the `tag:keyword` search operator. Simplenote will yield Notes whose tag names contain such *keyword*.
 
 4.15
 -----
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -144,6 +144,7 @@
 		B518899E1E0D5EF800E71B83 /* SPContactsManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = B518899D1E0D5EF800E71B83 /* SPContactsManager.swift */; };
 		B51889A01E0D5F2200E71B83 /* Contacts.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B518899F1E0D5F2200E71B83 /* Contacts.framework */; };
 		B51889A41E0DB01E00E71B83 /* ContactsUI.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = B51889A31E0DB01E00E71B83 /* ContactsUI.framework */; };
+		B51D7C7923D9CFA4005D4B77 /* StringSimplenoteTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B51D7C7823D9CFA4005D4B77 /* StringSimplenoteTests.swift */; };
 		B521A1961BC8446900E1CF2A /* SPAutomatticTracker.m in Sources */ = {isa = PBXBuildFile; fileRef = B521A1941BC8446900E1CF2A /* SPAutomatticTracker.m */; };
 		B524AE0F2352BE9200EA11D4 /* UITableView+Simplenote.swift in Sources */ = {isa = PBXBuildFile; fileRef = B524AE0E2352BE9200EA11D4 /* UITableView+Simplenote.swift */; };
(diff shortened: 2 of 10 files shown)


jleandroperez:
Thank you @bummytime !!!

- Autocapitalization is gone
- Spaces in between keywords are now ignored
- And there's a new unit test for the above

Whenever you have a few, miiind taking a quick peek? Thank you sir!!!

jleandroperez:
Thank you sir!!

SylvesterWilmott:
Just seeing this now.

>Thoughts on setting autocapitalizationType = .none in UISearchBar+Simplenote.swift? I don't think we should be auto-capitalizing search terms here.

I don't see a reason to turn off capitalization unless I'm missing something. All of the iOS searches are capitalized.

>This might be more of a question for @SylvesterWilmott — if a user enters two spaces after a search term, is it reasonable for this UX:

The user should be able to put as many spaces as they choose. Again, all iOS searches work in this way.

jleandroperez:
Added a note on the master issue, we'll restore autocapitalization in another PR.

Thanks for the feedback Sly!!

PR #597, opened 2020-01-27
Disabling custom transitions: Part 1

jleandroperez:
### Details
In this PR we're disabling Simplenote's custom transitions from List <> Editor.

Ref. #507 

cc @bummytime 
Thank you sir!!

### Test
1. Launch Simplenote
2. Press over any note 
3. Verify the Editor is presented with a Push transition
4. Verify that the Editor is dismissed when you press the top left **Back** button
5. Verify that the Editor is dismissed via Swipe to Dismiss gesture

### Known Issues:
- As a side effect of using the Standard Transition, the First Responder appears to be getting restored (when reopening the Editor). Issue #600
- The above appears to be an UIKit issue. Reproduced in a sample project, with no code (!!!)
- UX Glitches related to "First Responder Restoration" addressed via #607

### Release


jleandroperez:
Thanks for the review Bummy!!

PR #601, opened 2020-01-27
SearchBar: Hidden when visualizing Trash

jleandroperez:
### Details:
In this PR we're implementing a mechanism that hides the SearchBar when the user is visualizing the Trash 

Ref. #507 

cc @bummytime 
cc @SylvesterWilmott for sanity check! (Screenshot below!)

Thanks gentlemen!

### Test
1. Launch Simplenote
2. Trash some note
3. Press over the top left icon to reveal the Tags List
4. Press over Trash

- [x] Verify the SearchBar is no longer visible
- [x] Verify that scrolling the trashed note out of the visible area triggers the NavBar's Blur
- [x] Verify that opening **All Notes** again does show the Notes List (and that the first row is onScreen)

### Release
`RELEASE-NOTES.txt` was updated in c8b7526 with:

> The SearchBar will no longer be displayed when the Trash Filter is active.

---

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 4.16
 -----
- 
+-   The SearchBar will no longer be displayed when the Trash Filter is active.
+
 4.15
 -----
 -   Simplenote just got brand new App Icons!
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -86,6 +86,17 @@ extension SPNoteListViewController {
         tableView.contentInset.top = searchBarStackView.frame.height
         tableView.scrollIndicatorInsets.top = searchBarStackView.frame.height
     }
+
+    /// Scrolls to the First Row whenever the flag `mustScrollToFirstRow` was set to true
+    ///
+    @objc
+    func ensureFirstRowIsVisibleIfNeeded() {
+        guard mustScrollToFirstRow else {
+            return
+        }
+
+        ensureFirstRowIsVisible()
+        mustScrollToFirstRow = false
     }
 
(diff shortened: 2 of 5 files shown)


dmsnell:
do we want to prevent searching trashed noes?

SylvesterWilmott:
Looks good @jleandroperez 

@dmsnell with the addition of searching by tag, search has become global, so it doesn't matter where you are in the app, the search will always look at all your notes unless you specify a tag. Since we do omit trashed notes from the search results, there seems to be no reason to include our advanced search in this screen.

We may want to include a search specifically for the trash or an operator to search within trash later on if we find there is a demand for it.

dmsnell:
> search has become global

I haven't been following too closely so I assumed it worked as the web app worked where you still have the tag drawer. I guess this means that if we make parity we'll lose the ability to search through the trash until we add it back in.

Do you consider it bad to be able to limit the search via the tag drawer as the apps have worked for so long, to select a tag and then have the search bar limit the results within that category? I know it has some quirks, but I'm worried about removing a feature and not replacing it, especially when it's a reasonable feature (where was that note I accidentally deleted?)

SylvesterWilmott:
>Do you consider it bad to be able to limit the search via the tag drawer as the apps have worked for so long, to select a tag and then have the search bar limit the results within that category?

It's a little more subtle than either being good or bad, surfacing the `tag:` operator so prominently in search makes it tough to co-exist with search limiting. By adding a `tag:x` to your search while having a tag selected in the sidebar, you will effectively narrow your search by two tags without having a consistent way to see that was the case.

>I know it has some quirks, but I'm worried about removing a feature and not replacing it, especially when it's a reasonable feature (where was that note I accidentally deleted?)

I hear you and I do agree! I don't want to jump the gun on a solution before we understand how advanced search is being used.

jleandroperez:
Thanks for the review @bummytime !!! And thank you for stepping by (everyone!!) 

🔥

PR #607, opened 2020-01-31
Not restoring First Responder

### LINKED ISSUE #600: Editor: Don't restore First Responder
### Details:
0. Run a branch that contains #597 
1. Launch Simplenote
2. Open the editor
3. Press anywhere to trigger the First Responder
4. Swipe to dismiss
5. Re-open the Editor

As a result, the First Responder will be restored. Note that there'll also be an (unwanted) transparency glitch.

jleandroperez:
### Fix
In this PR we're neutralizing the UIKit's mechanism in charge of restoring the First Responder (whenever a ViewController is reused), since it's the known cause of a quite bad UI bug.

For the record, I've pulled together a [Sample Project demonstrating the First Responder bug](https://github.com/jleandroperez/ResponderRestorationKeyboardBug).

[Sample video of the glitch available here](https://github.com/jleandroperez/ResponderRestorationKeyboardBug/blob/master/Demo.mp4)

Closes #600

### Note:
Since we were in the neighborhood, an ancient UITextView workaround has been removed

Ref. **Legacy Repository** #248

### Scenario: First Responder Restoration
1. Log into your account
2. Make sure you've got at least one empty note, and one non-empty note

- [x] Verify that opening an Empty Note gets the keyboard onScreen without any UX glitch
- [x] Verify that opening a non empty note **does not** get the keyboard onScreen
- [x] Verify that force pressing over an Empty Note also gets the keyboard onScreen, without issues
- [x] Verify that force pressing over a non empty note **does not** get the keyboard onscreen
- [x] Verify that adding a new note from the List gets the editor onScreen, and opens the Keyboard
- [x] Verify that adding a new note from within the editor also gets the keyboard onScreen

### Scenario: Workaround Removal
1. Launch Simplenote
2. Enable the Dark Mode
3. Open a note that has links
4. Enter Edit mode

- [x] Verify that Links are readable with the keyboard dismissed (and get rendered in blue-ish)
- [x] Verify that Links are (also) readable when pressing the TextView (should turn the same color as the rest of the document!)

### Release

### DIFF
--- a/Simplenote/Classes/SPEditorTextView.h
+++ b/Simplenote/Classes/SPEditorTextView.h
@@ -18,7 +18,6 @@ extern NSString *const CheckListRegExPattern;
 
 @interface SPEditorTextView : SPTextView
 
-@property (nonatomic) BOOL editing;
 @property (nonatomic) BOOL lockContentOffset;
 @property (nonatomic) BOOL overideLockContentOffset;
 
--- a/Simplenote/Classes/SPEditorTextView.m
+++ b/Simplenote/Classes/SPEditorTextView.m
@@ -86,8 +86,11 @@ NSInteger const ChecklistCursorAdjustment = 2;
         tapGestureRecognizer.cancelsTouchesInView = NO;
         
         [self addGestureRecognizer:tapGestureRecognizer];
+
+        // Why: Data Detectors simply don't work if `isEditable = YES`
         [self setEditable:NO];
     }
+
     return self;
 }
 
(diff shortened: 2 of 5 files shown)


jleandroperez:
Thank you @bummytime !!! (And thank you for testing this everywhere!!)

New Note animation (from within the Editor) isn't slated for updates, for now. CC'ing @SylvesterWilmott since he always knows which way!

🔥

SylvesterWilmott:
>New Note animation (from within the Editor) isn't slated for updates, for now. CC'ing @SylvesterWilmott since he always knows which way!

Confirming the transition of new note from within the editor is as intended. 

However, I did notice a keyboard bug when tapping new note from the notes list in dark mode the keyboard appears black momentarily. This should not occur as the keyboard should only be displayed after the push transition is complete.

jleandroperez:
Thank you Sly!

PR #610, opened 2020-02-03
Preserving NavigationBar Hidden state

jleandroperez:
### Fix
In this PR we're fixing a glitch in which the Navigation Bar's state wouldn't be properly preserved when going from `List > Editor > List`, whenever the Search was active.

Ref. #507 

### Test
1. Launch Simplenote
2. Press over the Search Bar
3. Verify the Navigation Bar is hidden
4. Press over any row to open the Editor
5. Verify the Navigation Bar is hidden
6. Hide (or swipe) back
7. Verify the Navigation Bar remains hidden in the Notes List

### Release


jleandroperez:
Thank you for the review Bummy!!

PR #619, opened 2020-02-04
Exiting SearchMode when pressing `Done`

### LINKED ISSUE #618: Not exiting Search Mode when pressing Done
### Steps
1. Launch the app
2. Press over the Search Bar
3. Type any keyword that yields at least one note
4. Press over any of the note rows
5. In the editor, press **Done**
6. Go back to the Notes List

As a result, the Search Bar will still have the Keyword, and will still be active.

---

### Demo

Bugo

jleandroperez:
### Details:
In this PR we're fixing an issue in which the SearchMode wouldn't properly be dismissed, whenever the user presses over the **Done** button in the Editor.

Closes #618

### Test
1. Launch the app
2. Press over the Search Bar
3. Type any keyword that yields at least one note
4. Press over any of the note rows
5. In the editor, press Done
6. Go back to the Notes List

- [x] Verify the SearchBar is not the first responder
- [x] Verify there's no SearchText onScreen

### Release
`RELEASE-NOTES.txt` was updated in 4c1687d with:

> Fixed a bug in which the Search Mode wouldn't properly be dismissed from the Editor


jleandroperez:
Thank you sir!

PR #625, opened 2020-02-05
Restoring results when the Search Keyword is Empty

jleandroperez:
### Details:
In this PR we're fixing a few UX inconsistencies:

1. Whenever the Search Keyword is **empty**, we're now displaying, back again, the list of notes that was onScreen prior to entering the Search Mode.
2. I've spotted a **double FRC fetch** that was being performed when the Keyword was updated (potentially legacy code), and removed that.
3. **PLUS:** UITableView's Height Estimation APIs have been implemented, in order to prevent the TableView from jumping all over the place.

Ref. #507

### Test: Empty Keyword
1. Launch Simplenote
2. Press over the Search Bar
3. Enter any keyword
4. Delete such keyword

 - [x] Verify that the List onScreen is exactly what you had in (1) or (2)

### Test: Invalid Offset
1. Launch Simplenote
2. Scroll downwards (anywhere!)
3. Press over the Search Bar
4. Enter a partial keyword that yields results
5. Verify that the TableView was scrolled to the very first row
6. Scroll anywhere!
7. Continue typing the keyword, so that the Search Results are updated
8. Verify that the TableView is scrolled to the top, the second the Search Results are updated
9. Press over the **Cancel** button, by the search bar
10. Verify the TableView was scrolled to the top

### Release

### DIFF
--- a/Simplenote/Classes/NotesListController.swift
+++ b/Simplenote/Classes/NotesListController.swift
@@ -193,8 +193,15 @@ extension NotesListController {
 
     /// Refreshes the FetchedObjects so that they match a given Keyword
     ///
+    /// -   Note: Whenever the Keyword is actually empty, we'll fallback to regular results. Capisci?
+    ///
     @objc
     func refreshSearchResults(keyword: String) {
+        guard !keyword.isEmpty else {
+            state = .results
+            return
+        }
+
         state = .searching(keyword: keyword)
     }
 
--- a/Simplenote/Classes/NotesListController.swift
+++ b/Simplenote/Classes/NotesListController.swift
@@ -188,7 +188,7 @@ extension NotesListController {
     ///
     @objc
     func beginSearch() {
-// TODO: we should actually switch to `state = .history`
+        // NO-OP: Initially meant for History, keeping it around for both consistency and future purposes.
     }
 
     /// Refreshes the FetchedObjects so that they match a given Keyword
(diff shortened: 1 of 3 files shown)


Simplenote/Classes/NotesListController.swift
```diff
@@ -188,13 +188,20 @@ extension NotesListController {
     ///
     @objc
     func beginSearch() {
-// TODO: we should actually switch to `state = .history`
+        // NO-OP: Initially meant for History, keeping it around for both consistency and future purposes.
```
bummytime: 👍

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -306,6 +306,15 @@ extension SPNoteListViewController: UITableViewDataSource {
 //
 extension SPNoteListViewController: UITableViewDelegate {
 
+    public func tableView(_ tableView: UITableView, estimatedHeightForRowAt indexPath: IndexPath) -> CGFloat {
+        // Notes:
+        //  1.  No need to estimate. We precalculate the Height elsewhere, and we can return the *Actual* value
+        //  2.  We always scroll to the first row whenever Search Results are updated. If we don't implement this method,
+        //      UITableView ends up jumping off elsewhere!
+        //
+        return self.tableView(tableView, heightForRowAt: indexPath)
```
bummytime: Thank you for the descriptive comment! ❤️

jleandroperez:
Thank you Bummy!!

PR #630, opened 2020-02-07
Search Highlight Background

jleandroperez:
### Details
In this PR we're adding a background to the Search Keyword Highlight within the Notes List.

cc @SylvesterWilmott this woooould be:

**Foreground:** `light: .blue50` / `dark: .blue30`
**Background**: `light: .blue0`  / `dark: .darkGray8`

cc'ing @bummytime for the code review 

Thanks in advance!!

Ref. #507

### Test
1. Launch the app
2. Press over the Search Bar
3. Type a keyword that yields at least one result

- [x] Verify the matching keywords are highlighted with a FG / BG
- [x] Verify this also works super in dark mode!

### Release
These changes do not require release notes.

PR #634, opened 2020-02-10
Revert "Search Highlight Background"

jleandroperez:
### Details:

We're reverting PR #630 based on @SylvesterWilmott 's feedback.

**Plus:** Sly also spotted a bug in which, when in Search Mode, the Note Cells wouldn't properly render. Such issue was also tracked down to NSAttributedString's `.background` property usage (causing layout issues).

cc @bummytime (Thank you sir!)

### Testing:
1. Launch the app
2. Press over the Search Bar
3. Type a keyword that yields at least one result

- [x] Verify the matching keywords are highlighted without Background Highlight


jleandroperez:
Thank you sir!

PR #640, opened 2020-02-12
Implementing SortBar: Mark I

jleandroperez:
### Description:
In this PR we're implementing the SortBar's UI: A new component, visible only in Search Mode, which will allow users to quickly change the Search Results's Sort Mode.

This component will show up at the bottom of the screen, and will be visible whenever:

A. The user scrolls down (and thus, the keyboard is dismissed)
B. The **Search** button is pressed

**Note:** Logic will be added in a follow up PR, Mark 1 is all about the UI!

cc @bummytime 
cc @SylvesterWilmott 

🔥🔥🔥 

Ref. #507

### Details:
- New **SPSortBar** component
- We'll dismiss the Keyboard on scroll **only** when the Search Keyword is not empty

### Scenario: Keyboard Dragging
1. Launch Simplenote
2. Press over the Search Bar
3. Scroll over the Notes List
4. Verify the keyboard **does not** get dismissed the second you begin scrolling
5. Enter any keyword that yields at least one result
6. Scroll the Notes List again
7. Verify that this time the keyboard does get dismissed

### Scenario: SortBar
1. Launch Simplenote
2. Press over the Search Bar
3. Enter any keyword that yields at least one result
4. Scroll downwards to dismiss the keyboard
5. Verify that the SortBar shows up correctly at the bottom of the screen

### Scenario: Insets
1. Launch Simplenote
2. Press over the Search Bar
3. Scroll at the bottom of the list
4. Verify the last row can be scrolled (AKA "we get the right bottom insets")
5. Enter any keyword that yields **lots** of results
6. Scroll all the way down
7. Verify, again, that the TableView's insets allow for the last row to show up
8. Dismiss the SearchBar
9. Verify (for the very last time, promise!) that the bottom insets are right

### Release
`RELEASE-NOTES.txt` was updated in 2bb357b with:
 
> From now on, you'll be able alter the order of your Search Results right from the Notes List.

---


jleandroperez:
Thank you Sly!! For the record, we've:

- Updated the SortBar's Background color (in dark mode)
- And added a 1px separator (at the top of the sort bar)

Simplenote/Classes/SearchDisplayController.swift
```diff
@@ -53,9 +53,14 @@ class SearchDisplayController: NSObject {
     /// Dismissess the SearchBar
     ///
     func dismiss() {
+        // Set the inactive status first, and THEN resign the responder.
+        //
+        // Why: Because of the `keyboardWillChangeFrame` Notification. We could really, really use
+        // the actual status to be available when such note is posted. Capisci?
```
bummytime: Thanks for the descriptive comment 🥇

Simplenote/Classes/UIView+Animations.swift
```diff
@@ -6,15 +6,23 @@ import UIKit
 //
 extension UIView {
 
-    /// Animates a visibility switch, when applicable
+    /// Animates a visibility switch, when applicable.
+    /// - Note: We're animating the Alpha property, and effectively switching the `isHidden` property onCompletion.
+    ///         Whenever the animation is thru, we'll also restore Alpha to full (1.0).
     ///
-    func animateVisibility(isHidden: Bool, duration: TimeInterval = 0.3) {
+    func animateVisibility(isHidden: Bool, duration: TimeInterval = UIKitConstants.animationQuickDuration) {
```
bummytime: 👋 goodbye magic number!

jleandroperez:
Thank you Bummy!!

PR #643, opened 2020-02-13
Implementing SortBar: Mark II

jleandroperez:
### Details:
In this PR we're bringing the SortBar's to life:

- **NoteListController** now has a specific SortMode to be applied when retrieving search results
- **Notes List** now has a Feedback Generator, triggered whenever any of the SortBar buttons is pressed
- We're wiring the Sort Order button: active Sort Order will be inverted on press
- And we're also wiring the Sort Mode button: an Alert Controller will be presented, with the valid options.
- Last but not least, the current Search Sort Mode will now be displayed in the Sort Bar

**Any feedback will be very very oh so very much welcomed!!!**

Ref. #507

cc @bummytime

### Scenario: Sort Order
1. Launch Simplenote
2. Press over the Search Bar
3. Enter any keyword that yields at least one result
4. Scroll downwards to dismiss the keyboard

- [x] Verify that pressing over the Sort Order button (left corner, in the Sort Bar) inverts the Search Results

### Scenario: Sort Mode
1. Launch Simplenote
2. Press over the Search Bar
3. Enter any keyword that yields at least one result
4. Scroll downwards to dismiss the keyboard

- [x] Verify that pressing over the Sort Mode (center of the sort bar) presents an Alert with all of the available Sort Modes
- [x] Verify that the Search's Sort Mode is persisted among sessions (last option you've picked should always be the default)

### Release

### DIFF
--- a/SimplenoteTests/OptionsTests.swift
+++ b/SimplenoteTests/OptionsTests.swift
@@ -26,6 +26,14 @@ class OptionsTests: XCTestCase {
         XCTAssert(options.listSortMode == .alphabeticallyAscending)
     }
 
+    func testEmptySearchSortModeYieldsListSortMode() {
+        let options = Options(defaults: defaults)
+        XCTAssertEqual(options.listSortMode, options.searchSortMode)
+
+        options.listSortMode = .alphabeticallyDescending
+        XCTAssertEqual(options.searchSortMode, .alphabeticallyDescending)
+    }
+
     func testLegacyUnspecifiedThemeIsProperlyMigrated() {
         let options = Options(defaults: defaults)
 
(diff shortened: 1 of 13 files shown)


SylvesterWilmott:
Looking good!

One thing I noticed from this PR is that pinned notes always appear at the top of search results, this shouldn't be the case. Search should not take pinned notes into account when ordering.

jleandroperez:
@bummytime Crasher fixed (thank you!!!)

@SylvesterWilmott Not sorting by Pinned status in a follow up PR, for tracking purposes.

Thaaaank you both!!!

SimplenoteTests/OptionsTests.swift
```diff
@@ -26,6 +26,14 @@ class OptionsTests: XCTestCase {
         XCTAssert(options.listSortMode == .alphabeticallyAscending)
     }
 
+    func testEmptySearchSortModeYieldsListSortMode() {
```
bummytime: 🥇 More tests FTW!

PR #649, opened 2020-02-17
Search: Displaying Relevant Dates

jleandroperez:
### Details:
In this PR we're prefixing the **creation / modification** date to the Note's Body Preview, when we're in Search Mode, and the Sort Option is relevant (that is: we're sorting by Creation or Modification date).

Ref. #507

### Test
1. Launch Simplenote
2. Verify the Note List does not prepend any Date
3. Press over the Search Bar
4. Enter a keyword that yields results
5. Swipe downwards to dismiss the keyboard

- [x] Verify that **Alphabetical Sort** does not display dates
- [x] Verify that **Sort by Creation Date** (ascending / descending) prepend the Creation Date
- [x] Verify that **Sort by Modification Date** (ascending / descending) prepend the Modification Date

### Release

### DIFF
--- a/Simplenote/Classes/NSWritingDirection+Simplenote.swift
+++ b/Simplenote/Classes/NSWritingDirection+Simplenote.swift
@@ -1,0 +1,17 @@
+import Foundation
+
+
+// MARK: - NSWritingDirection
+//
+extension NSWritingDirection {
+
+    /// Returns the *current* writing direction
+    ///
+    static var current: NSWritingDirection {
+        guard UIApplication.shared.userInterfaceLayoutDirection == .rightToLeft else {
+            return .leftToRight
+        }
+
+        return .rightToLeft
+    }
+}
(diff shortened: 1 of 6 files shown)


bummytime:
@jleandroperez I'm still reviewing, but I noticed this while testing with a RTL system language:



I'm thinking we need to prepend the date on the opposite side for RTL languages...thoughts?

Simplenote/Classes/NSWritingDirection+Simplenote.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+
+// MARK: - NSWritingDirection
+//
+extension NSWritingDirection {
```
bummytime: Thank **you** @jleandroperez — every little bit of RTL ❤️  helps our cause 🙇

jleandroperez:
Thank you sir!!

PR #654, opened 2020-02-18
Search Mode: Adjusting Sort Descriptors

jleandroperez:
### Details:
We're introducing two adjustments to the Search Mode:

1. We're not sorting anymore Pinned notes on top
2. Tags are always sorted in a fixed way (ascending / case insensitive)

Ref. #507

### Test
1. Launch the app
2. Pin a few notes
3. Press over the Search Bar
4. Enter a keyword that yields some of the pinned notes

- [x] Verify that Pinned Notes are no longer on top of the Search Results
- [x] Verify that inverting the sort order leaves the Tags suggestions untouched

### Release

### DIFF
--- a/Simplenote/Classes/NSSortDescriptor+Simplenote.swift
+++ b/Simplenote/Classes/NSSortDescriptor+Simplenote.swift
@@ -5,7 +5,7 @@ import Foundation
 //
 extension NSSortDescriptor {
 
-    /// Returns a NSSortDescriptor, to be applied over Note collections, so that the resulting collection relects the specified `SortMode`
+    /// Returns a NSSortDescriptor, to be applied over Note collections, so that the resulting collection reflects the specified `SortMode`
     ///
     static func descriptorForNotes(sortMode: SortMode) -> NSSortDescriptor {
         let sortKeySelector: Selector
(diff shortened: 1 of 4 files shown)


bummytime:
@jleandroperez leaving a quick comment on something I noticed (unrelated to this PR) as I am reviewing it — on an iPad running either iOS 11 or 12, the popover is not anchored to the sort bar:



**Note:** This is **not** an issue on iOS 13

Simplenote/Classes/NSSortDescriptor+Simplenote.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+
+
+// MARK: - SortMode List Methods
+//
+extension NSSortDescriptor {
+
+    /// Returns a NSSortDescriptor, to be applied over Note collections, so that the resulting collection relects the specified `SortMode`
```
bummytime:
```suggestion
    /// Returns a NSSortDescriptor, to be applied over Note collections, so that the resulting collection reflects the specified `SortMode`
```

jleandroperez: Thank you!!

jleandroperez:
Thank you Bummy!! Popover glitch fixed via #656 (**Thank you for catching that one!!!**)

PR #673, opened 2020-02-26
Tag Operator: Not producing partial matches

### LINKED ISSUE #672: Search: Tag search returns notes with partial tag match
### Expected

When I search for a tag (using the `tag:` syntax) I expect to only get results for notes with an exact tag match.

### Observed

I get results for notes with a partial tag match, e.g. searching for `tag:test` returns notes with the tag `testing` even if they don't have the tag `test` (instead of only notes with the tag `test`).

### Reproduced

1. Create a tag `test`.
2. Create a tag `testing`.
3. Add the tag `testing` to a note (and _don't_ add the tag `test` to it).
4. Search for `tag:test`. Notice that your note with the tag `testing` is in the search results.

-|-|-|-
Apple|iPad Mini|iPadOS 13.3|4.16.0.0 TestFlight beta

jleandroperez:
### Fix
In this PR we're updating the `tag:keyword` operator's behavior, so that it does not produce partial match results.

@aerych Sir, may I bug you with a fun PR?

Closes #672

### Test
1. Create a tag `test`
2. Create a tag `testing`
3. Add the tag `testing` to a note (and _don't_ add the tag `test` to it).
4. Search for `tag:test`

- [x] Verify that the note you've tagged in (3) does not show up in the results list
- [x] Verify that the unit tests are ever green!

### Release

### DIFF
--- a/Simplenote/Classes/NSPredicate+Simplenote.swift
+++ b/Simplenote/Classes/NSPredicate+Simplenote.swift
@@ -22,7 +22,7 @@ extension NSPredicate {
                 continue
             }
 
-            output.append( NSPredicate(format: "tags CONTAINS[cd] %@", tag) )
+            output.append( NSPredicate(format: "tags CONTAINS[cd] %@", formattedTag(for: tag)) )
         }
 
         guard !output.isEmpty else {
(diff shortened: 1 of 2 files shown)


jleandroperez:
Thank you Eric!!!

PR #674, opened 2020-02-26
SPSidebarContainerViewController: RTL Support

jleandroperez:
### Fix
In this PR we're adding RTL Support to the Sidebar Container, so that the Tag List appears where it really should!

@etoledom May I bug you with a real quick one? 🔥 
**Thank you!!!**

### Test
1. Enable RTL on your phone
2. Log into the app
3. Press the top right button

- [ ] Verify the Tag List is revealed on the right hand side
- [ ] Verify you can Swipe towards the right edge to dismiss the list
- [ ] Verify you can swipe from the Right > Left to reveal the list
- [ ] And please run a smoke test in Non RTL mode!

### Release
`RELEASE-NOTES.txt` was updated in 0d9e5da with:

> Now the Sidebar will show up on the right hand side, whenever Right to Left is enabled on your phone.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -15,6 +15,7 @@
 -   Dark Mode is now darker than just dark, for an improved dark side experience.
 -   Simplenote's Launch Screen will now match iOS's Appearance Settings
 -   Simplenote now supports Password Autofill support. Logging in has never been faster!
+-   Now the Sidebar will show up on the right hand side, whenever Right to Left is enabled on your phone.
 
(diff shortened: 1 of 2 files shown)


jleandroperez:
@etoledom Eduardo!! May I bug you with a quick PR?

Thanks in advance!!

jleandroperez:
Yeah this would have been not required with Autolayout, definitely. Swifting and auto-layout-ing as we go!!!

Thank you Eduardo!! 🔥

PR #675, opened 2020-02-26
Filtering out (reserved) Tag Names

jleandroperez:
### Fix
In this PR we're fixing a bug which would cause an internal string, to be leaked to the user interface.

@aerych **Thank you** for spotting this one!!!

### Test: Unfiltered Notes
1. Launch the app
2. Select the `Untagged Notes` filter
3. Add a new note

- [x] Verify the Tag List is clean, and no strange strings show up

### Test: Regressions
1. Launch the app
2. Select any User-Y defined tags
3. Add a new note

- [ ] Verify that the tag you've selected in (3) gets pre-loaded to the tags list

### Release

### DIFF
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -1,0 +1,21 @@
+import Foundation
+
+
+// MARK: - Internal Methods
+//
+extension SPAppDelegate {
+
+    /// Returns the actual Selected Tag Name **Excluding** navigation tags, such as Trash or Untagged Notes.
+    ///
+    /// TODO: This should be gone... **the second** the AppDelegate is Swift-y. We should simply keep a `NoteListFilter` instance.
+    ///
+    @objc
+    var filteredTagName: String? {
+        guard let selectedTag = SPAppDelegate.shared().selectedTag,
+            case let .tag(name) = NotesListFilter(selectedTag: selectedTag) else {
+                return nil
+        }
+
+        return name
+    }
+}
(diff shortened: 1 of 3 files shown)


Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -0,0 +1,21 @@
...
+// MARK: - Internal Methods
+//
+extension SPAppDelegate {
+
+    /// Returns the actual Selected Tag Name **Excluding** navigation tags, such as Trash or Untagged Notes.
+    ///
+    /// TODO: This should be gone... **the second** the AppDelegate is Swift-y. We should simply keep a `NoteListFilter` instance.
+    ///
+    @objc
+    var filteredTagName: String? {
```
aerych: As a temp fix that will go away, I think this is fine.  If it was gonna last a while I wonder if it would be better to compose this as a method that accepted the a tag (the current tag in this case) and then performed the filter operation on it. Written that way it would be easier to have a unit test for the behavior.   And speaking of unit tests.... is it worth writing one for this?

jleandroperez: **Definitely** going away soon, I've already (nearly) Swifted all the new things. Okay, not everything, but a big majority 😄 

I actually tried that approach (filtering method). Thing is, I ended up with code duplication, because the enum `NoteListFilter` has associated values, and cannot be bridged to ObjC. 

Thank you for bringing this up Eric!! This (really simple PR) had two or three iterations, until I just picked the shortest one 🔥

jleandroperez:
Will revisit the whole `Selected Tag` thing in another iteration, I think it probably deserves a better home, than an AppDelegate ivar!!

Thank you @aerych !!!

PR #678, opened 2020-02-27
Support the unicode bullet as a list item bullet in unordered lists

ghost:
### Fix
Support the unicode bullet character `•` as a list item bullet because lists copied from HTML or word documents may contain this character.

### Test
1. Open Simplenote
2. Create an unordered list with unicode bullets •
  ```
    • list item a
    • list item b
  ```
3. Press enter after the first item
4. A new list item should be opened, starting with a unicode bullet •

### Review
Anyone can perform the review.


jleandroperez (approved):
Thank you for helping us improve Simplenote!!

:shipit:

PR #683, opened 2020-02-28
Fixing Dynamic Type Scaling

jleandroperez:
### Details
@SylvesterWilmott spotted an issue in which our **Dynamic Type** + **custom weight** mechanism would fail to properly scale, when the device's font size changes.

In this PR we're rebuilding such API, so that all the things look great!

@aerych may I bug you (yet again!!)? Thanks for the patience sir!!

cc @SylvesterWilmott sooolved

### Known Issues
The sort bar isn't reacting well to Dynamic Type changes. Somehow that slipped thru the cracks, will follow up in another PR!

### Test
1. Launch Simplenote
2. Press over the Search Bar
3. Type any keyword that yields results
4. Scroll downwards to hide the keyboard
5. Press over the Sort Bar
6. Select any of the Date-Y sort options

- [x] Verify that the date (prefixed in the note contents) looks great
- [ ] Verify that when changing the Dynamic Type settings, the Date + Contents scale accordingly
- [x] Please verify this in iOS 11 too!

### Release

---

### DIFF
--- a/Simplenote/Classes/UIFont+Simplenote.swift
+++ b/Simplenote/Classes/UIFont+Simplenote.swift
@@ -14,7 +14,7 @@ extension UIFont {
     ///
     @objc
     static func preferredFont(for style: TextStyle, weight: Weight) -> UIFont {
-        if let cachedFont = FontCache.cachedFont(for: style, weight: weight) {
+        if let cachedFont = FontCache.shared.cachedFont(for: style, weight: weight) {
             return cachedFont
         }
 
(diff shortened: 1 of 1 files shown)


SylvesterWilmott:
Looks good but just a side note:

Text that appears in bars, eg. search bar placeholder/input, navigation bar "Cancel" button, sort bar caption text, Settings "Done" button... etc should not react to system-level text size changes.

There is some particularly stubborn weirdness with the sort bar text scaling that is bordering on UI breaking if you relaunch the app after changing the text size but all of this can be addressed in a separate PR.

jleandroperez:
Thank you Eric!! Will follow up the Tags list issue in another PR!!

🔥

PR #685, opened 2020-03-02
Typography adjustments

SylvesterWilmott:
### Fix
Fixing minor typography issues:

- Extra space removed from note cell timestamp prefix
- Note cell accessory minimum size adjusted
- Fix the font size of sort bar text (in line with iOS standards)
- Fix the font size of search bar text (in line with iOS Standards)

### Test
1. Visit the notes list
2. Pin a note
3. Notice that the pin is now 16pt instead of 15pt
4. Perform a search
5. Switch search mode to "Created" or "Modified"
6. Notice that the extra space was removed from the timestamp in the note cell
7. Adjust text size settings in iOS system preferences
8. Return to Simplenote
9. Notice that the sort bar text does not change
10. Notice that the search bar placeholder and input text does not change

### Review

### Release

### DIFF
--- a/Simplenote/Classes/SPNoteTableViewCell.swift
+++ b/Simplenote/Classes/SPNoteTableViewCell.swift
@@ -179,7 +179,7 @@ class SPNoteTableViewCell: UITableViewCell {
     private func refreshBodyAttributedString() {
         let bodyString = NSMutableAttributedString()
         if let prefixText = prefixText {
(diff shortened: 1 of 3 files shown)


jleandroperez:
@jkmassel sir!! Would it be possible to get another beta build for Simplenote 4.16?

**Thank you!!!**

PR #696, opened 2020-03-11
Update rake to fix a CVE

jkmassel:
Fixes _another_ warning, this one caused by `rake`.

**To Test**
1) Ensure all tests are passing
2) Check out this branch and run `rake dependencies`
3) Everything should work great.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -198,7 +198,7 @@ GEM
       highline (>= 1.6, < 3)
       options (~> 2.3.0)
     public_suffix (2.0.5)
-    rake (12.3.2)
+    rake (12.3.3)
     rake-compiler (1.0.8)
       rake
     representable (3.0.4)


loremattei:
This works and fixes the warning, though `rake dependencies` (which is what we always do) runs the system wide `rake` and not the one installed by `bundler`.

I looked at the CI scripts and it seems that we don't use rake there. So I wonder if we should just remove it from the `Gemfile`.

jkmassel:
Hmm – maybe my system is configured differently 🤔 

When using `rbenv`, it automatically uses the bundled `rake` if possible (again, at least on my machine):

```
## in ~/.../simplenote-ios
rake --version #rake, version 12.3.3

## in ~
rake --version #rake, version 12.3.2
```

IMHO, ideally we'd avoid relying on system `rake` even being present, especially given that future macOS versions won't come bundled with it, but WDYT?

loremattei:
Mmm... this is interesting. 
```

jkmassel:
Replaced by https://github.com/Automattic/simplenote-ios/pull/699.

PR #701, opened 2020-03-18
Fixing Layout Issue(s)

### LINKED ISSUE #697: RTL letter are reversed in order in note list
### Expected
I would expect the letters to remain in the order ABC in the note list as well as in the note directly when using RTL language like Arabic.

### Observed
Letters are reversed from ABC to CBA in the note list in the iOS app but remain in the correct order when viewing the single note.

### Reproduced

> 1. Go to the iOS app and add a new note with RTL language (e.g. Arabic Unicode)
> 2. Letters stay in the correct order when viewed in the single note. 
> 3. Navigate back to the note list, the orders reverse themselves (ABC to CBA).

Image from iOS
Image from iOS (1)

-|-|-|-
iPhone X|MQAD2B/A|iOS 13.3.1|4.16
iPhone 7| |iOS 13.3.1|4.16
iPad Mini| | |

Reported by a user in 2777608-zen

### LINKED ISSUE #700: Note List: Layout Issue
### Details:
We're receiving reports of a layout issue, in which the body lines are capped to just one (when we should actually display (2x).

- Condensed mode **disabled**
- Replicated on iPhone SE / iPad mini 2 on 12.4.5

jleandroperez:
### Fix
In this PR we're nuking a `NSWritingDirection` override, which is causing (two) bugs, and turns out... it wasn't needed at all, to begin with.

For reference, this bug was introduced while attempting to [solve this glitch](https://github.com/Automattic/simplenote-ios/pull/649#issuecomment-587516100), which, in turn, is only triggered when using **Right to Left Pseudolanguage** iOS Sim's Application Setting (that is, only happens when testing on the simulator!).

@aerych (No rush at all sir!!) Thanks in advance!! 🔥 

Closes #697
Closes #700

### Note:
I've patched up (temporarily) a broken translation, for testing purposes. Ref. #702

### Scenario A: Right to Left List
1. Set the **Application Language** to a **Arabic** 
2. Fresh install the app
3. Log into your account
4. Find the (Arabic) welcome note

- [x] Verify that the way the note shows up in the Editor is exactly the same as in the Notes List.

**Note:** We've been receiving reports of "Mirrored" / "Inverted" text in the notes list. Ref. #697

### Scenario B: Search Date Location
1. Set the **Application Language** to a **Arabic** 
2. Launch the app and log into your account
3. Verify the UX looks right to left!
4. Press over the search bar
5. Enter any keyword
6. Scroll downwards to reveal the **Sort Bar**
7. Modify the **Sort Mode** to either Creation or Modification date (second or third option from the top)

- [ ] Verify the Date is displayed in the right hand side

### Scenario C: Preview Lines
1. Launch the app on an iPhone SE / iOS 12
2. Log into your account

- [x] Verify there are up to two preview lines per note (three in total: Title x1 + Body x2)

### Release
`RELEASE-NOTES.txt` was updated in 769856f with:
 
> Fixed an issue that caused the Notes List to show up incorrectly, in Right to Left languages.


aerych:
For my first pass at testing I just used the simulator. My understanding is the original glitch was just the pseudo setting and actually changing the application to Arabic would work as expected. If I'm misunderstood let me know!

I spotted two things related to Scenario B I wanted to ask about.
First, the date in the Arabic post appears as Latin numerals, not as Arabic script as shown in the example:
Simulator Screen Shot - iPhone X - 2020-03-18 at 20 50 26

Second, my other notes were in English and the date is showing up on the left, vs the right of the cell.  I think this is expected in this case but want to confirm:
Simulator Screen Shot - iPhone X - 2020-03-18 at 20 50 34

aerych:
So it looks like what I was seeing when testing B was due to using a sim running iOS 11 😱 when I thought it was running iOS 13. Yay for not having version labels on sims by default. 😬 

:shipit: after resolving the out of date branch :)

jleandroperez:
Thank you Eric!! wooo!!

PR #708, opened 2020-03-24
Fixing List Processing

### LINKED ISSUE #707: Impossible to add Newlines between List Items
```
"Loved this app up until the newest update that went into effect a day or so ago. It has this problem now where it doesn't want to let the user leave line breaks in between lists. If you make a list of bullets and then hit “enter” to create a line break and then start a new list it crushes the lists together as though they are one list.

Making a series of lists and spacing them however I want is literally all I use this app for. As it stands, if the problem doesn’t get fixed, I’ll have to switch to a new note taking app for my needs. Which is a shame, because I loved this app before the update. Hopefully it gets fixed soon. If the problem is fixed, I’ll update my review as five stars, because again, this was an awesome app before the problem started."
```

From an iOS review

jleandroperez:
### Fix
In this PR we're fixing a bug that removes whitespaces in between List Items.

cc @aerych sir, may I bug you with **yet another** Lists-Y issue? (iOS this time!)

cc @msilbers Shipping in 4.17!

**Thank you!!!**

Closes #707

### Test
1. Add a new note
2. Write a random word in the title and please **return** 4 times
3. Press the top right icon (Lists) to insert a List
4. Verify a new bullet gets added exactly where you had the cursor
5. Type a random word
6. Press **return** 4 times
7. Repeat steps 3 thru 5: Add a new List with a word
8. Hit Back

- [x] Verify the list looks **exactly** the way you left it in step (6)

### Release
`RELEASE-NOTES.txt` was updated in dbdeb7b with:
 
> Fixed a bug that removed whitelines in between List Items.

### DIFF
--- a/Simplenote/NSMutableAttributedString+Styling.m
+++ b/Simplenote/NSMutableAttributedString+Styling.m
@@ -3,10 +3,6 @@
 #import "Simplenote-Swift.h"
 
 
-// Note: Capture Group Mark I: We're only interested in replacing the `- [ ]` marker, not the leading spaces
-//
-static NSUInteger const ListRangeIndex = 1;
-
 
 @implementation NSMutableAttributedString (Checklists)
 
@@ -25,11 +21,11 @@ static NSUInteger const ListRangeIndex = 1;
                                           range:plainString.fullRange] reverseObjectEnumerator] allObjects];
 
     for (NSTextCheckingResult *match in matches) {
-        if (ListRangeIndex >= match.numberOfRanges) {
+        if (NSRegularExpression.regexForChecklistsExpectedNumberOfRanges != match.numberOfRanges) {
             continue;
         }
 
-        NSRange matchedRange = [match rangeAtIndex:ListRangeIndex];
+        NSRange matchedRange = [match rangeAtIndex:NSRegularExpression.regexForChecklistsMarkerRangeIndex];
         if (matchedRange.location == NSNotFound || NSMaxRange(matchedRange) > self.length) {
             continue;
         }
--- a/Simplenote/NSMutableAttributedString+Styling.m
+++ b/Simplenote/NSMutableAttributedString+Styling.m
@@ -3,6 +3,10 @@
 #import "Simplenote-Swift.h"
 
 
+// Note: Capture Group Mark I: We're only interested in replacing the `- [ ]` marker, not the leading spaces
+//
+static NSUInteger const ListRangeIndex = 1;
+
 
 @implementation NSMutableAttributedString (Checklists)
 
@@ -21,7 +25,11 @@
                                           range:plainString.fullRange] reverseObjectEnumerator] allObjects];
 
     for (NSTextCheckingResult *match in matches) {
-        NSRange matchedRange = match.range;
+        if (ListRangeIndex >= match.numberOfRanges) {
+            continue;
+        }
+
+        NSRange matchedRange = [match rangeAtIndex:ListRangeIndex];
         if (matchedRange.location == NSNotFound || NSMaxRange(matchedRange) > self.length) {
             continue;
         }
--- a/SimplenoteTests/NSMutableAttributedStringStylingTests.swift
+++ b/SimplenoteTests/NSMutableAttributedStringStylingTests.swift
@@ -6,11 +6,6 @@ import XCTest
 //
 class NSMutableAttributedStringStylingTests: XCTestCase {
 
-    /// Every Match is expected to have two ranges: one containing the leading spaces, and a second one containing *just* the checklist marker
-    ///
-    private let numberOfExpectedRangesPerMatch = 2
-
-
     /// Verifies that `NSRegularExpression.regexForChecklists` will not match checklists that are in the middle of a string
     ///
     func testRegexForChecklistsWillNotMatchChecklistsLocatedAtTheMiddleOfTheString() {
@@ -29,7 +24,6 @@ class NSMutableAttributedStringStylingTests: XCTestCase {
         let matches = regex.matches(in: string, options: [], range: string.fullRange)
 
         XCTAssertEqual(matches.count, 2)
-        XCTAssertTrue(verifyEveryMatch(in: matches, contains: numberOfExpectedRangesPerMatch))
     }
 
     /// Verifies that `NSRegularExpression.regexForChecklists` matches multiple spacing prefixes
@@ -40,7 +34,6 @@ class NSMutableAttributedStringStylingTests: XCTestCase {
         let matches = regex.matches(in: string, options: [], range: string.fullRange)
 
         XCTAssertEqual(matches.count, 1)
-        XCTAssertTrue(verifyEveryMatch(in: matches, contains: numberOfExpectedRangesPerMatch))
     }
 
     /// Verifies that `NSRegularExpression.regexForChecklistsEmbeddedAnywhere` matches multiple spacing prefixes
@@ -51,7 +44,6 @@ class NSMutableAttributedStringStylingTests: XCTestCase {
         let matches = regex.matches(in: string, options: [], range: string.fullRange)
 
         XCTAssertEqual(matches.count, 2)
-        XCTAssertTrue(verifyEveryMatch(in: matches, contains: numberOfExpectedRangesPerMatch))
     }
 
     /// Verifies that `NSRegularExpression.regexForChecklists` only matches corretly formed strings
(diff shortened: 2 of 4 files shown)


Simplenote/NSMutableAttributedString+Styling.m
```diff
@@ -21,7 +25,11 @@ - (void)processChecklistsWithColor:(UIColor *)color
                                           range:plainString.fullRange] reverseObjectEnumerator] allObjects];
 
     for (NSTextCheckingResult *match in matches) {
-        NSRange matchedRange = match.range;
+        if (ListRangeIndex >= match.numberOfRanges) {
```
aerych: Currently, we expect two matches and the last match is the list.  Under these circumstances comparing ListRangeIndex with the numberOfRanges in this way is sound. I'm wondering how it could be future proof.  Consider that at some point in the future more than two ranges are expected (who knows why) and maybe in some future scenario the ListRangeIndex is not 1, and is not the last match. 
My concern is we have a const with a specific purpose but we're using it for two different things. 
I'm fine if we want to keep it like this, but maybe we could add a comment about how its being used here for the benefit of future engineers?

SimplenoteTests/NSMutableAttributedStringStylingTests.swift
```diff
@@ -24,6 +29,7 @@ class NSMutableAttributedStringStylingTests: XCTestCase {
         let matches = regex.matches(in: string, options: [], range: string.fullRange)
 
         XCTAssertEqual(matches.count, 2)
+        XCTAssertTrue(verifyEveryMatch(in: matches, contains: numberOfExpectedRangesPerMatch))
```
dmsnell: would it make sense to choose more behavior-oriented tests vs. testing here the specifics of the implementation?

what's the goal? maybe the tests can demonstrate the intention behind the code instead of being tightly coupled to the implementation. to echo what @aerych wrote, what if we intentionally want to grab the `x` or `X` inside the check and that adds a new range? what if we replace the RegExp matcher for a more robust parsing function? should these tests fail? has the app been broken in any way?

anyway, that's kind of a rant to motivate my principle that tests should only fail when the app and its behaviors are wrong, not when we legitimately make intentional non-breaking changes. we're already strongly coupled here in the test name so that limits our ability to de-couple the tests to some extent, but we don't have to follow suit if we prefer not to.

```swift
func testAttributesChecklistsThatAppearInMiddleOfString() {
  let input = new NSMutableAttachmentString("This is a valid checklist - [ ] Buy avocados - []);
  input.processChecklistsWithColor(new NSColor(…), …, …);

  XCTAssertEqual(input.attributes(at:26), [SPTextAttachment.Key]);
}
```

obviously I don't know `NSMutableAttributedString` very well or how to detect if the attribute/attachment at a given offset is the checklist. my point is wondering how hard it would be to write tests that test for these behaviors which presumably shouldn't change regardless of how we detect the checklists.

SimplenoteTests/NSMutableAttributedStringStylingTests.swift
```diff
@@ -94,5 +104,22 @@ class NSMutableAttributedStringStylingTests: XCTestCase {
...
+private extension NSMutableAttributedStringStylingTests {
+
+    /// Verifies that all of the TextCheckingResults contain the specified number of ranges
+    ///
+    func verifyEveryMatch(in matches: [NSTextCheckingResult], contains numberOfRanges: Int) -> Bool {
+        for match in matches where match.numberOfRanges != numberOfRanges {
+            return false
+        }
+
+        return true
```
dmsnell: this is novel. did you try `.contains` already?

```swift
func verifyEveryMatch(in matches: [NSTextCheckingResult], contains numberOfRanges: Int) -> Bool {
  return matches.contains { match in match.numberOfRanges != numberOfRanges }
}
```

what which point, is this function necessary?

```swift
XCTAssertEqual(matches.count, 1)
XCTAssertTrue(matches.contains { match in match.numberOfRanges != numberOfExpectedRangesPerMatch } )
```

or alternatively in the positive…

```swift
XCTAssertEqual(matches.count, 1)
XCTAssertTrue(matches.allSatisfy({ $0.numberOfRanges == numberOfExpectedRangesPerMatch }))
```

Simplenote/NSMutableAttributedString+Styling.m
```diff
@@ -3,6 +3,10 @@
 #import "Simplenote-Swift.h"
 
 
+// Note: Capture Group Mark I: We're only interested in replacing the `- [ ]` marker, not the leading spaces
+//
```
metrue: Just curious,  Why two slashes `//`  here but three slashed `///` here https://github.com/Automattic/simplenote-ios/pull/708/files#diff-5ff7a2f45a87242564a5548454b30c2bR9

Simplenote/NSMutableAttributedString+Styling.m
jleandroperez: I'll add another constant with the expected number of ranges, just for future proof purposes, and will also add a comment detailing what's going on there. Thank you sir!!!

Simplenote/NSMutableAttributedString+Styling.m
jleandroperez: Should definitely document this somewhere! Overall goal is as follows (there are probably few inconsistencies here and there, though).

A. Top level declarations: double `//`
B. Method documentation: three `///`


Sample:

```
// MARK: - NSRegularExpression
//
extension NSRegularExpression {

    /// Matches Checklists at the beginning of each line
    ///
    @objc
    static let regexForChecklists: NSRegularExpression = {
//...
    }()

```

Thanks for stepping by @metrue!!

SimplenoteTests/NSMutableAttributedStringStylingTests.swift
jleandroperez: Hey there sir!! Yeah I'm familiar with those APIs, **but** personally I try to avoid them.

**Why:** Because you're passing over a closure (anonymous function) which will be executed N times. It's just all about performance.

I know it's a tiny difference, but in the long run it truly adds up.

Thanks for the heads up!!!

SimplenoteTests/NSMutableAttributedStringStylingTests.swift
jleandroperez: You know, I wasn't sure about the entire idea of adding a new check to every unit test. 

I've ended up adding a constant in the `NSRegularExpression` extension itself, documented (why it's there, and what for), and added, instead, two specific tests to ensure the RegEx(es) produce always exactly two ranges.

Thanks for stepping by sir!!

jleandroperez:
@aerych for sanity's purposes, mind taking a second peek sir?

**Thank you!!!**

SimplenoteTests/NSMutableAttributedStringStylingTests.swift
dmsnell: > I know it's a tiny difference, but in the long run it truly adds up.

But does it?
Did you measure it?

I mean, these aren't some rando library functions - they're part of the `NSArray` class in the standard library. Also, languages like Swift are going to have optimizations for closures, especially ones as trivial as this. It might not actually pass any closures after it compiles, and I'm pretty sure in either case it's significantly faster than the time it takes to run the RegExp.

So, it's your code - you make the call. I wanted to ask in case you weren't aware of these, whose purpose is to tersely communicate the intent of the check "does everything follow this rule?" vs. the _steps_ we're taking "call this function, iterate over all the items, and if any one doesn't follow the rule then return false" (which themselves could contain bugs but for which no explanation is readily at hand). And now the thought of avoiding them for performance seems dubious at best.

In other words, it would seem just as silly to use them for performance as to avoid them but do as you like. I'm sure the code already has a bunch of its own idiosyncrasies like this in it and there's value in consistency.

jleandroperez: Hey D!! let's take this one to Slack, since this PR fixes a quite annoying bug, and we've got a release scheduled next monday.

Thaaaaankyoouu!!

jleandroperez:
Thank you Eric!!

jleandroperez:
cc @loremattei sir! can we get a new Simplenote TestFlight build?

Thanks in advance!!

Simplenote/NSMutableAttributedString+Styling.m
metrue: @jleandroperez Thanks for the clear explanation here.

PR #722, opened 2020-03-30
Setup test target for screenshots automation

mokagio:
### What it does
Add a new UI tests target to run the script to take the screenshots for the App Store, SimplenoteScreenshot. The target is already set up with Fastlane's `snapshot` to take a sample screenshot.

I created a test user to use for the screenshots, and added the credentials to the secrets repo. Since this is the first time I do it, I haven't merged the changes to the repo adding the file to `master` yet, but rather pointed this one to the branch I made. That's all to say that I should probably wait and make sure I've done things right for the secrets, and if so point this branch to the commit on master, before merging.

**Things I left out:**

- Currently taking only one screenshot on one device. More screenshots and devices to come later.
- The test to drive the UI and take the screenshot is horrible. I plan to use adopt `BaseScreen` from the WooCommerce iOS next.
- Right not the screenshot is not edited. That'll come later too.

### Test
Run `bundle exec fastlane snapshot` and verify the screenshot is taken.

### Review
The only real thing in this PR is the scaffolding. I'd love input on it, as well as the naming of the folders and files. Thanks 🤓.

### Release


.configure
```diff
@@ -1,13 +1,18 @@
 {
   "project_name": "simplenote-ios",
-  "branch": "master",
-  "pinned_hash": "10c6ba49771ae477c2e9a7689d6fa808abab2fc6",
+  "branch": "simplenote-screenshot-user-test",
+  "pinned_hash": "72f07559c4624918965acf9eacde9b843c9edd26",
```
mokagio: Since I wasn't sure on what's the best approach to use for the secrets, I pushed my stuff to a branch.

.configure
```diff
@@ -1,13 +1,18 @@
...
   "files_to_copy": [
     {
       "file": "iOS/simplenote/SPCredentials.swift",
       "destination": "Simplenote/Credentials/SPCredentials.swift",
       "encrypt": true
     },
+    {
+      "file": "iOS/simplenote/ScreenshotsCredentials.swift",
+      "destination": "SimplenoteScreenshots/Credentials/ScreenshotsCredentials.swift",
```
mokagio: WooCommerce iOS puts these in the `Utils/` folder, but I felt like a dedicated folder made things clearer. 

I'm happy to move it to `Utils/` if we want to make that a convention.

Thoughts?

.configure
```diff
@@ -22,4 +27,4 @@
   "file_dependencies": [
 
   ]
-}
\ No newline at end of file
+}
```
mokagio: This is my Vim adding a new line at the end of a file by default. [Here's an interesting post on where the convention comes from](https://thoughtbot.com/blog/no-newline-at-end-of-file).

.gitignore
```diff
@@ -47,6 +48,7 @@ fastlane/README.md
 fastlane/test_output/
 Preview.html
 build/
+screenshots/
```
mokagio: Fastlane dumps this in the root folder. I left it there for now, but I think it shall go in the `fastlane/` folder, as per what WooCommerce iOS does.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -1872,6 +1947,24 @@
...
+			);
+			name = "Generate Screenshots Credentials";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "SOURCE_PATH=~/.mobile-secrets/iOS/simplenote/ScreenshotsCredentials.swift\nTARGET_FOLDER=${SRCROOT}/SimplenoteScreenshots/Credentials\nTARGET_PATH=${TARGET_FOLDER}/ScreenshotsCredentials.swift\n\nif [ ! -f ${SOURCE_PATH} ]; then\n    echo \"Unable to Copy Credentials =(\"\nelse\n    echo \"Copying Credentials...\"\n    mkdir -p ${TARGET_FOLDER}\n    cp ${SOURCE_PATH} ${TARGET_PATH}\nfi\n";
+		};
```
mokagio: I added the logic to copy the secret `ScreenshotsCredentials.swift` as run script build phase to the screenshots target only.

I noticed there's already a dedicated target that does the same for the `SPCredentials.swift`, but it seemed more appropriate to isolate this piece of work to the only target that needs it.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -7,6 +7,17 @@
...
+			isa = PBXAggregateTarget;
+			buildConfigurationList = 3F1BB4B4243199FF006D1A04 /* Build configuration list for PBXAggregateTarget "ScreenshotsSecrets" */;
+			buildPhases = (
+				3F1BB4B524319A04006D1A04 /* ShellScript */,
+			);
+			dependencies = (
+			);
+			name = ScreenshotsSecrets;
+			productName = ScreenshotsSecrets;
+		};
```
mokagio: I added the logic to copy the secret `ScreenshotsCredentials.swift` as dedicated aggregate target, rather than appending it to the one already in place for `SPCredentials.swift`. It seemed more appropriate to isolate this piece of work to the only target that needs it.

.gitignore
jkmassel: Yep definitely should – can we avoid the `.gitignore` change since we'll just undo it later, and for now you could use git exclude on it if it's in the way? :)

.configure
jkmassel: No issues with this :)

.configure
jkmassel: Yay! I think you're the first one to ever use the branch support I added to this tool 😄 

Feel free to just commit to `master` for mobile secrets – because everyone else's changes are pinned, it's unlikely you could break someone else's stuff – also, PRs don't work well with `git-crypt`, so it's easier just to use `master` 😄

fastlane/Snapfile
```diff
@@ -0,0 +1,13 @@
...
+   "iPhone 8 Plus"
+])
+
+languages([
+  "en-US"
+])
+
+scheme("SimplenoteScreenshots")
+
+output_directory("./screenshots")
```
jkmassel: This is your culprit for the directory thing – if you change this to `./fastlane/screenshots` it _should_ fix the issue, but that's off the top of my head 🙃

.gitignore
```diff
@@ -44,6 +45,7 @@ Carthage/Build
 fastlane/Preview.html
 fastlane/report.xml
 fastlane/README.md
+fastlane/screenshots/
```
mokagio: Rather than removing the `.gitignore` change, I went ahead and added the lane we'll use later on in CI, and configured it with the "proper" folder.

fastlane/Fastfile
```diff
@@ -372,6 +372,39 @@ end
...
+
+      output_directory: screenshots_directory,
+      clear_previous_screenshots: true,
+
+      # Retry a few times if something is a little flaky
+      number_of_retries: 3,
+
+      # But fail completely after those 3 retries
+      stop_after_first_error: true
+    )
```
mokagio: Not so loosely inspired by https://github.com/woocommerce/woocommerce-ios/blob/1cff1b31fcfa9809c9d852716a529dc20e9294a6/fastlane/Fastfile#L343-L403

mokagio:
Thanks for the review!

> For a future PR – lets ensure a scheme is specified so that the script can run without interaction 😄

🤦‍♂ 

> I had to run `bundle exec fastlane take_screenshots` to get them to work – `snapshot` didn't

I'll admit that, after adding the lane, I didn't test the behavior of `snapshot`.

Could it be that it's failing because there's no `Snapfile`? 🤔 

> For some reason running `bundle exec fastlane run configure_apply` told me it needed to update the screenshots file, but put an empty file in place 🤔

I see... I didn't try that command either, actually. I'll look into it 👌

PR #723, opened 2020-03-31
Add steps to take five more screenshots

mokagio:
### What it does
Writes just enough code to take all the screenshots required for the App Store.

Next steps:

- Support multiple devices.
- The test to drive the UI and take the screenshot is horrible. I plan to use adopt `BaseScreen` from the WooCommerce iOS next.
- Right now the screenshots are not edited. That'll come later too.

### Test & Review

Given the code will change, I could use help verifying that `bundle exec fastlane take_screenshots` does indeed take the screenshots we're expecting on a machine other than mine.

#### Light mode

`bundle exec fastlane take_screenshots`

Screen Shot 2020-03-31 at 16 06 39

#### Dark mode

`bundle exec fastlane take_screenshots mode:dark`

Screen Shot 2020-03-31 at 16 39 46

### Release

### DIFF
--- a/SimplenoteScreenshots/SimplenoteScreenshots.swift
+++ b/SimplenoteScreenshots/SimplenoteScreenshots.swift
@@ -4,6 +4,7 @@ class SimplenoteScreenshots: XCTestCase {
 
     override func setUp() {
         continueAfterFailure = false
+        UIPasteboard.general.strings = []
     }
 
     func testScreenshots() {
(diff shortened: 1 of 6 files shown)


SimplenoteScreenshots/SimplenoteScreenshots.swift
```diff
@@ -41,12 +42,104 @@ class SimplenoteScreenshots: XCTestCase {
...
+        type("123", onFirstKeyboardOf: app)
+
+        snapshot("6-passcode")
+
+        type("4", onFirstKeyboardOf: app)
+
+        // Now, disable the passcode so we're not blocked by it on next launch.
+        openMenu(using: app)
+        loadPasscodeScreen(using: app)
+        type(ScreenshotsCredentials.testUserPasscode, onFirstKeyboardOf: app)
```
mokagio: This might be overkill in CI, where we reset the Simulator all the time, but I think is worth keeping for the occurrence when devs run this locally. Otherwise, it'd be very annoying to have to type the passcode and then go and disable it after every run.

PR #724, opened 2020-04-01
Add screenshots composer configuration for iPhone 8 Plus

mokagio:
### What it does
Adds the config setup to run the promo screenshots generation (screenshots composer?) for iPhone 8 Plus.

I've added only this one so I can get feedback on the setup itself. If everything's good. Then I can replicate for the other device sizes.

### Test
1. Run `BUNDLE_WITH=screenshots bundle install`, please do not commit the resulting change in `.bundle/config`, if any. See comment on that file for rationale.
1. Run `bundle exec fastlane take_screenshots && bundle exec fastlane create_promo_screenshots`.

If everything works, you should have six screenshots in your `fastlane/promo_screenshots/en-US/` folder exactly like these.

#### Dark Mode

`bundle exec fastlane take_screenshots mode:dark && bundle exec fastlane create_promo_screenshots`

<table>
<tr>
<td>
iPhone 8 Plus-01
</td>
<td>
iPhone 8 Plus-02
</td>
<td>iPhone 8 Plus-03</td>
<td>iPhone 8 Plus-04</td>
<td>iPhone 8 Plus-05</td>
<td>iPhone 8 Plus-06</td>
</tr>
</table>

#### Light Mode

`bundle exec fastlane take_screenshots mode:light && bundle exec fastlane create_promo_screenshots`

<table>
<tr>
<td>
iPhone 8 Plus-01</td>
<td>iPhone 8 Plus-02</td>

<td>iPhone 8 Plus-03</td>

<td>iPhone 8 Plus-04</td>
<td>iPhone 8 Plus-05</td>
<td>iPhone 8 Plus-06</td>
</tr>
</table>

### Review
I think only @jkmassel  has enough context to review this PR. @SylvesterWilmott I'd love your input on the screenshots output: is this what you were expecting?

### Release


.bundle/config
```diff
@@ -2,3 +2,4 @@
 BUNDLE_PATH: "vendor/bundle"
 BUNDLE_RETRY: "3"
 BUNDLE_JOBS: "3"
+BUNDLE_WITHOUT: "screenshots"
```
mokagio, line 5: _Unfortunately, I can't leave this comment inline in the file, because it gets cleaned every time `bundle install` is called with some different option._

The `fastlane/Pluginfile` contains RMagick in an optional group called screenshots.

In order to run the screenshots composition, you need to `bundle install --with screenshots`, but that will add a `BUNDLE_WITH: "screenshots"` to the `.bundle/config` file.

If that change is committed, CI builds on CircleCI will fail, because the RMagick gem installation fails to build its native dependencies.

I'm not sure how to work around this yet, but know for a fact that WooCommerce iOS has somehow done it, as their builds pass. So, while I'm gathering more info, I'm just making we don't try to install RMagick.

Yes, it's sort of pointless to have a config to exclude an already optional group. My intent is to leave a trail in the repo for future me to remember.

fastlane/Fastfile
```diff
@@ -408,6 +408,19 @@ end
       dark_mode: options[:mode].to_s.downcase == "dark"
     )
   end
+
+  lane :create_promo_screenshots do |options|
+    unless Fastlane::Helper::GitHelper.has_git_lfs then
```
mokagio, line 413: Nitpick. I prefer `unless condition then` to `if !condition then` or `if condition == false then`.

I've attended a talk on code readability making the point that `!` are sneaky and easily overlooked when reading code. I don't have a paper with a real study to back this up, but it does make sense to me. From that day, I've stayed away from `!something` whenever possible.

fastlane/Fastfile
```diff
@@ -408,6 +408,19 @@ end
...
+      UI.user_error!("Git LFS not enabled – Unable to generate promo screenshots. Run `git lfs install && git lfs fetch && git lfs pull` to fix this.")
```
mokagio, line 414: I copied this [from the WooCommerce iOS repo](https://github.com/woocommerce/woocommerce-ios/blob/a9caacb9c738d363dee9c180733978d89d384055/fastlane/Fastfile#L506-L526).

@jkmassel , I'm not sure why this is required. But I can see `fastlane/appstroreres/assets` stores its masks with LFS. Good call, I'll do the same.

SylvesterWilmott:
Looking good!

I'm wondering if we can use a standard status bar so it always displays the same time and no "Carrier" similar to how it's done in the WP screenshots. If we simply cannot do this programmatically, maybe we can include another image layer for the status bars.

mokagio:
> I'm wondering if we can use a standard status bar so it always displays the same time and no "Carrier" similar to how it's done in the WP screenshots.

Yeah, good catch! We can do that. I'll get onto it before the final batch. 👍

fastlane/metadata/en-US/app_store_screenshot_6.txt
```diff
@@ -0,0 +1 @@
+Protect your notes with a psscode lock
```
jkmassel, line 1: Looks like there's a typo here
```suggestion
Protect your notes with a passcode lock
```

fastlane/metadata/en-US/app_store_screenshot_1.txt
```diff
@@ -0,0 +1,2 @@
+A simple,
```
jkmassel, line 1: I don't think this comma makes sense

```suggestion
A simple
```

fastlane/screenshots.json
```diff
@@ -0,0 +1,63 @@
...
+		},
+		{
+			"device": "iPhone 8 Plus",
+			"filename": "iPhone 8 Plus-02.png",
+			"background": "#FFFFFF",
+			"screenshot": "iPhone 8 Plus-2-all-notes.png",
+			"text": "metadata/{locale}/app_store_screenshot_2.txt"
+		},
+		{
+			"device": "iPhone 8 Plus",
```
jkmassel, line 35: We should probably dismiss the contacts popup for this screenshot :) 
iPhone 8 Plus-03

fastlane/screenshots.json
```diff
@@ -0,0 +1,63 @@
...
+			"device_frame_size": [1242, 2208],
+			"device_frame_offset": [0, 0]
+		}
+	],
+	"entries": [
+		{
+			"device": "iPhone 8 Plus",
+			"filename": "iPhone 8 Plus-01.png",
+			"background": "#FFFFFF",
+			"screenshot": "iPhone 8 Plus-1-note.png",
```
jkmassel, line 24: You'll probably want to namespace these for dark/not-dark mode. Could be in a future PR though.

fastlane/metadata/en-US/app_store_screenshot_6.txt
mokagio: Wooops. I did fix this in another branch, but I'll cherry pick it here 👍

fastlane/metadata/en-US/app_store_screenshot_1.txt
mokagio: Agreed, and with the forced new line, it looks weird too

```
A simple,
note taking experience
```

It was in the spec, though. I'll get rid of it here. Easy enough to restore anyways.

fastlane/screenshots.json
mokagio: Good catch. Thanks!

Some of my tests dismissed it automatically, but other didn't. I remember reading Fastlane should do this for us, but either I'm wrong or there's something escaping it...

All this to say. I've fixed this earlier, before reading this review. I'll cherry pick 🍒 👍

fastlane/screenshots.json
mokagio: Yep. Your comment actually prompted me to finally go look in the WooCommerce codebase for how it's done. I'll add it in a followup.

mokagio:
Adding @rachelmcr as a reviewer as Jeremy is off for the next few weeks.

rachelmcr:
I'm adding @loremattei as a reviewer, as well, since I'm a novice with Fastlane and `bundle` issues. I'm still taking a look at this one, though.

.bundle/config
rachelmcr: I'm also running into a problem with RMagick when I try to test the PR locally, probably because I've never done our screenshot composition locally and don't already have RMagick installed.

When I run `BUNDLE_WITH=screenshots bundle install`, or when I try to install the RMagick gem directly, it fails with this error:

> ERROR: Can't install RMagick 3.2.0. Can't find ImageMagick with pkg-config

Not sure if that's the same failure you were seeing with CI builds, but I'll see if I can at least find a local solution for my error. :)

rachelmcr: Actually, it looks like installing RMagick has been troublesome for both local installs and on CI, and is optional in both WooCommerce and WordPress iOS. I noticed that we made it optional in the release toolkit, as well: https://github.com/wordpress-mobile/release-toolkit/pull/60

My understanding is that in our other projects we are generating the raw screenshots but not doing the composition on CI, but perhaps @loremattei can check me on all of this. :)

loremattei: Yeah! So, installing `RMagick` requires to pre-install `image-magick` via `brew`, and it can still cause some issues sometimes. We didn't want this to be a problem for the developers, so we made it optional. 
Installing `image-magick` on CI is usually very simple, though I never dug into `RMagick` on CI. 

Please, note that running this on CI is not a requirement right now. We don't expect to update the screenshots a lot...

About `.bundle/config`, I actually think we should just remove it from the repository. We don't check it in git in other project, as `rake dependencies` takes care of configuring `bundle` properly. 

I'll open a PR to remove it :-)

mokagio: 🤔 I'm of two minds about removing the `.bundle/config`. I like how you can track stuff like the concurrency in there, but, at the same time, the fact that it tracks changes when we do `--with screenshots` makes it annoying to track.

[Heroku recommends not to track it](https://devcenter.heroku.com/articles/bundler-configuration), for example.

If we used it only for the concurrency and location maybe it would make sense to track it, but I can see how the `BUNDLE_WITH` and/or `BUNDLE_WITHOUT` business might make it annoying.

mokagio: @rachelmcr [this workflow](https://github.com/Automattic/simplenote-ios/blob/9a39c1d4529d49b0b3bec32d6fc6c6f3a3b23bd5/.github/workflows/screenshots.yml#L57-L62) does the composition in CI as well. The same code should work locally, it does for me at least.

rachelmcr: Ah yeah, thanks for that — I mixed up what exactly I was looking at and that got things working for me locally. 👍

loremattei: The PR to remove `.bundle/config` has been merged in `develop`. :-)

PR #729, opened 2020-04-06
Adding support for Persian Language

jleandroperez:
### Details
We got a group of users requesting Persian support. In this PR we're updating the Xcode project, adding the `FA` language (using English as the base).

This will be translated via GlotPress with the existing machinery.

@aerych may I bug you with a real quick one?
Thanks in advance sir!

### Test
Nothing to test, just verify the app builds correctly please!

### Release


jleandroperez:
Thank you Eric!!

PR #739, opened 2020-04-22
Merge Release/4.18 into develop

loremattei:
This PR merges the new 4.18.0.1 beta into develop.
It also adds a small change to pull the new Persian strings from GlotPress and the current set of translated strings which I pulled in order to be able to test the new language in the beta. 

*Note:* The "Release build" step on CI is red because I rerun the job a second time to test the error notification on Slack. 
The reason of the failure is that the last step of the  "Release Build" job (which runs only on release tags) tries to upload the binary to TestFlight and TestFlight doesn't allow to upload the same binary two times. 
The build works as expected and the binary has been successfully deployed to TestFlight during the first run.

### Release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,16 +86,13 @@ jobs:
             curl -sL https://sentry.io/get-cli/ | bash
       - run:
           name: Build
-          command: "bundle exec fastlane build_and_upload_release skip_confirm:true"
-          no_output_timeout: 60m
-      - run: 
-          name: Prepare Slack messages
-          command: |
-            APP_VERSION=cat ./config/Version.Public.xcconfig | grep "^VERSION_LONG" | cut -d "=" -f2
+          command: | 
+            APP_VERSION=$(cat ./config/Version.Public.xcconfig | grep "^VERSION_LONG" | cut -d "=" -f2)
             echo "export SLACK_FAILURE_MESSAGE=':red_circle: Build for Simplenote $APP_VERSION failed!'" >> $BASH_ENV
-            echo "export SLACK_SUCCESS_MESSAGE=':tada: Simplenote $APP_VERSION has been deployed!'"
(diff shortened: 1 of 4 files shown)


jleandroperez:
@loremattei sir, I've just noticed there's an error in the CircleCI output. Related to recent CI changes right?


```
[13:43:07]: [Transporter Error Output]: ERROR ITMS-90189: "Redundant Binary Upload. You've already uploaded a build with build number '4.18.0.1' for version number '4.18'. Make sure you increment the build string before you upload your app to App Store Connect. Learn more in Xcode Help (http://help.apple.com/xcode/mac/current/#/devba7f53ad4)."
```

loremattei:
@jleandroperez oh no, sir! Sorry for the confusion. I detailed it in the PR description: 

> Note: The "Release build" step on CI is red because I rerun the job a second time to test the error notification on Slack.
> The reason of the failure is that the last step of the "Release Build" job (which runs only on release tags) tries to upload the binary to TestFlight and TestFlight doesn't allow to upload the same binary two times.
> The build works as expected and the binary has been successfully deployed to TestFlight during the first run.

Basically, the "Release builds" steps runs only on tags in order to build the binary and upload it to the AppStore. 
In order to test the Slack notification on failures, I created a new branch and I manually forced a re-run of the job on CI, expecting a failure. Since I was on a test branch, I _thought_ the failure didn't impact the release branch, but I forgot that the hook between GitHub and CircleCI looks only at the commit hash, so the failure is reported on all the branches that have the same commit at their HEAD.

In fact, if you look at the error, it's "Redundant Binary Upload.", which is the error the AppStore throw when you try to upload a binary with the same version of one that is already in TestFlight. 

We could get rid of the error by adding a _fake_ empty commit just to move the HEAD to another commit hash if you want. I didn't do it because the cause of the error is clear and the failing check is not required, but I can do it if you want me to.

Issue #695, opened 2020-03-11
Drop UIWebView


jleandroperez:
- [x] Update Simperium's **SPWebViewController**
- [x] Update 1Password's dependency

jleandroperez:
Simperium 0.8.27 drops **UIWebView** in favor of **WebKit**.

Closing this one, since such dependency was updated via #747

PR #744, opened 2020-05-04
Updates OnePassword Dependency

jleandroperez:
### Details:
In this PR we're upgrading our 1Password dependency, because of UIWebView-deprecation reasons.

@aerych May I bug you with a realllly quick one?
Thank you!!

Ref. #695

### Test
1. Fresh install Simplenote on a device with 1P installed
2. Press over `Log In` > `Log in with email`
3. Press over the **1Password** icon in the Email field

- [x] Verify that cancelling the 1P sheet gets you back to the Login UI
- [x] Verify that selecting a proper set of credentials logs you in
### Release

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -20,7 +20,7 @@ abstract_target 'Automattic' do
 	target 'Simplenote' do
 		# Third Party
 		#
-		pod '1PasswordExtension', '1.8.5'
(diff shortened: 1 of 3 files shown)


jleandroperez:
Thank you Eric!!

Issue #20, opened 2016-08-14
Deleted Property Collision


kwonye:
Xcode 6.3 is printing out a collision between the `Note` entity's `deleted` attribute, and `NSManagedObject`'s `deleted` property.

Simplenote overrides `NSManagedObject`'s `deleted` getter, which is used internally by CoreData to mark entities for deletion, next time *save* is executed.

Let's investigate + implement a safe workaround.

![screen shot 2015-04-21 at 1 31 26 pm](https://cloud.githubusercontent.com/assets/1195260/7257271/c24ce83c-e82a-11e4-8edd-9226878e2d55.png)

The cleanest workaround to fix this issue, is to preserve the remote attribute name (deleted), and map it into a different local attribute.

Let's first implement the appropiate Simperium Mechanism first.

bummytime:
Still seeing this warning in XCode 10.2

PR #750, opened 2020-05-07
Spring Cleaning: Checklist Button

a8ck3n:
### Fix

This PR is to address the checklist button state. Design calls for button to show/hide based on focused state of note. Could also just disable if we prefer that behaviour.

Currently the button is always present and clicking it without a focused note (ie: ready for edits) inserts a checklist item at what feels like a random location. Most likely this is insertion point as tracked by the UITextView. Side effect is the note is marked as modified and triggers a sync.

If the insertion point is below the fold, it isn't clear the note has even been modified.

https://d.pr/i/Zy3exj — Having trouble uploading/embedding the GIF.

### Test

1. Open a note but do not click to activate the keyboard. Verify the Checklist button is not visible in the navigation.
2. Tap on the note to focus it for editing. The keyboard should appear and the Checklist button should become visible.
3. Tap the Keyboard button to dismiss the keyboard. The Checklist button should also disappear.

### Test with Hardware Keyboard

This one is harder to do in the simulator but you can do it by toggling on the hardware keyboard option. Select **Keyboard > Connect Hardware Keyboard** from the **I/O** menu then repeating the above steps. Verify that the Checklist button only appears when the note has focus, regardless of keyboard visibility.

### Review

cc @jleandroperez — Would you mind having a look at this please?!

### Release

The Checklist button is now shown only when a note is focused for editing.


a8ck3n:
Attaching an mp4 video showing the current behaviour.

This video shows how the Checklist button is always present and clicking it causes a new checkbox item to appear in the note. If the saved insertion point is below the fold, the insertion is hidden.

[Checklist-Example.mp4.zip](https://github.com/Automattic/simplenote-ios/files/4593163/Checklist-Example.mp4.zip)

a8ck3n:
And this video shows the updated behaviour. The Checklist button is now hidden unless the note is focused for editing.

[Checklist-Final.mp4.zip](https://github.com/Automattic/simplenote-ios/files/4593178/Checklist-Final.mp4.zip)

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -280,6 +283,28 @@ - (void)adjustFrameForSafeInsets
     _noteEditorTextView.frame = viewFrame;
 }
 
+- (void)handleNoteEditorDidBecomeFirstResponderNotification:(NSNotification *)notification {
+    
+    checklistButton.hidden = NO;
```
jleandroperez, line 288: (Not introduced in your PR!!) but this one should probably be a @property, WDYT?

a8ck3n: I agree the NSButtons should definitely be properties on the class. I'd probably handle orthogonal changes like that via a separate PR but I can add there here too.

a8ck3n:
Thanks @jleandroperez for the feedback!

I'm not sure how I overlooked the delegate approach here. I'll make another pass and do my best to address both your A & C notes.

a8ck3n:
Hello @jleandroperez !

I think this is ready for another code review. I've removed the notification bits and now hook into the delegate API instead, as you recommended. Works like a charm! I've also tied the UI updates to some local button state, so `isEditing` and `isKeyboardVisible`, which both trigger a refresh when set.

Let me know what you think!

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -969,8 +989,8 @@ - (void)keyboardWillChangeFrame:(NSNotification *)notification {
         [UIView animateWithDuration:duration animations:animations];
     }
     
-    [self setVisibleRightBarButtonsForEditingMode:isEditing];
-    _keyboardHeight = visibleHeight;
+    self.keyboardHeight = newKeyboardHeight;
+    self.keyboardVisible = newKeyboardHeight > 0;
```
jleandroperez, line 993: Same as [L963](https://github.com/Automattic/simplenote-ios/pull/750/files#diff-92ac1b7e27f673e8c8f34b3969f54fe1R963)

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -783,6 +773,35 @@ - (void)bounceMarkdownPreview
...
+    self.checklistButton.hidden = !self.isEditing;
+    
+    if ([UIDevice isPad]) {
+        self.keyboardButton.hidden = YES;
+        self.createNoteButton.hidden = NO;
+    } else {
+        self.keyboardButton.hidden = !self.isKeyboardVisible;
+        self.createNoteButton.hidden = self.isKeyboardVisible;
+    }
+}
```
jleandroperez, line 787: This one can be super polished. Note that both KeyboardButton and CreateNoteButton's state is always opposed, so it all boils down to the condition that determines if `mustHideKeyboardButton`

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -783,6 +773,35 @@ - (void)bounceMarkdownPreview
     }];
 }
 
+- (void)refreshNavigationBarButtons {
+    
+    self.checklistButton.hidden = !self.isEditing;
```
jleandroperez, line 778: Checked with Sly on this one, will get back to you as soon as I hear from him. We'll _probably_ want to disable, rather than hide.

Simplenote/Classes/SPEditorTextView.m
```diff
@@ -161,17 +161,24 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceive
     return [touch.view isKindOfClass:[SPTextView class]];
 }
 
-- (BOOL)becomeFirstResponder
-{
-    [self setEditable:YES];
+- (BOOL)becomeFirstResponder {
```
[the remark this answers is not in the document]
a8ck3n: Thanks for these links Jorge. I've been trying to take my cues from the surrounding code but as you noted, it's not super consistent in there at the moment! I'll do another pass based on those guidelines.

Simplenote/Classes/SPEditorTextView.m
```diff
@@ -161,17 +161,24 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceive
...
-{
-    BOOL response = [super resignFirstResponder];
-    [self setNeedsLayout];
-    return response;
+- (BOOL)resignFirstResponder {
+
+    // Invalidate view layout.
+    // This can fail so we capture and act only when successful.
+    BOOL result = [super resignFirstResponder];
+    if (result == YES) {
```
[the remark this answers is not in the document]
a8ck3n: My thinking was to only invalidate the view layout if we are actually resigning first responder status. When logging from the code, I was seeing this being called multiple times. If those all fall within one pass of the run loop, it would probably only result in one view update but I wasn't sure about that.

SylvesterWilmott:
>It's probably best to disable / enable the Checklists button. Touching base with @SylvesterWilmott, since it's definitely his call.

The plan was always to hide/show the checklist button, so the behaviour here looks perfect! Thanks @a8ck3n

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -646,6 +646,13 @@ extension SPNoteListViewController {
 
     @objc(keyboardWillChangeFrame:)
     func keyboardWillChangeFrame(note: Notification) {
+        
+        // No window means we aren't in the view hierarchy.
+        if self.view.window == nil {
```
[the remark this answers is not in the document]
a8ck3n: When the SPNoteListViewController class tries to handle this notification while not currently onscreen, it causes warnings in the console about laying out views/cells with potentially invalid layout data (rough summary). The basic idea from the console error is that we shouldn't be doing view layout while not in the view hierarchy, so this test prevents that from happening.

These console errors are pretty easily missed thanks to all the logging from Simperium but they should be there on the `develop` branch.

The other option would be to update SPNoteListViewController to only listen for notifications when onscreen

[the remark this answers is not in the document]
a8ck3n: This is the warning from `develop`:

Console-Warning

And the full console log message:

```
2020-05-12 18:50:24.329016+0800 Simplenote[62220:5302818] [TableView] Warning once only: UITableView was told to layout its visible cells and other contents without being in the view hierarchy (the table view or one of its superviews has not been added to a window). This may cause bugs by forcing views inside the table view to load and perform layout without accurate information (e.g. table view bounds, trait collection, layout margins, safe area insets, etc), and will also cause unnecessary performance overhead due to extra layout passes. Make a symbolic breakpoint at UITableViewAlertForLayoutOutsideViewHierarchy to catch this in the debugger and see what caused this to occur, so you can avoid this action altogether if possible, or defer it until the table view has been added to a window. Table view: <UITableView: 0x7f8bc7819400; frame = (0 0; 375 667); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x60000210e730>; layer = <CALayer: 0x600002f5e100>; contentOffset: {0, -120}; contentSize: {375, 348}; adjustedContentInset: {120, 0, 260, 0}; dataSource: <SPNoteListViewController: 0x7f8bd5415dc0>>
```

Simplenote/Classes/SPNoteEditorViewController.h
```diff
@@ -57,11 +52,23 @@
...
+@property (nonatomic, strong) UIButton *keyboardButton;
+@property (nonatomic, strong) UIButton *createNoteButton;
+
 @property (nonatomic, strong) Note *currentNote;
 @property (nonatomic, strong) SPEditorTextView *noteEditorTextView;
 @property (nonatomic, strong) SPTagView *tagView;
 @property (nonatomic, strong) NSString *searchString;
-@property (nonatomic, assign) BOOL isPreviewing;
+
+@property (nonatomic, getter=isEditing) BOOL editing;
```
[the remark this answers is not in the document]
a8ck3n: Thanks for pointing this out Jorge!

I'm curious if you got a compiler error for this or not. I tried enabling a handful of different warnings but I wasn't able to generate one.

Currently I'm redeclaring but not redefining so I thought that might prevent the error. But even if I change the declaration, no error…hmm.

In my reading of the docs, the editing property on UIViewController is specific to the system provided edit button, which we don't use. So this usage doesn't look to be a conflict. Just in case, I'll rename this BOOL to avoid any unexpected side-effects though.

Simplenote/Classes/SPEditorTextView.m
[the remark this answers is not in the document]
a8ck3n: Any thoughts on that? Is it worth keeping?

I'd lean towards yes as it's a bit more explicit about the desired behaviour but it's a minor optimization at best.

Simplenote/Classes/SPNoteEditorViewController.m
a8ck3n: Yes, we could set it based on isEditing from above.

I have it in my head not to conflate the keyboard being visible with us being in editing mode though, as that isn't always the case. So for me this feels better for clarity and doesn't hurt readability.

I'm happy to change it if you feel it's unnecessary though.

Simplenote/Classes/SPNoteEditorViewController.m
a8ck3n: Something like this?

```
    self.checklistButton.hidden = !self.isEditing;
    self.keyboardButton.hidden = [self mustHideKeyboardButton];
    self.createNoteButton.hidden = !self.keyboardButton.hidden;
```

I'll have a look at this!

Simplenote/Classes/SPNoteEditorViewController.m
a8ck3n: For sure. Let me know and I'll make the change!

a8ck3n:
Morning to you @jleandroperez !

Thank you for the excellent feedback here. I've left some comments inline where appropriate and now I'm pushing up some changes to address the following:

- Tweak the login in the -refreshNavigationBarButtons: code path based on your notes about the two buttons always being the inverse of each other.
- Fix property collision with UIViewController.
- Fix for code formatting to better match the guidelines you linked me to.

Thanks for looking!

Simplenote/Classes/SPEditorTextView.m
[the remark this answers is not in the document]
jleandroperez: Honestly, the change makes a lot of sense. **However** it's possible we're inadvertently introducing UX glitches.

There's an old saying that goes like this.. ```premature optimization is the source of all evil```.

Personally, and by experience, I'd probably avoid changing the original behavior, **unless** we can determine exactly why it was introduced in the first place (OR) alternatively we detail a good Testing Plan, to make absolutely sure nothing breaks.

**Both alternatives are super.**

There are many hacks in this specific class, that, at a glance, don't really make any sense. For instance, the `editable = YES / NO` snippets are related to the text linkification. Editable TextViews don't process Links, and don't react to taps events, whereas Non Editable TextViews work pretty much like a WebView.

Simplenote/Classes/SPNoteEditorViewController.h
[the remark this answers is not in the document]
jleandroperez: No side effects, **but** this happened a while ago #20

Sometimes the compiler don't bug you now, but might start complaining, eventually. It's highly likely if this class was in Swift, you'd get such error.

(Which is something that's in the plans, swiftify all the things!)

Simplenote/Classes/SPNoteEditorViewController.m
jleandroperez: Apologies, I wasn't really clear. What I meant to say is: the check is duplicated. So instead of doing this:

```
self.keyboardVisible = newKeyboardHeight > 0;
```

You can simply do this:

```
self.keyboardVisible = isEditing;
```

Simplenote/Classes/SPNoteEditorViewController.m
jleandroperez: Actually what I had in mind was: local variable (`BOOL mustHideKeyboardButton`).

That way you can also avoid accessing the KeyboardButton.hidden property, since it's all local

Simplenote/Classes/SPNoteEditorViewController.m
jleandroperez: Replied by Sly below, good to keep. Thank you!!

Simplenote/Classes/SPEditorTextView.m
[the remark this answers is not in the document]
a8ck3n: Sounds good Jorge. I'll revert this to the original behaviour and add a comment based on the discussion here.

And thanks for clarifying the editable hack. There is a comment about data detectors but I wasn't sure what that meant. Good to know how it relates to link handling.

Simplenote/Classes/SPNoteListViewController+Extensions.swift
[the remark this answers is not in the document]
a8ck3n: Done and updated!

Simplenote/Classes/SPNoteEditorViewController.h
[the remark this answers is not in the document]
a8ck3n: Yup, definitely seems like it's better to be safe than sorry!

Simplenote/Classes/SPNoteEditorViewController.m
a8ck3n: I understand what you mean here; that the test on line 993 is the same as the one on 963. My concern is that this is only true in the context of this method, which is why I'd rather not conflate the two states generally. To me, this is just a readability issue though.

What I've done is renamed the BOOL on line 963 to `isKeyboardVisible`, then made the change on line 993 to match.

Let me know if I'm making sense here! 😅

Simplenote/Classes/SPNoteEditorViewController.m
a8ck3n: I think I'm following you here. I've gone with this now:

```
self.checklistButton.hidden = !self.isEditingNote;
self.keyboardButton.hidden = shouldHideKeyboardButton;
self.createNoteButton.hidden = !shouldHideKeyboardButton;
```

a8ck3n:
Hello, hello @jleandroperez  !

A few quick changes to address comments from yesterday.

- Reverts the `-resignFirstResponder` changes to avoid unintended side effects.
- Uses guard syntax when blocking UITableView updates in notifications.
- Renames a local BOOL for clarity (and my peace of mind).
- Tweaks how the refresh of navigation button state is handled.

Thanks again!

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -411,35 +413,35 @@ - (void)sizeNavigationContainer {
         }
     }
     
-    backButton.frame = CGRectMake(leadingPadding,
+    self.backButton.frame = CGRectMake(leadingPadding,
```
jleandroperez, line 416: **Nipticky:** Mind reviewing the left indentation on the newly indented lines?

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -934,29 +959,32 @@ - (void)endSearching:(id)sender {
...
         if (self->bVoiceoverEnabled) {
             CGRect newFrame        = self->_tagView.frame;
-            newFrame.origin.y      = self.view.frame.size.height - self->_tagView.frame.size.height - visibleHeight;
+            newFrame.origin.y      = self.view.frame.size.height - self->_tagView.frame.size.height - newKeyboardHeight;
             self->_tagView.frame         = newFrame;
         }
     };
     
+    // TODO: Drop isInteractive code path.
+    // Should only be necessary on pre-iOS 9 devices.
```
jleandroperez, line 987: **Actually** this was related to the (old) custom transition between List <> Editor. But should definitely be reviewed / dropped

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -646,6 +646,13 @@ extension SPNoteListViewController {
 
     @objc(keyboardWillChangeFrame:)
     func keyboardWillChangeFrame(note: Notification) {
+        
+        guard self.view.window != nil else {
```
jleandroperez, line 650: Nipticky: no need for `self` in Swift (unless you're inside a Closure)
Extra nipticky: Swiftier way below!

```
guard let _ = view.window else {
   return
}
```

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -783,6 +772,42 @@ - (void)bounceMarkdownPreview
     }];
 }
 
+- (void)refreshNavigationBarButtons
+{
+    // Note: The keyboard and createNote button states are always inverse.
```
jleandroperez, line 777: Nipticky: No need to document this one (since the code is super readable)

Simplenote/Classes/SPNoteEditorViewController.m
a8ck3n: Sorry! I'm not exactly clear on what the preferred style is for this one. I've never manually indented my code files like this (just long lines and soft-wrap for me). I'm not seeing an example of this in the style guide either. Is this what we want here?

Example-Indents

jleandroperez: Thank you for fixing this!! Yeah as a rule of thumb (unless it's documented elsewhere) I just follow whatever Xcode's default behavior is, when you hit `return`.

Looks great!

Issue #741, opened 2020-04-24
Spring Cleaning: Note Options


jleandroperez:
### Details:
We're modernizing Simplenote's Note Options UI:

- [ ] Build a new ViewController, in charge of rendering the Note Options
- [ ] Build a new HistorySlider UI component
- [ ] Build a new Note Information UI component
- [ ] Review and delete old (unused!) code

All of the new components should be implemented in Swift

### New Options UI:
<img width="200" alt="Note Options - Unshared" src="https://user-images.githubusercontent.com/1195260/80241904-d0f28000-863a-11ea-84f0-cdfbbd1036ff.png"> <img width="200" alt="Note Options - Shared" src="https://user-images.githubusercontent.com/1195260/80241909-d354da00-863a-11ea-9af0-b0fcfe6c68cf.png">

### New History Slider
<img width="200" alt="Note Options - History" src="https://user-images.githubusercontent.com/1195260/80241911-d4860700-863a-11ea-9e7e-d7a03d0b82eb.png">

### New Information Sheet
<img width="200" alt="Information" src="https://user-images.githubusercontent.com/1195260/80241914-d5b73400-863a-11ea-92ef-ad2523283a1b.png">

jleandroperez:
Closing this one for now!

PR #751, opened 2020-05-11
Spring Cleaning: New Note Options View

a8ck3n:
### Fix

This PR adds the new note options button and wires up a basic view with controller and storyboard.

Sub-task of #741.

### Test

1. Select a note from notes list.
2. Confirm the new note options button displays correctly. Currently flagged with a red background.
3. Confirm the view presentation works as expected. Can dismiss with Done button or by swiping down.

**Note:** The various toggles and commands are not wired up yet.

Note-Options-New

PS: @jleandroperez — Nice small changeset for you. You expressed some concern about using storyboards over XIBs so I wanted to make sure this works for you.

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -126,6 +126,7 @@
 		46A3C9C217DFA81A002865AE /* InfoPlist.strings in Resources */ = {isa = PBXBuildFile; fileRef = E29ADD4417848E8500E55842 /* InfoPlist.strings */; };
 		46A3C9C317DFA81A002865AE /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = E29ADD4D17848E8500E55842 /* Images.xcassets */; };
 		46A3C9C417DFA81A002865AE /* Simplenote-DB5.plist in Resources */ = {isa = PBXBuildFile; fileRef = E2FCA5F21789118A00FC229D /* Simplenote-DB5.plist */; };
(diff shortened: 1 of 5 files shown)


a8ck3n:
Hello Jorge! Thanks for the feedback here.

I think I made a mistake basing this branch on kdl-checklist-button. The checklist work is largely independant from the new Note Options view so that could potentionally get merged to develop before any of the Note Options work is done. I'll set this up on a new feature branch so future PRs can be merged safely without concern for breaking develop.

> The UI on this one doesn't match the mockups

The UI here is based on the final blueprints from Sylvester so they don't match what is in the master issue. (plus the top four rows of note info, that will move before finalizing this view)

Let me know if I should work to the master issue or if we should update that to reflect the final blueprints.

> iPad UX needs some polishing. What do you think about using a Popover (like the other navbar buttons?)

For sure. I hadn't tested this at all on the iPad yet but if we currently use Popoves then that would be the preferred approach.

> Localization is probably gonna need some thought

I'm happy to add this to the view controller if we aren't able to pull strings from storyboards yet.

> The new Navigation Button's background is set to red
> There are several debug logs in the code

This is intentional. — 😄 I prefer to leave some details in the logs until the work is done but cleaning that up would definitely be part of the final stages before a merge. I can add a task to the master issue for this.

I'll have a think on the branching strategy here and then update this PR (or submit a fresh one). I'll look at testing the iPad flow as well and stub out localization points. I'll ping again when there's more for you to review!

PR #754, opened 2020-05-14
Perform screenshot composition in CI

mokagio:
### What it does
Runs the screenshot composition in CI.

### Test
Add the "Needs Screenshots" label to trigger a screenshot generation run, and check that there is an attachment with the composed screenshots. Like [here](https://github.com/Automattic/simplenote-ios/pull/754/checks?check_run_id=676696033).

Screen Shot 2020-05-15 at 11 52 59 am

You might notice that the composed screenshots don't use the custom font 😞 I don't know why that's the case as in the [build output](https://github.com/Automattic/simplenote-ios/pull/754/checks?check_run_id=676696033#step:11:266) it looks like the font installation was successful.

Screen Shot 2020-05-15 at 12 47 21 pm

_The font is just the system standard, Times New Roman, I think._

I'd still like to get this code merged, so that it's all there and ready to be used if we need. I'll followup with another PR that disables this step (to save CI time) and adds instruction of how to run the composition locally.

### Review

### Release

### DIFF
--- a/.github/workflows/screenshots.yml
+++ b/.github/workflows/screenshots.yml
@@ -57,6 +57,7 @@ jobs:
 
     - name: Compose
       run: |
+        git lfs fetch && git lfs checkout
         brew install pkg-config imagemagick@6
         BUNDLE_WITH=screenshots bundle install
         bundle exec fastlane create_promo_screenshots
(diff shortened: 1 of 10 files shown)


mokagio:
> Since this code doesn't run without the "Needs screenshots" label, this approach doesn't seem to have big downsides.

Fair point. It's also super fast to run, and it's good to have a check in CI to make sure that the composition doesn't break there, for when/if we'll move it there.

PR #767, opened 2020-06-01
Add lane to upload screenshots

mokagio:
### What it does
Adds a dedicated lane to upload the screenshots, and tweaks the `Fastfile` to make it possible.

**Feedback please 📣** Does it make sense as a standalone lane, or should I fold it into the settings for `deliver`? I went for a dedicated lane in the first place because I wanted to experiment. I kinda feel it's safer to have to call a dedicated lane for this destructive action, rather than _remembering_ to use `--skip-screenshots`. I'm not good at _remembering_. 😅

### Test
You can run `bundle exec upload_screenshots` and just kill it after seeing the report with the screenshots as expected.

### Review

### Release


fastlane/Fastfile
```diff
@@ -547,24 +547,53 @@ end
...
-  # bundle exec fastlane update_certs_and_profiles 
+  # bundle exec fastlane update_certs_and_profiles
   #####################################################################################
   lane :update_certs_and_profiles do | options |
     alpha_code_signing
     internal_code_signing
     appstore_code_signing
   end
 
+  lane :test_screenshots_upload do |options|
```
jkmassel, line 565: This would be great to extract into a shared lane over in the release toolkit! We always upload screenshots the same way across all apps (ie – upload them, overwrite the existing ones, but DON'T PUBLISH!!! so that we can click the button by hand) 😄 

Other comments in this PR are based on the idea of moving that over.

fastlane/Fastfile
```diff
@@ -626,7 +655,7 @@ def screenshots_directory()
   File.join(fastlane_directory, "screenshots")
 end
 
-def promo_screenshots_directory()
+def promo_screenshots_directory
```
jkmassel, line 658: I super like that this is extracted, but IMHO it should be moved to the release toolkit and used as a default anywhere we'd reference the value. It might not even need to be customizable at this point...we're pretty consistent with it.

(the end goal being to make the fastfiles very small and share as much as is reasonable without over abstracting 😄

mokagio: > (the end goal being to make the fastfiles very small and share as much as is reasonable without over abstracting 😄

Nice. I like it, it makes sense to me to extract with reasonable defaults and then add the ability to override where needed.

mokagio:
Thanks for the precious feedback @jkmassel.

Since having this lane here is not something urgent, I might leave the PR open for reference, but find some time to move the action to the release toolkit and eventually close this in favor of using that.

mokagio:
Actually, thinking about it, I'd like to merge this as it is @jkmassel. This way I can close the loop on this iteration of the screenshots work, as I don't know when I'll get the time to work on extracting it properly 🤔

PR #786, opened 2020-06-26
Merge release 4.20 into master

mokagio:
_This should be merged only after 4.20 has begun the phased rollout_


mokagio:
The build failure is due to a redundant build upload to TestFlight. It happened because the `finalize_release` lane failed on me locally a few times and I didn't realize that by running it again I would have pushed new commits 😳 .

I can either push a new build number to make CI happy, or use admin rights to merge, given that the [first upload was successful](https://app.circleci.com/pipelines/github/Automattic/simplenote-ios/1884/workflows/a058c774-3220-4c3c-a1f4-2f76e46f6a82).

Screen Shot 2020-06-26 at 4 24 31 pm

PR #792, opened 2020-07-01
Merge hotfix 4.20.1 into develop

mokagio:
This should be merged only after Apple approves the build and we ship it.


Simplenote/Resources/AppStoreStrings.pot
```diff
@@ -65,10 +65,8 @@ msgctxt "app_store_keywords"
...
 msgstr ""
 
-msgctxt "v4.20-whats-new"
+msgctxt "v4.20.1-whats-new"
 msgid ""
-"Fixed a bug that affected the Share Extension's Appearance.\n"
-"Fixed RTL support in the Authentication Screens.\n"
-"Fixed RTL support in the Tags Editor.\n"
+"New Privacy Notice for California Users\n"
 msgstr ""
```
mokagio, line 71: I'm not sure why, but this value didn't get updated on App Store Connect.

fastlane/metadata/fr-FR/release_notes.txt
```diff
@@ -0,0 +1,3 @@
+Résolution d’un bogue qui affectait l’apparence de l’extension Partage.
+Résolution de la prise en charge RTL sur les écrans d’authentification.
+Résolution de la prise en charge RTL dans l’éditeur d’étiquettes.
```
mokagio, line 3: These are from 4.20.0. I'm guessing they weren't yet localized when I run the script at the time of submitting it?

The same goes for the `id` and `ru` versions.

fastlane/metadata/de-DE/description.txt
```diff
@@ -30,6 +30,7 @@ ORGANISIEREN UND SUCHEN
 
 Datenschutzerklärung: https://automattic.com/privacy/
 Geschäftsbedingungen: https://simplenote.com/terms/
+California users privacy notice: https://wp.me/Pe4R-d/#california-consumer-privacy-act-ccpa
```
mokagio, line 33: I touched this and all the other ones manually.

I'm not sure whether it's appropriate to bypass the translation process? I thought it would be handy to have the links in already given this an hotfix.

jkmassel: I think this only needs to be present in `en-US`?

But it's probably not hurting anything for it to be there – as long as it goes for translation. I guess the other question is do we have localized versions of that content at the URL? If it's only even in English, this probably doesn't help much

mokagio: My rationale for adding it to other locales too was that a user in California might not necessarily have their app in English. Although I do see your point and I think it's fair.

"What's the point of having it in `<language>` if it's not translate in `<language>`" is a valid question, too. 🤔 I wonder if there's a rule about this in CCPA?

The privacy policy in the Automattic page is translated only in a few languages right now. Also, while a user in California might have their phone in a language other than English, they should be able to _read_ English.

All in all, maybe I was bit rushed in adding those. @belcherj @jorgeatorres what do you think? I can easily open a PR against `develop` to remove them.

Issue #762, opened 2020-05-25
Spring Cleaning: Nü History Slider


jleandroperez:
### Guidelines
- New code should always be Swift
- Autolayout **is always preferred**
- Please drop outdated components when appropriate! 🔥🔥🔥

### Details
- [ ] Implement a new History Slider component, to replace the current one

### Mockups: iPhone
<img width="150" alt="History Slider" src="https://user-images.githubusercontent.com/1195260/82848103-5c746080-9ec8-11ea-8e58-aa82f6e5062f.png">

### Mockups: iPad
<img width="400" alt="History Slider iPad" src="https://user-images.githubusercontent.com/1195260/82848105-6007e780-9ec8-11ea-9d61-3b2ae4eb7ca0.png">

jleandroperez:
### Note:
Sly brought up an important note. We need to disable Checklists Interaction while the slider is UP.

PR #793, opened 2020-07-01
New component to embed a view controller in a card-like view

eshurakov:
### Details

This PR add a new `SPCardViewController` component, which can embed another view controller in a view with rounded corners, shadow and a little bit transparent background.
This component will be used to wrap and show new history slider on the note editor screen, but not used anywhere at the moment.

### Notes
This is part of https://github.com/Automattic/simplenote-ios/issues/762
Will follow up with another PR for History Slider

### Screenshots

These are the example of how the component looks like.

<p float="left">

</p>

### DIFF
--- a/Simplenote/Classes/SPCardViewController.swift
+++ b/Simplenote/Classes/SPCardViewController.swift
@@ -16,6 +16,10 @@ final class SPCardViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
+    deinit {
+        stopListeningToNotifications()
+    }
+
     override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -24,6 +28,10 @@ final class SPCardViewController: UIViewController {
         addShadowView()
         setupContainerView()
         addChildViewController()
+
+        refreshStyle()
+
+        startListeningToNotifications()
     }
 }
 
@@ -39,7 +47,6 @@ private extension SPCardViewController {
     func setupContainerView() {
         view.addFillingSubview(containerView)
 
-        containerView.backgroundColor = UIColor.simplenoteCardBackgroundColor.withAlphaComponent(Constants.backgroundAlpha)
         containerView.layer.cornerRadius = Constants.cornerRadius
         containerView.layer.maskedCorners = [.layerMaxXMinYCorner, .layerMinXMinYCorner]
         containerView.layer.masksToBounds = true
@@ -49,6 +56,27 @@ private extension SPCardViewController {
         addChild(viewController)
         containerView.addFillingSubview(viewController.view)
         viewController.didMove(toParent: self)
+    }
+
+    func refreshStyle() {
+        containerView.backgroundColor = UIColor.simplenoteCardBackgroundColor.withAlphaComponent(Constants.backgroundAlpha)
+    }
+}
+
+// MARK: - Notifications
+//
+private extension SPCardViewController {
+    func startListeningToNotifications() {
+        let nc = NotificationCenter.default
+        nc.addObserver(self, selector: #selector(themeDidChange), name: .VSThemeManagerThemeDidChange, object: nil)
+    }
+
+    func stopListeningToNotifications() {
+        NotificationCenter.default.removeObserver(self)
+    }
+
+    @objc func themeDidChange() {
+        refreshStyle()
     }
 }
 
--- a/Simplenote/Classes/SPCardViewController.swift
+++ b/Simplenote/Classes/SPCardViewController.swift
@@ -39,7 +39,7 @@ private extension SPCardViewController {
     func setupContainerView() {
         view.addFillingSubview(containerView)
 
-        containerView.backgroundColor = UIColor.simplenoteCardBackgroundColor.withAlphaComponent(0.97)
+        containerView.backgroundColor = UIColor.simplenoteCardBackgroundColor.withAlphaComponent(Constants.backgroundAlpha)
         containerView.layer.cornerRadius = Constants.cornerRadius
         containerView.layer.maskedCorners = [.layerMaxXMinYCorner, .layerMinXMinYCorner]
         containerView.layer.masksToBounds = true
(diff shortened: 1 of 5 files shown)


Simplenote/Classes/SPCardViewController.swift
```diff
@@ -0,0 +1,61 @@
...
+    func addShadowView() {
+        let shadowView = SPShadowView(cornerRadius: Constants.cornerRadius,
+                                      roundedCorners: [.topLeft, .topRight])
+        view.addFillingSubview(shadowView)
+    }
+
+    func setupContainerView() {
+        view.addFillingSubview(containerView)
+
+        containerView.backgroundColor = UIColor.simplenoteBackgroundColor.withAlphaComponent(0.97)
```
jleandroperez, line 42: Nipticky: Maybe `0.97` > constant?

jleandroperez: Detail: You may need to listen to `VSThemeManagerThemeDidChangeNotification` and refresh the `.backgroundColor`.

TL;DR: Simplenote has supported, for a few years, dark mode. As per iOS 13 you don't need to do anything, for legacy support would require a manual refresh.

Simplenote/Classes/SPCardViewController.swift
```diff
@@ -0,0 +1,61 @@
...
+        containerView.layer.cornerRadius = Constants.cornerRadius
+        containerView.layer.maskedCorners = [.layerMaxXMinYCorner, .layerMinXMinYCorner]
+        containerView.layer.masksToBounds = true
+    }
+
+    func addChildViewController() {
+        addChild(viewController)
+        containerView.addFillingSubview(viewController.view)
```
jleandroperez, line 50: Out of curiosity, do you think it's possible to avoid using `containerView` entirely, and add the `viewController.view` as a direct subview?

Simplenote/UIView+Constraints.swift
```diff
@@ -0,0 +1,22 @@
...
+
+extension UIView {
+    func addFillingSubview(_ view: UIView,
+                           edgeInsets: UIEdgeInsets = .zero,
+                           useSafeArea: Bool = false) {
+
+        view.translatesAutoresizingMaskIntoConstraints = false
+
+        addSubview(view)
+        view.frame = bounds.inset(by: edgeInsets)
```
jleandroperez, line 11: Since the insets are set via constraints, maybe we can get rid of this one?

Simplenote/SPShadowView.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    override var bounds: CGRect {
+        didSet {
+            updatePath()
+        }
+    }
+
+    /// Designated Initializer
+    ///
+    ///  - cornerRadius: The radius of each corner oval
```
jleandroperez, line 20: Nipticky: If you add `// - Parameters:` before enumerating the method's parameters, Xcode's **Option + Click** will display them under the title **Parameters**

Simplenote/SPShadowView.swift
```diff
@@ -0,0 +1,97 @@
...
+        configureShadow()
+        configureMask()
+
+        updatePath()
+    }
+
+    /// Configuration of the shadow
+    ///
+    func configureShadow() {
+        layer.shadowColor = UIColor.black.cgColor
```
jleandroperez, line 55: Maybe it'd be useful to parametrize this one? 🤔 

(In the lines of `var shadowColor = UIColor.black`?)

Simplenote/SPShadowView.swift
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+// MARK: - Constants
+//
+private extension SPShadowView {
+    struct Constants {
+        static let shadowOpacity: Float = 0.1
+        static let shadowRadius: CGFloat = 4.0
+        static let shadowOffset = CGSize(width: 0, height: -2)
```
jleandroperez, line 95: You can use UIOffset, instead, which might be more appropriate, in this particular case!

Simplenote/Classes/SPCardViewController.swift
```diff
@@ -0,0 +1,61 @@
+import UIKit
+
+// MARK: - SPCardViewController
```
jleandroperez, line 3: Nipticky: Mind adding a brief description of what's the purpose of this VC?

Simplenote/Classes/SPCardViewController.swift
eshurakov: Thanks for the info about `VSThemeManagerThemeDidChangeNotification `. I wasn't aware of it, will add support for it!

Simplenote/Classes/SPCardViewController.swift
eshurakov: I believe it's not possible. Right now card consists of 2 subviews: 
* shadow view, which draws shadow on the outside of bounds rect
* content view, which has rounded corners and clips inside content, so that child view controller doesn't need to worry about any of those

If we would add child view controller directly to the main view, we would need to apply rounded corners and clip subviews. This would clip and remove shadow.

Simplenote/SPShadowView.swift
eshurakov: Nice! Didn't know about that, thanks!

Simplenote/SPShadowView.swift
eshurakov: Good point

Simplenote/SPShadowView.swift
eshurakov: I agree UIOffset would fit better there, but it is used to configure `layer.shadowOffset`, which is `CGSize` :-(

Simplenote/Classes/SPCardViewController.swift
eshurakov: Good point! Will add a description 👍

Simplenote/UIView+Constraints.swift
eshurakov: Indeed 👍

eshurakov:
Thank you for the feedback, @jleandroperez. I've addressed some of the comments. Please take a look when you have time.

jleandroperez:
Merging this one, thank you @eshurakov !!

PR #795, opened 2020-07-06
History slider

eshurakov:
### Info
This PR replaces old note history component with a new history slider as described in #762.
Along with the main task the following improvements were made:
* Support rotation while history slider is shown
* Allow scrolling the content of the note editor
* History slider component waits till all versions are loaded before showing the slider
* Error message is shown inline inside the card instead of in a popup alert view

### Pending changes
The following changes are expected in the next PR:

* Put node editor in read only mode and disable interactions with navigation bar while history slider is visible
* Add voice over support 
* Update dismiss button according to mockups
* Support theme change notifications and improve light / dark mode support
* Add unit tests

### Test
Common steps:
1. Open a note
2. Open note options
3. Tap on the history button

What can be tested after that:
* If no internet connection, inline error message is shown
* If current version is selected, restore button is disabled
* On devices with haptic feedback, dragging a slider triggers haptic feedback
* Screen can be rotated to landscape / portrait 
* Selecting a version updates note editor with that version
* Selecting a version and dismissing a history slider restores editor to the current version
* Selecting a version and saving it should persist the changes

<p float="left">

</p>


Simplenote/Classes/SPNoteEditorViewController.h
```diff
@@ -66,9 +58,12 @@
 @property (nonatomic, strong) SPTagView *tagView;
 @property (nonatomic, strong) NSString *searchString;
 
+@property (nonatomic, strong) SPNoteEditorViewControllerSwiftCollaborators *swiftCollaborators;
```
jleandroperez, line 61: I'm not 100% sure about this pattern, because of a few reasons:

1. The name is a bit confusing. It's `swiftCollaborators` but it actually is meant to retain the HistoryViewController
2. The purpose of keeping this reference is to determine if the HistoryViewController is being presented
3. HistoryLoader is used to process the versions, received by the AppDelegate, passed along to the Editor, and then forwarded to the HistoryLoader, which, in turn, is retained by `(..)SwiftCollaborators`, to be consumed by HistoryViewController.

Do you thing this can be simplified a bit?

Simplenote/Classes/SPNoteHistoryViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+    func styleDateLabel() {
+        dateLabel.textColor = .simplenoteNoteHeadlineColor
+    }
+
+    func styleErrorMessageLabel() {
+        errorMessageLabel.textColor = .simplenoteTextColor
+    }
+
+    func styleSlider() {
+        let color = UIColor.simplenoteGray50Color.withAlphaComponent(0.2)
```
jleandroperez, line 62: We should probably move this to a UIColor property (or drop the 0.2 inline!)

Simplenote/SPNoteHistoryController.swift
```diff
@@ -0,0 +1,140 @@
...
+    ///
+    var observer: ((State) -> Void)? {
+        didSet {
+            observer?(state)
+        }
+    }
+
+    /// Delegate
+    ///
+    var delegate: ((Event) -> Void)?
```
jleandroperez, line 57: May I suggest replacing the closure / enum pattern, with an old school `protocol`?

Reason is just avoiding the switch in the handling closure. Since handlers for each one of the Event cases are mandatory, the code gets much simpler, and improves readability as well.

Simplenote/SPHistoryLoader.swift
```diff
@@ -0,0 +1,109 @@
...
+// MARK: - SPHistoryLoader: Request and aggregate versions of Simperium object
+//
+final class SPHistoryLoader {
+
+    // MARK: - Item: Represents a version object
+    //
+    struct Item {
+        let version: Int
+        let data: [String: Any]
+    }
```
jleandroperez, line 12: Suggestion: Both `SPHistoryLoader.Item` and `SPNoteHistoryController.Presentable` end up encapsulating the same entity: a Note Version.

How about having a single non nested entity, and parsing the fields you need?
That way you can avoid parsing the data collection in multiple spots

Simplenote/Classes/SPNoteHistoryViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+    func setActivityIndicatorVisible(_ isVisible: Bool) {
+        if isVisible {
+            activityIndicator.startAnimating()
+        } else {
+            activityIndicator.stopAnimating()
+        }
+    }
+
+    func setErrorMessageVisible(_ isVisible: Bool) {
+        errorMessageLabel.alpha = isVisible ? UIKitConstants.alpha1_0 : UIKitConstants.alpha0_0
```
jleandroperez, line 152: (Unless you're planning on animating this one), you can simplify things up by:

```
errorMessageLabel.isHidden = !isVisible
```

Simplenote/SPHistoryLoader.swift
```diff
@@ -0,0 +1,109 @@
...
+    ///     - version: version of an object
+    ///
+    func process(data: [String: Any], forVersion version: Int) {
+        self.data[version] = data
+        checkIfFinished()
+    }
+
+    private func checkIfFinished() {
+        guard data.count == amountOfVersionsToLoad else {
+            return
```
jleandroperez, line 74: Question: If, for whatever reasons, not all of the requested versions make it back (networking, new document?), would this API get stuck?

Simplenote/Classes/SPCardViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+            UIView.animate(withDuration: UIKitConstants.animationShortDuration,
+                           animations: animations,
+                           completion: animationCompletion)
+        } else {
+            animationCompletion(true)
+        }
+    }
+
+    private func setupConstraints(withContainerView containerView: UIView) {
+        let bottomToBottomConstraint = view.bottomAnchor.constraint(equalTo: containerView.bottomAnchor, constant: 0)
```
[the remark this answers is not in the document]
eshurakov: Nice, will do 👍

Simplenote/Classes/SPCardViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+
+        view.translatesAutoresizingMaskIntoConstraints = false
+        let containerView = viewController.view!
+        containerView.addSubview(view)
+
+        setupConstraints(withContainerView: containerView)
+
+        containerView.layoutIfNeeded()
+
+        let animations = { () -> Void in
```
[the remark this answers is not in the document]
eshurakov: Indeed, thanks 👍 
Funny thing is that compiler doesn't always correctly infer return type in case it's a one-liner.
For example in the following code compiler thinks that the signature is `() -> ()?` because of the optional `snapshot` variable...
```
let animations = {
    snapshot?.alpha = 0.0
}
```

Simplenote/Classes/SPCardViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+            self.bottomToBottomConstraint?.isActive = false
+            self.topToBottomConstraint?.isActive = true
+            self.view.superview?.layoutIfNeeded()
+        }
+
+        let animationCompletion: (Bool) -> Void = { _ in
+            self.view.removeFromSuperview()
+            self.removeFromParent()
+            completion?()
```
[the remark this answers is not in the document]
eshurakov: As per Apple's doc `removeFromParent` is calling internally `didMove(toParent: nil)`
Similar when adding a view controller `willMove` will be called during `addChild` call.
https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html

Simplenote/Classes/SPCardViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+//
+extension SPCardViewController {
+
+    /// Presents card from another view controller animated
+    ///
+    func present(from viewController: UIViewController) {
+        viewController.addChild(self)
+
+        view.translatesAutoresizingMaskIntoConstraints = false
+        let containerView = viewController.view!
```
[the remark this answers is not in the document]
eshurakov: Sounds good 👍

Simplenote/Classes/SPNoteHistoryViewController.swift
eshurakov: That's taken care of in the next PR (#796), where I have extracted this to a UIColor property :-)

Simplenote/SPHistoryLoader.swift
```diff
@@ -0,0 +1,109 @@
...
+                  simperiumKey: note.simperiumKey,
+                  currentVersion: version)
+    }
+}
+
+// MARK: - Constants
+//
+private extension SPHistoryLoader {
+    struct Constants {
+        static let maxNumberOfVersions = 30
```
dmsnell, line 107: is this new or copied from the old code?
30 is way too few. we keep up to 60 revisions of a note in the "primary archive" and provide more in the "long-term archive"

at least 60 would be nice

Simplenote/Classes/SPNoteHistoryViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+
+            errorMessageLabel.text = text
+        }
+    }
+
+    func update(withSliderValue value: Float) {
+        let index = Int(value)
+        let item = items[index]
+
+        dateLabel.text = item.date
```
dmsnell, line 109: At least in the Electron app we've had a lot of confusion relating to the display of the date for revisions. While we're in here we might want to think about clearing some of that up. cc: @SylvesterWilmott 

In the work I've been doing, for example, I show something like "Originally created: June 30, 2020 8:31 PM" to show that the revisions don't follow a linear ordering in time so much as a linear sequence of edits which may have happened out-of-order. This helps too when sliding back over restored versions, showing that they might have been a restore from a much earlier version of the note.

Simplenote/SPNoteHistoryController.swift
eshurakov: I think both ways have their benefits. 
Some thoughts:
* With closure it is possible to have a private implementation of delegate methods. When a class implements a protocol all methods are exposed to the "outside", which clutters the code completion a bit.
* Traditionally names of the functions in a protocol are quite verbose and long, including contextual information, like  name and instance of the object. For example `func noteHistoryControllerDidCancel(_ controller: SPNoteHistoryController)`. With closure contextual information is often inferred and names can be more concise.

I changed the current implementation from the closure to a protocol to follow the project 👍

Simplenote/SPHistoryLoader.swift
eshurakov: Yes, it was in the old code: https://github.com/Automattic/simplenote-ios/blob/fbc0f800b1c0cad99374671ec93eb50af063c309/Simplenote/Classes/SPNoteEditorViewController.m#L1967

My concern here is that we switched from a scrollable list to a slider and fitting that many versions may result in a very small distance between them on a slider. This may cause difficulties while trying to pick one specific version.

Simplenote/SPHistoryLoader.swift
eshurakov: I think it may stuck, but...
Now we do:
* show history only if current version of the note > 1, which should mean that it's not a new document anymore
* in case simperium authenticator is not connected when we're about to fetch versions, error message is shown and history loader is not invoked
* I see that web socket reconnects automatically on errors and should resume fetching versions. So in case of some brief network error things should get resolved
* User can always dismiss the history view (in case loading takes to long or stuck for some reason)

Maybe we can:
* Set a timeout and show a message if loading takes too long
* Listen to reachability notifications and update UI in some way to reflect possible connectivity issues

What do you think?

Simplenote/SPHistoryLoader.swift
eshurakov: I agree about having an object with strongly typed fields instead of just having a dictionary. I can definitely use `Codable` here.

But I think it is beneficial to have different entities. Even though they're both encapsulating a note version, the purpose and the fields are different: `SPHistoryLoader.Item` represents data that is received from Simperium and `SPNoteHistoryController.Presentable` is created from `SPHistoryLoader.Item` with only fields that are required to render the UI. For example, they both have `date` field, but `SPNoteHistoryController.Presentable` store it already formatted according to UI needs.

In other words, If I would create a framework for Simplenote to separate data management from UI, `SPHistoryLoader` would go to the framework, but `SPNoteHistoryController` would stay in the app as it is part of UI and transforms framework data to the format suitable for the user.

eshurakov: I've removed the dictionary from `SPHistoryLoader.Item` in https://github.com/Automattic/simplenote-ios/pull/795/commits/708dab0155e1c014b087ccf78f62b6568248e4f5 and I think it became better now 👍

Simplenote/Classes/SPNoteHistoryViewController.swift
eshurakov: Interesting! 
On iOS `modificationDate` is shown when user picks a version and they seem to be in linear order.
If I restore a note to some previous version and open history again, the last entry will have a current date. I also noticed that `creationDate` for each version is the same...

Something worth investigating?

Simplenote/Classes/SPNoteEditorViewController.h
eshurakov: It's not ideal, but what I like in this pattern is that only `SPNoteEditorViewControllerSwiftCollaborators` has to be declared as `@objc` and all fields inside it can be pure swift. It's not only for history, basically any object that counterpart swift extension wants to use, can go there. Classic way here is to put all those objects as a separate properties in the header file `SPNoteEditorViewController.h`.
I can remove `SPNoteEditorViewControllerSwiftCollaborators` and go for a more classic approach 👍 

> The purpose of keeping this reference is to determine if the HistoryViewController is being presented
But that's not a bad approach. I think keeping a weak var to the view controller to determine if it's still alive and on screen is better than having a separate Boolean variable, for example.
The same reference is also used to dismiss `HistoryViewController `

> HistoryLoader is used to process the versions, received by the AppDelegate, passed along to the Editor, and then forwarded to the HistoryLoader, which, in turn, is retained by (..)SwiftCollaborators, to be consumed by HistoryViewController.
With `HistoryLoader` I had to compromise as I didn't want to introduce many changes at once. `SPHistoryLoader` asks bucket to load versions, but can't directly listen for events coming from the bucket. `AppDelegate` receives all the events from Simperium. I went with the weak var to the history loader in the editor because editor was already receiving versions from the app delegate.

Some object has to reference `SPHistoryLoader` to provide versions to it. It can be AppDelegate, but I'm not sure it's better :-) I also don't consider using Notification Center as a way to notify about new versions as I think it makes it too loosely coupled. 

What do you think, @jleandroperez? Maybe you have some ideas how to make it better?

eshurakov: I have an idea on how to improve it.
The issue is that we have to forward versions from the AppDelegate to the HistoryLoader and bucket can have only one delegate, which is AppDelegate now (I think technically it's not a delegate, but rather an observer, getting notified about events that happened in a bucket).
What if we make a wrapper around SPBucket which would allow to set several observers. This wrapper can be injected into Editor and HistoryLoader, which would enable them to directly get updates without the need to proxy things from the AppDelegate.

Simplenote/Classes/SPCardViewController.swift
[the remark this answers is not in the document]
jleandroperez: Ahhhhh TIL! 

Right! Single statement closures are evaluated, and the compiler infers you're actually returning the value.

Simplenote/Classes/SPCardViewController.swift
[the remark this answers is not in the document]
jleandroperez: TIL Thank you!

Simplenote/Classes/SPNoteEditorViewController.h
jleandroperez: @eshurakov 😃  That's exactly what we've got on [the mac here](https://github.com/Automattic/simplenote-macos/blob/develop/Simplenote/VersionsController.swift)

Simplenote/Classes/SPNoteHistoryViewController.swift
jleandroperez: Thank you!

Simplenote/SPHistoryLoader.swift
jleandroperez: (For the record!) replying in a comment, to "unthread" things up a bit!

Simplenote/SPNoteEditorViewController+Extensions.swift
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            let completion: (Bool) -> Void = { _ in
+                snapshot?.removeFromSuperview()
+            }
+
+            UIView.animate(withDuration: UIKitConstants.animationShortDuration,
+                           animations: animations,
+                           completion: completion)
+        }
```
jleandroperez, line 102: Do you think this API could be decoupled a bit? Current structure is:

```
if A {
   // Do 1.1
}

// Do 2

// if A {
   // Do 1.2
}
```

Simplenote/SPNoteHistoryController.swift
```diff
@@ -0,0 +1,120 @@
...
+    func selectVersion(atIndex index: Int) {
+        let item = historyItems[index]
+        delegate?.noteHistoryControllerDidSelectVersion(with: item.content)
+    }
+
+    /// Invoked when view is loaded
+    ///
+    func onViewLoad() {
+        guard SPAppDelegate.shared().simperium.authenticator.connected else {
+            state = .error(NSLocalizedString("version-alert-message", comment: "Error alert message shown when trying to view history of a note without an internet connection"))
```
jleandroperez, line 97: Totally fine to just add the default / fallback string here.
Localizations are all handled via GlotPress (on the web), and downloaded by the release scripts.

Simplenote/SPHistoryLoader.swift
jleandroperez: I do appreciate separation of concerns, **but** (for the time being, at least) splitting into a Framework is not on the pipeline. 

Keeping a single entity would be beneficial because:

1. It'd be easier to read (`Item` / `Presentable` are super generic, though namespaced)
2. The codebase would shrink (no need to map Item > Presentable, less code overall)
3. The `Int(note.version() ?? "1") ?? 1` snippet could go in a single spot
4. And both, Item and Presentable represent the same entity

What do you think?

Simplenote/Classes/SPNoteEditorViewController.h
eshurakov: 👍 I will make changes

Simplenote/SPHistoryLoader.swift
eshurakov: I didn't mean that there is a need for a framework, but I think that it's nice to separate things.

> It'd be easier to read (Item / Presentable are super generic, though namespaced)

I can move `Item` out of `SPHistoryLoader`, but I think `Presentable` is best being namespaced as it's only purpose it to convey information from history controller to history view controller and won't be used by any other code.

> The codebase would shrink (no need to map Item > Presentable, less code overall)

I don't believe it will as there is still need to map history Item to UI. In case of using one entity, the code to format date and check if note's version is current will move to the view controller.

> The Int(note.version() ?? "1") ?? 1 snippet could go in a single spot

This actually would be nice to have in an extension of Note. But either way we would still need to access it from both loader (to check available versions) and in a view controller (to check if selected version is current)

> And both, Item and Presentable represent the same entity

I would say it's a different representation of the same entity. One is how it is handled by Simperium API and the other one is a representation for the UI.

I think the difference is not that big and by not using `Presentable` we shift code from controller to view controller, but is it a good thing?

Simplenote/SPNoteHistoryController.swift
eshurakov: This string was already used in the project, so I thought to reuse it without introducing a new one

Simplenote/SPNoteEditorViewController+Extensions.swift
eshurakov: Will see what I can do 👍

Simplenote/SPHistoryLoader.swift
dmsnell: It's not essential to change, though I would say we already have enough confusion because of the fact that it hides versions. In totality there are up to 160 revisions for a note and I'm not sure why we'd intentionally hide them, unless we unintentionally did it because we were unaware.

Just pointing it out here since we're redoing the history loader overall. Might have some solution to each the selection, such as next/previous buttons…

Simplenote/Classes/SPNoteHistoryViewController.swift
dmsnell: > they seem to be in linear order.

This is a source of confusion because _they aren't_ 😉 

The only way to linearize the versions is to list them in the order that the server accepted them. This is actually independent from the `modificationDate` and `creationDate`.

`creationDate` should never change because it's an intrinsic property of the note throughout its lifetime.

`modificationDate` is something each device sets and should do that whenever edits are made on that device. They can be out of order in the history. For example, open a note on device A and on device B. With device B offline make an edit to the note. Then, _after_ that make an edit to the same note on device A. Device A's change will hit the server first. Bring device B back online and its change will hit the server second but the `modificationDate` timestamps will now be backwards.

I may not have realized that iOS is changing the modification date when restoring but personally I like leaving it in tact _so that_ we can show "this version of the note history was a revert from some previous point in time and this is that point in time."

It's all confusing; acting like it's a chain of edits in `modificationDate` time order though is one of the most-confusing ways to represent it.

eshurakov: Thanks for the explanation! 
Behavior is quite confusing indeed, especially if different platforms handle it slightly different 🤔

eshurakov:
Hi @jleandroperez

I have removed `SPNoteEditorViewControllerSwiftCollaborators` and `Presentable`. Also extracted `SPHistoryLoader.Item` into top level object.

I will take care of bugs and stuck spinner in a follow up pull requests.

Please take a look at the changes! Thanks!

Simplenote/SPManagedObject+Version.swift
```diff
@@ -0,0 +1,14 @@
...
+
+// MARK: - SPManagedObject extension
+//
+extension SPManagedObject {
+
+    /// Version as an Int
+    ///
+    @objc
+    var versionInt: Int {
+        return Int(version() ?? "1") ?? 1
```
jleandroperez, line 12: There's a bit of duplication here. (`"1"` / `1`) How about:

```
guard let version = version() else {
    return 1
}

return Int(version)
```

(It's a bit longer, but no duplication at all!)

Simplenote/SPNoteEditorViewController+Extensions.swift
```diff
@@ -0,0 +1,98 @@
...
+        let controller = SPNoteHistoryController(note: currentNote, loader: loader)
+        let historyViewController = SPNoteHistoryViewController(controller: controller)
+        let cardViewController = SPCardViewController(viewController: historyViewController)
+
+        controller.delegate = self
+
+        historyLoader = loader
+        historyCardViewController = cardViewController
+
+        return cardViewController
```
jleandroperez, line 33: This API is a bit coupled: 

- Returns a `SPCardViewController` instance
- But also initializes `historyLoader` and `historyCardViewController` ivars

And those actions aren't explicited in the method signature. Do you think this could be cleared up a bit?

Simplenote/SPNoteEditorViewController+Extensions.swift
jleandroperez: Thank you!

Simplenote/Classes/SPCardViewController.swift
```diff
@@ -35,6 +38,85 @@ final class SPCardViewController: UIViewController {
...
+
+    /// Presents card from another view controller animated
+    ///
+    func present(from viewController: UIViewController) {
+        viewController.addChild(self)
+
+        view.translatesAutoresizingMaskIntoConstraints = false
+        guard let containerView = viewController.view else {
+            fatalError("Can't load view")
+        }
```
jleandroperez, line 53: Nipticky: since this whole method depends on the guard being successful, perhaps that should be the very first thing we run?

Simplenote/SPHistoryLoader.swift
```diff
@@ -0,0 +1,111 @@
...
+// MARK: - Convenience init
+//
+extension SPHistoryLoader {
+    /// Convenience init
+    ///
+    /// - Parameters:
+    ///     - note: a note
+    ///
+    convenience init(note: Note) {
+        let bucket = SPAppDelegate.shared().simperium.bucket(forName: Note.classNameWithoutNamespaces)!
```
jleandroperez, line 85: `note.bucket` should be the same!

Simplenote/SPHistoryLoader.swift
```diff
@@ -0,0 +1,111 @@
...
+private extension SPHistoryLoader {
+    struct Constants {
+        static let maxNumberOfVersions = 30
+    }
+}
+
+// MARK: - SPHistoryVersion: init with raw data
+//
+private extension SPHistoryVersion {
+    init(version: Int, data: [String: Any]) {
```
jleandroperez, line 104: Since the purpose of `SPHistoryVersion` is to only be used by `SPHistoryLoader` (and parse this payload), perhaps move this to `SPHistoryVersion.swift` ?

Simplenote/SPManagedObject+Version.swift
eshurakov: Ended up like this, cause initializer of `Int` from `String` is failable (in case `String` doesn't contain a number)
```
guard let versionStr = version(), let version = Int(versionStr) else {
    return 1
}

return version
```

Simplenote/Classes/SPCardViewController.swift
eshurakov: Right now we first add view controller as a child and then check / add view. 
If we move guard to the top, we would first access view (which triggers viewDidLoad) and only after that add view controller as a child.
My concern here that if for some reason we will want to access `parent` view controller from viewDidLoad, it won't work in the second case.

Simplenote/SPHistoryLoader.swift
eshurakov: Cool, didn't know that, thanks 👍

Simplenote/SPNoteEditorViewController+Extensions.swift
eshurakov: Couldn't think of a good way to modify existing `newHistoryViewController`, but moved the code to `showHistory`, as that's the method to setup and configure everything related to history.

PR #796, opened 2020-07-07
History Slider - VoiceOver and UI improvements

### LINKED ISSUE #670: Link highlighting is gone after viewing History
### Expected

I expect links in the editor to be highlighted in blue.

### Observed

After viewing the History for a note, the blue highlighting is gone for links and they appear with a regular text color.

### Reproduced

1. Create or open a note with at least one link.
2. In the editor, note that the link text is highlighted in blue.
3. Tap the (i) button in the top right and select "History..."
4. Notice that the blue highlighting is gone and the link has a regular text color.

Screencast here: https://cloudup.com/cSRbXcFcXAQ

-|-|-|-
Apple|iPad Mini|iPadOS 13.3|4.16.0.0 TestFlight beta

eshurakov:
### Details
This PR is a follow up to #813. It adds small fixes and improvements to the history slider component:
* Dismiss button now looks according to the mockup
* If note was opened from search, presenting history slider would clear text highlights in the editor
* VoiceOver support
* Theme change notification support (< iOS 13)
* Use readable content guide to limit size of the card for regular horizontal size class

Original issue #762

### Test
**With an opened history slider**
- [ ] Check that all elements are accessible with VoiceOver

**With note opened from search**
- [ ] Check that opening history slider clears text highlights in the editor


Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -1899,6 +1900,10 @@ -(void)togglePinStatusAction:(id)sender
 }
 
 - (void)viewVersionAction:(id)sender {
+    if (bSearching) {
+        [self endSearching:_noteEditorTextView];
+    }
```
jleandroperez, line 1905: Annoying comment: worth moving to a method? dunno, `ensureSearchIsDismissed` or something like that?

Simplenote/Classes/UIColor+Studio.swift
```diff
@@ -7,23 +7,34 @@ extension UIColor {
...
     }
 
     /// Initializes a new UIColor instance with a given ColorStudio Dark / Light set.
     /// Note: in `iOS <13` this method will always return a UIColor matching the `Current` Interface mode
     ///
-    convenience init(lightColor: ColorStudio, darkColor: ColorStudio) {
+    convenience init(lightColor: ColorStudio,
+                     darkColor: ColorStudio,
+                     lightColorAlpha: CGFloat = UIKitConstants.alpha1_0,
+                     darkColorAlpha: CGFloat = UIKitConstants.alpha1_0) {
```
jleandroperez, line 20: Nice!!

Simplenote/Classes/UIColor+Studio.swift
```diff
@@ -7,23 +7,34 @@ extension UIColor {
...
+        let colorProvider: (_ isDark: Bool) -> (value: ColorStudio, alpha: CGFloat) = { isDark in
+            if isDark {
+                return (darkColor, darkColorAlpha)
+            } else {
```
jleandroperez, line 24: No need for `else` (since you'll always return in the first branch!)

Simplenote/Classes/SPNoteEditorViewController.m
eshurakov: Good point! Indeed, it will make it more clear 👍

eshurakov:
Thank you @jleandroperez for the feedback! I've made suggested changes.

PR #803, opened 2020-07-13
History Slider - Show loading progress for each version

eshurakov:
### Details
This PR is a follow up to #796 and changes the loading behavior of History Slider.
History Slider is now showing loading progress for a selected version and doesn't wait for all versions to be loaded before showing main interface. After selected version is successfully loaded, restore button becomes enabled and editor is updated with related content.

Original issue #762

### Test
1. Using slow network, open Note History
2. As you pick different versions, observe progress indicator as versions are being loaded


jleandroperez (approved):
Thank you Evgeny!

Merging this one, we'll deal with any pendings in the master PR. 🔥

PR #804, opened 2020-07-13
Improve side menu accessibility via VoiceOver

eshurakov:
### Fix
Currently text fields inside cells act as an accessibility elements and provide accessibility label. This makes it impossible to open linked screens via double tap as the action is connected to the cell.

This PR sets accessibility label on for a cell based on a text field value and prevents text field to become an accessibility element. Now cell acts as an accessibility element, can be selected and linked screen can be opened via double tap.

### Test
1. Turn VoiceOver on
2. Open side menu and try to open linked screen and edit tags

### DIFF
--- a/Simplenote/Classes/SPTagsListViewController.m
+++ b/Simplenote/Classes/SPTagsListViewController.m
@@ -370,9 +370,6 @@ static const NSInteger SPTagListEmptyStateSectionCount  = 1;
             [self configureBottomCell:cell atIndexPath:indexPath];
             break;
     }
-
-    cell.accessibilityLabel = cell.textField.text;
-    cell.textField.isAccessibilityElement = NO;
 }
 
 - (void)configureSystemCell:(SPTagListViewCell *)cell atIndexPath:(NSIndexPath *)indexPath {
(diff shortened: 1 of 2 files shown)


Simplenote/Classes/SPTagsListViewController.m
```diff
@@ -370,6 +370,9 @@ - (void)configureCell:(SPTagListViewCell *)cell atIndexPath:(NSIndexPath *)index
             [self configureBottomCell:cell atIndexPath:indexPath];
             break;
     }
+
+    cell.accessibilityLabel = cell.textField.text;
+    cell.textField.isAccessibilityElement = NO;
```
jleandroperez, line 375: Since this would essentially apply to every single instance, do you think it'd be best to make the cell perform these calls?

eshurakov: Good point. Also this way can directly return `textField.text` from an `accessibilityLabel` accessor. 👍 
Thanks!

jleandroperez:
@mokagio Gio!! is it possible to skip / bypass the CircleCI check? (since Circle doesn't run in forks!)

(Thank you!)

mokagio:
🤔 I'm sure I marked my PR as draft, not this one... 🤔 sorry if this resulted in spam in your inbox.

mokagio:
@jleandroperez because of #812 running on the same `HEAD` as this, you can now merge 👍 Well done @eshurakov 👏

jleandroperez:
Thank you @mokagio !!

PR #806, opened 2020-07-13
Editor: Autolayout FTW

### LINKED ISSUE #64: iPad: Split Keyboard Weirdness
A user reported this UI glitch, on iPad devices, with Split Keyboard.

Notice that the word `Tomatoes` gets vertically cut.

img_0185

Opened by @jleandroperez

### LINKED ISSUE #396: iPadOS 13: Changing keyboard size temporarily clears editor
On iPadOS 13 Beta 5, we are seeing the following when changing the keyboard size while in the editor:

ios13b5_keyboard

### To repro:
1. Open a note and start editing (using the soft key keyboard only)
2. Pinch the keyboard to make it smaller
3. Drag the keyboard around the screen (notice how the note text clears)
4. Expand the keyboard to full size
5. Notice how the text comes back

This issue is unverified on iPadOS 13 Beta 6...we should test it on there as well before digging in!

### LINKED ISSUE #658: Tags View not relocated after dismissing Search
### Details
- iPhone 11 Pro (iOS 13.3)
- The bottom tags bar would remain "high" after dismissing the search footer

[Originally reported here](https://github.com/Automattic/simplenote-ios/pull/605#issuecomment-581549280) by @bummytime

tags_iOS13_iPhone11Pro

### LINKED ISSUE #769: Bug: Note editor doesn't take into account bottom safe area insets when opened for the first time
### Expected
Expected note editor to respect bottom safe area insets and have some space between bottom edge of my device and text or tags view.

### Observed
Opening a note after cold start of the app shows text or tags behind the home indicator. Which means that safe area wasn't taken into account. 
On subsequent opening of any of the notes or after editing an opened one the issue fixes itself.

On a screenshot red is a frame of the editor, green is a frame of the tags view.
Simulator Screen Shot - iPhone Xs - 2020-06-04 at 23 18 43

-|-|-|-
iPhone|Xs|13.5|4.19

jleandroperez:
### Description:
In this PR we're implementing Autolayout in the Note Editor. **Finally!!**

Closes #64
Closes #396
Closes #658
Closes #769

### Details:
- **KeyboardObserver** was updated to handle **frame notifications* rather than the old school willShow / willHide 
- **Nuked all of the manual layout code** in the EditorViewController
- Wired the new KeyboardObserver in the Editor.

@aerych @danielebogo Gentlemen, may I bug the both of you with this one?

I promise it's way simpler than it looks like. Not so much code, but loooots of flows!!

**Thank you!!!**

---

### Test: Editor :: @aerych 👇👇👇 (Thank you!!!)
- [x] Verify that whenever you tap over the note editor bottom insets are adjusted: you should be able to scroll the whole document, while the keyboard is onScreen.
- [x] Verify that scrolling over a large document causes the navigation bar to shrink
- [x] Verify that the Tags View is fixed at the bottom when VoiceOver is enabled

### Test: Voiceover :: @aerych 👇👇👇 (Thank you!!!)
0. Launch Simplenote on a device with **Safe Area** (iPhone X / 11)
1. Add a note with (lots) of text, so that it causes scroll
2. Enable voiceover (`hey siri voiceover on`!)

- [x] Verify that the Tags View now shows up locked at the bottom
- [x] Verify that scrolling upwards / downwards does not cause the Tags Editor to move
- [x] Verify that whenever the keyboard is dismissed, and you're at the top of the document, no text shows up between the TagsEditor and the bottom edge (we're actually covering that up!!)

### Test: iPad :: @danielebogo  👇👇👇 (Thank you!!!)
0. Launch Simplenote on an iPad Device
1. Toggle a Split Keyboard 

- [x] Verify that text shows up **all** over, and does not clip above the Keyboard anymore
- [x] Verify that switching to **floating keyboard**  and moving it all over does not cause the Text to disappear

### Test: Search :: @danielebogo  👇👇👇 (Thank you!!!)
0. Launch the app on an iPhone device
1. Enter a search keyword that yields any results
2. Press over any search result
3. Scroll all the way down, until the *Tags Editor* is revealed
4. Press `Done` to dismiss

### Test: Extension :: @danielebogo  👇👇👇 (Thank you!!!)
0. Launch Simplenote
1. Switch to Safari
2. Trigger Simplenote's Share Sheet

- [x] Verify that the Share Extension properly updates its sheet size when the keyboard is triggered

### Release
`RELEASE-NOTES.txt` was updated in ac69bee with:
 
> Fixed several layout issues that affected the Editor


aerych:
Spotted a few odd bits while testing. Not sure if they are all related to the changes here but wanted to share them all just in case.

### First
The editor can be a little jumpy when switching between the note editor and the tag editor.
Screen Recording 2020-07-15 at 3 23 02 PM 2020-07-15 15_31_43

### Second
The position of the tags can be a little jumpy when gaining or resigning first responder.
Screen Recording 2020-07-15 at 3 25 12 PM 2020-07-15 15_33_00

### Third
Try this: On an iPad, open a very long note that scrolls in both landscape and portrait orientation. Scroll to the bottom and tap into the Tag Editor like you're going to add a new tag.  Now rotate the device.  The tag editor has scrolled away.  I'd expect it to remain visible with the content pinned to the bottom.  Try this on an iPhone with a long list of tags so that the tag bar scrolls horizontally.  The new tag you're editing will have scrolled off the side of the screen. 

### Fourth
If I've scrolled down a long post half way and left my cursor there, then scroll to the bottom and tap to edit a tag, when I dismiss the keyboard rather than keeping its scroll offset the editor scrolls back to where my cursor previously was.  This seems similar to the first example but I'm not 100% sure it's the same thing.
Screen Recording 2020-07-15 at 3 38 42 PM 2020-07-15 15_42_09

### Fifth
While testing voice over mode I found a scenario where it was kind of hard to read tag suggestions when the suggestion control overlaps a lot of text. 
IMG_1582

Simplenote/Classes/SPEditorTextView.m
```diff
@@ -70,7 +69,11 @@ - (instancetype)init
                forKeyPath:@"contentOffset"
                   options:NSKeyValueObservingOptionNew
                   context:NULL];
-        
+        [self addObserver:self
+               forKeyPath:@"contentInset"
```
aerych, line 73: Maybe constants for some of these strings that are being used in multiple places?

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -322,11 +280,12 @@ - (void)highlightSearchResultsIfNeeded
...
             NSString *searchDetailFormat = count == 1 ? NSLocalizedString(@"%d Result", @"Number of found search results") : NSLocalizedString(@"%d Results", @"Number of found search results");
             self->searchDetailLabel.text = [NSString stringWithFormat:searchDetailFormat, count];
-            
+            self->searchDetailLabel.alpha = UIKitConstants.alpha0_0;
+
             [UIView animateWithDuration:0.3
                              animations:^{
                                  
-                                 self->searchDetailLabel.alpha = 1.0;
+                                 self->searchDetailLabel.alpha = UIKitConstants.alpha1_0;
```
danielebogo, line 288: Did you mean to access its ivar here? If not can you use dot notation instead?

Simplenote/Classes/SPNoteEditorViewController.h
```diff
@@ -66,9 +71,14 @@
 @property (nonatomic, strong) SPTagView *tagView;
 @property (nonatomic, strong) NSString *searchString;
 
+// Voiceover
+@property (nonatomic, strong) UIView *bottomView;
+
+// Keyboard!
+@property (nonatomic, assign) BOOL isKeyboardVisible;
```
danielebogo, line 78: Nitpick, you can omit `assign` here

Simplenote/Classes/SPNoteEditorViewController.m
jleandroperez: This one doesn't have a property (yet), **lots** of refactors on the way. Apologies about the ugly (temporary!!) snippet!!!

jleandroperez:
All of the glitches should be gone:

1. Scroll position when switching from / to the Tags Editor should be the same
2. We're ensuring that the Tags Editor is visible on rotation. (**Note:** vertical scroll ins not guarranteed, may address that elsewhere!)
3. Suggestions should always be readable (no matter what!)
4. "Apply Bottom Insets" animation should be **TERRIBLY** smooth. Not 100% perfect, but perhaps 97% (lots of tricks to get there)
5. "Undocked keyboard" yields no bottom padding at all. Checked with Sly, we might revisit this.

@aerych Mind taking another peek? (This one has been literally a while in the works, LOTS of tiny bugs interconnected!!!)

**Thank you both!!!**

aerych:
Hola @jleandroperez 👋 
Latest code changes look ✨ .   I've retested the glitches I spotted before and they all are either fixed or look soooooooo much better that they aren't worth fussing more over.   Nice work yo!
Ship it with great shipping :shipit:

jleandroperez:
Thanks a million @aerych !! Woooo 🔥

PR #813, opened 2020-07-15
History Slider - Swipe to dismiss

eshurakov:
### Details
This PR adds swipe to dismiss interaction to the History Slider and tinted overlay to prevent interaction with the editor while the slider is on screen.

* History Slider is now presented over the editor using standard view controller presentation mechanism (before if was added to the editor as a child view controller)
* History Slider is presented using custom transition delegate. `SPCardPresentationController` takes care of adding tinted overlay and decorating presented view controller with shadow and rounded corners to make it look like a card. It also handles swipe to dismiss transparently to the presented view controller.
* `SPCardViewController` is nuked 💥 

I believe that #670 is also fixed now.

### Test
**With opened Note History**

- [ ] Swipe to dismiss
- [ ] Pull card a little down and back up. It shouldn't dismiss the card.
- [ ] Check that overlay blocks interaction with the editor


Simplenote/SPNoteEditorViewController+Extensions.swift
```diff
@@ -8,50 +8,58 @@ extension SPNoteEditorViewController {
...
+        let transitioningManager = SPCardTransitioningManager()
+        transitioningManager.observer = self
+        historyTransitioningManager = transitioningManager
+
+        present(viewController, with: transitioningManager)
     }
 
-    private func newHistoryViewController(with loader: SPHistoryLoader, delegate: SPNoteHistoryControllerDelegate) -> SPCardViewController {
+    private func newHistoryViewController(with loader: SPHistoryLoader,
+                                          delegate: SPNoteHistoryControllerDelegate) -> UIViewController {
```
jleandroperez, line 32: Q: Do you think it'd be cleaner to return a `SPNoteHistoryViewController`, and allow the caller to set the delegate?

Why: One less parameter, you get to simply `return SPHistoryViewController(...`, and it's cleaner!

Simplenote/SPNoteEditorViewController+Extensions.swift
```diff
@@ -65,6 +73,27 @@ extension SPNoteEditorViewController: SPNoteHistoryControllerDelegate {
...
+        cleanUpAfterHistoryDismissal()
+        restoreOriginalNoteContent()
+    }
+}
+
+// MARK: - Transitioning
+//
+private extension SPNoteEditorViewController {
+    func present(_ viewController: UIViewController,
+                 with transitioningManager: UIViewControllerTransitioningDelegate) {
```
jleandroperez, line 89: Nipticky: No need to go multiline 😄

Simplenote/SPCardTransitioningManager.swift
```diff
@@ -0,0 +1,35 @@
+import UIKit
+
+// MARK: - SPCardTransitioningManager: Manages card-like presentation of a view controller
+//
+final class SPCardTransitioningManager: NSObject, UIViewControllerTransitioningDelegate {
+    private weak var presentationController: SPCardPresentationController?
+
+    /// Observer for transition related events
+    ///
+    weak var observer: SPCardTransitionObserver?
```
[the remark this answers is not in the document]
eshurakov: Naming is hard! :-) 
Let's call it `SPCardPresentationControllerDelegate`. I think it fits, thanks!

Simplenote/SPNoteEditorViewController+Extensions.swift
eshurakov: Delegate is actually for the controller (not view controller) and is created inside this method. 
But isn't it nice to return a plain UIViewController and hide the actual creation process as the caller doesn't really care what class it is as long as it's UIViewController and can be presented :-)

Simplenote/SPNoteEditorViewController+Extensions.swift
```diff
@@ -8,50 +8,58 @@ extension SPNoteEditorViewController {
...
-        historyLoader = nil
-
+        cleanUpAfterHistoryDismissal()
         viewController.dismiss(animated: true, completion: nil)
     }
+
+    private func cleanUpAfterHistoryDismissal() {
+        historyViewController = nil
+        historyLoader = nil
+        historyTransitioningManager = nil
```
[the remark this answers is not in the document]
eshurakov: I first did it without nullifying weak vars and what happened is: `save` method in the note editor is checking if the history slider is on screen. If it's on screen, save is skipped. My guess it's done to prevent accidental save happening while user is browsing through the history. To check if history slider is on screen I'm using `historyViewController` var. If I don't nullify it here and user tries to restore to another version of a note, save won't work as the var will still be there (and will be gone only after view controller is fully dismissed).

PR #816, opened 2020-07-21
Fixes Interactive Scrollbar

### LINKED ISSUE #733: Scrollbar within notes does not work as expected
### Expected
In the Notes List, you can tap onto the scrollbar on the right side and it will get a bit bigger, so that you can move it up and down more easily. I expected this behavior to stay the same within notes.

### Observed
The scrollbar within each note is not working that way.

### Reproduced
1. Go to a note
2. Try to scroll down the same way as the notes list

RPReplay_Final1587060345 2020-04-16 14_14_24

Make & Model|iOS Version|App Version
-|-|-
iPhone X | iOS 13.4 | Simplenote 4.17
iPad Pro 12.9" | iOS 13.4.1 | Simplenote 4.17

jleandroperez:
### Fix
In this PR we're fixing a bug that caused the Scroll Indicator not to accept long press gestures, and thus, made it impossible to move along the document super quickly.

@guarani Paul!! Can I bug you with a super simple PR?
Thanks in advance!!

Closes #733

### Test
1. Open a long document
2. Scroll a tiny bit, so that the Scroll Bar appears
3. Press over the Scroll Indicator / Scroll Bar

- [ ] Verify that after a second, the Indicator increases its size, and accepts drag events
- [ ] Verify that pressing over the Tags Editor works as expected
- [ ] Verify that you can add / delete tags, from the Tags Editor

### Release
`RELEASE-NOTES.txt` was updated in 5b0e92a with:

> Now it's possible to interactively drag the Scroll Bar within the Editor

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 4.22
 -----
 -   Fixed several layout issues that affected the Editor
+-   Now it's possible to interactively drag the Scroll Bar within the Editor
  
 4.21
 -----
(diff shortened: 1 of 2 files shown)


jleandroperez:
Thank you Paul!!

PR #817, opened 2020-07-23
Update Fastlane iOS Simulator version for screenshots to 13.5

mokagio:
This is due to us using macos-latest in the GitHub Actions workflow. It might be better to switch to a targeted macOS image instead.

Also notice we had to update the snapshot helper file. That was prompted by a build failure which I think is related to the new Simulator version only, since we didn't update Fastlane since the last screenshots build.

See https://github.com/Automattic/simplenote-ios/runs/815164553#step:9:1103

### Test

Add the "Needs Screenshots" label and verify the screenshots build starts on GitHub Actions. Then you might want to go take a break or work on something else because the build will take a while to finish.

If everything works, the build should be green with downloadable screenshots attachments, like [here](https://github.com/Automattic/simplenote-ios/pull/817/checks?check_run_id=901217752)

image

### Review

### Release

### DIFF
--- a/SimplenoteScreenshots/SnapshotHelper.swift
+++ b/SimplenoteScreenshots/SnapshotHelper.swift
@@ -38,22 +38,13 @@ func snapshot(_ name: String, timeWaitingForIdle timeout: TimeInterval = 20) {
 }
 
 enum SnapshotError: Error, CustomDebugStringConvertible {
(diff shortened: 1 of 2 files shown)


ali112a (approved):

mokagio:
Closing in favor of #1066

PR #818, opened 2020-07-24
Merge release 4.21.0 into develop

mokagio:
This should be merged only **after** App Store approval and release.


jleandroperez:
Definitely not intentional. Are those strings pulled from GlotPress?

mokagio:
Good question... Yeah I think they come from GlotPress, or at least, I certainly haven't edited them manually 😅

mokagio:
So, I just went back the history and looked around a bit and I think what happened is that I added the 

```
California users privacy notice: https://wp.me/Pe4R-d/#california-consumer-privacy-act-ccpa
```

link _manually_ (😱) to all those metadata files, I'd say because I did't take the time to understand how the process works.

So when I run `bundle exec fastlane finalize_release`, it pulled the data from GlotPress and deleted my manual additions. In fact, on GlotPress there is no CCPA link in the description, see for example [the German page](https://translate.wordpress.com/projects/simplenote/ios/release-notes/de/default/)

Screen Shot 2020-07-27 at 8 46 58 pm

I think it's too late to fix this branch, but I'll make sure it's done properly for 4.22 (which at this point will be one day late, I'm afraid).

PR #825, opened 2020-07-28
Merge 4.22.0-beta1 into develop

mokagio:
Worth pointing out that in this PR I also added the fix for the missing CCPA link in some of the localized descriptions.

I decided to put the fix in the release branch rather than opening a dedicated PR for it first because of time constraints; I'm already one day late on the code freeze.


fastlane/appstoreres/metadata/source/description.txt
```diff
@@ -30,6 +30,7 @@ ORGANIZE AND SEARCH
 
 Privacy Policy: https://automattic.com/privacy/
 Terms of Service: https://simplenote.com/terms/
+California Users Privacy Notice: https://wp.me/Pe4R-d/#california-consumer-privacy-act-ccpa
```
mokagio, line 33: I'm pretty sure this is why in #818 some of the localized CCPA strings were deleted. Because I didn't add the link to this file but to the one in `fastlane/metadata/en-US/`, the change didn't go upstream to GlotPress and when I downloaded the newly localized strings, my manual edits from 0095223f4b8556ac8886a589df95b3ce9b9b0bf7 got overridden.

PR #824, opened 2020-07-27
Tags Edition: Standard Validation!

jleandroperez:
### Details:
Since forever, the Tag Entity has had its **simperiumKey** set to the **URL Encoded ( Tag Name )**.

In this PR we're upgrading the encode invocation. Rather than using a standard `URLEncode` behavior, we'll percent escape anything that's not alphanumeric. This effectively fixes several Simperium-Y issues, caused by invalid **simperiumKey(s)**.

We're also:

1. Ensuring Tags can only have up to 256 characters
2. Simplifying the Tags Edition logic. A lot.
3. And... adding a few Unit Tests!! 🔥🔥

@aerych Sir!! May I bug you with this one?
Thanks in advance!!

Closes #823

### Scenario: Note Editor
1. Checkout commit b4bd0aae80c3180cf431856d315534024b0e77df
2. Press over the top right button to add a new Note
3. Press over the Tags Editor

- [x] Verify that new tags can have at most 5 characters (**this is for testing!** actual limit is 256 char)
- [x] Verify that special characters are allowed (parentheses, comma)
- [x] Verify that pressing a space "commits" the text you've typed

### Scenario: Tags Editor
1. Checkout commit b4bd0aae80c3180cf431856d315534024b0e77df
2. Press over the top left button to reveal the Tags Editor
3. Long press over any tag

- [x] Verify that pressing **Delete** effectively removes the new Tag
- [x] Verify that pressing Rename allows you to edit the Tag Name (**up to 5 characters**, for testing purposes!)
- [x] Verify that during a rename OP pressing spacebar "Commits" the edit

### Release
`RELEASE-NOTES.txt` was updated in 1bfee244aeedc4bf84857febb8fdd2e3a89fd06c with:
 
> Updated Tags limit to up to 256 characters

PR #828, opened 2020-07-29
Tags Normalization

jleandroperez:
## Details:
In this PR we're bringing over several enhancements introduced [in this macOS Issue](https://github.com/Automattic/simplenote-macos/pull/617)

- Encapsulates **Tag Hash** generation in a NSString extension
- Encapsulates **Tag Name** validation in a (new) NSString extension
- Updates Tag Lookup mechanism, so that comparison is based on Hash rather than on Strings. [Details here!](https://github.com/Automattic/simplenote-ios/pull/828/files#diff-502ce6a4227df26c1d3c682811133e70R42)

@aerych Sir!! since you've validated the first part, mind taking a quick peek? (nothing big should have changed!!)
**Thank you!!**

/cc @dmsnell **NOW** we're talkin'! 😃 

Ref. #824

## Scenario: Note Editor
1. Checkout commit cd4f251
2. Press over the top right button to add a new Note
3. Press over the Tags Editor

- [ ] Verify that new tags can have at most 5 characters (**this is for testing!** actual limit is 256 char).
- [x] Verify that special characters are allowed (parentheses, comma).
- [x] Verify that pressing a space "commits" the text you've typed.

> **IMPORTANT:** Length is enforced against **encoded** tag name. Using non alphanumeric characters will yield a much lower limit!

## Scenario: Tags Editor
1. Checkout commit cd4f251
2. Press over the top left button to reveal the Tags Editor
3. Long press over any tag

- [x] Verify that pressing **Delete** effectively removes the new Tag.
- [x] Verify that pressing Rename allows you to edit the Tag Name (**up to 5 characters**, for testing purposes!).
- [x] Verify that during a rename OP pressing spacebar "Commits" the edit.

> **IMPORTANT:** Length is enforced against **encoded** tag name. Using non alphanumeric characters will yield a much lower limit!

## Release

### DIFF
--- a/Simplenote/Classes/NSString+Simplenote.swift
+++ b/Simplenote/Classes/NSString+Simplenote.swift
@@ -19,10 +19,12 @@ extension NSString {
         components(separatedBy: .space).first ?? String(self)
     }
 
-    /// Percent Encodes all of the non alphanumeric characters in the receiver
+    /// Encodes the receiver as a `Tag Hash`
     ///
     @objc
-    var byEncodingNonAlphanumerics: String? {
-        addingPercentEncoding(withAllowedCharacters: .alphanumerics)
+    var byEncodingAsTagHash: String {
+        precomposedStringWithCanonicalMapping
+            .lowercased()
+            .addingPercentEncoding(withAllowedCharacters: .alphanumerics) ?? self as String
     }
 }
--- a/Simplenote/Classes/NSString+Simplenote.swift
+++ b/Simplenote/Classes/NSString+Simplenote.swift
@@ -33,6 +33,6 @@ extension NSString {
     ///              For that reason we must check on the `encoded` lenght (and not the actual raw string length)
     @objc
     var isValidTagName: Bool {
-        byEncodingAsTagHash.count < SimplenoteConstants.maximumTagLength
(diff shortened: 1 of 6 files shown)


Simplenote/Classes/NSString+Simplenote.swift
```diff
@@ -19,10 +19,20 @@ extension NSString {
...
+            .lowercased()
+            .addingPercentEncoding(withAllowedCharacters: .alphanumerics) ?? self as String
+    }
+
+    /// Indicates if the receiver is a valid Tag Name
+    /// - Important: `Tag.name` is used as the entity's `simperiumKey`, and the backend imposes a length.
+    ///              For that reason we must check on the `encoded` lenght (and not the actual raw string length)
+    @objc
+    var isValidTagName: Bool {
+        byEncodingAsTagHash.count < SimplenoteConstants.maximumTagLength
```
aerych, line 36: Should this be `<=` ?

jleandroperez: YES!!! Thank you!!!

jleandroperez:
Thank you sir!!!

PR #833, opened 2020-08-10
Merge 4.22.0 into develop

mokagio:
I'm late on this, sorry. I noticed Simplenote 4.21 was ready for release over the weekend, but when I went to prepare 4.22, I noticed the release notes were missing, see 31380e4d, and I didn't have time to fix it.


loremattei:
I'm not sure what's going on here... how is merging the `release/4.22` branch into `develop` related to 4.21?
Also, the version in `glotPressWhatsNewKey` looks weird... is that change intentional?

mokagio:
Hey @loremattei 

> I'm not sure what's going on here... how is merging the `release/4.22` branch into `develop` related to 4.21?

Because 4.21 was still rejected in App Store Connect on Friday when I started finalizing 4.22, I wasn't able to create the new version for 4.22 there. So I couldn't move forward with the release process, i.e. running `deliver` to upload the new metadata.

I wanted to wait to do that successfully before opening the PR just in case I'd had to do something extra in the release branch. And actually turned out to be the case, which leads to your next question.

> Also, the version in `glotPressWhatsNewKey` looks weird... is that change intentional?

Yeah, it was intentional in 31380e4.

I must have done something wrong during the code freeze, as the script to download the metadata was expecting "4.22" for the release notes key, but on GlotPress the value was "4.22.0".

---

I'm going to merge this PR in order to start with the 4.23 code freeze, even though I don't have an explicit approval from you. I think you'd be okay with that because your comment doesn't raise any blocking issue. Let me know if I've made a bad call on this one.

PR #835, opened 2020-08-10
Nü History Slider

### LINKED ISSUE #762: Spring Cleaning: Nü History Slider

jleandroperez:
### Details:
In this PR we're implementing a brand new History Slider. This is @eshurakov 's work, opening the PR in order to give it a bit of more visibility.

Depends on #803 (and further testing!).

Closes #762

### Test
1. Open a note
2. Open note options
3. Open note history

- [ ] If no internet connection, inline error message is shown
- [x] If current version is selected, restore button is disabled
- [x] On devices with haptic feedback, dragging a slider triggers haptic feedback
- [x] Screen can be rotated to landscape / portrait
- [x] Selecting a version updates note editor with that version (with alpha transition animation)
- [x] Selecting a version and dismissing a history slider restores editor to the current version
- [x] Selecting a version and saving it should persist the changes
- [x] Dismiss the card by swiping down
- [x] Dismiss the card by pressing "X"
- [x] Pull the card a little down and back up. It shouldn't dismiss the card.
- [x] Check that overlay blocks interaction with the editor
- [x] Tap outside of the card to dismiss
- [x] Tap inside of the card, view should not get dismissed
- [x] Check that all elements are accessible with VoiceOver

### Test (from search)
1. Search notes
2. Open one of the note

- [x] Check that opening history clears text highlights in the editor

### Test (slow network)
1. Setup slow network
2. Open a note
3. Open note options
4. Open note history

- [x] As you pick different versions, observe progress indicator as versions are being loaded
- [x] Select one of the versions that wasn't loaded yet. Wait till the version is loaded and check that UI of Version Slider is updated (date, button) and editor content is updated to reflect the selected version.

### Release
`RELEASE-NOTES.txt` was updated in 36046a3 with:
> New Note History interface #762


Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -256,6 +362,45 @@ extension SPNoteEditorViewController {
 }
 
 
+// MARK: - Editor
+//
+private extension SPNoteEditorViewController {
+
+    func updateEditor(with content: String, animated: Bool = true) {
+        let contentUpdateBlock = {
```
jleandroperez, line 397: Might be worth looking into cleaning things up a bit. 

I feel like this API might be adding a bit of code implemented by `newButtonAction` (which is due up for a decoupling refactor!!).

Maybe we could add a `// TODO` (or file another issue) and follow up?

Simplenote/SPAppDelegate.m
```diff
@@ -562,7 +562,7 @@ - (void)bucket:(SPBucket *)bucket didReceiveObjectForKey:(NSString *)key version
 {
     if ([bucket.name isEqualToString:@"Note"]) {
         if ([key isEqualToString:self.noteEditorViewController.currentNote.simperiumKey]) {
-            [self.noteEditorViewController didReceiveVersion:version data:data];
+            [self.noteEditorViewController.historyLoader processData:data forVersion:[version integerValue]];
```
jleandroperez: Problem with this approach is: iPad + Multiple Windows (not supported yet, but it'd be good if we could adjust things up a bit!).

Please take a look at [VersionsController](https://github.com/Automattic/simplenote-macos/blob/develop/Simplenote/VersionsController.swift) in the Mac project, that'd help us convert this into a Publisher / Subscriber (which should work with N editor instances).

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -191,6 +191,98 @@ extension SPNoteEditorViewController {
...
+    @objc
+    var isShowingHistory: Bool {
+        return historyViewController != nil
+    }
+
+    /// Shows note history
+    ///
+    @objc
+    func showHistory() {
+        let loader = SPHistoryLoader(note: currentNote)
```
jleandroperez: (Added a note below as well!) Please take a look at the [VersionsController](https://github.com/Automattic/simplenote-macos/blob/develop/Simplenote/VersionsController.swift) publisher/subscriber mechanism we've got in the Mac App.

That'd help us support multiple Editors out of the box (iPad Multiwindow), and will simplify the History Initialization

jleandroperez (commented):
@eshurakov  Added few notes! and before I forget!!

Can we also get rid of `SPActionSheet` ? (I've noticed there are a couple leftovers in the EditorViewController from the legacy Versions mechanism).

Specifically `actionSheetVisible` + `versionActionSheet` (which may have been gone **before**, and came back by a merge glitch? (if so, apologies about that!!!).

**Plus:** Please feel free to take over the PR (could use few test steps) and Release Notes!

Thaaank you sir!

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -191,6 +191,98 @@ extension SPNoteEditorViewController {
...
+    ///
+    @objc
+    var isShowingHistory: Bool {
+        return historyViewController != nil
+    }
+
+    /// Shows note history
+    ///
+    @objc
+    func showHistory() {
```
jleandroperez: Please take a peek at how `presentOptionsController` or `presentShareController` work. Might be worth shuffling a couple things around (So that we take the `note` as a parameter, and get rid of the IUO usage).

eshurakov: I actually think passing `note` as a parameter is confusing in this scenario. This code is inside the context of `SPNoteEditorViewController` and looking at this code I would assume that it should always work with the current note. If I see that a `note` is passed as a param I think that maybe it is a different note? But why note editor shows one note and presents options / share / history of another note? 🤔

Simplenote/SPAppDelegate.m
eshurakov: We will probably need to adjust VersionController as well to support multiple subscribers for the same simperium entity.

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: Added TODO for now as animations are different 🤔

RELEASE-NOTES.txt
```diff
@@ -1,5 +1,6 @@
 4.27
 -----
+-   New Note History interface #762
```
jleandroperez, line 3: Thank youuuu 😄

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -227,6 +227,96 @@ extension SPNoteEditorViewController {
...
+    @objc
+    func presentHistoryController() {
+        guard SPAppDelegate.shared().simperium.authenticator.connected else {
+            let title = NSLocalizedString("version-alert-message", comment: "Error alert message shown when trying to view history of a note without an internet connection")
+            let cancelTitle = NSLocalizedString("OK", comment: "")
+            let alertController = UIAlertController(title: title,
+                                                    message: nil,
+                                                    preferredStyle: .alert)
+            alertController.addCancelActionWithTitle(cancelTitle)
+            present(alertController, animated: true, completion: nil)
```
jleandroperez: Maybe move the AlertController snippet to another method, and decouple things? (OR was that gonna be embedded in the history?)

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
jleandroperez: Thank you!

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -972,209 +961,6 @@ - (void)insertChecklistAction:(id)sender {
...
-                             snapshot.alpha = 0.0;
-                             
-                         } completion:^(BOOL finished) {
-                             [snapshot removeFromSuperview];
-                         }];
-	}
-    
-}
-
-
```
jleandroperez, line 1177: 🔥

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -682,11 +680,14 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView
...
-    if (self.searching)
-        [self endSearching:textView];
-    
+- (BOOL)textViewShouldBeginEditing:(UITextView *)textView
+{
+    if ([self isShowingHistory]) {
+        return NO;
+    }
+
+    [self ensureSearchIsDismissed];
```
jleandroperez, line 689: Is it possible this `ensureSearchIsDismissed` invocation came thru a bad merge? can we drop it? (since it's a different behavior)

jleandroperez (commented):
@eshurakov  Apologies I cannot reply to this one above, since GH closed the conversation!

```I actually think passing note as a parameter is confusing in this scenario. This code is inside the context of SPNoteEditorViewController and looking at this code I would assume that it should always work with the current note. If I see that a note is passed as a param I think that maybe it is a different note? But why note editor shows one note and presents options / share / history of another note?
```

Only note there is that `currentNote` is an implicitly unwrapped optional. **Ideally**, when we go full swift on this, such property will not be optional nor IUO.

So, thinking aloud!! since we're already aiming there, no need to change anything!!

---

Last note I've got is: **UX is superb!!**

I find myself (by accident) swipping down the History Card by 1mm, while trying to grab the slider. Lifting your finger just gets it back to the right location, so no big deal.

Do you think it's worth it making the Card dismissable only if swiped from the title area?

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: I asked Sylvester about it and we're gonna remove this alert completely and will show an error message directly in the card. We will also adjust a copy to `Couldn't retrieve history`

Simplenote/Classes/SPNoteEditorViewController.m
eshurakov: Dismissing search (bottom bar and highlights) is a current behaviour when starting editing a note while searching. I think the main goal is to remove the highlights and let the user edit text without distractions.
Same code from develop:
https://github.com/Automattic/simplenote-ios/blob/16853f8fde4a20595edba4c0e8e1b32ccf45ac94/Simplenote/Classes/SPNoteEditorViewController.m#L860-L866

I extracted this code to `ensureSearchIsDismissed` method so that we can use it from `showHistoryController` as well.

eshurakov:
> I find myself (by accident) swipping down the History Card by 1mm, while trying to grab the slider. Lifting your finger just gets it back to the right location, so no big deal.
> Do you think it's worth it making the Card dismissable only if swiped from the title area?

That's a very good point! Thanks!
I played around with it and thought that we only need to ignore swipe-to-dismiss if it originates from the thumb of the slider. 
It is implemented in https://github.com/Automattic/simplenote-ios/pull/835/commits/4f8ecefa408ca9a2a78ecf724a18f4e6f2fa3f55

Simplenote/Classes/SPNoteEditorViewController.m
jleandroperez: Woooo I tried this yesterday (AppStore) and I was entirely sure it remained in search mode. I'll take another look, apologies about the noise!!

jleandroperez (commented):
@eshurakov Thank you for updating the recognizer behavior, works perfecty!!

(Please hold merging until #926 makes its way back into develop!)

:shipit: !!!

eshurakov (approved):

Issue #763, opened 2020-05-25
Spring Cleaning: Nü Note Options


jleandroperez:
### Guidelines
- New code should always be Swift
- Autolayout **is always preferred**
- Please drop outdated components when appropriate! 🔥🔥🔥
- [Simplenote Icons Here](https://github.com/Automattic/Simplecons)

### Details:
- [x] Implement a new Note Options component, and replace the legacy UI

### Mockups
<img width="300" src="https://user-images.githubusercontent.com/1195260/95608947-4a88f980-0a34-11eb-8aa1-e59b538bad9c.png">

mattdc28:
Currently, the note options menu is activated by tapping the `i` button. These mocks show the new menu being presented from a new ellipsis button but the `i` button is still visible. 

Do we want to retain the `i` button for displaying things like word count that aren't displayed in the new menu or should it be removed?

jleandroperez:
(1 + 3) Word Count [is actually another issue](https://github.com/Automattic/simplenote-ios/issues/836). 

For the purpose of this particular task, please, feel free to leave the `( i )` button setup as a placeholder. We'll merge back into *develop* only once all of the pieces are in place.

By the way, every icon you might possibly need [is in this repository](https://github.com/Automattic/Simplecons)! (description also updated with the link).

(2) 100% your call, no preferences there! Whatever approach you think produces the best quality code 😃 

(4) Detail there is: not all of those items will push. Will get back to you on this one!

(5) Since the parent ViewController will have to present the History ([Work In Progress here!](https://github.com/Automattic/simplenote-ios/pull/835)), yes, you'd need to relay a message back.

The approach you take for (2) and (5) will probably be related, if you go for consistency.

Thank you @mattdc28 !!

jleandroperez:
**Addendum:**
I've noticed there was a missing mockup, with an alert (indicating if the Publish OP failed). Such component is (allegedly) similar to this [WordPress iOS component](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift), but bringing all of the tooling we've got there might be an overkill.

PR #846, opened 2020-08-15
New note options view UI

mattdc28:
### Details
Adds a new `NoteOptionsViewController` containing the UI as part of the work for #763. This PR only contains the layout code and placeholder methods for handling interaction with the switches and tapped cells. Logic and behaviour will be in a future pull request. Eventually, this new view will replace the `SPActivityView` that currently serves these actions.

The views added are
1. `NoteOptionsViewController`
2. `SwitchTableViewCell`
3. `Value1TableViewCell`

### Screenshots
Simulator Screen Shot - iPhone 11 Pro - 2020-08-15 at 12 40 23
Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2020-08-15 at 12 39 57-2

### Notes
There are nested classes for `Row` and `Section` which are used to render the view and can be easily moved around in the future with minimal effort (For consistency, I took this approach after looking through similar table views in WordPress and Simplenote).

After discussion with @jleandroperez I decided not to subclass `SPTableViewController` and instead set the few styles manually as there are only a few lines.

### Known Issues
Currently, the ellipsis icon is not compliant with the design, which has a circular border around it. I made attempts to add it programmatically but was having layout issues due to how the `SPOutsideTouchView` logic works. I'd be thankful for advice on how to go about the implementation if we want to avoid including a new asset that includes the circle.

Please check that this is visually correct on both iPhone and iPad against the supplied mocks in #763.

### Review
Please review the code and design against #763.

### DIFF
--- a/Simplenote/Classes/NoteOptionsViewController.swift
+++ b/Simplenote/Classes/NoteOptionsViewController.swift
@@ -139,6 +139,7 @@ class NoteOptionsViewController: UITableViewController {
                     cell.textLabel?.text = NSLocalizedString("Copy Link", comment: "Note Options: Copy Link")
                     cell.textLabel?.textColor = .simplenoteGray20Color
                     cell.accessibilityHint = NSLocalizedString("Tap to copy link", comment: "Accessibility hint on cell that copies public URL of note")
+                    cell.isUserInteractionEnabled = false
                 },
                 handler: { [weak self] in
                     self?.handleCopyLink()
--- a/Simplenote/Classes/NoteOptionsViewController.swift
+++ b/Simplenote/Classes/NoteOptionsViewController.swift
@@ -1,11 +1,3 @@
-//
-//  NoteOptionsViewController.swift
-//  Simplenote
-//
-//  Created by Matthew Cheetham on 11/08/2020.
-//  Copyright © 2020 Automattic. All rights reserved.
-//
-
 import UIKit
 
 /// A class used to display options for the note that is currently being edited
--- a/Simplenote/Classes/NoteOptionsViewController.swift
+++ b/Simplenote/Classes/NoteOptionsViewController.swift
@@ -126,7 +126,7 @@ class NoteOptionsViewController: UITableViewController {
             var cellType: UITableViewCell.Type {
                 switch self {
                 case .Value1:
-                    return UITableViewCell.self
+                    return Value1TableViewCell.self
                 case .Switch:
                     return SwitchTableViewCell.self
                 }
--- a/Simplenote/Classes/NoteOptionsViewController.swift
+++ b/Simplenote/Classes/NoteOptionsViewController.swift
@@ -26,7 +26,7 @@ class NoteOptionsViewController: UITableViewController {
     }
 
     //MARKL - View Setup
-    /// Configures a dismiss 
+    /// Configures a dismiss button for the navigation bar
     func setupDoneButton() {
         let doneButton = UIBarButtonItem(title: NSLocalizedString("Done", comment: "Note options: Done"),
                                          style: .done,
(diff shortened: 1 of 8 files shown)


Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -0,0 +1,289 @@
+//
+//  NoteOptionsViewController.swift
+//  Simplenote
+//
+//  Created by Matthew Cheetham on 11/08/2020.
+//  Copyright © 2020 Automattic. All rights reserved.
+//
```
jleandroperez, line 7: Mind dropping all of the headers? (I still need to go thru the whole project, but by convention, all of our files should be header-less).

Thank you!!

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+
+    /// Contains all the data required to render a row
+    fileprivate struct Row {
+        /// Determines what cell is used to render this row
+        let style: Style
+
+        /// Called to set up this cell. You should do any view configuration here and assign targets to elements such as switches.
+        let configuration: ((UITableViewCell, Row) -> Void)?
```
jleandroperez, line 221: Potentially speaking, `UITableViewCell` could be replaced with generics.

**BUT** potentially speaking, you'd end up with type erasure here and there, and **potentially** speaking, we'd end up with a severe headache 😄

mattdc28: Yeah I did briefly start down the generics route then wasn't sure it was the right option for this particular view. I can revisit if you think it's how we want to go about it.

Simplenote/Classes/NoteOptionsViewController.swift
mattdc28: No problem

mattdc28:
@jleandroperez 

> ### Note B:
> 
> Navigation Bar's BG color doesn't match the one in the `Options` ViewController

So it is! The colour difference is subtle but I found the cause. Using `SPNavigationController`, the controller itself switches the background colour of the navigation bar depending on whether the presentation style is `formSheet` or not (which for `.popover` is isn't). It switches between `simplenoteNavigationBarModalBackgroundColor` and `simplenoteNavigationBarBackgroundColor`.

Options I can see:
1. Override the colour in `NoteOptionsViewController`
2. Add `.popover` to the case in `SPNavigationController` 

The second option obviously affects other popover views that use a navigation bar, but from a search, I'm pretty confident that no other views currently use `SPNavigationController` in a popover context, so it's my preferred option currently.

Thoughts?

Simplenote/Classes/NoteOptionsViewController.swift
jleandroperez: Not really, honestly, nearly everytime I pick up Generics, I end up regretting doing so. I think it's all good!

jleandroperez:
@mattdc28 Ahhhhh **yes** feel free to extend SPNavigationController. We can make a sweep pass all over the iPad UI and check if we didn't miss anything.

Thank you!

mattdc28:
All the comments above should be resolved now. The row selection behaviour will be adjusted a little once I implement the logic for things like the publishing and copy link rows but for now, switch cells can't be tapped/highlighted but value 1 rows can.

Dark mode looking much healthier now. 

As per discussions on Slack, the ellipsis will stay as the non-outlined version.

Simulator Screen Shot - iPad Pro (11-inch) (1st generation) - 2020-08-19 at 15 18 22

mattdc28:
@jleandroperez Thank you for that, you actually made me spot that I'd passed the wrong view type into a method. Fixed and code is simplified now too.

mattdc28:
Screenshot 2020-08-20 at 15 13 18
Screenshot 2020-08-20 at 15 13 01

iPad and iPhone, respectively.

@jleandroperez If you would be so kind as to merge, I'll get on with the functionality behind the rows/switches.

jleandroperez:
Thanks for all the hard work @mattdc28 !!

Issue #842, opened 2020-08-12
Selecting Text in iOS app 4.22 very finicky


rezzap:
A user has reported having difficulty selecting text in the latest release of the iOS app. I've replicated the issue with text often jumping multiple lines with selection and being difficult to target only a few lines of text. 

### Expected

I would expect that when I try to select text within a note it goes slowly enough that I can control which text is selected.

### Observed

This jumps around a lot making it very challenging to select only a portion of text. 

### Reproduced
> 1. Go to any note in Simplenote 4.22
> 2. Long push on the screen to get the cursor in select mode. 
> 3. Move the cursor down a few lines to try and select it. Notice the issue. 

![Screen Recording 2020-08-12 at 04 26 pm](https://user-images.githubusercontent.com/10325627/90034749-ea3a3d80-dcb8-11ea-8887-185eb84f0ad3.gif)

Make|Model|iOS Version|App Version
-|-|-|-
Apple|iPhone X|13.6|4.22
Apple|iPad Air 2|13.6|4.22

msilbers:
reported in 3217374-zen

iOS 13.3.1
Simplenote ver.4.22

msilbers:
reported in 3220146-zen

asked for iOS and Simplenote versions

msilbers:
reported in 3223123-zen

Simplenote 4.22
asked for iOS version

msilbers:
reported in 3228480-zen

iPhone
Simplenote 4.22

rezzap:
Reported in 3232729-zen

iPhone 8S

jleandroperez:
Closed via #847. Shipping soon in 4.22.1 (hotfix).

Thanks everyone!!

PR #847, opened 2020-08-17
Fixing Text Selection

jleandroperez:
### Fix
In this PR we're updating our TextView supersmooth animation, and making sure it's only enabled for Keyboard-Y events.

As far as I can tell, the superclass implementation would appear to have a special behavior, whenever `UIAutoscroll` (private class) is involved, which is out of the question for us.

@aerych sir!! mind taking a look?
Thanks in advance!!

Ref. #842

### Test
1. Open a long note
2. Double tap over any word
- [x] Verify expanding the selected text upwards / downwards works as expected

**Important:** I've spotted a glitch in which the autoscroll wouldn't follow to the Heading / bottom of the document. Addressing that in a follow up, for atomicity's sake!

### Release
`RELEASE-NOTES.txt` was updated in 7edfc20 with:

> Fixed a bug that caused text selection to unexpectedly scroll to the top (or bottom) of the document

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,10 @@
 4.23
 -----
- 
+
+4.22.1
+------
+-   Fixed a bug that caused text selection to unexpectedly scroll to the top (or bottom) of the document
+
 4.22
 -----
 -   Fixed several layout issues that affected the Editor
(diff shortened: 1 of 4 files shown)


jleandroperez:
Thank you Eric!!

PR #848, opened 2020-08-19
Use the `main` mobile secrets branch

jkmassel:
### Fix
Changes the mobile secrets branch.

### Test
- In the project's root directory, run `rm -rf configure_files && git reset`
- In the project's Scripts directory, run `bundle exec fastlane run configure_apply`, then build the app in Xcode.
- If it succeeds, this PR is good to go!

### Review
Anyone can review!

### Release

### DIFF
--- a/.configure
+++ b/.configure
@@ -1,6 +1,6 @@
 {
   "project_name": "simplenote-ios",
-  "branch": "master",
+  "branch": "main",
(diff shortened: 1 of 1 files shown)


jkmassel:
Ah yes, that just means the version you have locally isn't the same as the version in the mobile secrets and it would like you to update – it's just asking you nicely what you want it to do before it clobbers your change (which may actually be the more recent one?).

Thank you!!

PR #851, opened 2020-08-20
[WIP] Nü Note Options

mattdc28:
### Details
In this PR we're implementing the new note options view. The work has been split into smaller pull requests which are being merged into this branch. I'm opening this request for visibility of ongoing work on the issue (#763) and it's eventual approval for merging.

Closes #763 

### Test
[TBD]

### Release
[TBD]

PR #856, opened 2020-08-22
New Note Options Logic

mattdc28:
### Details
This PR is part of the work required for #763 and will join the work in #851 when merged. It adds the logic for handling each of the table rows in the new note options view and their switches. This includes:
1. Pin to top 
2. Markdown
3. Share note
4. History
5. Publishing
6. Link copying
7. Collaboration handling
8. Trash

I've replaced some of the accessibility localisation strings that I had set with some of the ones from the existing view as they serve the same purpose and creating new strings would be a time waste. 

This PR also renames `SPAcitivitySafari` to `SPActivitySafari` which I believe was a typo.

### Screenshots
Simulator Screen Shot - iPad Pro (12 9-inch) - 2020-08-22 at 12 11 09

### Notes
I took the liberty of adding a loading indicator to the public link row as it seemed sensible to show that something was happening while the sync happens. I have tested running two simulators and toggling the switches on both and they are reflected on both devices quickly.

### Issues & Questions
1. The designs show an alert that displays at the bottom of the view if the note fails to publish. With the current publish mechanism I can see no way to subscribe to error notifications. If there is no internet connection then it looks like Simperium queues the action and waits for a connection before syncing. If any other type of error occurs, it doesn't look like the there is a way of getting details passed back from the save operation as there is no closure or completion block for the action.

2. I have not yet removed all of the legacy methods that this code replaces. It adds several hundred lines of deletion so will be moved to a separate request to avoid sending this over 500LOC

3. I've been unable to fully test the publishing behaviour as I'm currently using the demo credentials which I don't believe allows me to publish. I may need setting up with some proper credentials to test this one on.

4. I've been considering moving the collaboration view logic into the note options view. It would mean presenting another popover from our current one but I feel that the UX would feel nicer. You would edit your collaborators and be able to see the count increase in the options view. Currently, the options view is dismissed in favour of a modal view for collaboration.

### Feedback
Any feedback on this code is very welcome.

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -62,7 +62,7 @@
 		46A3C96317DFA81A002865AE /* NSString+Metadata.m in Sources */ = {isa = PBXBuildFile; fileRef = 467D9C771788D0FF00785EF3 /* NSString+Metadata.m */; };
 		46A3C96417DFA81A002865AE /* SPVersionPickerViewCell.m in Sources */ = {isa = PBXBuildFile; fileRef = E2ACA88217A7728E00EF801C /* SPVersionPickerViewCell.m */; };
 		46A3C96517DFA81A002865AE /* SPActivityView.m in Sources */ = {isa = PBXBuildFile; fileRef = E22BCDF61794F53600A17ED2 /* SPActivityView.m */; };
-		46A3C96617DFA81A002865AE /* SPAcitivitySafari.m in Sources */ = {isa = PBXBuildFile; fileRef = E283709917D7B46300AB562D /* SPAcitivitySafari.m */; };
+		46A3C96617DFA81A002865AE /* SPActivitySafari.m in Sources */ = {isa = PBXBuildFile; fileRef = E283709917D7B46300AB562D /* SPActivitySafari.m */; };
 		46A3C96717DFA81A002865AE /* NSManagedObjectContext+CoreDataExtensions.m in Sources */ = {isa = PBXBuildFile; fileRef = E22A17CF17A2E1CE00383575 /* NSManagedObjectContext+CoreDataExtensions.m */; };
 		46A3C96A17DFA81A002865AE /* SPTagStub.m in Sources */ = {isa = PBXBuildFile; fileRef = E277B5FA179F714C0095CD24 /* SPTagStub.m */; };
 		46A3C96C17DFA81A002865AE /* SPOptionsViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = E26AC90F179DA25F00C4BFB6 /* SPOptionsViewController.m */; };
@@ -768,8 +768,8 @@
 		E277B5FF17A063020095CD24 /* SPTagEntryField.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = SPTagEntryField.h; path = Classes/SPTagEntryField.h; sourceTree = "<group>"; };
 		E277B60017A063020095CD24 /* SPTagEntryField.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; name = SPTagEntryField.m; path = Classes/SPTagEntryField.m; sourceTree = "<group>"; };
 		E280453D180DAE0200670073 /* en */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.plist.strings; name = en; path = en.lproj/Localizable.strings; sourceTree = "<group>"; };
-		E283709817D7B46300AB562D /* SPAcitivitySafari.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = SPAcitivitySafari.h; path = Classes/SPAcitivitySafari.h; sourceTree = "<group>"; };
-		E283709917D7B46300AB562D /* SPAcitivitySafari.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; name = SPAcitivitySafari.m; path = Classes/SPAcitivitySafari.m; sourceTree = "<group>"; };
+		E283709817D7B46300AB562D /* SPActivitySafari.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = SPActivitySafari.h; path = Classes/SPActivitySafari.h; sourceTree = "<group>"; };
+		E283709917D7B46300AB562D /* SPActivitySafari.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; name = SPActivitySafari.m; path = Classes/SPActivitySafari.m; sourceTree = "<group>"; };
 		E28A760D178CBE6D008659DE /* SPNoteEditorViewController.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = SPNoteEditorViewController.h; path = Classes/SPNoteEditorViewController.h; sourceTree = "<group>"; };
 		E28A760E178CBE6D008659DE /* SPNoteEditorViewController.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; lineEnding = 0; name = SPNoteEditorViewController.m; path = Classes/SPNoteEditorViewController.m; sourceTree = "<group>"; xcLanguageSpecificationIdentifier = xcode.lang.objc; };
 		E2911E7C17B49F4600F13AD9 /* DTPinDigitView.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = DTPinDigitView.h; path = DTPinLock/DTPinDigitView.h; sourceTree = "<group>"; };
@@ -954,8 +954,8 @@
 				B5AEC38523FAE70600D24221 /* Note+List.swift */,
 				E25C39D417A42C4900B2591A /* PersonTag.h */,
 				E25C39D517A42C4900B2591A /* PersonTag.m */,
-				E283709817D7B46300AB562D /* SPAcitivitySafari.h */,
-				E283709917D7B46300AB562D /* SPAcitivitySafari.m */,
+				E283709817D7B46300AB562D /* SPActivitySafari.h */,
+				E283709917D7B46300AB562D /* SPActivitySafari.m */,
 				E277B5F9179F714C0095CD24 /* SPTagStub.h */,
 				E277B5FA179F714C0095CD24 /* SPTagStub.m */,
 				467D9C631788A54900785EF3 /* Tag.h */,
@@ -2203,7 +2203,7 @@
 				46A3C96517DFA81A002865AE /* SPActivityView.m in Sources */,
 				B5E4082B235DDFC800D4E1DF /* ColorStudio.swift in Sources */,
 				B552AB8724B8E75E00E5E115 /* SPNoteEditorViewController+Extensions.swift in Sources */,
-				46A3C96617DFA81A002865AE /* SPAcitivitySafari.m in Sources */,
+				46A3C96617DFA81A002865AE /* SPActivitySafari.m in Sources */,
 				B5BE05461AB7522F002417BF /* JSONKit+Simplenote.m in Sources */,
 				B5095DC924632E3300812711 /* SimplenoteConstants.swift in Sources */,
 				B56E763422BD394C00C5AA47 /* UIImage+Simplenote.swift in Sources */,
--- a/Simplenote/Classes/NoteOptionsViewController.swift
+++ b/Simplenote/Classes/NoteOptionsViewController.swift
@@ -13,6 +13,10 @@ class NoteOptionsViewController: UITableViewController {
     /// The note from the editor that we will change settings for
     fileprivate var note: Note
 
+    /// The delegate to notify about
+    /// chaanges made here
+    weak var delegate: NoteOptionsViewControllerDelegate?
+
     init(with note: Note) {
         self.note = note
         super.init(style: .grouped)
@@ -264,6 +268,7 @@ class NoteOptionsViewController: UITableViewController {
     func handleMarkdown(sender: UISwitch) {
         note.markdown = sender.isOn
         save()
+        delegate?.didToggleMarkdown(toggle: sender, sender: self)
     }
 
     func handleShare(from indexPath: IndexPath) {
@@ -319,4 +324,10 @@ class NoteOptionsViewController: UITableViewController {
         SPTracker.trackEditorNoteEdited()
         CSSearchableIndex.default().indexSearchableNote(note)
     }
+}
+
+// MARK: - Action protocol
+//
+protocol NoteOptionsViewControllerDelegate: class {
+    func didToggleMarkdown(toggle: UISwitch, sender: NoteOptionsViewController)
 }
--- a/Simplenote/Classes/NoteOptionsViewController.swift
+++ b/Simplenote/Classes/NoteOptionsViewController.swift
@@ -11,17 +11,17 @@ final class NoteOptionsViewController: UITableViewController {
     }
 
     /// The note from the editor that we will change settings for
-    private var note: Note
+    private let note: Note
 
     /// The delegate to notify about
     /// chaanges made here
     weak var delegate: NoteOptionsViewControllerDelegate?
 
     /// Formats number of collaborators to respect locales
-    private var collaboratorNumberFormatter = NumberFormatter()
+    private let collaboratorNumberFormatter = NumberFormatter()
 
     /// Activity indicator that displays when note is publishing or unpublishing
-    private var publishActivityIndicator = UIActivityIndicatorView(style: SPUserInterface.isDark ? .white : .gray)
+    private let publishActivityIndicator = UIActivityIndicatorView(style: SPUserInterface.isDark ? .white : .gray)
 
     /// Initialises the options view for a specific note
     /// - Parameter note: The note to configure options for
--- a/Simplenote/Classes/NoteOptionsViewController.swift
+++ b/Simplenote/Classes/NoteOptionsViewController.swift
@@ -21,7 +21,7 @@ final class NoteOptionsViewController: UITableViewController {
     private let collaboratorNumberFormatter = NumberFormatter()
 
     /// Activity indicator that displays when note is publishing or unpublishing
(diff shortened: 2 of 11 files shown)


jleandroperez:
> 1. The designs show an alert that displays at the bottom of the view if the note fails to publish. With the current publish mechanism I can see no way to subscribe to error notifications. If there is no internet connection then it looks like Simperium queues the action and waits for a connection before syncing. If any other type of error occurs, it doesn't look like the there is a way of getting details passed back from the save operation as there is no closure or completion block for the action.

Android is using [this mechanism](https://github.com/Automattic/simplenote-android/blob/bc8b6d38d364e8d072f78e6e6788a301ec4895f8/Simplenote/src/main/java/com/automattic/simplenote/NoteEditorFragment.java#L1247-L1261). If you do go that road, please keep it super encapsulated 😄 

> 2. I have not yet removed all of the legacy methods that this code replaces. It adds several hundred lines of deletion so will be moved to a separate request to avoid sending this over 500LOC

👍 

> 3. I've been unable to fully test the publishing behaviour as I'm currently using the demo credentials which I don't believe allows me to publish. I may need setting up with some proper credentials to test this one on.

I'll get back to you on this one!

> 4. I've been considering moving the collaboration view logic into the note options view. It would mean presenting another popover from our current one but I feel that the UX would feel nicer. You would edit your collaborators and be able to see the count increase in the options view. Currently, the options view is dismissed in favour of a modal view for collaboration.

Not sure exactly what you mean by the `collaboration logic` (NoteEditorVC only has a delegate). But I'd vote you pick the approach that leads to beautiful code 😄 

(And I think you're aiming for consistency all over), which is great.

Thank you!

mattdc28:
Thanks for all the above!

I decided to go ahead and move the collaboration handling/delegate into the note options view so now (on iPad) we get the view presented from the collaborator row which feels nicer than being kicked out of the note options and into a modal view.

I'll be able to remove the old code from the editor view but again, holding off for a separate PR to do all the cleaning out as there is going to be a fair amount that can go.

Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2020-08-25 at 15 52 56

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -1,14 +1,39 @@
...
+
+    /// The delegate to notify about
+    /// chaanges made here
+    weak var delegate: NoteOptionsViewControllerDelegate?
+
+    /// Formats number of collaborators to respect locales
+    private var collaboratorNumberFormatter = NumberFormatter()
+
+    /// Activity indicator that displays when note is publishing or unpublishing
+    private var publishActivityIndicator = UIActivityIndicatorView(style: SPUserInterface.isDark ? .white : .gray)
```
jleandroperez, line 24: Would this one react to traits changes?

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -1,14 +1,39 @@
...
+final class NoteOptionsViewController: UITableViewController {
 
     /// Array of `Section`s to display in the view.
     /// Each `Section` has `Rows` that are used for display
     fileprivate var sections: [Section] {
         return [optionsSection, linkSection, collaborationSection, trashSection]
     }
 
+    /// The note from the editor that we will change settings for
+    private var note: Note
```
jleandroperez, line 14: Would this property change at any time? (Same note for nearly all of the new properties!)

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -83,38 +115,46 @@ class NoteOptionsViewController: UITableViewController {
...
+                        NSLocalizedString("Enable Markdown formatting", comment: "Accessibility hint for enabling markdown mode")
+                    cell.cellSwitch.isOn = note.markdown
                 }
             ),
             Row(style: .Value1,
                 configuration: { (cell: UITableViewCell, row: Row) in
                     let cell = cell as! Value1TableViewCell
                     cell.textLabel?.text = NSLocalizedString("Share", comment: "Note Options: Show Share Options")
-                    cell.accessibilityHint = NSLocalizedString("Tap to open share options", comment: "Accessibility hint on cell that activates share sheet")
+                    cell.accessibilityHint = NSLocalizedString("share-accessibility-hint", comment: "Accessibility hint on share button")
```
jleandroperez, line 145: There are a few localizations like this, but the standard all across the apps is your former approach. Mind switching back? (thank you!)

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -241,42 +292,177 @@ class NoteOptionsViewController: UITableViewController {
...
+            activityVC.modalPresentationStyle = .popover
+
+            let presentationController = activityVC.popoverPresentationController
+            presentationController?.permittedArrowDirections = .any
+            presentationController?.sourceRect = tableView.rectForRow(at: indexPath)
+            presentationController?.sourceView = tableView
+        }
+        present(activityVC,
+                animated: true,
+                completion: nil)
```
jleandroperez, line 342: No need to go multiline 😄

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -712,7 +711,7 @@ - (void)bounceMarkdownPreview
                     self.noteEditorTextView.hidden = NO;
                     [snapshot removeFromSuperview];
 
-                    self->bounceMarkdownPreviewOnActivityViewDismiss = NO;
+                    self->_bounceMarkdownPreviewOnActivityViewDismiss = NO;
```
jleandroperez, line 714: Since we're there, and you've already added a property, mind dropping the direct ivar usage? 😄

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -62,7 +62,7 @@
 		46A3C96317DFA81A002865AE /* NSString+Metadata.m in Sources */ = {isa = PBXBuildFile; fileRef = 467D9C771788D0FF00785EF3 /* NSString+Metadata.m */; };
 		46A3C96417DFA81A002865AE /* SPVersionPickerViewCell.m in Sources */ = {isa = PBXBuildFile; fileRef = E2ACA88217A7728E00EF801C /* SPVersionPickerViewCell.m */; };
 		46A3C96517DFA81A002865AE /* SPActivityView.m in Sources */ = {isa = PBXBuildFile; fileRef = E22BCDF61794F53600A17ED2 /* SPActivityView.m */; };
-		46A3C96617DFA81A002865AE /* SPAcitivitySafari.m in Sources */ = {isa = PBXBuildFile; fileRef = E283709917D7B46300AB562D /* SPAcitivitySafari.m */; };
+		46A3C96617DFA81A002865AE /* SPActivitySafari.m in Sources */ = {isa = PBXBuildFile; fileRef = E283709917D7B46300AB562D /* SPActivitySafari.m */; };
```
jleandroperez, line 65: Nice 🥳

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -153,13 +153,76 @@ extension SPNoteEditorViewController {
...
+        let noteView = NoteOptionsViewController(with: currentNote)
+        noteView.delegate = self
         let noteNavigation = SPNavigationController(rootViewController: noteView)
         noteNavigation.displaysBlurEffect = true
         noteNavigation.modalPresentationStyle = .popover
         noteNavigation.popoverPresentationController?.sourceRect = button.bounds
         noteNavigation.popoverPresentationController?.sourceView = button
         noteNavigation.popoverPresentationController?.backgroundColor = .simplenoteNavigationBarModalBackgroundColor
+        noteNavigation.popoverPresentationController?.delegate = self
         present(noteNavigation, animated: true, completion: nil)
```
jleandroperez, line 172: We could, perhaps, encapsulate a bit, and split this method in two, WDYT?

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -241,42 +292,177 @@ class NoteOptionsViewController: UITableViewController {
...
+        if sender.isOn {
+            SPTracker.trackEditorNotePublished()
+        } else {
+            SPTracker.trackEditorNoteUnpublished()
+        }
+
+        note.published = sender.isOn
+
+        if (note.published && note.publishURL.isEmpty ||
+            !note.published && !note.publishURL.isEmpty) {
```
jleandroperez, line 360: Perhaps it's worth moving this to an ivar?

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -241,42 +292,177 @@ class NoteOptionsViewController: UITableViewController {
...
+        sender.accessibilityHint = note.pinned ?
+            NSLocalizedString("Unpin note", comment: "Action to mark a note as unpinned") :
+            NSLocalizedString("Pin note", comment: "Action to mark a note as pinned")
+
+        if sender.isOn {
+            SPTracker.trackEditorNotePinned()
+        } else {
+            SPTracker.trackEditorNoteUnpinned()
+        }
     }
```
jleandroperez, line 307: There's a bit of code duplication in all of the event handlers. Mind taking a peek sir?

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -241,42 +292,177 @@ class NoteOptionsViewController: UITableViewController {
...
+        SPTracker.trackEditorNoteEdited()
+        CSSearchableIndex.default().indexSearchableNote(note)
+    }
+}
+
+// MARK: - Action protocol
+//
+/// A protocol to pass row actions to the parent view controller for handling
+protocol NoteOptionsViewControllerDelegate: class {
+    func didToggleMarkdown(toggle: UISwitch, sender: NoteOptionsViewController)
```
jleandroperez, line 444: Would the delegate need direct access to the internal UISwitch?

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
mattdc28: Yeah it's getting a bit busy. Perhaps I could leave the save method call and tracking here and then move the rest into a `presentNoteOptions` method?

Simplenote/Classes/NoteOptionsViewController.swift
mattdc28: I think I passed it just in case we needed to do anything to the switch but that's not the case in the end. I can just read the markdown state directly from `currentNote` and remove everything passed to the delegate

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -241,42 +302,160 @@ class NoteOptionsViewController: UITableViewController {
...
+        present(viewController, animated: true, completion: nil)
+    }
+}
+
+// MARK: - Action protocol
+//
+/// A protocol to pass row actions to the parent view controller for handling
+protocol NoteOptionsViewControllerDelegate: class {
+    func didToggleMarkdown()
+    func didTapHistory(sender: NoteOptionsViewController)
```
mattdc28, line 438: @jleandroperez Do you have an opinion on delegates? I've always passed a sender of the view, just in case, but it's not actually used in the current delegate methods. My feeling here is that I can probably just remove the sender parameters

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -57,6 +57,7 @@ final class NoteOptionsViewController: UITableViewController {
     func setupViewStyles() {
         tableView.backgroundColor = .simplenoteTableViewBackgroundColor
         tableView.separatorColor = .simplenoteDividerColor
+        publishActivityIndicator.color = UIColor.color(name: .tableViewDetailTextLabelColor)
```
mattdc28, line 70: @jleandroperez Tinting the activity indicator is better as it respects the theme changes. Is it worth adding a dedicated colour for it to UIColor+Studio or is using another colour like this okay? I went for the detail text label as the indicator is in the same place that label would be so it keeps the same styling.

Simplenote/Classes/NoteOptionsViewController.swift
mattdc28: Updated to use a colour now that will switch with themes.

Simplenote/Classes/NoteOptionsViewController.swift
mattdc28: A good spot, I've updated most to `let` now.

Simplenote/Classes/NoteOptionsViewController.swift
mattdc28: Thanks, tidied this up now to stop duplication of if statements and moved the presentation code into a new method.

Simplenote/Classes/NoteOptionsViewController.swift
jleandroperez: Honestly, I try not to be biased by personal opinions!! (note the **try** word!!). But the detail is:

By exposing the internal UISwitch reference, the API would imply that it's okay for a delegate to access / tamper with the UISwitch right? (Since it's not hidden / protected anyhow).

For that reason, sending over just the property that is needed is a slightly better design, since it encapsulates the internal component, and only exposes what's 100% needed.

Simplenote/Classes/NoteOptionsViewController.swift
jleandroperez: Oh not really!! I was actually thinking about iOS 13 dynamic colors. The `tableViewDetailTextLabelColor` case isn't iOS 13 friendly. **However, just tested, and it's reacting well to trait changes**, so no actual need to change anything!!

Apologies about the noise!!

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -1,14 +1,49 @@
...
+    private let note: Note
+
+    /// Determines if a note is currently awaiting publishing
+    /// changes to synchronise
+    private var noteIsChangingPublishState: Bool {
+        if (note.published && note.publishURL.isEmpty ||
+            !note.published && !note.publishURL.isEmpty) {
+            return true
+        }
+        return false
```
jleandroperez, line 23: Nipticky!

```
private var noteIsChangingPublishState: Bool {
    return (note.published && note.publishURL.isEmpty ||
                !note.published && !note.publishURL.isEmpty)
}
```

And you can take this one step further, and drop the `return` statement, since it's a single line, and returns its value implicitly.

Simplenote/Classes/NoteOptionsViewController.swift
jleandroperez: Thank you @mattdc28 !! Just one last note. The accessibilityHint assignment itself is duplicated (so the strings are x2, and it's slightly harder to maintain).

Same note for the other API(s)!

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -241,42 +302,160 @@ class NoteOptionsViewController: UITableViewController {
...
+            SPTracker.trackEditorNotePublished()
+            sender.accessibilityHint = NSLocalizedString("Unpublish note", comment: "Action which unpublishes a note")
+        } else {
+            SPTracker.trackEditorNoteUnpublished()
+            sender.accessibilityHint = NSLocalizedString("Publish note", comment: "Action which published a note to a web page")
+        }
+
+        note.published = sender.isOn
+
+        if (noteIsChangingPublishState) {
```
jleandroperez, line 357: No need to use `()` 😄

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
jleandroperez: Looks good! Yeah it was a bit too overloaded (single API doing lots of things). Looking good!

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -153,13 +153,82 @@ extension SPNoteEditorViewController {
...
+        trashNoteAction(self)
+    }
+}
+
+// MARK: - Note options state handling
+//
+extension SPNoteEditorViewController {
+    /// A convenience variable for accessing the currently presented note options view
+    /// This is used to pass on note updates from Simperium
+    @objc var presentedNoteOptionsViewcontroller: NoteOptionsViewController? {
```
jleandroperez, line 231: Lowercased `c` in `Viewcontroller`

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -153,13 +153,82 @@ extension SPNoteEditorViewController {
...
         present(noteNavigation, animated: true, completion: nil)
     }
 }
+
+// MARK: - Popover presentation delegate
+//
+extension SPNoteEditorViewController: UIPopoverPresentationControllerDelegate {
+    public func popoverPresentationControllerDidDismissPopover(_ popoverPresentationController: UIPopoverPresentationController) {
+        if bounceMarkdownPreviewOnActivityViewDismiss {
+            self.bounceMarkdownPreview()
```
jleandroperez, line 187: `self` is not needed here

Simplenote/Classes/NoteOptionsViewController.swift
```diff
@@ -155,13 +215,14 @@ class NoteOptionsViewController: UITableViewController {
     fileprivate var collaborationSection: Section {
         let rows = [
             Row(style: .Value1,
-                configuration: { (cell: UITableViewCell, row: Row) in
+                configuration: { [note, collaboratorNumberFormatter] (cell: UITableViewCell, row: Row) in
                     let cell = cell as! Value1TableViewCell
                     cell.textLabel?.text = NSLocalizedString("Collaborate", comment: "Note Options: Collaborate")
-                    cell.accessibilityLabel = NSLocalizedString("Tap to open collaboration menu", comment: "Accessibility hint on cell that opens collaboration menu")
+                    cell.detailTextLabel?.text = collaboratorNumberFormatter.string(from: NSNumber(value: note.emailTagsArray.count))
```
jleandroperez, line 221: Note here: When there are no collaborators, this would effectively display a `0` on the right hand side.

Do you think we could either hide the number if it's zero, or display a `-` instead perhaps?

Simplenote/Classes/NoteOptionsViewController.swift
jleandroperez: Last nipticky note of the day!!!

Since the delegate needs, pretty much, a loot of visibility, would it be possible to move the protocol to the very top of the file?

Simplenote/Classes/NoteOptionsViewController.swift
mattdc28: Thanks!

mattdc28:
@jleandroperez I believe I've tackled all the above now. If you're happy with the changes please merge them in for me.

jleandroperez:
Closing this PR for now.

Thank you Matt!!

PR #881, opened 2020-09-16
Drops iOS 11 Support

jleandroperez:
### Details:
In this PR we're dropping iOS 11 support 🔥

- Drops several availability checks that aren't really needed anymore
- Links CoreServices (rather than MobileCoreServices!)

@guarani YES Paul!! It's **THAT** time of the year!! Can I bug you with a quick PR?

Thanks in advance!!

### Test
- [ ] Verify the app builds correctly!

### Release
`RELEASE-NOTES.txt` was updated in 4efe804 with:
 
> iOS 12 is now the minimum supported OS #881

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@
 -   Fixed a bug that prevented adding collaborators with email addresses that contain new TLDs #872
 -   Swiping a Note in the List now reveals the Copy Internal Link Action #864
 -   Long Pressing over an Internal Link now pushes the target Note #868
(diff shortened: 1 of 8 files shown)


guarani (approved):
LGTM! 
Looks like there's still a few checks for iOS 12: https://github.com/Automattic/simplenote-ios/search?q=iOS+12&unscoped_q=iOS+12, not sure if they can be removed.

jleandroperez:
Huh I missed a few checks I'll get rid of those elsewhere 😄 

Thanks for spotting that!! (And thank you for testing this one Paul!!!)

Issue #880, opened 2020-09-16
Find replacement for archived SAMKeychain library


mokagio:
I noticed that the [`SAMKeychain`](https://github.com/soffes/SAMKeychain) library, which we use in the app, has been archived.

While this doesn't stop us from using it, it does ring a warning bell about the state of the library and its future. In fact, [the last commit](https://github.com/soffes/SAMKeychain/commit/d3d64f8c5fa14bf2adb2d51a18ee8b6639232ca5) is from December 2018.

mokagio:
🎉

PR #889, opened 2020-09-18
Drops SAMKeychain: Mark 1

jleandroperez:
### Details:
- New **SimplenoteKeychain** namespace, which is meant to hold all the keychain constants
- Switches **SAMKeychain** >> **KeychainManager** for PIN Lock code access operations

@aerych Can I bug you with this one sir?
Thank you!!!

P.s.: I've been going back and forth with the idea of adding log statements. WDYT? 
Perhaps it's best to patch the Keychain Wrapper to log errors, rather than adding `do / catch` all over?

Ref. #880

---

### Test: Pin Lock!
1. Install **develop** and log into your account
2. Open **Settings** > **Passcode**
3. Set a Pin Lock code
4. Set the `Lock Timeout` to 15 seconds
5. Install this branch on top

- [x] Verify the Pin Lock code you've used in (3) unlocks the app
- [x] Verify that the Pin Lock is requested only if the app stays for +15 seconds backgrounded 
- [x] Verify that the app remains unlocked if you reopen in under 15 seconds

### Release


Simplenote/Classes/KeyboardObservable.swift
```diff
@@ -1,4 +1,6 @@
 import Foundation
+import UIKit
```
aerych, line 2: Is this import needed?  Looks like there's nothing in UIKit being used here (or I'm missing it) and the app builds with out it.

Simplenote/Classes/UIKitConstants.swift
```diff
@@ -1,4 +1,5 @@
 import Foundation
+import UIKit
```
aerych, line 2: Is this import needed?

aerych (approved):
Hola @jleandroperez 👋 
Tested as described and it worked perfectly. :) 
Code looks good, I just had a nitpick about some imports.  
As far as logging goes... yeah I would really be a fan of switching away from `try?` and instead wrapped calls in `do catch` blocks and logged any errors.  Up to you sir.
Otherwise :shipit:

Simplenote/Classes/KeyboardObservable.swift
jleandroperez: Actually it's required by the follow up PR (where we get rid of SAMKeychain!!). Xcode started compaining about `UIResponder` and `CGRect` not being in scope.

Apparently SAMKeychain (somehow) got UIKit imported for us, and now that's gone, we appear to need few extra imports.

Thanks for checking sir!!

Simplenote/Classes/UIKitConstants.swift
jleandroperez: (Just for the record!!) Xcode actually complains if the import is missing, in the follow up PR.

Apparently SAMKeychain imported UIKit for us, and we didn't need the explicit import declarations. Error involves `CGFloat` not being in scope.

Thank you sir!!

jleandroperez:
@aerych 100% with you on the `do / catch` front, but I wasn't sure about duplicating that snippet all over.

**SO!!** Fresh week, fresh start... I remembered `@propertyWrapper` exists!!. 

Would it be possible to take a quick peek, for sanity?

Thanks in advance!!!

aerych:
Niiiiiice. Very nice.  Tested and it looks good to me! :shipit: :)

jleandroperez:
Thank you Eric!!

PR #898, opened 2020-10-02
Merge release 4.25 into develop

mokagio:

PR #899, opened 2020-10-02
Merge release 4.25 into trunk

mokagio:
I'm not sure why there are so many commits in the log. It looks like there are also commits related to v4.24. 

I tried a merged locally and confirmed it works, so I guess we're still good? 🤔


jleandroperez (approved):
Looks super, thank you Gio!!

:shipit:

mokagio:
image

I wouldn't normally do this, but I already merged `release/4.25` into `develop`, #898, and there already is a tag for 4.25 so little harm is done by this other than making the log a bit messy in the interest of avoiding an history rewrite which would make it confusing if we ever went back and looked at this and its sister PR against `develop`.

jkmassel (approved):

mokagio:
That Release Build CircleCI task has been stuck there for _days_... I'd really like to chase down the reason why before merging this.

mokagio:
I haven't had the time to look into this yet.

Merging just to avoid things getting stuck later this week for 4.26

Issue #863, opened 2020-09-01
Interlinking


jleandroperez:
### Pasting Links:
- [x] Swipeable Rows should display Images
- [x] New **Copy Link** swipeable row
- [x] Long Press Preview Mechanism: **UIContextMenuInteraction**
- [x] New **Copy Internal Link** (Note Options) Row

### Navigation:
- [x] Editor / Keyboard Onscreen:: Regular Press over a markdown link should do nothing
- [x] Editor / Keyboard Hidden: Long Press over a markdown link should push the target note
- [x] Markdown Preview: Pressing a link should push the target note's editor
- [x] Back should always lead you back to the Notes List

### Mockups
![blueprint_ios](https://user-images.githubusercontent.com/1195260/91865578-e8c0bd00-ec47-11ea-8ed7-1a827e392f68.png)

jleandroperez:
Closing this one since the main functionality has been already shipped

Addressing New Metrics UI (with Interlink References) via #836, and Interlink Autocomplete via #914.

Thank you 😄

PR #902, opened 2020-10-06
Contextual Menu Support

jleandroperez:
### Details:
In this PR we're implementing the contextual UITableView / UIMenu API, which replaced the legacy 3d Touch API(s).

@etoledom Sir!! would you be game for a quick PR?
Thanks in advance!!

Ref. #863

### Test: All Notes
1. Log into your account
2. Select All Notes
3. Long press over any Note in your List

- [x] Verify that a Preview with actions at the bottom shows up
- [x] Verify that pressing the Preview pushes the Editor
- [x] Verify that **Copy Link** places in the Pasteboard a link to the selected note (paste elsewhere!)
- [x] Verify that pressing **Share** opens the Share Sheet
- [x] Verify that pressing **Delete** trashes the selected note

### Test: Trash
1. Log into your account
2. Open the Trash
3. Long press over any trash ed notes

- [x] Verify nothing happens!

### Release


etoledom (approved):
Nice PR @jleandroperez ! Working great 🎉 

One small detail I noticed while deleting notes:
The behaviour seems to be different than deleting the note with the normal cell buttons, it will remove the cell and it seems to be replaced by other random note, without animations. While the expected behaviour would be to see the deletion animation and see the cell disappear.

deleting


Approving since everything (else) I working as expected ✅

jleandroperez:
Thank you for testing this one @etoledom !!!

Quick Q sir!! is it possible that video was taken on the iOS Sim? (can't repro on the actual device, only happens in the simulator on my end).

etoledom:
@jleandroperez - it was on an iPhone 7 device running iOS13.
I hope that helps!

jleandroperez:
Thank you @etoledom !!! Bug confirmed in iOS 13 (simulator). 

For the time being, I've disabled the `Delete` action on iOS 13, since this has the shape of a SDK / animation glitch (afaik solved in 14!).

**Again, thank you sir!!**

PR #909, opened 2020-10-09
Nü Note Options: Mark IV

jleandroperez:
### 🛑  Important
Don't merge until #906 #907 #909 and #910 are ready: everything must be merged in a single shot!

### Details:
In this PR we're wiring support for few missing Note Option Actions:

- Share
- History
- Trash

@aerych Sir!! can I trouble you with this one?
Thanks in advance!!

Ref. #763

### Test
1. Log into your account
2. Open any note
3. Press over the top `( i )` button

- [x] Verify pressing **Share** dismisses the Options and presents the Share Sheet
- [x] Verify pressing **History** dismisses the Options and presents the History Slider
- [x] Verify pressing **Trash** dismisses both, Options and the Editor, and trashes the Note
- [x] Verify the same works super on an iPad device please!!

### Release
These changes do not require release notes.

PR #907, opened 2020-10-09
Nü Note Options: Mark III

jleandroperez:
### 🛑  Important
Don't merge until #906 #907 #909 and #910 are ready: everything must be merged in a single shot!

### Details
This PR definitely drops **a lot** of dead code. No new behavior is introduced!

- **SPActivityView** + **SPActionButton** + **SPToggle**: Legacy Actions Sheet (replaced by OptionsViewController!)
- And several LOC(s) in the EditorViewController, that were already commented out

@danielebogo Sir!! May I bug you with **the coolest PR ever?**
Thanks in advance!!

Ref. #763

### Test
- [x] Verify the app builds correctly
- [x] Please run a smoke test over the editor!

### Release
These changes do not require release notes.

PR #906, opened 2020-10-08
Nü Note Options: Mark II

jleandroperez:
### 🛑  Important
Don't merge until #906 #907 #909 and #910 are ready: everything must be merged in a single shot!

### Details
- Implements **EntityObserver** in the Options UI: we'll dynamically refresh everything whenever the associated Note is updated.
- Implements handlers several actions:
  - **Pin**
  - **Markdown**
  - **Copy Interlink**
  - **Publish** + **Copy Public Link**
- Dead code has been commented out (and removed in a follow up PR!)

@etoledom @guarani Gentlemen, may I bug you with this one?
**Thanks a looooot!!!**

Closes #715
Ref. #763
Ref. #863 

---

### Test: Dynamic Refresh      :::      @guarani 👇👇👇 **Thank you Paul!!!**
1. Open the same Note in iOS + another client (web or macOS!)
2. In iOS: Press the top `( i )` icon to reveal the Options

- [x] Verify that toggling the **Pin** / **Markdown** / **Published** flags in your second client causes iOS to refresh its UI

### Test: Markdown      :::      @guarani 👇👇👇 **Thank you Paul!!!**
1. Open any note and press over the top `( i )` icon
2. Disable Markdown and press **done** to dismiss
3. Reopen the Note Options
4. Enable Markdown and press **done** to dismiss

- [x] Verify that as soon as the Options UI is gone, the view bounces horizontally (revealing the Markdown Preview UI)
- [x] Verify that if you **add a new Note**, the **Markdown** flag of your last selection will be applied (On or Off by default!)

### Test: Copy Interlink      :::      @guarani 👇👇👇 **Thank you Paul!!!**
1. Open any note and press over the top `( i )` icon
2. Press over the **Copy Interlink** row

- [x] Verify that **Pasting** over the editor inserts a link to the note you're currently editing!

---

### Test: Pinning      :::      @etoledom  👇👇👇 **Thank you Eduardooooo!!!**
1. Open any note and press over the top `( i )` icon
2. Press over the **Pin to Top** row

- [x] Verify the Note ends up pinned!

### Test: Publishing      :::      @etoledom  👇👇👇 **Thank you Eduardooooo!!!**
1. Open any note and press over the top `( i )` icon
2. Enable the `Publish` switch

- [ ] Verify the row below reads as `Publishing...` and cannot be selected
- [ ] Verify that after a while the row belows reads as `Copy Link`
- [ ] Verify that pressing `Copy Link` copies the associated Note URL into the Pasteboard
[...]

PR #910, opened 2020-10-09
Nü Note Options: Mark V

jleandroperez:
### 🛑  Important
Don't merge until #906 #907 #909 and #910 are ready: everything must be merged in a single shot!

### Details:
- **Wires Collaboration support** in the Note Options Interface
- **OptionsViewController** now implements the AddCollaborator delegate protocol
- **Editor:** Drops the **blankNote** flag
- **Note:** New **isBlank** and **emailTags** properties

@frosty Sir!! Can I bug you with this one?
Thanks in advance!!!

Ref. #763

### Test: Collaboration
1. Open any note
2. Press over the top `( i )` icon
3. Press over **Collaborate**

- [ ] Verify you can add / remove collaborators
- [ ] Verify that newly added collaborators get access to the shared document

### Test: Empty Notes
1. Launch Simplenote and press the top right `+` button
2. **Do not type anything**
3. Press back

- [ ] Verify that the newly added note is deleted

### Test: Non Empty Notes
1. Launch Simplenote and press the top right `+` button
2. Type anything
3. Delete the contents
4. Press back

- [ ] Verify that the newly added is **not** deleted

### Release
`RELEASE-NOTES.txt` was updated in cdedacd with:
 
> New Note Options interface #763

### DIFF
--- a/Simplenote/Classes/Note+Properties.swift
+++ b/Simplenote/Classes/Note+Properties.swift
@@ -5,6 +5,13 @@ import Foundation
 //
 extension Note {
 
+    /// Indicates if the receiver is a blank document
+    ///
+    @objc
+    var isBlank: Bool {
+        objectID.isTemporaryID && content?.count == .zero && tagsArray?.count == .zero
+    }
+
     /// Returns the Creation / Modification date for a given SortMode
     ///
     func date(for sortMode: SortMode) -> Date? {
--- a/Simplenote/Classes/Note+Properties.swift
+++ b/Simplenote/Classes/Note+Properties.swift
@@ -19,4 +19,16 @@ extension Note {
             return modificationDate
         }
     }
+
+    /// Returns the collection user emails with who we're sharing this document
+    ///
+    var emailTags: [String] {
+        guard let tags = tagsArray as? [String] else {
+            return []
+        }
+
+        return tags.filter {
+            $0.isValidEmailAddress
+        }
+    }
 }
--- a/Simplenote/Classes/Note.h
+++ b/Simplenote/Classes/Note.h
@@ -21,7 +21,6 @@
     NSString        *tags;
     NSString        *systemTags;
     NSMutableArray  *tagsArray;
-    NSMutableArray  *emailTagsArray;
     NSMutableArray  *systemTagsArray;
     NSString        *remoteId;
     int             lastPosition;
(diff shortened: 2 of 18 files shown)


Simplenote/Classes/Note+Properties.swift
```diff
@@ -5,6 +5,13 @@ import Foundation
 //
 extension Note {
 
+    /// Indicates if the receiver is a blank document
+    ///
+    @objc
+    var isBlank: Bool {
+        objectID.isTemporaryID && content?.count == .zero && tagsArray?.count == .zero
```
frosty, line 12: A note can only be blank if it hasn't been saved?

Simplenote/Classes/Note.h
```diff
@@ -34,20 +33,19 @@
...
+@property (nonatomic, strong) NSMutableArray<NSString*> *tagsArray;
+@property (nonatomic,   copy) NSString                  *shareURL;
+@property (nonatomic,   copy) NSDate                    *creationDate;
+@property (nonatomic,   copy) NSString                  *systemTags;
+@property (nonatomic,   copy) NSString                  *modificationDatePreview;
+@property (nonatomic,   copy) NSString                  *creationDatePreview;
 
-@property (nonatomic,   copy) NSString          *titlePreview;
-@property (nonatomic,   copy) NSString          *bodyPreview;
+@property (nonatomic,   copy) NSString                  *titlePreview;
```
frosty, line 47: :chef-kiss:

frosty (approved):
**Test: Collaboration**

- [x] Verify you can add / remove collaborators
- [x] Verify that newly added collaborators get access to the shared document

**Test: Empty Notes**

- [x] Verify that the newly added note is deleted

**Test: Non Empty Notes**

- [x] Verify that the newly added is not deleted

Looks great! One small comment, but I think I understand the rationale now I've run through the testing steps – isBlank is to determine if it's a new, blank document? (and therefore delete it)

Simplenote/Classes/Note+Properties.swift
jleandroperez: Yes sir, that would be accurate!. Perhaps we should consider renaming `isBlank` to something more clear!

If it was never saved (and there is no content), it's safe to delete. The second you save it, `isTemporaryID` returns false, so it doesn't matter if it has (or doesn't have) any contents.

jleandroperez:
@frosty **Thank you!!!!**

PR #965, opened 2020-10-29
Show relevant excerpts while searching

### LINKED ISSUE #956: Relevant Excerpts
### Details:
When searching, Search Results should display excerpts relevant to the active search.

```
The notes list excerpt can be used to show a more relevant part of the note, for example, the first matching keyword for that search (see design below). I think we should be careful about defining the logic for this since it could cause some jumpiness in the text as you type.
```

Please sync with the team about the extraction algorithm, so that the behavior is consistent.
Ref. `p2XJRt-2bO-p2`

eshurakov:
In this PR we're updating note preview to show relevant excerpt while searching. 

Excerpt is generated from the text around the first encountered keyword. Text before and after the keyword is trimmed and only full words are used.

@jleandroperez Please take a look when you have time. Thank you!

Closes #956 

### Test
1. Open notes list
2. Search

- [x] You can see the word you're searching for in the excerpt
    - [x] Even if it's the first word in the body
    - [x] Even if it's the last word in the body
    - [x] Even if it's somewhere in the middle of the body
- [x] Searching for words in the beginning of the text doesn't prepend "..." to the excerpt
- [x] Searching for words in the middle of the text prepends "..." to the excerpt
- [x] Diactric insensitive search. For example, if you search for Sao, it finds and highlights São.

### Release
`RELEASE-NOTES.txt` was updated in https://github.com/Automattic/simplenote-ios/pull/965/commits/3f23cbd2850de6b0d362c36e84e2cb5c8df07d30 with:
> Searching for a note is showing relevant excerpts #956


Simplenote/NoteContentHelper.swift
```diff
@@ -0,0 +1,89 @@
...
+                                                                    options: [],
+                                                                    range: untrimmedBodyRange)
+        guard firstCharacterRange.location != NSNotFound else {
+            return firstCharacterRange
+        }
+
+        return rangeByTrimmingTrailingWhitespacesAndNewlines(in: content,
+                                                             firstCharacterLocation: firstCharacterRange.location,
+                                                             endRangeLocation: content.fullRange.length)
+    }
```
jleandroperez: I've noticed both, `titleRange` + `bodyRange` look **almost** the same.

1. Do you think we could merge them?
2. How about if, instead of returning a trimmed range, you just return the untrimmed range (and trim before sending the String to the UITableViewCell?, that'd be a bit simpler perhaps?)

jleandroperez: Second note: (IF possible), it'd be neat to replace `NSString+Condensing: generatePreviewStrings` with your new API, so that a single component builds the Title / Body

Simplenote/ExcerptMaker.swift
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        self.keywords = keywords
+
+        guard let keywords = keywords, !keywords.isEmpty else {
+            regexp = nil
+            return
+        }
+
+        regexp = NSRegularExpression.regexForExcerpt(withKeywords: keywords)
```
jleandroperez: Sugestion! You can decouple / split things a bit:

```
private var keywords: [String]? {
   didSet {
      guard oldValue != keywords else {
         return
      }
      refreshRegex()
   }
}
```

Simplenote/ExcerptMaker.swift
```diff
@@ -0,0 +1,61 @@
...
+                return content.nsString.substring(with: excerptRange)
+            }
+            return foldedContent.nsString.substring(with: excerptRange)
+        }()
+        excerpt = excerpt.nsString.replacingNewlinesWithSpaces()
+        if excerptRange.location > bodyRange.location {
+            excerpt = "…" + excerpt
+        }
+        return excerpt
+    }
```
jleandroperez: The resulting string will not contain diacritics (unless I'm missing something). Is that expected?

Simplenote/ExcerptMaker.swift
```diff
@@ -0,0 +1,61 @@
...
+            regexp = nil
+            return
+        }
+
+        regexp = NSRegularExpression.regexForExcerpt(withKeywords: keywords)
+    }
+
+    /// Generate and return excerpt from the note. Excerpt is based on keywords
+    ///
+    func excerpt(from note: Note) -> String? {
```
jleandroperez: Nipticky: `bodyExcerpt` perhaps?

Simplenote/ExcerptMaker.swift
```diff
@@ -0,0 +1,61 @@
...
+        let foldedContent = content.folding(options: [.diacriticInsensitive], locale: .current)
+
+        let bodyRange = NoteContentHelper.structure(of: foldedContent).body
+        guard bodyRange.location != NSNotFound else {
+            return nil
+        }
+
+        let excerptRange = regexp.rangeOfFirstMatch(in: foldedContent,
+                                                    options: [],
+                                                    range: bodyRange)
```
jleandroperez: Idea: how about if, instead of using `NoteContentHelper` to calculate the bodyRange we just do...

```
let bodyText = note.bodyPreview
let excerptRange = regexp.rangeOfFirstMatch(in: bodyText, options: [], range: bodyText.fullRange)
```

Simplenote/ExcerptMaker.swift
jleandroperez: Or `excerptForBody` ?

jleandroperez: (Thinking) Should we cache these strings somewhere?

Simplenote/ExcerptMaker.swift
```diff
@@ -0,0 +1,61 @@
...
+        guard excerptRange.location != NSNotFound else {
+            return note.bodyPreview
+        }
+
+        var excerpt: String = {
+            if foldedContent.nsString.length == content.nsString.length {
+                return content.nsString.substring(with: excerptRange)
+            }
+            return foldedContent.nsString.substring(with: excerptRange)
+        }()
```
jleandroperez: Nipticky: `nsString` builds a new Foundation String everytime it gets called. How about:

```
let foundationFolded = foldedContent.nsString
let foundationContent = content.nsString
let sourceString = foldedContent.length == content.length ? content : foldedContent
var excerpt = sourceString.substring(with: excerptRange)
```

(Or so!)

jleandroperez (changes requested):
Looking great sir!! few nipticky comments, LMK what you think!!

Simplenote/ExcerptMaker.swift
eshurakov: Good point!

Simplenote/NoteContentHelper.swift
eshurakov: 1. Managed to merge them :-)
2. My reason: I wanted to keep all the logic regarding what we consider "title" and "body" in one place
3. Agree with that!

Simplenote/ExcerptMaker.swift
eshurakov: It will return substring from the original string (containing diacritics), unless folded and original strings have different lengths.
I believe that in our case folding will produce the string of the same length as the original one, but just in case it won't I added a check.

It's a pity that NSRegularExpression doesn't support diacritic insensitive matching and the approach is not ideal.
Another options that we have:
1) include all possible diactric symbols when creating a pattern from keywords. For example: `test` will become `t[eèéêëēėę]st`
2) Don't use regexp. Split string by words and compare keywords to each word (using case and diactric insensitive mode). Stop when we have the first match. Get some words around it.

What do you think?

Simplenote/NoteContentHelper.swift
```diff
@@ -0,0 +1,82 @@
...
+
+        return (
+            title: titleRange,
+            body: bodyRange
+        )
+    }
+
+    private static func trimmedTextRange(in content: String?, startingFrom startLocation: Int, endAtNewline: Bool) -> NSRange {
+        guard let content = content else {
+            return .notFound
```
jleandroperez: How about guarding in `structure(of content:`? (That would save quite a lot of processing!)

Simplenote/NoteContentHelper.swift
```diff
@@ -0,0 +1,82 @@
...
+        }
+        let fullRange = content.nsString.fullRange
```
jleandroperez: Since `content.nsString` is invoked in few other spots, maybe assigning that to a constant is worthy?

Simplenote/NoteContentHelper.swift
```diff
@@ -0,0 +1,82 @@
...
+                let newlineRange = content.nsString.rangeOfCharacter(from: .newlines,
+                                                                     options: [],
+                                                                     range: newlineSearchRange)
+                if !newlineRange.isNotFound {
+                    return newlineRange.location
+                }
+            }
+
+            return fullRange.length
+        }()
```
jleandroperez: Q: Do you think it'd be worthy extracting this to an API?

```
extension NSString {
    func rangeOfFirstCharacter(_ character: Character) -> NSRange?
}
```

Simplenote/NoteContentHelper.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    /// Returns structure of the content: range of title and body
+    ///
+    /// - Parameters:
+    ///     - content: note content
+    ///
+    static func structure(of content: String?) -> (title: NSRange, body: NSRange) {
+        let titleRange = trimmedTextRange(in: content, startingFrom: 0, endAtNewline: true)
+        let bodyRange: NSRange
+        if titleRange.isNotFound {
```
jleandroperez: Q: How about `NSRange?` (and drop the sentinel?)

Simplenote/NoteContentHelper.swift
```diff
@@ -0,0 +1,82 @@
...
+
+        // Look for the first character ignoring whitespaces and newlines
+        let firstCharacterSearchRange = NSRange(location: startLocation,
+                                                length: fullRange.length - startLocation)
+        let firstCharacterRange = content.nsString.rangeOfCharacter(from: CharacterSet.whitespacesAndNewlines.inverted,
+                                                                    options: [],
+                                                                    range: firstCharacterSearchRange)
+        guard !firstCharacterRange.isNotFound else {
+            return .notFound
+        }
```
jleandroperez: Q: Is this trimming improving performance?

jleandroperez (commented):
Sending few nipticky notes!! 

Regarding **RegEx**, for the record, discussed over Slack.
Approach entirely up to you!

Simplenote/NoteContentHelper.swift
eshurakov: Good point 👍

Simplenote/NoteContentHelper.swift
eshurakov: I've extracted it to a String extension, I think it became better. Thanks for the suggestion!

Simplenote/NoteContentHelper.swift
eshurakov: Do you mean to use and return an optional NSRange? I thought about it but wanted to be consistent with other APIs. From what I see NSRange is never used as an optional value, even in swift.

Simplenote/NoteContentHelper.swift
eshurakov: Probably not, but I also think it's not affecting performance much. And if we do it here, we wouldn't need to do it when showing the string in a label.

eshurakov:
Hey @jleandroperez, could you take another look at this PR please? Thanks!

Simplenote/Classes/NSString+Condensing.h
```diff
@@ -4,10 +4,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface NSString (Condensing)
 
-/// Generates string with fewer whitespaces and special characters. Good for previews in UITableViews
-///
-- (void)generatePreviewStrings:(void (^)(NSString *titlePreview, NSString * _Nullable bodyPreview))block;
-
```
jleandroperez, line 10: 🖖

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -245,15 +245,16 @@ - (void)themeDidChange
...
         return;
     }
     
     NSString *searchText = _noteEditorTextView.text;
     
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
-        
-        self.searchResultRanges = [searchText rangesForTerms:self.searchString];
+
+        self.searchResultRanges = [self searchResultRangesIn:(searchText ? searchText : @"")
```
jleandroperez:
```suggestion
        self.searchResultRanges = [self searchResultRangesIn:(searchText ?: @"")
```

jleandroperez: Old school trick! 🔥

Simplenote/Classes/String+Simplenote.swift
```diff
@@ -45,4 +45,125 @@ extension String {
...
+            return nil
+        }
+
+        let range: Range<String.Index> = {
+            if backwards {
+                return startIndex..<startLocation
+            }
+
+            return startLocation..<endIndex
+        }()
```
jleandroperez: Hooow about:

```
let range = backwards ? startIndex..<startLocation : startLocation..<endIndex
```

(Less lines, one less context switch)

jleandroperez: Huh I'm not 100% sure if an inline anonymous function like this one would or would not produce a stack dump? I'd say yes, buuut I haven't checked the ASM!!

Simplenote/Classes/String+Simplenote.swift
```diff
@@ -45,4 +45,125 @@ extension String {
...
+
+        let upperBound: String.Index = {
+            if trailingLimit == 0 {
+                return range.upperBound
+            }
+            return trailingWordsRange.last?.upperBound ?? lastMatch.upperBound
+        }()
+
+        return ContentSlice(content: self, range: lowerBound..<upperBound, matches: matchingWordsRange)
+    }
```
jleandroperez, line 172: Would it be possible to add unit tests for the APIs in this file?

- droppingPrefix
- locationOfFirstCharacter 
- contentSlice

Simplenote/ContentSlice.swift
```diff
@@ -0,0 +1,52 @@
...
+        let newRange = newContent.startIndex..<newContent.endIndex
+        let offset = content.distance(from: content.startIndex, to: range.lowerBound)
+        let newMatches: [Range<String.Index>] = matches.map {
+            let lower = content.index($0.lowerBound, offsetBy: -offset)
+            let upper = content.index($0.upperBound, offsetBy: -offset)
+            return lower..<upper
+        }
+
+        return ContentSlice(content: newContent, range: newRange, matches: newMatches)
+    }
```
jleandroperez: Would it be possible to unit test this one?

Simplenote/NoteContentHelper.swift
```diff
@@ -0,0 +1,56 @@
+import Foundation
+
+// MARK: - NoteContentHelper
+//
+struct NoteContentHelper {
+
+    private init() {
+
+    }
```
jleandroperez: Trick! (standard across the division!):

```
enum NoteContentHelper { }
```

That way it'd be impossible to initialize a NoteContentHelper (unless... you forget not to add a `case`!).

That'd be why there are so many `enum Metrics` / `enum Constants` / `enum Settings` here and there!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -834,6 +834,30 @@ extension SPNoteListViewController {
...
+        let excerpt = content.contentSlice(matching: keywords,
+                                           in: bodyRange,
+                                           leadingLimit: Constants.excerptLeadingLimit,
+                                           trailingLimit: Constants.excerptTrailingLimit)
+
+        let shouldAddEllipsis = excerpt.range.lowerBound > bodyRange.lowerBound
+        let excerptString = (shouldAddEllipsis ? "…" : "") + excerpt.normalized.content
+
+        return excerptString.replacingNewlinesWithSpaces()
+    }
```
jleandroperez: Question: (Thinking aloud!) do you think it'd make sense to move this API into a Note Extension?

```
extension Note {
    func bodyExcerpt(keywords: [String]?) -> String? { }
}
```

jleandroperez: + Would it be possible to unit test this one?

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -834,6 +834,30 @@ extension SPNoteListViewController {
...
+//
+private extension SPNoteListViewController {
+
+    func bodyExcerpt(from note: Note, withKeywords keywords: [String]?) -> String? {
+        guard let keywords = keywords, !keywords.isEmpty, let content = note.content else {
+            return note.bodyPreview
+        }
+
+        guard let bodyRange = NoteContentHelper.structure(of: content).body else {
+            return note.bodyPreview
```
jleandroperez: Q: in this case, wouldn't it be the same as doing `return nil` ?

jleandroperez (approved):
Sending few comments your way sir! Some of them are just ideas of (potentially) code relocation.

If possible, please add as many unit tests as possible for all the new extension API(s).

Other than that, this PR looks awesome. Performance wise it's super smooth (tested on an iPhone 6s in an account with +500 notes).

Feel free to :shipit: when ready!!

---

(Edit) Please also make sure that all of the new code is inside the `/Simplenote/Classes` folder! (there are few files one level above!)

Simplenote/ContentSlice.swift
jleandroperez: (Plus) please relocate this file to `Simplenote/Classes`

Simplenote/NoteContentHelper.swift
jleandroperez: (Plus) please relocate this file to `Simplenote/Classes`

Simplenote/Classes/SPNoteEditorViewController.m
eshurakov: Indeed, thanks!

Simplenote/Classes/String+Simplenote.swift
eshurakov: Not sure why I went with an anonymous function here, doesn't make much sense 🤦 
Thank you for the suggestion, looks beautiful!

Simplenote/NoteContentHelper.swift
eshurakov: Good point, thanks!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
eshurakov: > do you think it'd make sense to move this API into a Note Extension?

Looks better, thank you!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
eshurakov: Yes! Changed :-)

Simplenote/ContentSlice.swift
eshurakov: I removed most of the logic from it and it became a quite simple model :-) Don't think it is needed to unit test this class anymore.

Simplenote/Classes/String+Simplenote.swift
eshurakov: Sure, good point! Added tests 👍

eshurakov:
Thank you for the review @jleandroperez! Added a bunch of unit tests and fixed files locations 👍 
Do you mind taking another peek? Thanks!

jleandroperez (approved):
Nice work sir!! go for it!!

:shipit:

eshurakov:
Thank you @jleandroperez !

Issue #921, opened 2020-10-15
Editor: Odd Text Selection


jleandroperez:
### Steps:
1. Open a long document
2. Tap anywhere to reveal the keyboard
3. Tap anywhere at the bottom of the visible text

As a result, the cursor will be relocated, and the pressed line will be slighly scrolled upwards.
Thank you @aerych for spotting this!!

I've tracked this issue down to the [SPEditorTextView.adjustedContentInsets overridden API](https://github.com/Automattic/simplenote-ios/blob/develop/Simplenote/Classes/SPEditorTextView.m#L174).

References: #806 #916

jleandroperez:
### Update:
Text Selection is still a bit odd. Autoscroll should only happen whenever the selected row is *partially* not visible.

PR #973, opened 2020-11-03
Fixing Text Selection: Mark II

jleandroperez:
### Fix
In this PR we're fixing a glitch that caused the TextView's scroll offset to jump when pressing anywhere in the last ~5 lines of text.

@eshurakov May I bug you with this one?
**Small PR**... looots of things to check (apologies about that!!)

Ref #921

### Test: Selection
1. Open a long document
2. Tap anywhere to reveal the keyboard
3. Tap anywhere at the bottom of the visible text

 - [x] Verify that the contentOffset doesn't get pushed upwards **unlesss it's the last 1-2 lines!**

**Note**
Please compare the behavior against develop: the Content Offset jumps when pressing the last ~5 lines!

### Test: Regressions
- [x] Verify dragging the cursor all over works as expected (and it's not impossible to control!)
- [x] Verify that scrolling all the way down and pressing over the Tags Editor does not cause the Scroll Offset to jump
- [x] Verify that the Scroll Indicator clips right before the area where Tags Editor begin

### Test: Search
1. Launch the app on an iPhone device
2. Enter a search keyword that yields any results
3. Press over any search result
4. Scroll all the way down, until the Tags Editor is revealed
5. Press Done to dismiss

- [ ] Verify that Keyword Matches are properly highlighted (and that they don't look off!)

### Test: Voiceover
1. Launch Simplenote on a device with Safe Area (iPhone X / 11)
2. Add a note with (lots) of text, so that it causes scroll
3. Enable voiceover (hey siri voiceover on!)

- [x]  Verify that the Tags View now shows up locked at the bottom

### Test: iPad
1. Launch Simplenote on an iPad Device
2. Toggle a Split Keyboard

- [x]  Verify that text shows up all over, and does not clip above the Keyboard
- [x]  Verify that switching to floating keyboard and moving it all over does not cause the Text to disappear

### Release
`RELEASE-NOTES.txt` was updated in 1428131 with:
 
> Fixed a bug that affected Text Selection #973

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,6 @@
 4.28
 -----
+-   Fixed a bug that affected Text Selection #973
  
 4.27
 -----
--- a/Simplenote/Classes/SPEditorTextView.h
+++ b/Simplenote/Classes/SPEditorTextView.h
@@ -22,15 +22,6 @@
 @property (nonatomic, strong) UIFont *checklistsFont;
 @property (nonatomic, strong) UIColor *checklistsTintColor;
 
-/// Allows you to set `contentInset.bottom`
-///
-@property (nonatomic, assign) CGFloat bottomContentInset;
-
-/// Allows you to set `scrollIndicatorInsets` while considering (automatically) an extra padding, to be allocated by the TagsEditor
-///
-@property (nonatomic, assign) CGFloat bottomScrollerInset;
-
-
 - (void)scrollToBottomWithAnimation:(BOOL)animated;
 - (void)scrollToTop;
 - (void)processChecklists;
--- a/Simplenote/Classes/SPEditorTextView.h
+++ b/Simplenote/Classes/SPEditorTextView.h
@@ -22,6 +22,15 @@
 @property (nonatomic, strong) UIFont *checklistsFont;
 @property (nonatomic, strong) UIColor *checklistsTintColor;
 
+/// Allows you to set `contentInset.bottom`
+///
+@property (nonatomic, assign) CGFloat bottomContentInset;
+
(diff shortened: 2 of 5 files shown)


eshurakov:
Looks good! Some comments:

> Verify that the contentOffset doesn't get pushed upwards unlesss it's the last 1-2 lines!  

It still doesn't feel right to me :-) I think I would prefer if it is pushed up only to make partially visible line fully visible.


> Verify that the Scroll Indicator clips right before the area where Tags Editor begin

This looks very strange to me. Shouldn't scroll indicator go to the keyboard / safe area. I think the version from develop is better :-)


> Verify that Keyword Matches are properly highlighted (and that they don't look off!)

The baseline is a bit off. And there is an issue with highlights when note has checklists. Both issues are in develop as well.
Screenshot 2020-11-04 at 13 00 05

eshurakov (approved):

jleandroperez:
@eshurakov Let's go with your instinct on the Scroll Indicator, I wasn't 100% sure about it!

Regarding the autoscroll behavior, I'll check with @SylvesterWilmott! (current behavior should match 100% the way the system works, or other editors).

As per the Highlight baseline, I'll file another ticket, thank you!

Mind taking another peel whenever you have a few? (Scroll Indicator bottom inset should be gone)
(Thanks in advance!!)

eshurakov:
> current behavior should match 100% the way the system works, or other editors

I was comparing with Notes, but maybe they do something different. Anyway, the changes are great and autoscroll behaviour is much better than before.

🚢 :shipit:

jleandroperez:
Merging this one (BUT) keeping the main issue open. I'll do my best to smoothen out the pendings!

**Thank you** @eshurakov !!

PR #975, opened 2020-11-03
Remove canOpenURL check for https scheme

### LINKED ISSUE #963: Help button not working on iOS if Default Browser is not Safari
The Help option does not work if you don't have the Safari Browser set as your primary browser

### Expected
I  expected the Help option to open the Help page on the browser regardless of which browser you have set as your primary. 

### Observed
The help page will only open if Safari is the main browser. Tested with Chrome and FF as the default browser and in both cases, the Help option does not do anything. 

### Reproduced

> 1. Open the Simplenote App
> 2. Tap on the Settings option
> 3. Scroll down and tap on the Help option

I tested this on an iPad 7th Gen using iOS 14.0.1
Simplenote version 4.26

@reginabally  has tested this on iPhone SE, iOS 14.1 and Simplenote iOS 4.26.

The user has tried on the following:

Iphone 8
iOS 14.1
Simple Note App Version 4.26

3439709-zen

eshurakov:
### Fix
In this PR we remove `canOpenURL` check for `https` urls. This check was preventing opening urls if default browser is not Safari.

@jleandroperez Quick one! Thanks.

Closes #963 

### Test
1. Change default browser
2. Open Simplenote -> Settings
3. Tap Help

- [x] Help page is opened in a browser from step 1

### Release

### DIFF
--- a/Simplenote/Classes/SPSettingsViewController.m
+++ b/Simplenote/Classes/SPSettingsViewController.m
@@ -581,10 +581,7 @@ typedef NS_ENUM(NSInteger, SPOptionsDebugRow) {
             switch (indexPath.row) {
                 case SPOptionsHelpRowTitle: {
                     NSURL *url = [NSURL URLWithString:@"https://simplenote.com/help"];
-
-                    if ([[UIApplication sharedApplication] canOpenURL:url]) {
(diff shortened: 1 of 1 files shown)


jleandroperez (approved):
Nice catch!!

:shipit:

PR #982, opened 2020-11-04
Fixing Text Selection: Mark III

### LINKED ISSUE #921: Editor: Odd Text Selection

jleandroperez:
### Fix
In this PR we're (further) adjusting the Autoscroll Behavior, so that **only partially visible lines** cause the TextView to autoscroll.

@eshurakov **Last round!!!** Please LMK if this time autoscroll feels right. I *think* this is what we were after.

Thanks in advance!!

Closes #921

### Test: Regressions
- [x] Verify dragging the cursor all over works as expected (and it's not impossible to control!)
- [x] Verify that scrolling all the way down and pressing over the Tags Editor does not cause the Scroll Offset to jump
- [ ] Verify that the Scroll Indicator clips right before the area where Tags Editor begin

### Test: Search
1. Launch the app on an iPhone device
2. Enter a search keyword that yields any results
3. Press over any search result
4. Scroll all the way down, until the Tags Editor is revealed
5. Press Done to dismiss

- [x] Verify that Keyword Matches are properly highlighted (and that they don't look off!)

### Test: Voiceover
1. Launch Simplenote on a device with Safe Area (iPhone X / 11)
2. Add a note with (lots) of text, so that it causes scroll
3. Enable voiceover (hey siri voiceover on!)

- [x]  Verify that the Tags View now shows up locked at the bottom

### Test: iPad
1. Launch Simplenote on an iPad Device
2. Toggle a Split Keyboard

- [x]  Verify that text shows up all over, and does not clip above the Keyboard
- [x]  Verify that switching to floating keyboard and moving it all over does not cause the Text to disappear

### Release

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,7 +1,6 @@
 4.28
 -----
 -   Fixed a bug that affected Text Selection #973
- 
 -   Fixed a bug that toggled Edition while dragging the Editor #972
 
 4.27
--- a/Simplenote/Classes/SPEditorTextView.m
+++ b/Simplenote/Classes/SPEditorTextView.m
@@ -168,8 +168,9 @@ NSInteger const ChecklistCursorAdjustment = 2;
     CGFloat width       = self.bounds.size.width - self.safeAreaInsets.left - self.safeAreaInsets.right;
     CGFloat height      = CGRectGetHeight(self.tagView.frame);
 
-    // When the Keyboard is absent we'll always honor the Safe Area Insets
-    CGFloat paddingY    = MAX(self.contentInset.bottom, self.safeAreaInsets.bottom);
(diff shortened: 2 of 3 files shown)


eshurakov (approved):
It's awesome!

:shipit:

jleandroperez:
Thank you sir!!

PR #997, opened 2020-11-09
Remove unused "Distribution AdHoc" build configuration

mokagio:
While looking into extracting the build settings in `xcconfig`, I noticed this build configuration was unused: there is no mention of it in the `Fastfile` or in CI.

Note that the Pods xcconfig files were deleted manually, running `bundle exec pod install` didn't do it.

### Test
This is all unused code, so if the CI build passes we're good to merge

### Review

### Release


eshurakov (approved):
Looks good, thank you!

:shipit:

jleandroperez (approved):
Thank you for doing this @mokagio!!

:shipit: as well!

PR #998, opened 2020-11-09
Remove orphan reference to SPTagsListViewController from project

mokagio:
When Xcode touched the project file to remove the unused ad hoc build configuration in 605014e (#997), it produced this diff.

There is no other reference of SPTagsListViewController in the project, nor any usage of the type in the codebase. I guess it's something left over from when the view controller was removed 🤷‍♂️.

### Test
This is dead code; if the build on CI passes we're good

### Review

### Release


eshurakov (approved):
Thank you for cleaning it up! 

:shipit:

Issue #938, opened 2020-10-21
In-App Notifications: Snackbar


jleandroperez:
### Details:
We'd need to build an In-App Snackbar, in charge of displaying quick notifications which may or may not contain an associated action.

### Requirements:
- Should be visible even if the presenting ViewController gets dismissed
- Only one Snackbar can be displayed at a time.
- May contain a single action.
- Should disappear when not interacted with.
- They should appear on screen sliding up from the bottom and then fade out.
- Notifications not containing an action should be displayed for 1.5s.
- Notifications containing an action should be displayed for 2.75s.

### Notes:
Implementing an In-App Snackbar -which runs unaffected by the ViewControllers onScreen- can be solved by means of a secondary UIWindow instance.

That's precisely the technique used by [WordPress's Notice Presenter](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift).

Suggestion is to split this task into few steps steps:

1. Design the Public API this component would have. A good lead would be [UIAlertController](https://developer.apple.com/documentation/uikit/uialertcontroller) itself, please bear in mind our needs are much simpler!
2. Build a proof of concept that displays a simple Snackbar, on top of everything (and analyze the helper UIWindow mechanism)
3. Add support for associated actions
4. Skin all the things: Dark Mode support, ensure Dynamic Type works well 
5. Analyze the "Device Rotation" Scenario

### Mockups:
![with-action](https://user-images.githubusercontent.com/1195260/96787154-abff7f80-13c7-11eb-958a-c4a0530cb3ad.gif)
![without-action](https://user-images.githubusercontent.com/1195260/96787162-af930680-13c7-11eb-84e4-5c1fb6344b5d.gif)

PR #999, opened 2020-11-09
Snackbar: First steps at potential implementation

a8ck3n:
### Details
Part of #938, **In-App Notifications: Snackbar**. This is the initial plumbing demonstrating one possible way to present Snackbars. Covers the view lifecycle, display, and clean up. Basic support for dynamic text.

Still needs:

- [x] 1. Proper theming for light & dark mode.
- [x] 2. Re-sizing the view based on message length. The original Issue shows the containing view only taking up the width necessary to show the message + the action button's title. Need some clarity on clipping vs wrapping longer messages.
- [x] 3. Additional initializer to wire up the action button. Currently the button just logs when clicked. Additional initializer would include a closure and execute that when clicked.
- [x] 4. Prevents additional notices if one is currently displayed.
- [x] 5. Needs to set duration based on presence of action button. Currently just puts up the notice for 2 seconds regardless of type.
- [ ] 6. Final colours for Dark & Light mode.
- [ ] 7. Haptic feedback.
- [ ] 8. Rotation support (appears to work but not throughly tested yet).
- [ ] 9. Remove debug logs showing presentation flow.

Not implemented but might be nice to have:

- [ ] Queuing mechanism if we wanted to chain notices.

### Test
1. Build and run.
2. With a note visible, click the Note Info button to see a simple message. Click the Note Options button to see a message with an action button.
3. See the Snackbar animate in then fade out.

### Release
**Do not merge!** This is a draft of work in progress.

### Comments

The basic architecture is built around the `Snackbar` and `SnackbarPresenter` classes with `SnackbarViewController` for XIB-based loading of the view. Only the `Snackbar` class is needed to present a notification.

```
let snackbar = Snackbar(message: "Some message")
snackbar.show()
```

For a Snackbar with action button:
```
let snackbar = Snackbar(message: "Snackbar with action!", actionTitle: "Boom!") {
    print("Callback code executed!")
}

```

**Light Mode example:**

**Dark Mode with larger text:**

### DIFF
--- a/Simplenote/Snackbar.swift
+++ b/Simplenote/Snackbar.swift
@@ -10,9 +10,9 @@ import Foundation
 
 class Snackbar {
     
-    private var message: String
+    fileprivate let message: String
     
-    init(message: String = "Test message") {
+    init(message: String) {
         self.message = message
     }
     
@@ -81,13 +81,14 @@ class SnackbarPresenter {
     private func prepareViewFromXIB() -> UIView {
         
         let vc = SnackbarViewController()
-        let view = vc.view!
         
+        let view = vc.view!
         let frame = view.frame
         let newFrame = CGRect(x: 0, y: 0, width: frame.width - 30, height: frame.height)
         vc.view.frame = newFrame
         
         viewController = vc
+        vc.messageLabel.text = snackbar?.message
         
         return view
     }
--- a/Simplenote/Snackbar.swift
+++ b/Simplenote/Snackbar.swift
@@ -112,19 +112,43 @@ class SnackbarPresenter {
 			view.center.y = view.center.y - offset - viewHeight
 		}) { _ in
 			print("Animation complete.")
+			
+			// Using the delay param results in the UIButton not receiving inputs.
+			// Using the asyncAfter allows the button to work correctly.
+			
+//			self.removeView(view)
+			self.removeViewAsync(view)
+		}
+	}
+	
+	func removeView(_ view: UIView) {
+		
+		// Results in action button not working.
+		UIView.animate(withDuration: 0.66, delay: 2.4, options: []) {
+			view.alpha = 0
+		} completion: { (Bool) in
+			view.removeFromSuperview()
+			print("Done fade.")
 
-			DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) {
-				print("Hiding snackbar")
+			self.snackbar = nil
+			self.viewController = nil
+		}
+	}
+	
+	func removeViewAsync(_ view: UIView) {
+		
+		// Allows action button to accept inputs.
+		DispatchQueue.main.asyncAfter(deadline: .now() + 2.4) {
+			print("Hiding snackbar")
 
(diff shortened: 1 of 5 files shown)


Simplenote/Snackbar.swift
```diff
@@ -0,0 +1,128 @@
...
+				} completion: { (Bool) in
+					view.removeFromSuperview()
+					print("Done fade.")
+					
+					self.snackbar = nil
+					self.viewController = nil
+				}
+			}
+		}
+	}
```
jleandroperez: Few notes!

1. Please review encapsulation: It'd be neat if properties not meant for external usage were protected
2. Would it be possible to convert to **Autolayout**?
3. Please review Tabs / Spaces (the project itself uses Spaces rather than Tabs!)
4. Please use constants when appropriate (**UIKitConstants** may be of help!)
5. Rather than animating `asyncAfter` please consider using the `UIView.animate`'s **delay** parameter
6. What would happen if the Window rotates?
7. Please bear in mind that the action itself is optional

Simplenote/Snackbar.swift
```diff
@@ -0,0 +1,128 @@
...
+//  Copyright © 2020 Automattic. All rights reserved.
+//
+
+import Foundation
+
+class Snackbar {
+	
+	var message: String
+	
+	static var presenter = SnackbarPresenter()
```
jleandroperez: Please review this singleton (a better spot would be, perhaps, the SnackbarPresenter class itself)

jleandroperez (changes requested):
Hi there @a8ck3n,

Thanks for the PR! Sending your way few notes. Public API looks good BTW!

Simplenote/Snackbar.swift
a8ck3n: Not sure what I was thinking here. I've moved it to the appropriate place now!

Simplenote/Snackbar.swift
a8ck3n: Thank you for these! I was able to address some of them in my changes today with the exceptions being:

- 2. **Move to Auto Layout** — I haven't had a chance to work on this yet. I thought it would make sense to have the view properly configured (action vs no action) before digging into this. One question, do you recommend moving away from UIStackView as part of tackling this?
- 5. **Try without `asyncAfter`** — I tested this but using `UIView.animate`'s **delay** parameter caused the button to stop accepting input. Additionally, after wiring up the button, the code path changed a little to allow short-circuiting the timeout. If I've overlooked or misunderstood something here, please let me know and I'll have another look.
- 6. **Rotation** — Currently the Snackbar doesn't respond to rotation at all. I'm thinking this would be handled by adding some Auto Layout constraints to the view once the zoom-in animation completes but I haven't tried this yet.

I've updated the issue details to show the new API (new initializer) and added an updated GIF to show both styles in action. This still needs Auto Layout, theming for Dark/Light modes, and possibly a queuing mechanism if we want to allow queuing up of Snackbar notifications.

jleandroperez:
Closing this one for now. Thank you Kevin!

Issue #914, opened 2020-10-12
Interlink Autocomplete


jleandroperez:
### Details
- [x] Display Interlink Suggestions whenever the user enters `[`
- [x] Support Rotation / Size Class changes
- [x] TableView Padding: `textView.textContainer.lineFragmentPadding`
- [x] Resize Interlink View dynamically
- [x] Insert selected Interlink at the cursor's location 

### Pendings:
- [x] Autoscroll when typing a new character might cause the Autocomplete UI to flip Top / Bottom
- [x] When switching from Bottom / Top, prefer the same position if possible
- [x] Dismiss when pressing the Navigation Bar
- [x] Well actually we can always remove the bottom separator from the last child
- [x] iPad: Align cursor
- [x] Screen Coordinates
- [x] Landscape Mode
- [x] Dismiss on scroll
- [x] Drop KVO!
- [x] Tracks
- [x] LTR
- [x] Bugfix: Select won't dismiss Interlink
- [x] Remote Updates
- [x] Review Blur
- [x] Adjust Height based on available area
- [x] Lock Scroll when possible

### Bugs:
- [x] Bugfix: Stability Issues / NSRange conversion
- [x] Bugfix: TagView not scrolling the second the bottom text hits the topY

### Mockups
<img width="300" src="https://user-images.githubusercontent.com/1195260/95781586-83c7a080-0ca4-11eb-93a8-f430935be0ee.png">

PR #1002, opened 2020-11-10
Interlinks Autocomplete: Mark I

jleandroperez:
### Fix
In this PR we're implementing **InterlinkResultsController**. This component will be in charge of:

1. Filtering Notes that contain a specified Interlink keyword
2. Excluding a specified Entity ID (useful to exclude the "current" document, and thus, prevent loop links)

We're also adding capabilities to limit the number of results.

Please bear in mind this is part of a larger feature, splitting this into review friendly pieces.

Ref. #914
cc @eshurakov (Thanks in advance!!)

### Test
- [x] Verify the code makes sense
- [x] Verify the new unit tests are green!
- [x] Yay!

### Release

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -88,7 +88,6 @@
 		46A3C99B17DFA81A002865AE /* Tag.m in Sources */ = {isa = PBXBuildFile; fileRef = 467D9C641788A54900785EF3 /* Tag.m */; };
 		46A3C99C17DFA81A002865AE /* NSString+Bullets.m in Sources */ = {isa = PBXBuildFile; fileRef = E2C5C73F17CAD2FD003B9795 /* NSString+Bullets.m */; };
 		46A3C99D17DFA81A002865AE /* SPAddCollaboratorsViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = E25C39CF17A41F3400B2591A /* SPAddCollaboratorsViewController.m */; };
-		46A3C9A117DFA81A002865AE /* SPTagsListViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = E277B5E8179EDAFB0095CD24 /* SPTagsListViewController.m */; };
 		46A3C9A217DFA81A002865AE /* DTPinDigitView.m in Sources */ = {isa = PBXBuildFile; fileRef = E2911E7D17B49F4600F13AD9 /* DTPinDigitView.m */; };
 		46A3C9A317DFA81A002865AE /* SPTagEntryField.m in Sources */ = {isa = PBXBuildFile; fileRef = E277B60017A063020095CD24 /* SPTagEntryField.m */; };
 		46A3C9A717DFA81A002865AE /* DTPinLockController.m in Sources */ = {isa = PBXBuildFile; fileRef = E2911E8117B49F4600F13AD9 /* DTPinLockController.m */; };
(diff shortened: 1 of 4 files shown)


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -88,7 +88,6 @@
 		46A3C99B17DFA81A002865AE /* Tag.m in Sources */ = {isa = PBXBuildFile; fileRef = 467D9C641788A54900785EF3 /* Tag.m */; };
 		46A3C99C17DFA81A002865AE /* NSString+Bullets.m in Sources */ = {isa = PBXBuildFile; fileRef = E2C5C73F17CAD2FD003B9795 /* NSString+Bullets.m */; };
 		46A3C99D17DFA81A002865AE /* SPAddCollaboratorsViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = E25C39CF17A41F3400B2591A /* SPAddCollaboratorsViewController.m */; };
-		46A3C9A117DFA81A002865AE /* SPTagsListViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = E277B5E8179EDAFB0095CD24 /* SPTagsListViewController.m */; };
```
jleandroperez: I've just noticed Xcode did this. I'll take care of fixing merge conflicts (if they do arise!).

Thank you!

Simplenote/Classes/InterlinkResultsController.swift
```diff
@@ -0,0 +1,77 @@
...
+        resultsController.predicate = NSPredicate.predicateForNotes(deleted: false)
+        try? resultsController.performFetch()
+    }
+
+    /// Returns the collection of Notes filtered by the specified Keyword in their title, excluding a specific ObjectID
+    /// - Important: Returns `nil` when there are no results!
+    ///
+    func searchNotes(byTitleKeyword keyword: String, excluding excludedID: NSManagedObjectID?) -> [Note]? {
+        filter(notes: resultsController.fetchedObjects, byTitleKeyword: keyword, excluding: excludedID)
+    }
```
eshurakov, line 34: Does it search with only one keyword? I though maybe it's nice to have a behaviour similar to the main search: split by words and search for all of them.

Simplenote/Classes/InterlinkResultsController.swift
```diff
@@ -0,0 +1,77 @@
...
+    ///     - It's easier and anyone can follow along!
+    ///
+    func filter(notes: [Note], byTitleKeyword keyword: String, excluding excludedID: NSManagedObjectID?) -> [Note]? {
+        let normalizedKeyword = keyword.folding(options: [.diacriticInsensitive, .caseInsensitive], locale: nil)
+        var output = [Note]()
+
+        for note in notes where note.objectID != excludedID {
+            note.ensurePreviewStringsAreAvailable()
+            guard let normalizedTitle = note.titlePreview?.folding(options: [.diacriticInsensitive, .caseInsensitive], locale: nil),
+                  normalizedTitle.contains(normalizedKeyword)
```
eshurakov: I believe you can do it without folding keyword and title.
```
guard note.titlePreview.range(of: keyword, options: [.caseInsensitive, .diacriticInsensitive], range: note.titlePreview.startIndex..<note.titlePreview.endIndex, locale: Locale.current) != nil else {
    continue
}
```

eshurakov (approved):
Looks good! 👍

Simplenote/Classes/InterlinkResultsController.swift
jleandroperez: You're **SO** right!! updating shortly!! I can't believe I missed that. Thank you for pointing it out!!

Simplenote/Classes/InterlinkResultsController.swift
jleandroperez: MUCH better!! thanks for the tip!

jleandroperez:
@eshurakov Mind taking another peek, just for safety?

Thanks in advance, loved the feedback!!

jleandroperez:
### Note:
I've missed a huge detail. When searching for a document, if we search something that contains **any** of the words the user entered, the filtering is a bit shaky.

Why: Suppose your note's title contains `It's a long way to the top`.
When your document looks like `[it's a`, we'd be showing documents that contain both `it's` and `a` (in my case that translates to... way above 10 documents!).

Let's revert to the "single word" behavior (plus it matches macOS!). Thank you!

Simplenote/Classes/String+Simplenote.swift
```diff
@@ -23,6 +23,12 @@ extension String {
 //
 extension String {
 
+    /// Returns the Range enclosing all of the receiver's contents
+    ///
+    var fullRange: Range<String.Index> {
+        self.startIndex ..< self.endIndex
```
eshurakov: Do you need `self` here? 🤔

jleandroperez: Thank you!!

jleandroperez:
Merging this one. Thank you @eshurakov!!

PR #1003, opened 2020-11-10
Interlinks Autocomplete: Mark II

jleandroperez:
### Details
In this PR we're implementing the core logic of Interlinks:

- **InterlinkProcessor:** In charge of determining when the Interlink Autocomplete must be shown / hidden
- **InterlinkViewController:** Renders Autocomplete suggestions above / below the cursor

**Note:** Nothing is expected to happen, at this stage, when you press over any row.

**Apologies about the size of the PR!!!** I've tried splitting this further, but testing without an actual UI makes no sense.

cc @eshurakov **Thank yoou sir!!!**

Ref. #914

---

### Test: Orientation
1. Open the Editor and add a note with 6 lines
2. Place the cursor in the 6th line
3. Type `[keyword` (with `keyword` being a word contained in any of your note's titles)

- [x] Verify Autocomplete shows up below the cursor's location
- [x] Verify that scrolling over the TextView below causes the Autocomplete window to follow along
- [x] Verify that if you scroll enough, the Autocomplete UI shows **below the Navigation Bar**
- [x] Verify that if your (updated) results yield just one row, the Autocomplete View doesn't jump above the keyword
- [x] Verify that if there are (way more) lines above, typing `[keyword` displays Autocomplete **above** the cursor

### Test: Filtering
1. Add a new note
2. In the new note's title, enter a word that's not present anywhere else. Suggestion `TEST999` perhaps?
3. Place the cursor in a newline
4. Type `[TEST999`

- [x] Verify Autocomplete does not show up (we intend to prevent circular references!)

### Test: Adaptivity
1. Open the Editor and type `[keyword` to trigger Autocomplete

- [x] Verify that rotating the device causes Autocomplete to disappear / reappear in the right location

### Test: UX
1. Open the Editor and type `[keyword` to trigger Autocomplete

- [x] Verify that a maximum of `3.5` rows always show up (we intend to avoid huge tables!!)
- [x] Verify Autocomplete reacts well to Dark / Light mode
- [ ] Verify Autocomplete reacts well to Dynamic Fonts
- [x] Verify that, on highlight, Autocomplete's First and Last rows get rounded corners

### Test: Dismissal
Please trigger the Autocomplete window, and verify it can be dismissed as follows:

- [x] Verify that pressing anywhere else in the TextView causes Autocomplete to get dismissed
- [x] Verify that pressing over any of the Navigation Bar buttons causes interlink to dismiss
- [x] Verify that pressing Back causes Interlink to get dismissed

### Release


Simplenote/Classes/InterlinkProcessor.swift
```diff
@@ -0,0 +1,179 @@
...
+    func parentViewControllerForInterlinkProcessor(_ processor: InterlinkProcessor) -> UIViewController
+}
+
+
+// MARK: - InterlinkProcessorDatasource
+//
+protocol InterlinkProcessorDatasource: NSObjectProtocol {
+    var interlinkExcudedEntityID: NSManagedObjectID? { get }
+    var interlinkViewContext: NSManagedObjectContext { get }
+}
```
eshurakov: I was thinking, do we really need to have Context Provider and Data Source here? From what I can tell the only variable thing is `interlinkExcudedEntityID`, but it changes only when note changes. So maybe we can inject all of these dependencies instead of asking every time? I think it would simplify things and also less protocols to deal with. What do you think?

Simplenote/Classes/InterlinkProcessor.swift
```diff
@@ -0,0 +1,179 @@
...
+
+
+// MARK: - Presenting
+//
+private extension InterlinkProcessor {
+
+    func ensureInterlinkControllerIsOnScreen() {
+        if let _ = presentedViewController {
+            return
+        }
```
eshurakov: What do you think about checking `if presentedViewController != nil` or `guard presentedViewController == nil`. I feel it better expresses the intention.

Simplenote/Classes/InterlinkProcessor.swift
```diff
@@ -0,0 +1,179 @@
...
+
+        presentInterlinkController()
+    }
+
+    func presentInterlinkController() {
+        let interlinkViewController = InterlinkViewController()
+        interlinkViewController.attachWithAnimation(to: parentViewController, below: parentOverlayView)
+        self.presentedViewController = interlinkViewController
```
eshurakov: Nitpick: don't need `self` here :-)

Simplenote/Classes/UITextView+Simplenote.swift
```diff
@@ -13,16 +14,41 @@ extension UITextView {
...
+    ///
+    var interlinkKeywordAtSelectedLocation: (Range<String.Index>, Range<String.Index>, String)? {
+        guard let text = text else {
+            return nil
         }
 
-        return layoutManager.boundingRect(forGlyphRange: selectedRange, in: textContainer)
+        return text.indexFromLocation(selectedRange.location).flatMap { index in
+            text.interlinkKeyword(at: index)
+        }
```
eshurakov: Maybe we can do something like:
```
let range = Range(selectedRange, in: text)
let location = range.lowerBound
text.interlinkKeyword(at: location)
```

Simplenote/Classes/UITextView+Simplenote.swift
```diff
@@ -34,3 +60,56 @@ extension UITextView {
...
+
+
+// MARK: - Geometry
+//
+extension UITextView {
+
+    /// Returns the Bounding Rect for the specified `Range<String.Index>`
+    ///
+    func boundingRect(for range: Range<String.Index>) -> CGRect {
+        let nsRange = text.utf16NSRange(from: range)
```
eshurakov: I believe `NSRange(range, in: text)` should also work!

Simplenote/Classes/UITextView+Simplenote.swift
```diff
@@ -34,3 +60,56 @@ extension UITextView {
...
+        let editingHeight = frame.height - paddingTop - paddingBottom
+
+        return CGRect(x: .zero, y: paddingTop, width: frame.width, height: editingHeight)
+    }
+
+    /// Returns the Window Location for the text at the specified range
+    ///
+    func locationInSuperviewForText(in range: Range<String.Index>) -> CGRect {
+        let rectInEditor = boundingRect(for: range)
+        return superview?.convert(rectInEditor, from: self) ?? rectInEditor
```
eshurakov, line 100: To convert this rect to window coordinates you can do: 
```
convert(rectInEditor, to: nil)
```
I think it's safer to convert it to window coordinates here and let the receiver to convert it from window coordinates to local coordinates.

Simplenote/Classes/UITextView+Simplenote.swift
```diff
@@ -34,3 +60,56 @@ extension UITextView {
...
+
+    /// Returns the Selected Text's bounds
+    ///
+    @objc
+    var selectedBounds: CGRect {
+        guard selectedRange.length > 0 else {
+            return .zero
+        }
+
+        return layoutManager.boundingRect(forGlyphRange: selectedRange, in: textContainer)
```
eshurakov: I think we need to convert `selectedRange` to glyph range first with 
```
open func glyphRange(forCharacterRange charRange: NSRange, actualCharacterRange actualCharRange: NSRangePointer?) -> NSRange
```

Simplenote/Classes/UIViewController+Simplenote.swift
```diff
@@ -26,4 +26,21 @@ extension UIViewController {
...
+    }
+
+    /// Detaches the receiver from its parent
+    ///
+    func detachWithAnimation() {
+        view.fadeOut { _ in
+            self.view.removeFromSuperview()
+            self.removeFromParent()
+        }
+    }
```
eshurakov: When adding a view controller as a child we need to call didMove after the view is added. Similar in removing.
I always forget the sequence and often consult with: 

https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html

:-)

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -17,6 +17,13 @@ extension SPNoteEditorViewController {
 
         ensureEmptyNoteIsDeleted()
     }
+
+    /// Whenever a ViewController is presented, let's ensure Interlink is dismissed!
+    ///
+    override public func present(_ viewControllerToPresent: UIViewController, animated flag: Bool, completion: (() -> Void)? = nil) {
+        super.present(viewControllerToPresent, animated: flag, completion: completion)
+        interlinkProcessor.dismissInterlinkLookup()
+    }
```
eshurakov, line 26: Nice! Love the idea

Simplenote/Classes/InterlinkViewController.swift
```diff
@@ -0,0 +1,184 @@
...
+    }
+}
+
+
+// MARK: - Initialization
+//
+private extension InterlinkViewController {
+
+    func setupRootView() {
+        view.frame = UIScreen.main.bounds
```
eshurakov: Do we need to set a frame here?

Simplenote/Classes/InterlinkViewController.swift
```diff
@@ -0,0 +1,184 @@
...
+    ///
+    func startObservingContentOffset(in textView: UITextView) {
+        kvoOffsetToken = textView.observe(\UITextView.contentOffset, options: [.old, .new]) { [weak self] (_, value) in
+            guard let oldY = value.oldValue?.y, let newY = value.newValue?.y, oldY != newY else {
+                return
+            }
+
+            self?.tableTopConstraint.constant += oldY - newY
+        }
+    }
```
eshurakov, line 139: I vote for listening for scrollViewDidScroll in the editor and passing it here. KVO gives me chills :-)

Simplenote/Classes/InterlinkViewController.swift
```diff
@@ -0,0 +1,184 @@
...
+    /// Returns the target Origin.Y
+    /// -   Parameters:
+    ///     - height: The new target height
+    ///     - range: Anchor's Range
+    ///     - textView: Sibling TextView
+    ///
+    /// - Important: We'll always prefer either the "location above the cursor" whenever the **Maxed Out TableView** can be fit (that is: +3.5 results!)
+    ///
+    func calculateLocation(for height: CGFloat, around range: Range<String.Index>, in textView: UITextView) -> CGFloat {
+        let anchorFrame = textView.locationInSuperviewForText(in: range)
```
eshurakov, line 150: Do we need to convert it to a local view coordinates?

Simplenote/Classes/InterlinkProcessor.swift
```diff
@@ -0,0 +1,179 @@
...
+    ///     2. The Editor is the first responder
+    ///     3. There is an interlink `[keyword` at the current location
+    ///     4. There are Notes with `keyword` in their title
+    ///
+    ///  Otherwise we'll simply dismiss the Autocomplete View, if any.
+    ///
+    @objc
+    func processInterlinkLookup() {
+        guard mustProcessInterlinkLookup,
+              let (_, keywordRange, keywordText) = parentTextView.interlinkKeywordAtSelectedLocation,
```
eshurakov, line 59: Idea: maybe we can already pass bounds for keywords instead of range?

Simplenote/Classes/InterlinkProcessor.swift
jleandroperez: I was gonna mention that `if let _ = ...` is **way faster** [if you need to check for null values](https://albinek.com/to-if-let-or-to-if-not-nil)... but [Google agrees with you!!](https://google.github.io/swift/).

Updating, thank you for the feedback sir!

Simplenote/Classes/InterlinkProcessor.swift
jleandroperez: Thank you!

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
jleandroperez: Feels hacky but ... single spot to catch them all! Thank you!

Simplenote/Classes/InterlinkViewController.swift
jleandroperez: We do need to do this manually, **but perhaps** it's best for the snippet that instantiates this VC to initialize it's view size. I'll relocate this over to the InterlinkProcessor, thank you!

Simplenote/Classes/InterlinkViewController.swift
jleandroperez: If it's okay with you, let's address that in a follow up PR! Added to my list of pendings in #914

Simplenote/Classes/UITextView+Simplenote.swift
jleandroperez: You're a genius, looks SO MUCH better!!

Simplenote/Classes/InterlinkProcessor.swift
jleandroperez: I've moved the `viewContext` into a required initializer parameter. However, the `excuded entity ID` is super cleaner to be fetched this way: the Note ID might change (IE if you press the + button).

I'll look into merging all the things into a single protocol (if I manage to come up with a proper name!!). Following up in another PR if you don't mind!  (Added a note in #914)

Simplenote/Classes/UITextView+Simplenote.swift
jleandroperez: Sort of! `Range(selectedRange, in: text)` returns an optional. However, I love the idea of using an SDK API instead.

Thank you for the heads up!

Simplenote/Classes/InterlinkProcessor.swift
jleandroperez: Ranges are used for both, keyword positioning (And in a follow up PR) text replacement, I'm afraid.

Even if we do remove the positioning Range, we'd still need the full text range (for replacement purposes), soooo for consistency, I'd vote to keep them both the same. 

WDYT?

Simplenote/Classes/InterlinkViewController.swift
jleandroperez: The TextView itself would be a sibling of this ViewController's view (both live in the same superview).

Since this VC is expected to assume the exact same size as it's parent, there should be no need for conversion (AFAIK!!), please correct me if I'm wrong!

jleandroperez:
@eshurakov Thanks **a lot** for the awesome feedback!!. Addressed (most) of it, added some of them to my pendings list in #914 (will address them in a follow up PR).

Ready for another look. Thanks in advance!!

Simplenote/Classes/InterlinkProcessor.swift
eshurakov: TIL that there is this different is performance between `if let _` and `!= nil` 👍

Simplenote/Classes/InterlinkViewController.swift
eshurakov: I would like to make as little assumptions as possible. I think safest approach would be if `textView.locationInSuperviewForText(in: range)` returns frame in Window coordinates and we would convert it back to our local coordinates here (similar to how we do it for keyboard frame).

Simplenote/Classes/InterlinkProcessor.swift
eshurakov: I see! Let's see how it goes 👍

eshurakov (approved):
Looks good! Thank you @jleandroperez

jleandroperez:
Thanks a loot for reviewing this one!!

PR #1004, opened 2020-11-10
Search Minimap

### LINKED ISSUE #953: Search Minimap
### Details:
- [ ] Render a touchable minimap on the right hand side of the editor
- [ ] Quickly navigate to search hits when pressing dent marks on the right hand side
- [ ] Refresh the Minimap whenever remote updates come thru (ensure the map integrity!)

eshurakov:
### Details
This PR adds a touchable minimap to the Note when in search mode. 

Closes #953 

@jleandroperez Can I bug you with this one? Thanks!
cc @SylvesterWilmott 

### Test
1. Have both big and small notes on hand
2. Search for something
3. Open a note

- [x] Search minimap is shown on the right side
- [x] Tapping on a bar would scroll to the result and highlight it
- [x] Dragging finger on the minimap snaps moves to the result close to the finger
- [x] Dragging finger on the minimap has haptic feedback on supported devices
- [x] Dismissing search dismisses minimap
- [ ] Starting editing a note dismisses minimap

### Release
`RELEASE-NOTES.txt` was updated in https://github.com/Automattic/simplenote-ios/pull/1004/commits/a95205cf8c781900c36a11fbd1e248b8b478e6fb with:
 > Added Search Minimap to quickly navigate search hits #953


Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -616,6 +616,67 @@ extension SPNoteEditorViewController {
...
 }
 
+// MARK: - Search Map
+//
+extension SPNoteEditorViewController {
+
+    /// Show search map keyword ranges
+    ///
+    @objc
+    func showSearchMap(with searchRangeValues: [NSValue]) {
```
jleandroperez, line 693: Would it be possible to add a note to convert this into `[Range]` (or `[NSRange]` ?)

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -616,6 +616,67 @@ extension SPNoteEditorViewController {
...
+        createSearchMapViewIfNeeded()
+        // trigger layout to get correct height
+        searchMapView.superview?.layoutSubviews()
```
jleandroperez: perhaps `view.setNeedsLayout()` instead? (Plus: would it be possible to add a note indicating why is this one needed?)

Simplenote/SPEditorTextView+Simplenote.swift
```diff
@@ -30,3 +30,12 @@ extension SPEditorTextView {
         UIPasteboard.general.string = text
     }
 }
+
+extension SPEditorTextView {
+    /// Bounding rect for a character range
+    ///
+    func boundingRectForCharacterRange(_ range: NSRange) -> CGRect {
+        let glyphRange = layoutManager.glyphRange(forCharacterRange: range, actualCharacterRange: nil)
+        return layoutManager.boundingRect(forGlyphRange: glyphRange, in: textContainer)
```
jleandroperez: Perhaps this [API could be of use?](https://github.com/Automattic/simplenote-ios/blob/develop/Simplenote/Classes/UITextView+Simplenote.swift#L148)?

(It's available in develop!)

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -616,6 +616,67 @@ extension SPNoteEditorViewController {
...
+            return
+        }
+
+        // Value from 0 to 1
+        let searchBarPositions: [CGFloat] = searchRangeValues.map {
+            let boundingRect = noteEditorTextView.boundingRectForCharacterRange($0.rangeValue)
+            return max(boundingRect.midY / editorHeight, CGFloat.leastNormalMagnitude)
+        }
+
+        searchMapView?.update(with: searchBarPositions)
```
jleandroperez: Idea: How about splitting this API into smaller methods? 

1. Editor Height calculation (potentially a UITextView extension). Maybe [this could help?](https://github.com/Automattic/simplenote-ios/blob/develop/Simplenote/Classes/UITextView+Simplenote.swift#L160)

2. Range + Height > Position conversion. Potentially another UITextView extension?

This approach would allow you to return `[]` whenever `editorHeight > CGFloat.leastNormalMagnitude`, and you could kill the guard from `L635`.

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -616,6 +616,67 @@ extension SPNoteEditorViewController {
...
+    ///
+    @objc
+    func showSearchMap(with searchRangeValues: [NSValue]) {
+        createSearchMapViewIfNeeded()
+        // trigger layout to get correct height
+        searchMapView.superview?.layoutSubviews()
+
+        let editorRect = noteEditorTextView.layoutManager.usedRect(for: noteEditorTextView.textContainer)
+        let editorHeight = max(editorRect.size.height,
+                               searchMapView?.frame.size.height ?? 0)
```
jleandroperez: `searchMapView` is used without `?` in L629 (and is guarranteed to be available). Perhaps a `guard let` below `createSearchMap..()`?

jleandroperez (changes requested):
Working super smoothly!! Sending few cosmetic comments (we can share few APIs from Interlinking).

Plus one note: we get an "overlap" between the Minimap and Scrollbar. @SylvesterWilmott are we good with that? (ooor should we display the Minimap in the opposed edge?)

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: Definitely!

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: I'm triggering the layout to get correct height of the search map. It is used as a minimum editor height in search terms positions calculation.
I thought to trigger layout and use the height is easier than calculating all margins. That's also why `setNeedsLayout` not gonna work, because the height is used right in that method.

eshurakov: Actually got rid of it and now using `editingRect` from the UITextView extension

Simplenote/SPEditorTextView+Simplenote.swift
eshurakov: Thanks, will use it!

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: 1. Var name was confusing (renamed it to `textContainerHeight`), it's actually rect for a text container, because I'm interested in the total height of the content of the text container. It is quite specific to the search map, so I wouldn't put in a UITextView extension.
2. I don't think it's good to put it in an extension as it is very specific to the search map. 

I split it in a several methods, but kept them in the `SPNoteEditorViewController`. Thanks for the suggestion 👍

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: Fixed, thanks!

eshurakov:
Thank you for the feedback @jleandroperez. 
I've made some changes, do you mind taking another look? Thank you!

jleandroperez (approved):
Thank you for making those changes!! looks great!!

:shipit:  🎉  🚀

eshurakov:
Merging, thank you @jleandroperez !

PR #1005, opened 2020-11-10
Interlinks Autocomplete: Mark III

jleandroperez:
### Details
In this PR we're wiring the Interlinks Autocomplete UI, so that whenever the user selects a row, an Interlink for the selected note gets inserted.

cc @eshurakov Yet another piece of the puzzle!
Thanks in advance!!

Ref. #914

### Test
1. Open any note
2. Type `[keyword` with `keyword` being a word contained within any of your Note's title

- [x] Verify that selecting any of the rows that show up causes the App to insert an Interlink to the target note
- [x] Verify that **Shake > Undo** reverts the Interlink Insertion OP

### Release

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
@@ -99,6 +99,7 @@ extension SPNoteEditorViewController {
     func configureInterlinksProcessor() {
         interlinkProcessor = InterlinkProcessor()
         interlinkProcessor.datasource = self
+        interlinkProcessor.delegate = self
         interlinkProcessor.contextProvider = self
     }
 }
@@ -602,6 +603,17 @@ extension SPNoteEditorViewController: InterlinkProcessorDatasource {
 
     var interlinkViewContext: NSManagedObjectContext {
         SPAppDelegate.shared().managedObjectContext
+    }
+}
+
+
+// MARK: - InterlinkProcessorDelegate
+//
+extension SPNoteEditorViewController: InterlinkProcessorDelegate {
+
+    func interlinkProcessor(_ processor: InterlinkProcessor, insert text: String, in range: Range<String.Index>) {
+        noteEditorTextView.insertText(text: text, in: range)
+        processor.dismissInterlinkLookup()
     }
 }
 
--- a/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
@@ -105,7 +105,6 @@ extension SPNoteEditorViewController {
     @objc
     func configureInterlinksProcessor() {
         interlinkProcessor = InterlinkProcessor(viewContext: SPAppDelegate.shared().managedObjectContext)
-        interlinkProcessor.datasource = self
         interlinkProcessor.delegate = self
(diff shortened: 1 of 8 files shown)


Simplenote/Classes/UITextView+Simplenote.swift
```diff
@@ -28,6 +28,76 @@ extension UITextView {
...
+// MARK: - Updating!
+//
+extension UITextView {
+
+    /// Inserts the specified Text at a given range.
+    /// - Note: Resulting selected range will end up at the right hand side of the newly inserted text.
+    ///
+    func insertText(text: String, in range: Range<String.Index>) {
+        registerUndoCheckpointAndPerform { storage in
+            let range = self.text.utf16NSRange(from: range)
```
eshurakov: `NSRange(range, in: self.text)` and can maybe completely remove `utf16NSRange`?

Simplenote/Classes/UITextView+Simplenote.swift
```diff
@@ -28,6 +28,76 @@ extension UITextView {
...
+            storage.replaceCharacters(in: range, with: text)
+            self.selectedRange = NSRange(location: range.location + text.count, length: .zero)
```
eshurakov: I believe it is incorrect to add `text.count` to NSRange. `text.count` is based on the representation of the string (utf8 / utf16) and can result in a wrong adjustment.

I would try to do:
```
let insertedTextRange = NSRange(text.fullRange, in: text)
self.selectedRange = NSRange(location: range.location + insertedTextRange.length, length: .zero)
```

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -599,17 +599,22 @@ extension SPNoteEditorViewController: InterlinkProcessorPresentationContextProvi
...
 
-    var interlinkExcudedEntityID: NSManagedObjectID? {
+    func excludedEntityIdentifierForInterlinkProcessor(_ processor: InterlinkProcessor) -> NSManagedObjectID? {
         currentNote?.objectID
     }
+
+    func interlinkProcessor(_ processor: InterlinkProcessor, insert text: String, in range: Range<String.Index>) {
+        noteEditorTextView.insertText(text: text, in: range)
+        processor.dismissInterlinkLookup()
+    }
```
eshurakov, line 613: Idea: do not to pass markdown range to the Processor and back here. Instead we can figure it out right here based on the current selection in the editor.

Simplenote/Classes/UITextView+Simplenote.swift
jleandroperez: You're 💯 on spot. I need to fix mac as well (this code was borrowed from there!!)

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
jleandroperez: (For the record!!) Discussed over Slack. Let's keep the Range (for now at least!).

Thank you for bringing this up!!

jleandroperez:
@eshurakov Ready for another peek!! thanks in advance!!

eshurakov (approved):
Looks great!

:shipit:

jleandroperez:
Thank you sir!!

PR #1006, opened 2020-11-11
Update SwiftLint to 0.41.0

eshurakov:
### Fix
This PR updates SwiftLint to 0.41.0. This will allow us to use inclusive language in the configuration file. 

There were already quite some warnings while running previous version of SwiftLint. I checked and it doesn't look like new version added new warnings, which is good :-)

### Test
- [x] Test that SwiftLint runs via `rake lint`

### Release


jleandroperez (approved):
:shipit: sir!!

PR #1013, opened 2020-11-13
Interlinks Autocomplete: Mark V

jleandroperez:
### Details:
1. We're using **Window coordinates** when exchanging frames between Editor / Interlink
2. **Dropped KVO in InterlinkViewController**: Explicit scroll invocations
3. Autocomplete is **dismissed when the scrolled offset exceeds a threshold** (30pt)
4. We've overhauled the geometry calculations for Autocomplete

#### Geometry:
1. Autocomplete will always be displayed above the cursor, when possible 📏 
2. Whenever displaying above **and** below cause clipping, we'll pick the orientation that clips the least
3. Horizontal position will be placed to match the Keyword onscreen
4. And we'll definitely never allow horizontal clipping 😄 

@eshurakov One step closer!
cc @SylvesterWilmott Violently improved positioning logic (exactly the same as macOS!)

**Note!!**
RTL support coming up in a follow up!

Ref. #914

### Test: Geometry
1. Open the Editor and add a note with 6 lines
2. Place the cursor in the 6th line
3. Type `[keyword` (with `keyword` being a word contained in any of your note's titles)

- [ ] Verify Autocomplete shows up below the cursor's location
- [ ] Verify that if you type another keyword (and narrow the results) the UI will shrink **but won't jump above**
- [ ] Verify that scrolling over the TextView below causes the Autocomplete window to follow along
- [ ] Verify that if you scroll over 30pt the Autocomplete UI gets dismissed
- [ ] Verify that if you repeat this in the 7th line, the Autocomplete UI shows up above
- [ ] Verify this also works super in an iPad device please!

### Test: Horizontal Overflow
1. Open the Editor
2. Type a long line
3. Type `[keyword` (with `keyword` being a word contained in any of your note's titles)

- [ ] Verify that hte Autocomplete UI never gets clipped offscreen!

### Test: Adaptivity
1. Open the Editor and type `[keyword` to trigger Autocomplete

- [ ] Verify that rotating the device causes Autocomplete to disappear / reappear in the right location

### Release
`RELEASE-NOTES.txt` was updated in bf0d2199 with:
 
> Typing `[` will now allow you to pick Interlinking Notes right from the editor

### DIFF
--- a/Simplenote/Classes/InterlinkProcessor.swift
+++ b/Simplenote/Classes/InterlinkProcessor.swift
@@ -156,10 +156,10 @@ private extension InterlinkProcessor {
 extension InterlinkProcessor {
 
     /// Relocates the Interlink UI (whenever it's visible) to match the new TextView's Content Offset.
-    /// - Important: Whenever the **Maximum Allowed Scroll Offset** is exceeded, we'll just dismiss the UI
+    /// - Important: Whenever the **Maximum Allowed Scroll Offset** is exceeded (and the scroll event is user initiated), we'll just dismiss the UI
     ///
-    @objc(refreshInterlinkControllerWithNewOffset:)
-    func refreshInterlinkController(contentOffset: CGPoint) {
+    @objc(refreshInterlinkControllerWithNewOffset:isDragging:)
+    func refreshInterlinkController(contentOffset: CGPoint, isDragging: Bool) {
         guard let interlinkViewController = presentedViewController else {
             return
         }
@@ -173,7 +173,9 @@ extension InterlinkProcessor {
         }
 
         interlinkViewController.relocateInterface(by: oldY - contentOffset.y)
-        dismissInterlinkLookupIfNeeded(initialY: initialY, currentY: contentOffset.y)
+        if isDragging {
+            dismissInterlinkLookupIfNeeded(initialY: initialY, currentY: contentOffset.y)
+        }
     }
 
     /// Dismisses the Interlink Lookup whenever the **Maximum Allowed Scroll Offset** is exceeded
--- a/Simplenote/Classes/InterlinkProcessor.swift
+++ b/Simplenote/Classes/InterlinkProcessor.swift
@@ -121,7 +121,11 @@ private extension InterlinkProcessor {
     }
 
     func relocateInterlinkController(around range: Range<String.Index>) {
-        presentedViewController?.anchorView(around: range, in: parentTextView)
+        let keywordFrame = parentTextView.locationInWindowForText(in: range)
+        let editingFrame = parentTextView.editingRectInWindow()
+
+        presentedViewController?.relocateInterface(around: keywordFrame, in: editingFrame)
+        lastKnownEditorOffset = parentTextView.contentOffset
     }
 
     func refreshInterlinkController(notes: [Note]) {
--- a/Simplenote/Classes/InterlinkProcessor.swift
+++ b/Simplenote/Classes/InterlinkProcessor.swift
@@ -40,6 +40,8 @@ class InterlinkProcessor: NSObject {
 
     private let viewContext: NSManagedObjectContext
     private var presentedViewController: InterlinkViewController?
+    private var lastKnownEditorOffset: CGPoint?
+    private var initialEditorOffset: CGPoint?
     private lazy var resultsController = InterlinkResultsController(viewContext: viewContext)
 
     weak var contextProvider: InterlinkProcessorPresentationContextProvider?
@@ -73,6 +75,7 @@ class InterlinkProcessor: NSObject {
         ensureInterlinkControllerIsOnScreen()
         refreshInterlinkController(notes: notes)
         relocateInterlinkController(around: keywordRange)
+        trackLastKnownScrollOffsets()
         setupInterlinkEventListeners(replacementRange: markdownRange)
     }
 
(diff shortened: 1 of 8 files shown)


Simplenote/Classes/InterlinkProcessor.swift
```diff
@@ -140,6 +151,45 @@ private extension InterlinkProcessor {
...
+    ///
+    @objc(refreshInterlinkControllerWithNewOffset:isDragging:)
+    func refreshInterlinkController(contentOffset: CGPoint, isDragging: Bool) {
+        guard let interlinkViewController = presentedViewController else {
+            return
+        }
+
+        defer {
+            lastKnownEditorOffset = contentOffset
+        }
```
eshurakov, line 169: Nice!

Simplenote/Classes/InterlinkViewController.swift
```diff
@@ -15,14 +15,9 @@ class InterlinkViewController: UIViewController {
     /// Layout Constraints: Inner TableView
     ///
     @IBOutlet private var tableLeadingConstraint: NSLayoutConstraint!
-    @IBOutlet private var tableTrailingConstraint: NSLayoutConstraint!
     @IBOutlet private var tableTopConstraint: NSLayoutConstraint!
     @IBOutlet private var tableHeightConstraint: NSLayoutConstraint!
 
-    /// KVO
-    ///
-    private var kvoOffsetToken: NSKeyValueObservation?
```
eshurakov, line 24: 🔥 🔥 🔥

Simplenote/Classes/InterlinkViewController.swift
```diff
@@ -52,12 +47,26 @@ class InterlinkViewController: UIViewController {
...
 extension InterlinkViewController {
 
-    /// Relocates the receiver so that it shows up around a given Keyword in a **Sibling TextView**
-    /// - Important: We'll start listening for Content Offset changes, and the UI will be automatically repositioned
+    /// Relocates the receiver so that it shows up **around** the specified Anchor Frame, in a given ViewPort.
+    /// - Important: Both frames must be expressed in Window Coordinates. Capisce?
+    ///
+    func relocateInterface(around anchor: CGRect, in viewport: CGRect) {
+        let anchorFrame = view.convertFromWindowCoordinates(anchor)
+        let editingRect = view.convertFromWindowCoordinates(viewport)
```
eshurakov: I'm not sure if we need a category for it, cause it's just
```
let anchorFrame = view.convert(anchor, from: nil)
let editingRect = view.convert(viewport, from: nil)
```
Specifying `nil` will convert given rect from a window coordinate system

Simplenote/Classes/UITextView+Simplenote.swift
```diff
@@ -157,19 +157,19 @@ extension UITextView {
...
     }
 
     /// Returns the Window Location for the text at the specified range
     ///
-    func locationInSuperviewForText(in range: Range<String.Index>) -> CGRect {
+    func locationInWindowForText(in range: Range<String.Index>) -> CGRect {
         let rectInEditor = boundingRect(for: range)
-        return superview?.convert(rectInEditor, from: self) ?? rectInEditor
+        return window?.convert(rectInEditor, from: self) ?? rectInEditor
     }
```
eshurakov, line 174: An easier way is to do `convert(rectInEditor, to: nil)`

Simplenote/Classes/UITextView+Simplenote.swift
```diff
@@ -157,19 +157,19 @@ extension UITextView {
     ///
     /// - Note: `contentInset.bottom` is expected to contain the bottom padding required by the keyboard. Capisce?
     ///
-    func editingRect() -> CGRect {
-        let paddingTop = safeAreaInsets.top
+    func editingRectInWindow() -> CGRect {
+        let paddingTop = frame.minY + safeAreaInsets.top
```
eshurakov, line 161: That would only work if textView is added to the view that occupies full screen. I think the safest approach is to convert the original rect to window coordinates using UIView API.

eshurakov (approved):
Works smooth!

eshurakov:
Small UI comments:
1. On iPad in horizontal compact mode and maybe on small screen phones the popup goes quite close to the left edge.
Screenshot 2020-11-13 at 10 19 01

2. Sorry about this one, it's an edge case :-) Type `[`, then switch to another position in the text and type `[keyword` so that you see autocomplete. Now tap right after the first bracket, the cursor will move there, but autocomplete window will stay where it was.
Screenshot 2020-11-13 at 10 16 13

Simplenote/Classes/InterlinkViewController.swift
jleandroperez: I forgot about that one!!! thank youuu!!

Simplenote/Classes/UITextView+Simplenote.swift
jleandroperez: And that also kills the optional. **Thank you!!** Absolutely forgot about those APIs!!

Simplenote/Classes/UITextView+Simplenote.swift
jleandroperez: Updated, thank youu!!

jleandroperez:
Thanks for the awesome review @eshurakov !!

1. All comments addressed, thank you for the CGRect conversion help!

2. I've checked with Sly, and updated the layout constraints so that there's **always** a minimum of 16pt on Leading / Trailing edges (Autolayout will slightly compress the view, when appropriate!)

3. I couldn't trigger the `[` bug on my end!! Will check with you over Slack. Do you mind if we address that one as a bugfix?

Thanks in advance!!

jleandroperez:
### Addendum:
Please run a quick check over Search Minimap (looks good on my end!). Ref:
`editingRect >> editingRectInWindow`

Should be 100% unaffected since the Minimap checked the "Viewport"'s height.

Thank yooouu!!

eshurakov (approved):
Yay! 🎉 
:shipit:

jleandroperez:
@eshurakov thanks a loot sir!!

PR #1030, opened 2020-11-17
Interlinking: Dynamic Height

### LINKED ISSUE #914: Interlink Autocomplete

jleandroperez:
### Details
In this PR we're limiting the Interlinking UI's height to the available area.

**Note:** The algorithm is overly simplified, and assumes the cursor is either at the top or bottom of the viewport. Overflow in any other case is expected!

cc @eshurakov (Thank youuu!!)
Closes #914

### Test
1. Open the editor
2. Turn the app in landscape mode
3. Place the cursor either at the top or bottom of the editor
4. Type `[keyword`

- [x] Verify the Interlink UI doesn't overflow. (As long as the cursor is anchored at the top / bottom!!)

### Release

### DIFF
--- a/Simplenote/Classes/CGRect+Simplenote.swift
+++ b/Simplenote/Classes/CGRect+Simplenote.swift
@@ -10,18 +10,18 @@ extension CGRect {
     /// - Note: We rely on this API to determine the available editing area above and below the cursor
     /// - Important: **For simplicity's sake** we won't return an optional. If the parameters are invalid, we'll just pass along the "non split receiver"
     ///
-    func split(by rect: CGRect) -> (upperSlice: CGRect, lowerSlice: CGRect) {
+    func split(by rect: CGRect) -> (aboveSlice: CGRect, belowSlice: CGRect) {
         guard contains(rect) else {
             return (.zero, self)
         }
 
-        var lowerSlice = self
-        lowerSlice.size.height = rect.minY - minY
+        var belowSlice = self
+        belowSlice.size.height = rect.minY - minY
 
-        var upperSlice = self
-        upperSlice.origin.y = rect.maxY
-        upperSlice.size.height = height - lowerSlice.height - rect.height
+        var aboveSlice = self
+        aboveSlice.origin.y = rect.maxY
+        aboveSlice.size.height = maxY - rect.maxY
 
-        return (upperSlice, lowerSlice)
+        return (aboveSlice, belowSlice)
     }
 }
--- a/Simplenote/Classes/CGRect+Simplenote.swift
+++ b/Simplenote/Classes/CGRect+Simplenote.swift
@@ -1,0 +1,27 @@
+import Foundation
+
+
+// MARK: - CGRect Methods
+//
+extension CGRect {
+
+    /// Returns the resulting Rectangles by splitting the receiver in two, by the specified Rectangle's **minY** and **maxY**
+    ///
+    /// - Note: We rely on this API to determine the available editing area above and below the cursor
+    /// - Important: **For simplicity's sake** we won't return an optional. If the parameters are invalid, we'll just pass along the "non split receiver"
+    ///
+    func split(by rect: CGRect) -> (upperSlice: CGRect, lowerSlice: CGRect) {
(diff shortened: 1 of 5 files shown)


Simplenote/Classes/CGRect+Simplenote.swift
```diff
@@ -0,0 +1,27 @@
...
+        guard contains(rect) else {
+            return (.zero, self)
+        }
+
+        var lowerSlice = self
+        lowerSlice.size.height = rect.minY - minY
+
+        var upperSlice = self
+        upperSlice.origin.y = rect.maxY
+        upperSlice.size.height = height - lowerSlice.height - rect.height
```
eshurakov: Hmmm, is it the same as `upperSlice.size.height = maxY - rect.maxY`? To make it similar to `lowerSlice` calculations.

Simplenote/Classes/CGRect+Simplenote.swift
```diff
@@ -0,0 +1,27 @@
...
+
+        return (upperSlice, lowerSlice)
+    }
+}
```
eshurakov: What do you think about naming it `aboveSlice` and `belowSlice`?
I was trying to wrap my head around `lower` and `upper` and got confused, cause `lower` is actually higher on screen than `upper` :-)

Simplenote/Classes/InterlinkViewController.swift
```diff
@@ -54,7 +54,7 @@ extension InterlinkViewController {
         let anchorFrame = view.convert(anchor, from: nil)
         let editingRect = view.convert(viewport, from: nil)
 
-        let targetHeight = calculateHeight()
+        let targetHeight = calculateHeight(around: anchorFrame, in: editingRect)
         let targetTop = calculateTopLocation(for: targetHeight, around: anchorFrame, in: editingRect)
```
eshurakov: From what I see these two methods do quite similar thing. 
In `calculateHeight(around: anchorFrame, in: editingRect)` we already know if there is more space above / below the `anchorFrame`. We calculate the available space again in a different way in `calculateTopLocation(for: targetHeight, around: anchorFrame, in: editingRect)`.
Do you think we can combine / optimise it?

eshurakov (commented):
Hey @jleandroperez, sending a few comments your way!

Simplenote/Classes/CGRect+Simplenote.swift
jleandroperez: You're right, thank you!!

Simplenote/Classes/CGRect+Simplenote.swift
jleandroperez: Sounds good!

jleandroperez:
@eshurakov Ready for another look sir! Thank you!!

eshurakov (approved):
Yeeees! 

:shipit:

jleandroperez:
@eshurakov Thanks **a lot** for the super awesome review!!!

PR #1031, opened 2020-11-17
Empty states

### LINKED ISSUE #951: Empty States
### Details
Let's review and implement new Empty State placeholders, app wide:

- [ ] Empty State: All Notes
- [ ] Empty State: Tag
- [ ] Empty State: Trash
- [ ] Empty State: No Search Results. **Support for quick actions:** prompt to create a note (screenshot below!)

eshurakov:
### Details
This PR updates Empty States for the following screens:
* All notes
* Tag
* Untagged notes
* Trash
* No Search Results. On this screen added quick action to create a new note with a search term as a title.

@jleandroperez May I bug you with this PR? Thank you!

Closes #951 

### Test
Check empty state is shown when there are no notes on:
- [x] All notes
- [x] Tag
- [x] Untagged notes
- [x] Trash
- [x] Unsuccessful search (from all notes and tag)
- [x] Tapping on Empty State for unsuccessful search creates a new note with search term as a title

### Release
`RELEASE-NOTES.txt` was updated in https://github.com/Automattic/simplenote-ios/pull/1031/commits/c1e70016f697f06b4ed2e4f3a287d57a7a9929af with:
 
> Updated Empty State placeholders #951


Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -293,21 +298,40 @@ extension SPNoteListViewController {
...
+            switch notesListController.filter {
+            case .everything:
+                return .pictureAndText(imageName: .allNotes, text: Localization.EmptyState.allNotes)
+            case .deleted:
+                return .pictureAndText(imageName: .trash, text: Localization.EmptyState.trash)
+            case .untagged:
+                return .pictureAndText(imageName: .untagged, text: Localization.EmptyState.untagged)
+            case .tag(name: let name):
+                return .pictureAndText(imageName: .tag, text: Localization.EmptyState.tagged(with: name))
+            }
```
jleandroperez, line 314: For readability purposes, would it be possible to split this into a smaller method?

Perhaps `var placeholderDisplayMode: DisplayMode` ?

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -293,21 +298,40 @@ extension SPNoteListViewController {
...
         }()
+
+        if case .text = placeholderView.displayMode {
+            placeholderViewVerticalCenterConstraint.isActive = false
+            placeholderViewTopConstraint.isActive = true
+        } else {
+            placeholderViewTopConstraint.isActive = false
+            placeholderViewVerticalCenterConstraint.isActive = true
+        }
```
jleandroperez, line 339: Yet another reason to drop the custom SearchDisplayController!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -341,6 +365,18 @@ extension SPNoteListViewController {
...
+    /// Creates and opens new note with a given text
+    ///
+    func openNewNote(with content: String) {
+        SPTracker.trackListNoteCreated()
+        let note = NoteFactory.newNote()
+        note.content = content
+        if case let .tag(name) = notesListController.filter {
+            note.addTag(name)
+        }
+        open(note, ignoringSearchQuery: true, animated: true)
```
jleandroperez, line 383: WDYT if instead of bypassing the search query, we simply exit search mode (and add a new note?)

(Just thinking if we could simplify a bit!)

Simplenote/NoteFactory.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+enum NoteFactory {
+
+    static func newNote() -> Note {
+        let viewContext = SPAppDelegate.shared().managedObjectContext
```
jleandroperez: Idea!!

Rather than querying the NSMOC, how about if we have a property, that must be set in the AppDelegate (before anything else runs)?

Check out [EditorFactory](https://github.com/Automattic/simplenote-ios/blob/develop/Simplenote/Classes/EditorFactory.swift), we do something _slightly_ similar (but not with CoreData precisely)

jleandroperez (changes requested):
Sending you few annoying notes! Looks amazing!!

Please LMK what you think, thank you!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
eshurakov: That would be simpler, but not what we want :-) (discussed with Sly)

Simplenote/NoteFactory.swift
eshurakov: Moved it to `SPObjectManager` 🙈

Simplenote/Classes/SPNoteListViewController+Extensions.swift
eshurakov: Thank you, done!

eshurakov:
@jleandroperez, this PR is ready for another look. Thank you!

jleandroperez (approved):
Awesome work @eshurakov !!!

:shipit:

PR #1032, opened 2020-11-18
Home screen quick actions

### LINKED ISSUE #950: Quick Actions
### Details:
Let's implement iOS's Quick Actions support for the following items:

- [ ] Search 
- [ ] New Note
- [ ] Recent

quick_actions_light

eshurakov:
### Details
This PR adds home screen quick actions:
* Search: opens note list with active search
* New Note: opens new note
* Recent: opens recent note

@jleandroperez Another day another PR! Thank you!

Closes #950 

### Test
1. Long press on the app icon. 

- [x] Check that Search opens Note List and search bar is active
- [x] Check that New Note opens an empty editor 
- [x] Check that Recent opens a recently modified note

Please check when the app is in memory / killed.
Please also check when the app is in memory and some random screen is opened.

### Test (Log out)
1. Log out of the app
2. Long press on the app icon

- [x] Check that there are no quick actions available

### Release
`RELEASE-NOTES.txt` was updated in https://github.com/Automattic/simplenote-ios/pull/1032/commits/45dd28e13ef6325debded3730d9f5f841dc8b6f0 with:

> Home screen quick actions #950


Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -69,6 +75,28 @@ extension SPAppDelegate {
...
+        sidebarViewController.hideSidebar(withAnimation: false)
+        if navigationController.viewControllers.contains(noteListViewController) {
+            navigationController.popToViewController(noteListViewController, animated: false)
+        }
+
+        DispatchQueue.main.async {
+            // Switch from trash to all notes as trash doesn't have search
+            if self.selectedTag == kSimplenoteTrashKey {
+                self.selectedTag = nil
+            }
```
jleandroperez, line 82: Since we're coming from the springboard, and the user is triggering Search, how about if we simply always ensure `All Notes` is selected?

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -69,6 +75,28 @@ extension SPAppDelegate {
...
     }
+
+    func presentSearch() {
+        dismissAllModals(animated: false, completion: nil)
+        sidebarViewController.hideSidebar(withAnimation: false)
+        if navigationController.viewControllers.contains(noteListViewController) {
+            navigationController.popToViewController(noteListViewController, animated: false)
+        }
+
+        DispatchQueue.main.async {
```
jleandroperez: Do we 100% need to do this async? (is it for animation purposes?)

Simplenote/SPAppDelegate.m
```diff
@@ -412,18 +428,6 @@ - (NSPersistentStoreCoordinator *)persistentStoreCoordinator
...
-{
-    [self.navigationController dismissViewControllerAnimated:animated
-                                                  completion:^{
-                                                      
-                                                      if (completion) {
-                                                          completion();
-                                                      }
-                                                  }];
-    
-}
```
jleandroperez, line 425: 🖖

jleandroperez (changes requested):
Sending few nipticky notes sir!

1. **Shortcut:** Search feels a bit broken, when the editor was the last screen. Video sent!
2. **Shortcut:** New note doesn't have an animation
3. (Checking with Sly over GH) perhaps it's best if the Recent showed the last opened note. WIll confirm on this one!

Other than that, changes look super!

Simplenote/SPAppDelegate+Extensions.swift
eshurakov: I checked this with Sly. It is fine to open the app wherever it was before. If tag was selected, we keep the tag selected. Search is global, so it won't be affected by the selected tag.
The only place that doesn't have search is Trash, that's why we switch it to All Notes.

Simplenote/SPAppDelegate+Extensions.swift
eshurakov: I did it similar to what we already had in app delegate. I will check if it can be done in a different way.

eshurakov:
Thank you for the review, @jleandroperez!

I've updated the code that handles quick actions, there should be less animations now and the user should just directly see the destination screen.

Also added quick actions cleanup when user logs out.

Ready for another look, thanks!

Simplenote/SPAppDelegate.m
```diff
@@ -691,39 +686,6 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
...
-    [self.sidebarViewController hideSidebarWithAnimation:NO];
-    
-    // Little trick to postpone until next run loop to ensure controllers have a chance to pop
-    double delayInSeconds = 0.05;
-    dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC));
-    dispatch_after(popTime, dispatch_get_main_queue(), ^(void){
-        [self.noteListViewController openNote:note animated:NO];
-        [self showPasscodeLockIfNecessary];
-    });
-}
```
jleandroperez, line 724: Thank you for doing this!!

jleandroperez (approved):
Works super smoothly, @eshurakov thank you for fixing all the animation things!!

:shipit:

PR #1038, opened 2020-11-20
Update `notesWithTag` method to use `SimplenoteFoundation`

eshurakov:
### Fix
Updating `notesWithTag` method to use `SimplenoteFoundation` and be in line with macOS.

cc @jleandroperez mind taking a quick peek? Thank you!

### Test
1. Create tags `aaaa` and `aa`
2. Assign those tags to some notes

- [x] Rename tag `aa` to `aaa` and check that the changes are reflected correctly inside notes 
- [x] Delete tag `aaa` and check that the changes are reflected correctly inside notes (and that tag `aaaa` is not affected)

### DIFF
--- a/Simplenote/Classes/SPObjectManager+Simplenote.swift
+++ b/Simplenote/Classes/SPObjectManager+Simplenote.swift
@@ -1,4 +1,5 @@
 import Foundation
+import SimplenoteFoundation
 
 // MARK: - SPObjectManager
 //
@@ -23,4 +24,18 @@ extension SPObjectManager {
         return note
     }
 
+    @objc
+    func notesWithTag(_ tag: Tag?) -> [Note] {
+        guard let tagName = tag?.name else {
+            return []
+        }
+
+        let request = NSFetchRequest<Note>(entityName: Note.entityName)
+        request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
+            .predicateForNotes(tag: tagName),
+            .predicateForNotes(deleted: false)
+        ])
+
+        return (try? managedObjectContext.fetch(request)) ?? []
+    }
 }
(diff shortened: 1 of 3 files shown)


jleandroperez (approved):
:shipit: !!!

Thank you for doing this!!

Simplenote/Classes/SPObjectManager+Simplenote.swift
```diff
@@ -23,4 +24,18 @@ extension SPObjectManager {
         return note
     }
 
+    @objc
+    func notesWithTag(_ tag: Tag?) -> [Note] {
+        guard let tagName = tag?.name else {
```
dmsnell, line 29: I realize that this code is existing but it's looks very suspicious as we shouldn't be comparing the tag names to determine if a note has the tag. This extends beyond case-comparison because we have a few levels of transformation in the `tagHash` to normalize the strings.

We can use the tag's entity ID and compare against the `tagHash` transforms of the tags in the note object but we can't [perform a string search of the tags list to find matching notes](https://github.com/Automattic/SimplenoteSearch-Swift/blob/main/Sources/SimplenoteSearch/Extensions/NSPredicate%2BSearch.swift#L20) because we'll get the wrong results if the tag name is written differently

the _display_ of the tag names in the note editor UI should present the canonical tag name to the customer, but the text inside the `tags` property of the note object _will_ contain _variations_ of the tag name which all normalize to the same value through the `tagHash`.

this doesn't need to happen here in this PR but I noticed it and wanted to point out the issue. the effect is that you select a tag and only find a subset of the notes containing that tag and it can be different on different devices.

eshurakov: That's a very good point, thanks!
I did a quick test and it seems that when comparing strings using `NSPredicate` the framework internally normalises strings. 
```
let s1 = "\u{00e9}"
let s2 = "\u{0065}\u{0301}"
let predicate = NSPredicate(format: "self CONTAINS[c] %@", s1)
predicate.evaluate(with: s2)
```

But in any case we should switch to using tagHash to compare tags.

jleandroperez: +1 great catch Dennis, we need to fix that API all over

PR #1046, opened 2020-11-25
Nü passcode screen: I

### LINKED ISSUE #761: Nü Passcode
### Guidelines

### Details
- [ ] Encapsulate all of the Pincode support (currently spread in the AppDelegate)
- [ ] Encapsulate our FaceId + TouchID support
- [ ] Unit Test all the things
- [ ] Implement the new Passcode UI
- [ ] Please keep the Attempts Limit logic

eshurakov:
### Fix
Add new passcode screen with a keypad that is always on screen and doesn't rely on built-in keyboard.
New passcode screen is not wired with the current code and it is an initial effort to create the UI. Wiring will be done in a follow up PR. 

cc @jleandroperez Thank you!

Related issue: #761 

Before             |  After
:-------------------------:|:-------------------------:
Simulator Screen Shot - iPhone 11 - 2020-11-25 at 08 18 05 |  Simulator Screen Shot - iPhone 11 - 2020-11-25 at 13 03 42
Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-11-25 at 08 11 43 |  Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-11-25 at 13 04 41

### Test
As the screen is not wired with the current passcode code, the best way to test the UI is to present the view controller from the app delegate:

```
DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
    let vc = PinLockViewController()
    self.window.rootViewController?.present(vc, animated: true, completion: nil)
}
```

- [x] Check that all buttons can be pressed
- [x] Pressing on a number key on the keypad fills a `dot`
- [x] Pressing `delete` on a keypad removes the last filled `dot`
- [x] If noting is filled, the bottom right button reads `cancel`

### Review

### DIFF
--- a/Simplenote/Classes/PinLock/PinLockViewController.swift
+++ b/Simplenote/Classes/PinLock/PinLockViewController.swift
@@ -1,18 +1,39 @@
 import UIKit
 
+// MARK: - PinLockViewController
+//
 class PinLockViewController: UIViewController {
+    override var preferredStatusBarStyle: UIStatusBarStyle {
+        return .lightContent
+    }
+
     @IBOutlet private var keypadButtons: [UIButton] = [] {
         didSet {
             for button in keypadButtons {
                 button.setBackgroundImage(UIColor.simplenoteLockScreenButtonColor.dynamicImageRepresentation(), for: .normal)
                 button.setBackgroundImage(UIColor.simplenoteLockScreenHighlightedButtonColor.dynamicImageRepresentation(), for: .highlighted)
+                button.addTarget(self, action: #selector(handleTapOnKeypadButton(_:)), for: .touchUpInside)
             }
         }
     }
+
     @IBOutlet private var cancelButton: UIButton! {
         didSet {
-            cancelButton.setTitle(Localization.cancelButton, for: .normal)
             cancelButton.setTitleColor(.white, for: .normal)
+            updateCancelButton()
+        }
+    }
+
+    @IBOutlet private var progressView: PinLockProgressView! {
+        didSet {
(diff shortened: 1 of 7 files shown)


Simplenote/Classes/PinLock/PinLockViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+        setupKeypadButtons()
+    }
+
+    func setupCancelButton() {
+        cancelButton.setTitleColor(.white, for: .normal)
+        updateCancelButton()
+    }
+
+    func setupProgressView() {
+        progressView.length = Constants.pinLength
```
jleandroperez, line 56: This one could be the default value (and you'd save a bit of code here and there!)

jleandroperez (approved):
**STUNNING** UI!!!

Awesome work @eshurakov !! :shipit:

Simplenote/Classes/PinLock/PinLockViewController.swift
eshurakov: I thought about that and to use the default value we would need to store `pinLength` as a global variable (it is used in the VC and progress view). I thought it is nicer if it is scoped to the view controller, because it seems that `pinLegth` doesn't really matter for non-views.

jleandroperez: Sounds good! Absolutely no need to change anything, just a cosmetic / subjective note.

Thank you Evgeny!!

pachlava:
[UI Testing] Verified on iPhone XR | iOS 14.3 | SN 4.30

PR #1051, opened 2020-11-27
Add Arabic to release tooling

loremattei:
### Fix
We added Arabic to the languages which Simplenote iOS is available in. Because of that, we have also to add it to the languages of the metadata. 

### Review

### Release


jleandroperez (approved):
Looks good!! Thank you sir!

:shipit:

PR #1066, opened 2020-12-30
Update screenshots automation with new details and iOS version – Local only

mokagio:
### Fix

Updates the screenshots automation to work with ~~Xcode 12.3 and the iOS 14.3 Simulators~~ Xcode 12.4 and the iOS 14.4 Simulators.

It also replaces the screenshot for the collaborate screen with one for the inter-note linking features. Internal ref p2XJRt-2wR.

~~This is off `release/4.30` because I want the new screenshots to go in with that release.~~

### Test

I'm having issues on CI, because of the [known `xcodebuild` bug with SPM and SSH repositories](https://discuss.bitrise.io/t/xcode-11-resolving-packages-fails-with-ssh-fingerprint/10388/2). To keep the work on this PR from taking too long, I decided to drop the CI update part. I think this is acceptable, albeit undesirable because we rarely updates the screenshot.

To test locally, you can either:

- Running the "SimplenoteScreenshots" scheme; it should pass
- Running `bundle exec fastalne take_screenshots`; this will take a while and the result should be:

screencapture-file-Users-gio-Developer-a8c-snios-fastlane-screenshots-screenshots-html-2021-01-15-20_54_45

In particular, notice the new screenshot with the inter note linking:

image

and the one with the new passcode screen:

image

If you run the screenshots composition with 

```
bf create_promo_screenshots source:$(pwd)/fastlane/screenshots/
```

You should see up-to-date copy and background, as per internal discussion, ref p2XJRt-2wR.

### Review
Only one platform developer required to review these changes, but anyone can perform the review.

### Release

### DIFF
--- a/SimplenoteScreenshots/SimplenoteScreenshots.swift
+++ b/SimplenoteScreenshots/SimplenoteScreenshots.swift
@@ -54,25 +54,6 @@ class SimplenoteScreenshots: XCTestCase {
 
         actionButton.tap()
 
-        // The collaborators screen will ask to access our contacts. Before loading the screen,
-        // let's setup an handler to dismiss the alert, so it doesn't come in the screenshots.
-        // Fastlane's snapshot _should_ do this for us, but it seems to happen unreliably.
-        //
-        // Not sure when this happened, but at least since Xcode 12.3, it doesn't seem necessary
-        // to dismiss the system dialog anymore in this test suite.
-        //
-        // Leaving the code to do it here for future reference.
-        //
-        // This logic is super rough, but for the context of this test script where we know we'll
-        // only get one alert, it'll do us. Obviously, when the time comes to refine the script by
-        // adopting the `BaseScreen` pattern from the WooCommerce iOS repo, this should logic should
-        // be improved as well.
-        //
-//         addUIInterruptionMonitor(withDescription: "Any system dialog") { alert in
-//            alert.buttons.firstMatch.tap()
-//            return true
-//        }
-
         let collaborateButton = app.staticTexts["Collaborate"].firstMatch
         XCTAssertTrue(collaborateButton.waitForExistence(timeout: 3))
 
@@ -82,14 +63,15 @@ class SimplenoteScreenshots: XCTestCase {
         let doneButton = app.buttons["Done"]
         XCTAssertTrue(doneButton.waitForExistence(timeout: 3))
 
-        // Now that we know the collaborators picker screen is presented, we also know that a system
-        // dialog to grant access to the contacts might have been shown. If it has, interacting with
-        // the app will trigger the UI interruption monitor. If it hasn't, this interaction won't
-        // result in any UI change (with how the UI is laid out at the time of writing this).
+        // The collaborators screen will ask to access our contacts the first time it's presented.
         //
-        // This code is unnecessary, but it's here for future reference. See not in the
-        // `addUIInterruptionMonitor` call above.
-//        app.tap()
+        // Let's wait to see if a system dialog appeared and dismiss it if so.
+        let systemAlert = XCUIApplication(bundleIdentifier: "com.apple.springboard").alerts.firstMatch
+        if systemAlert.waitForExistence(timeout: 2) {
+            // It doesn't matter whether we allow or deny access in the Simulator for the kind of
+            // screenshots we want to make.
+            systemAlert.buttons.firstMatch.tap()
+        }
 
         // The index 3 is _intentional_ as that's the desired position of the screenshot in the App
         // Store
--- a/SimplenoteScreenshots/SimplenoteScreenshots.swift
+++ b/SimplenoteScreenshots/SimplenoteScreenshots.swift
@@ -58,14 +58,20 @@ class SimplenoteScreenshots: XCTestCase {
         // let's setup an handler to dismiss the alert, so it doesn't come in the screenshots.
         // Fastlane's snapshot _should_ do this for us, but it seems to happen unreliably.
         //
+        // Not sure when this happened, but at least since Xcode 12.3, it doesn't seem necessary
+        // to dismiss the system dialog anymore in this test suite.
+        //
+        // Leaving the code to do it here for future reference.
+        //
         // This logic is super rough, but for the context of this test script where we know we'll
         // only get one alert, it'll do us. Obviously, when the time comes to refine the script by
         // adopting the `BaseScreen` pattern from the WooCommerce iOS repo, this should logic should
         // be improved as well.
-        addUIInterruptionMonitor(withDescription: "Any system dialog") { alert in
-            alert.buttons.firstMatch.tap()
-            return true
-        }
+        //
+//         addUIInterruptionMonitor(withDescription: "Any system dialog") { alert in
+//            alert.buttons.firstMatch.tap()
+//            return true
+//        }
 
         let collaborateButton = app.staticTexts["Collaborate"].firstMatch
         XCTAssertTrue(collaborateButton.waitForExistence(timeout: 3))
@@ -80,7 +86,10 @@ class SimplenoteScreenshots: XCTestCase {
         // dialog to grant access to the contacts might have been shown. If it has, interacting with
         // the app will trigger the UI interruption monitor. If it hasn't, this interaction won't
         // result in any UI change (with how the UI is laid out at the time of writing this).
-        app.tap()
+        //
+        // This code is unnecessary, but it's here for future reference. See not in the
+        // `addUIInterruptionMonitor` call above.
+//        app.tap()
 
         // The index 3 is _intentional_ as that's the desired position of the screenshot in the App
         // Store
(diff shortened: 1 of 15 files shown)


.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
   Test:
     executor:
       name: ios/default
-      xcode-version: "12.0"
+      xcode-version: "12.3.0"
```
mokagio: Use Xcode 12.3, so we can run on the latest iOS Simulators – 14.3.

Using the same version between local development and CI is helpful because there are quirks in UI tests between iOS Simulator versions.

Gemfile.lock
```diff
@@ -99,24 +99,26 @@ GEM
...
       multipart-post (>= 1.2, < 3)
+      ruby2_keywords
     faraday-cookie_jar (0.0.7)
       faraday (>= 0.8.0)
       http-cookie (~> 1.0.0)
     faraday_middleware (1.0.0)
       faraday (~> 1.0)
     fastimage (2.2.0)
-    fastlane (2.162.0)
+    fastlane (2.170.0)
```
mokagio: Routine Fastlane update, since we're doing some work related to it.

SimplenoteScreenshots/SimplenoteScreenshots.swift
```diff
@@ -30,69 +30,24 @@ class SimplenoteScreenshots: XCTestCase {
 
         let password = app.secureTextFields["Password"]
         XCTAssertTrue(password.waitForExistence(timeout: 10))
-        // -.-'
-        password.tap()
-        password.pasteText(text: ScreenshotsCredentials.testUserPassword)
+        password.typeSecureText(ScreenshotsCredentials.testUserPassword, using: app)
```
mokagio, line 43: In hindsight, that comment wasn't that useful. Extracting this unexpected logic into a dedicated function is a better option.

SimplenoteScreenshots/SnapshotHelper.swift
```diff
@@ -38,22 +38,13 @@ func snapshot(_ name: String, timeWaitingForIdle timeout: TimeInterval = 20) {
 }
 
 enum SnapshotError: Error, CustomDebugStringConvertible {
```
mokagio, line 40: The changes here were all part of the `snapshot` update process.

fastlane/metadata/en-US/app_store_screenshot_1.txt
```diff
@@ -0,0 +1,2 @@
+A simple
+note taking experience
```
mokagio: These keep disappearing after I run a code freeze. I need to move them to a different location 😳

mokagio:
Even though the code freeze for 4.30 has started today, I'm leaving this draft PR on that milestone because its changes are relevant for the release branch.

fastlane/metadata/en-US/app_store_screenshot_1.txt
```diff
@@ -0,0 +1,2 @@
+A simple note taking
+experience
```
mokagio, line 2: These files keep disappear every time I run the lane to fetch localizations from GlotPress.

I remember having to put them here to get the screenshots composition to work back when I first worked on it, but I was such a noob then 😅 I'll see if I can move them in a place where they won't disappear as a followup.

SimplenoteScreenshots/SimplenoteScreenshots.swift
```diff
@@ -30,69 +30,24 @@ class SimplenoteScreenshots: XCTestCase {
...
-        // Store
-        takeScreenshot("3-collaborators")
-
-        // Tapping done in the collaborate view dismisses the collaborate screen _and_ the action
-        // menu.
-        doneButton.tap()
-
-        let backButton = app.buttons.matching(NSPredicate(format: "label = %@", "Notes")).firstMatch
-        XCTAssertTrue(backButton.waitForExistence(timeout: 3))
-        backButton.tap()
```
mokagio, line 95: We don't take a screenshot of the collaboration screen anymore. Instead, we take one of the note inter-linking, as you can see in the new code below.

SimplenoteScreenshots/SimplenoteScreenshots.swift
```diff
@@ -135,36 +129,31 @@ class SimplenoteScreenshots: XCTestCase {
         openMenu(using: app)
         loadPasscodeScreen(using: app)
         // Set the passcode
-        // Writing the value here in clear because anyways one can see it being typed.
-        type("1234", onFirstKeyboardOf: app)
+        // Writing the value here in clear because one can see it being typed anyways.
+        typeOnPassCodeScreen(1234, using: app)
```
mokagio, line 145: The passcode screen now has custom buttons instead of loading the iOS keyboard.

mokagio:
Moved this to 4.31 because I didn't get a review in time for 4.30. 

That's not a big deal because whether the code is in the release branch is irrelevant from the automation point of view. Nothing stops us from running it and since it doesn't go into the user-facing app I think it's acceptable for the code not to be there.

mokagio:
Reverted to draft because there's a little experiment we want to try to make the screenshots more readable.

Internal ref p2XJRt-2wR#comment-5014.

SimplenoteScreenshots/SimplenoteScreenshots.swift
```diff
@@ -135,38 +141,40 @@ class SimplenoteScreenshots: XCTestCase {
...
+        guard app.staticTexts["Verify Your Email"].waitForExistence(timeout: 10) else { return }
+        app.buttons["icon cross"].tap()
+    }
 
-        menu.tap()
+    func logout(using app: XCUIApplication) {
+        getMenuButtonElement(from: app).tap()
 
-        let settings = app.textFields["Settings"]
+        let settings = app.staticTexts["Settings"]
```
mokagio, line 177: This started failing when moving from the 14.3 to the 14.4 Simulator... 🤷‍♂️

SimplenoteScreenshots/SimplenoteScreenshots.swift
```diff
@@ -204,11 +254,44 @@ class SimplenoteScreenshots: XCTestCase {
...
+        //
+        // - iPhone 8 Plus
+        // - iPhone Xs
+        // - iPad Pro 12.9" 2nd generation
+        // - iPad Pro 12.9" 3rd generation
+        //
+        // Of those, the only one requiring the secure text field workaround is the iPhone 8 Plus
+        // one.
+        return app.isDeviceIPhone8Plus()
+    }
```
mokagio, line 292: This is not true anymore, because we now run on iPhone 8 instead of 8 Plus and X instead of Xs (Max).

I think it's worth keeping this code in here even if technically unused, just in case we'll have problems with the secure text field in the future, a component which has proven to be fiddly over the years...

fastlane/Fastfile
```diff
@@ -493,6 +493,13 @@ end
       # But fail completely after those 3 retries
       stop_after_first_error: true,
 
+      # Run one Simulator at a time. One of the screenshots requires adding
+      # text to a note but because we're using the same account across all the
+      # Simulators, editing the note on one will result in the changes
+      # appearing on the other making for unexpected and inconsistent
+      # screnshots
+      concurrent_simulators: false,
```
mokagio, line 501: Took me a while to figure this out 😳

mokagio:
@jkmassel this is finally ready to merge. It turned out to be a bit long and with a lot of commits that go back and forth with the different experiments and fix attempts, so definitely best to look at the end result only.

I think I ironed out all of the flakes, but who knows... The last issue I had which I'm not sure I fixed or not was the passcode screen looking like it tapped the first passcode button, the number 1, but somehow not registering it: the dot in the progress view wouldn't update and when the automation proceeded to press 2, 3, and 4 the passcode would end up short of one number and the test would fail.

If you experience that, I'd ask you to run the tests again. I think it's an Xcode UI testing quirk rather than a genuine issue on our end. _I could obviously be wrong about that, so if you have any insights please do let me know_ 😄 

Also, in case you want to try the promo screenshots composition, the masks for the promo screenshots don't include the notch. I had to "hack" the action to draw the device frame on the canvas last. I plan to open an RFC/PR to let the caller specify some kind of composition order. It would be best to get properly cut masks, but if that's not possible and we're in a rush, I think my hacky approach is alright.

mokagio:
> but if that's not possible and we're in a rush, I think my hacky approach is alright.

Although... We shouldn't be in a rush for something like screenshots 🤔

SimplenoteScreenshots/SimplenoteScreenshots.swift
jkmassel: Hey – this is the passcode to my luggage!

jkmassel (approved):
This LGTM – I wasn't able to test the composition because `drawText` doesn't work in Big Sur 🤦 

But the screenshots ran very nicely on the first try, and the composition file looks good. 

I think we can land this to avoid merge issues – in the end we should generate these in CI anyway, and there's Catalina images there that can run the compositor 😅

mokagio:
> I wasn't able to test the composition because drawText doesn't work in Big Sur 🤦

Ouch! I'm still on Catalina on my main machine. This is something we might want to look into (_but who's got time for that_ 🤣) since with Xcode 12.5 we'll all be forced to upgrade to Big Sur. Of course, we could leave the CI for the screenshots on Catalina, but who knows what other issues with the divergent SDKs we might face 🙃

PR #1069, opened 2021-01-05
Nü passcode screen: IV

### LINKED ISSUE #761: Nü Passcode

eshurakov:
### Fix
In this PR we're finally removing `DTPinLockController` and replacing it with `PinLockViewController`. The final piece is the passcode screen to unlock the app.
The app should behave exactly as it was, but with new shiny passcode UI.

Related issue: #761

|Before|After|
|-|-|
|Simulator Screen Shot - iPhone 12 - 2021-01-05 at 21 24 18|Simulator Screen Shot - iPhone 12 - 2021-01-05 at 21 17 59|
|Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-01-05 at 21 25 42|Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-01-05 at 21 18 08|

### Test (Passcode)
1. Set a passcode
2. Move the app to the background
3. Open multitasking view
4. Check that passcode screen is visible and covers the whole screen
5. Open the app

- [x] Check that there is no "Cancel" button and it is not possible to dismiss the screen
- [x] Entering incorrect passcode shakes the header and shows an error message with the number of attempts
- [x] Entering correct passcode dismisses the screen and gives access to the app

### Test (Biometry)
1. Set a passcode
2. Enable Touch ID / Face ID
3. Move the app to the background
4. Open the app

- [x] If it's the first time using Touch ID / Face ID, system permission dialog will be shown
- [x] Using Touch ID / Face ID let you unlock the app
- [x] Providing wrong finger / face let you cancel Touch ID / Face ID flow and enter a passcode

### Test (Timeout)
1. Set a passcode
2. Set a timeout
3. Move the app to the background
4. Open the app

- [x] Opening the app before timeout unlocks the app without the need to enter passcode
- [x] Opening the app after timeout blocks the app with passcode screen

### Review

### Release

`RELEASE-NOTES.txt` was updated in 37d767e4 with:
 
> New passcode interface


Simplenote/SPPinLockManager.swift
```diff
@@ -47,12 +50,117 @@ class SPPinLockManager: NSObject {
...
+        }
+
+        switch context.biometryType {
+        case .none:
+            return nil
+        case .touchID:
+            return .touchID
+        case .faceID:
+            return .faceID
+        @unknown default:
```
jleandroperez, line 140: I think you can drop the `case .none` by removing the `@unknown` modifier here

Simplenote/SPPinLockManager.swift
```diff
@@ -47,12 +50,117 @@ class SPPinLockManager: NSObject {
...
+    @objc
+    static var isEnabled: Bool {
+        pin?.isEmpty == false
+    }
+
+    /// Set pin
+    ///
+    static func setPin(_ pin: String) {
+        self.pin = pin
+    }
```
jleandroperez, line 80: Q: Since we have the `static var pin: String` below, would it be possible to drop this one?

Simplenote/SPPinLockManager.swift
```diff
@@ -47,12 +50,117 @@ class SPPinLockManager: NSObject {
...
+
+    /// Remove pin
+    ///
+    @objc
+    static func removePin() {
+        pin = nil
+        shouldUseBiometry = false
```
jleandroperez, line 87: Ahhh perhaps this is why you kept the methods?

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -252,3 +228,59 @@ extension SPAppDelegate: SimperiumDelegate {
...
+    }
+}
+
+
+// MARK: - PinLockVerifyControllerDelegate
+//
+extension SPAppDelegate: PinLockVerifyControllerDelegate {
+    func pinLockVerifyControllerDidComplete(_ controller: PinLockVerifyController) {
+        UIView.animate(withDuration: UIKitConstants.animationShortDuration) {
+            self.pinLockWindow?.alpha = 0.0
```
jleandroperez:
```suggestion
            self.pinLockWindow?.alpha = UIKitConstants.alpha0_0
```

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -252,3 +228,59 @@ extension SPAppDelegate: SimperiumDelegate {
...
+        pinLockWindow = UIWindow(frame: UIScreen.main.bounds)
+        pinLockWindow?.rootViewController = viewController
+        pinLockWindow?.makeKeyAndVisible()
+    }
+
+    /// Dismiss the passcode lock window if the user has returned to the app before their preferred timeout length
+    ///
+    @objc
+    func ensurePasscodeLockIsDismissed() {
+
```
jleandroperez: Nipticky: Extra newline here

jleandroperez (approved):
Looks good!!

:shipit:

Simplenote/SPPinLockManager.swift
eshurakov: I thought to keep `pin` private and only provide methods to update or verify it. Unfortunately forgot to make the var private here, but did it in the next PR.

Simplenote/SPPinLockManager.swift
eshurakov: True, but I think it's better to keep it, so that if there are new cases, we would see an error during compilation.

Simplenote/SPPinLockManager.swift
eshurakov: Answered above :-)

mokagio:
In #1074, I verified that updating Fastlane to the latest version, 2.171.0, fixes the Installable Build CI failure.

I opened #1075 with the fix itself. We can merge that and rebase this PR stack on top of the updated `develop`, cherry pick the fix, or simply run the following command on this branch (i.e. cherry pick 00a8827f27d0781ad3820da45869641a6e4b5374)

```
bundle update fastlane
```

Up to you 👍

eshurakov:
Thank you @mokagio! 👍

SylvesterWilmott (approved):
Looks great and works as expected. Thanks!

eshurakov:
Thank you @jleandroperez and @SylvesterWilmott!

pachlava:
[UI Testing] Verified on iPhone XR | iOS 14.3 | SN 4.30

PR #1088, opened 2021-01-14
Prevent bypassing login screen

### LINKED ISSUE #1035: Bypassing login / signup screen
Opening the app using `simplenote://new` deep link bypasses onboarding (login / signup) and takes you directly to the app.

### Expected
Deep link is ignored and onboarding is still on screen.

### Observed
Onboarding is dismissed.

### Reproduced
1. Open the app
2. Type `simplenote://new` in Safari

eshurakov:
### Fix
This PR adds a check if user is authenticated before processing incoming URLs, shortcuts and activities.

Closes #1035 

### Test
1. Logout or fresh install
2. Put the app in the background

- [x] Check that typing `simplenote://new` in Safari doesn't dismiss login screen
- [x] Check that log in with Wordpress works

### Review

### Release

### DIFF
--- a/Simplenote/SPAppDelegate.m
+++ b/Simplenote/SPAppDelegate.m
@@ -618,6 +618,21 @@
 
 - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options
 {
+    if (!self.simperium.user.authenticated) {
+        if ([WPAuthHandler isWPAuthenticationUrl: url]) {
+            SPUser *newUser = [WPAuthHandler authorizeSimplenoteUserFromUrl:url forAppId:[SPCredentials simperiumAppID]];
+            if (newUser != nil) {
+                self.simperium.user = newUser;
+                [self.navigationController dismissViewControllerAnimated:YES completion:nil];
+                [self.simperium authenticationDidSucceedForUsername:newUser.email token:newUser.authToken];
+
+                [SPTracker trackWPCCLoginSucceeded];
+            }
+        }
+
+        return YES;
+    }
+
     // URL: Open a Note!
     if ([self handleOpenNoteWithUrl:url]) {
         return YES;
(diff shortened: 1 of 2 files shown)


Simplenote/SPAppDelegate.m
```diff
@@ -618,6 +618,21 @@ - (void)indexSpotlightItems
...
+                self.simperium.user = newUser;
+                [self.navigationController dismissViewControllerAnimated:YES completion:nil];
+                [self.simperium authenticationDidSucceedForUsername:newUser.email token:newUser.authToken];
+
+                [SPTracker trackWPCCLoginSucceeded];
+            }
+        }
+
+        return YES;
+    }
```
jleandroperez, line 624: REALLY Nipticky!! Maybe avoid the nested `if(s)` (by moving the first block to another method?)

```
    if (!self.simperium.user.authenticated) {
         [self performDotcomAuthenticationWithURL:url];
         return YES;
     }
```

(or so?)

Simplenote/SPAppDelegate.m
```diff
@@ -653,22 +668,6 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
         [_simperium save];
         
         [self presentNote:newNote];
-    } else if ([WPAuthHandler isWPAuthenticationUrl: url]) {
-        if (self.simperium.user.authenticated) {
-            // We're already signed in
-            [[NSNotificationCenter defaultCenter] postNotificationName:kSignInErrorNotificationName
```
jleandroperez, line 659: Just making 100% sure we want this one gone?

jleandroperez (approved):
Nipticky comments in, feel free to disregard!!
Works exactly as advertised 🔥 

:shipit:

Simplenote/SPAppDelegate.m
eshurakov: It seems to be a very edge case to me, to try to log in with WP when you already logged in 🤔 
The only thing this code was doing before is triggering a notification that would show an alert, so I think it can safely go away. 
WDYT? :-)

Simplenote/SPAppDelegate.m
eshurakov: I have extracted it to a separate method (https://github.com/Automattic/simplenote-ios/pull/1088/commits/404f7df92e762a19d269d501d1c2ad64e2b029ef), looks better! Thank you for the suggestion 👍

jleandroperez (approved):
Thank you for doing those changes!! looks super!!

:shipit:

Issue #1037, opened 2020-11-19
"Empty" trash button remains enabled when notes are restored\deleted individually


pachlava:
### Expected
"Empty" button in Trash should be disabled when there are no notes

### Observed
With scenario below, "Empty" button remains enabled even if there are no notes

### Reproduced
> 1. Make sure there's at least 1 note in Trash
> 2. In Trash, restore\delete all present notes individually by swiping
> 3. "Empty" button is still enabled

![Simulator Screen Shot - iPhone 11 - 2020-11-19 at 16 48 13](https://user-images.githubusercontent.com/73365754/99681670-13702400-2a87-11eb-97d9-16ae82ced388.png)

Make|Model|iOS Version|App Version
-|-|-|-
iPhone Sim. | 11 | 14.1 | 4.28

jleandroperez:
Closed via #1093 by @charliescheer 

Thanks everyone!!

PR #1093, opened 2021-01-17
Issue #1037 Empty Trash Button Remains Enabled

charliescheer:
### Fix
This PR fixes issue #1037 

When viewing the notes trash list, if you remove all of the notes either by restoring them or deleting them one at a time, the empty notes button remains active.  

With this fix if when the list becomes empty by removing items either using the empty button or by removing single items, the empty trash button will become disabled.

Example:

### Test
1. Create multiple new notes and then trash them
2. Go to the trash menu and restore them all to the notes list.  Make sure that the empty trash button stays enabled until the last note is restore
3. Trash the notes again, return to the notes list and delete each note individually by swiping to the left and selecting the red trash button.  Make sure that the trash button stays enabled till the last note is deleted
4. Create a few new notes and trash them
5. Return to the trash list and use the empty trash button.  Make sure the trash button disables after the list is cleared.

### Review

### Release

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -184,6 +184,7 @@ extension SPNoteListViewController {
 
             self.tableView.performBatchChanges(sectionsChangeset: sectionsChangeset, objectsChangeset: objectsChangeset) { _ in
                 self.displayPlaceholdersIfNeeded()
+                self.refreshEmptyTrashState()
             }
         }
     }
@@ -670,7 +671,6 @@ private extension SPNoteListViewController {
             UIContextualAction(style: .destructive, image: .image(name: .trash), backgroundColor: .simplenoteDestructiveActionColor) { (_, _, completion) in
                 SPTracker.trackListNoteDeleted()
                 SPObjectManager.shared().permenentlyDeleteNote(note)
-                self.setEmptyTrashEnabled()
                 completion(true)
             }
         ]
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -391,7 +391,7 @@ extension SPNoteListViewController {
         let isTrashOnScreen = self.isDeletedFilterActive
         let isNotEmpty = !self.isListEmpty
         
-        self.emptyTrashButton.isEnabled = isTrashOnScreen && isNotEmpty
+        emptyTrashButton.isEnabled = isTrashOnScreen && isNotEmpty
     }
 }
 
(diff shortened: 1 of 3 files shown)


Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -500,16 +500,23 @@ - (void)update
...
     [self hideRatingViewIfNeeded];
 }
 
+- (void)setEmptyTrashEnabled
+{
+    BOOL isTrashOnScreen = self.isDeletedFilterActive;
+    BOOL isNotEmpty = !self.isListEmpty;
+
+    self.emptyTrashButton.enabled = isTrashOnScreen && isNotEmpty;
+}
```
jleandroperez: Two notes!

1. Can we rename this method to something like `refreshEmptyTrashState` or `refershEmptyTrashButton`?
   Why: by convention, in ObjC, properties would "hide" access to methods named `setProperty:value` or `property` (getter). Although this API does update the trash button, it doesn't accept values.

2. We're moving all new code over to Swift. Mind placing this one in `SPNoteListViewController+Extensions`?

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -660,6 +660,7 @@ private extension SPNoteListViewController {
             UIContextualAction(style: .destructive, image: .image(name: .trash), backgroundColor: .simplenoteDestructiveActionColor) { (_, _, completion) in
                 SPTracker.trackListNoteDeleted()
                 SPObjectManager.shared().permenentlyDeleteNote(note)
+                self.setEmptyTrashEnabled()
```
jleandroperez: A better sport for this one would be `startDisplayingEntities`. If you empty the `trash` in a 2nd device / web, the Trash button would remain enabled otherwise.

jleandroperez (commented):
Looking good @charliescheer !! two nipticky notes, and one enhancement regarding remote updates.

Thank you!!

charliescheer:
Great suggestions!  I have made these changes.  Let me know what you think!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -382,6 +383,16 @@ extension SPNoteListViewController {
...
     }
+    
+    /// Sets the state of the trash button
+    ///
+    @objc
+    func refreshEmptyTrashState() {
+        let isTrashOnScreen = self.isDeletedFilterActive
+        let isNotEmpty = !self.isListEmpty
+        
+        self.emptyTrashButton.isEnabled = isTrashOnScreen && isNotEmpty
```
jleandroperez: No need for `self` anywhere 😄

Simplenote/Classes/SPNoteListViewController.h
```diff
@@ -29,6 +29,8 @@
 @property (nonatomic) CGFloat                                               keyboardHeight;
 @property (nonatomic) BOOL                                                  firstLaunch;
 @property (nonatomic) BOOL                                                  mustScrollToFirstRow;
+@property (nonatomic, strong) UIBarButtonItem
+    *emptyTrashButton;
```
jleandroperez: Extra newline here!

jleandroperez (approved):
Two cosmetic notes! One extra newline, and a few extra `self` calls.

Other than that, please :shipit: when ready!

Simplenote/Classes/SPNoteListViewController.h
charliescheer: Updated these two suggestions!

Issue #33, opened 2016-08-14
Swipe to dismiss keyboard


kwonye:
Most of Apple's built in apps allow you to swipe the keyboard down off the screen when being displayed over a UIScrollView (or descendent of).

Could be as simple as flipping a boolean property on the UIScrollView in question.

Opened by @beaucollins

jleandroperez:
Closed via #1094 by @charliescheer 

Thank youuu!!

PR #1094, opened 2021-01-17
Adds ability to swipe to dismiss SPTextView

charliescheer:
### Fix
This PR fixes issue #33 enabling iOS' built in swipe to dismiss keyboard action.

With this fix, when working on a note or a tag, the keyboard can now be dismissed with a swipe down motion.

example:

### Test

1. open up a note and tap into the text to make the keyboard appear
2. tap slightly above the keyboard and swipe down to dismiss the keyboard
3. tap into the ta area to make the keyboard appear
4. tap above the tag area and swipe down to dismiss the keyboard

### Release

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
@@ -113,7 +113,7 @@ extension SPNoteEditorViewController {
     ///
     @objc
     func configureTextViewKeyboard() {
-        noteEditorTextView?.keyboardDismissMode = UIScrollView.KeyboardDismissMode.interactive
+        noteEditorTextView?.keyboardDismissMode = .interactive
     }
 }
 
--- a/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
@@ -108,6 +108,13 @@ extension SPNoteEditorViewController {
         interlinkProcessor.delegate = self
         interlinkProcessor.contextProvider = self
     }
+    
+    /// Sets up the Keyboard
+    ///
+    @objc
+    func configureTextViewKeyboard() {
+        noteEditorTextView?.keyboardDismissMode = UIScrollView.KeyboardDismissMode.interactive
+    }
 }
 
 
--- a/Simplenote/Classes/SPTextView.m
+++ b/Simplenote/Classes/SPTextView.m
@@ -48,6 +48,8 @@ CGFloat const TextViewHighlightCornerRadius = 3;
         caret position.
      */
         self.textContainer.heightTracksTextView = NO;
+
+        self.keyboardDismissMode = UIScrollViewKeyboardDismissModeInteractive;
     }
     return self;
 }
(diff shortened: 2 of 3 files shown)


Simplenote/Classes/SPTextView.m
```diff
@@ -48,6 +48,8 @@ - (instancetype)init {
         caret position.
      */
         self.textContainer.heightTracksTextView = NO;
+
+        self.keyboardDismissMode = UIScrollViewKeyboardDismissModeInteractive;
```
jleandroperez: Suggestion:

1. `SPNoteEditorViewController+Extensions.swift` new `configureTextView` method (there's a category that has a bunch of configure methods)
2. Set the dismissMode in the `EditorViewController` instead

Why: Since this is a concrete configuration, which depends on the way the component needs to be used, it'd be best to set in place.

jleandroperez (commented):
Leaving you a quick note up there!!

Plus:
1. Perform the Swipe to Dismiss gesture
2. Pay **close close close** attention to the bottom edge of the screen

The Keyboard Autocorrect Suggestion will get stuck for a second. This appears to be an iOS bug (last time I checked, happens in an empty project with a UITextView).

Checked with Sly, we're good to ship with this glitch, and maybe improve down the road (it'll hopefully be fixed by Apple soon-ish?).

https://user-images.githubusercontent.com/1195260/105048606-84619700-5a4a-11eb-9c7b-9d26ae3d1790.MP4

Simplenote/Classes/SPTextView.m
charliescheer: >Why: Since this is a concrete configuration, which depends on the way the component needs to be used, it'd be best to set in place.

Gotcha.  I moved it to SPNoteEditorViewController+Extensions.swift

One thing I am not sure about.  In the SPNoteEditorViewController.m file there is already an objective-c method for configureTextView.  I saw two options for setting this up and wasn't sure which to go with.
1. Create a new method on the swift file configureTextViewKeyboard which contains the change to the keyboard dismiss mode and then call that method when the view controller is setup.  This is the way I have done it currently because it is makes the fewest changes to how the app is currently setup
2. Or, I could see moving the entire configureTextViewKeyboard into the swift file instead.  If the longer term goal of SNiOS is to get as much of the app into swift as possible, this might be the better solution, but didn't want to make that change without discussing if this was the way we wanted to go.
3. Possible other option I haven't thought of

What do you think would be best?

Have added a commit that follows option one, which can be reviewed.

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -108,6 +108,13 @@ extension SPNoteEditorViewController {
         interlinkProcessor.delegate = self
         interlinkProcessor.contextProvider = self
     }
+    
+    /// Sets up the Keyboard
+    ///
+    @objc
+    func configureTextViewKeyboard() {
+        noteEditorTextView?.keyboardDismissMode = UIScrollView.KeyboardDismissMode.interactive
```
jleandroperez: Swift's Type Inference allows you to do just this:

```
noteEditorTextView?.keyboardDismissMode = .interactive
```

jleandroperez (approved):
Looks good @charliescheer !!

:shipit:

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
charliescheer: Ah cool!  Updated.  Will ship once CI checks are complete

Issue #1053, opened 2020-12-01
Tags: Unit Test hashing


eshurakov:
Add unit tests for tag hashing function.

Use the following scenarios:

```
1.
Name: TáßvĒёи兔子@#$%^&*+-=_`~/?., ><{}[]\()"|':;
Hash: t%C3%A1%C3%9Fv%C4%93%D1%91%D0%B8%E5%85%94%E5%AD%90%40%23%24%25%5E%26%2A%2B%2D%3D%5F%60%7E%2F%3F%2E%2C%20%3E%3C%7B%7D%5B%5D%5C%28%29%22%7C%27%3A%3B

2. 
Name: \u00E9
Hash: %C3%A9

3.
Name: \u0065\u0301
Hash: %C3%A9

The names for 2 and 3 are the same (é), but represented differently and in the end should have the same hash.
```

Ref: p2XJRt-2tq

charliescheer:
Hey @eshurakov I was looking into this and had a few questions for you.

First, the function that we are looking to test here is the String function byEncodingAsTagHash correct?

Also, for your first example.  The hash you provided here has an additional character that is not in the name.

`t%C3%A1%C3%9Fv%C4%93%D1%91%D0%B8%E5%85%94%E5%AD%90%40%23%24%25%5E%26%2A%2B%2D%3D%5F%60%7E%2F%3F%2E%2C%20%3E%3C%7B%7D%5B%5D_%5C_%28%29%22%7C%27%3A%3B`

That backslash character does not appear in the name that you used as your example.  Should the backslash be part of the name, or should the backslash not be in the hash output?

eshurakov:
Hey @charliescheer! Good catch, `\` should be in the name, I missed it.

Correct, what we want to test is `byEncodingAsTagHash`, but I see that it already has some tests. It's good to check if we need to extend those tests, possibly they already cover everything from here. (I should've checked before creating this task 🙈)

Thank you!

charliescheer:
Closed via #1096

PR #1096, opened 2021-01-18
Extended unit tests for string hashing

charliescheer:
### Fix
This PR responds to issue #1053 and adds additional unit tests to confirm the hashing done by the string method byEncodingAsTagHash

Some tests for this method already existed, so in this PR I have extended the test "testByEncodingTagAsHashAllowsUsToProperlyCompareStringsThatEvaluateAsNotEqualOtherwise" to include some additional use cases, the cases specifically called for in #1053, and added a test to confirm the encoding of the special character set defined in the issue.

The new test that I created may be repeating the test before it, so may be unnecessary but take a look, I thought specifically checking the special characters would prove useful.  There may also be a more efficient way to get a similar character set than writing out the encoded hash as I did.

### Test
1. in Xcode go to NSStringSimplenoteTests.swift
2. Run the unit tests for the project
3. All tests should pass

### Review

### DIFF
--- a/SimplenoteTests/NSStringSimplenoteTests.swift
+++ b/SimplenoteTests/NSStringSimplenoteTests.swift
@@ -47,7 +47,7 @@ class NSStringSimplenoteTests: XCTestCase {
 }
 
 extension NSStringSimplenoteTests {
-    func testNonEqualStringsCreateSameHash(_ samples: [NSString]) {
+    private func testNonEqualStringsCreateSameHash(_ samples: [NSString]) {
         var count = 0
         
         while count + 1 < samples.count {
--- a/SimplenoteTests/NSStringSimplenoteTests.swift
+++ b/SimplenoteTests/NSStringSimplenoteTests.swift
@@ -40,8 +40,9 @@ class NSStringSimplenoteTests: XCTestCase {
             ["\u{0065}\u{0301}", "\u{00E9}"] // é
         ]
         
-        testNonEqualStringsCreateSameHash(differentStringsOneHashSamples[0])
-        testNonEqualStringsCreateSameHash(differentStringsOneHashSamples[1])
+        for sample in differentStringsOneHashSamples {
+            testNonEqualStringsCreateSameHash(sample)
+        }
     }
 }
 
--- a/SimplenoteTests/NSStringSimplenoteTests.swift
+++ b/SimplenoteTests/NSStringSimplenoteTests.swift
@@ -34,7 +34,7 @@ class NSStringSimplenoteTests: XCTestCase {
     ///
     /// - Note: When using the `Swift.String` class, the same comparison is actually correct.
     ///
-    func testByEncodingTagAsHashAllowsUsToProperlyCompareStringsThatEvaluateAsNotEqualOtherwise() {
+    func testNonEqualStringsCreateSameHash() {
         let sampleA = NSString(stringLiteral: "\u{0073}\u{0323}\u{0307}")
         let sampleB = NSString(stringLiteral: "\u{0073}\u{0307}\u{0323}")
         let sampleC = NSString(stringLiteral: "\u{1E69}")
--- a/SimplenoteTests/NSStringSimplenoteTests.swift
+++ b/SimplenoteTests/NSStringSimplenoteTests.swift
@@ -16,16 +16,6 @@ class NSStringSimplenoteTests: XCTestCase {
 
         XCTAssertTrue(expectedSet.isSuperset(of: escapedSet))
     }
-    
-    /// Verifies that `byEncodingAsTagHash` effectively escapes special characters
-    ///
-    func testByEncodingAsTagHashEncodesAllOfSpecialCharacters() {
-        let string = String("TáßvĒёи兔子@#$%^&*+-=_`~/?., ><{}[]\\()\"|':;")
-        let hash = string.byEncodingAsTagHash
-        let expected = "t%C3%A1%C3%9Fv%C4%93%D1%91%D0%B8%E5%85%94%E5%AD%90%40%23%24%25%5E%26%2A%2B%2D%3D%5F%60%7E%2F%3F%2E%2C%20%3E%3C%7B%7D%5B%5D%5C%28%29%22%7C%27%3A%3B"
-        
-        XCTAssertEqual(hash, expected)
-    }
 
     /// Verifies that `byEncodingAsTagHash` allows us to properly compare Unicode Strings that would otherwise evaluate as not equal.
     /// Although our (three) sample strings yield the exact same character`ṩ`, regular `isEqualString` API returns `false`.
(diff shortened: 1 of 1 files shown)


SimplenoteTests/NSStringSimplenoteTests.swift
```diff
@@ -28,17 +38,29 @@ class NSStringSimplenoteTests: XCTestCase {
         let sampleA = NSString(stringLiteral: "\u{0073}\u{0323}\u{0307}")
         let sampleB = NSString(stringLiteral: "\u{0073}\u{0307}\u{0323}")
         let sampleC = NSString(stringLiteral: "\u{1E69}")
+        let sampleD = NSString(stringLiteral: "\u{0065}\u{0301}")
+        let sampleE = NSString(stringLiteral: "\u{00E9}")
```
eshurakov: I think it's better to separate this test in two, because it is testing 2 sets of input data which are not related to each other. 
We can extract a function that would accept an array of strings and test them according to existing logic: check that strings in array are not equal and check that hashes are equal. After that we can call this function with different sets of data from our test.

Example pseudo code:
```
func testNonEqualStringsHaveTheSameHash(_ samples: [String]) { ... }
func test...() {
    testNonEqualStringsHaveTheSameHash(["\u{0073}\u{0323}\u{0307}", "\u{0073}\u{0307}\u{0323}", ""\u{1E69}""])
    testNonEqualStringsHaveTheSameHash(["\u{0065}\u{0301}", "\u{00E9}"])
}
```

eshurakov (changes requested):
Hey @charliescheer! 
Sending you a suggestion that I think will make this test better! Let me know what do you think.

SimplenoteTests/NSStringSimplenoteTests.swift
charliescheer: That is a good idea for separating the different values that should test the same.  I updated the tests, can you check it out again when you get a chance and see what you think?

I'm not sure it is the clearest on what it is testing, if you have any suggestions for improving the readability that would be great!

eshurakov: Looks good! But we can do better! :-)

First we can make `func testNonEqualStringsCreateSameHash(_ samples: [NSString])` private, because it's our internal helper method.

To make it more readable, we can combine samples in array like this:
```
let differentStringsOneHashSamples = [
    ["\u{0073}\u{0323}\u{0307}", "\u{0073}\u{0307}\u{0323}", "\u{1E69}"], // ṩ
    ["\u{0065}\u{0301}", "\u{00E9}"] // é
]
```

And in the test we can walk through our samples and call helper method `testNonEqualStringsCreateSameHash`. I think we can also rename the original test `testByEncodingTagAsHashAllowsUsToProperlyCompareStringsThatEvaluateAsNotEqualOtherwise` to `testNonEqualStringsCreateSameHash` but without parameters.

In `differentStringsOneHashSamples` I used swift strings instead of NSString, I think we don't really need to stick to NSString for most of the test, so it's fine to use Swift strings. The only "issue" is that comparing our samples using swift strings would return `true` as it seems that swift strings compare "characters" and not  underlying representation of the string. To solve this we can temporarily cast our strings to NSString: `XCTAssertFalse((samples[count] as NSString).equal(to: samples[testCount]))`

Another thing is that in original test we were comparing every sample to every sample, but I don't think it's needed. I think we can compare the first sample to all other samples and get the same coverage. If A == B and A == C, then B == C. This way we will need to walk the samples array only once.

p.s. you can use for loop instead of while + separate var: `for i in 0..<samples.count`

charliescheer:
>First we can make func testNonEqualStringsCreateSameHash(_ samples: [NSString]) private, because it's our internal helper method.

Complete

>testByEncodingTagAsHashAllowsUsToProperlyCompareStringsThatEvaluateAsNotEqualOtherwise to testNonEqualStringsCreateSameHash but without parameters.

much clearer, updated

>To solve this we can temporarily cast our strings to NSString: XCTAssertFalse((samples[count] as NSString).equal(to: samples[testCount]))

complete

>Another thing is that in original test we were comparing every sample to every sample, but I don't think it's needed. I think we can compare the first sample to all other samples and get the same coverage. If A == B and A == C, then B == C. This way we will need to walk the samples array only once.

That's a good point!  I have updated the helper function quite a bit to reflect this change.  I think it makes the whole test much easier to read

>p.s. you can use for loop instead of while + separate var: for i in 0..<samples.count
I had started with a for in loop, but for some reason found it cleaner to use the while when checking each property against each other, but after removing the b==c check I like the for in loop much better.

Let me know what you think!

SimplenoteTests/NSStringSimplenoteTests.swift
```diff
@@ -24,21 +34,26 @@ class NSStringSimplenoteTests: XCTestCase {
...
-        XCTAssertNotEqual(sampleA, sampleC)
-        XCTAssertNotEqual(sampleB, sampleC)
+    func testNonEqualStringsCreateSameHash() {
+        let differentStringsOneHashSamples = [
+            ["\u{0073}\u{0323}\u{0307}", "\u{0073}\u{0307}\u{0323}", "\u{1E69}"], // ṩ
+            ["\u{0065}\u{0301}", "\u{00E9}"] // é
+        ]
+        
+        testNonEqualStringsCreateSameHash(differentStringsOneHashSamples[0])
+        testNonEqualStringsCreateSameHash(differentStringsOneHashSamples[1])
```
eshurakov: Nitpick: this can be done in a loop, this way if we ever need to add more samples later, we can just add them to `differentStringsOneHashSamples`

SimplenoteTests/NSStringSimplenoteTests.swift
```diff
@@ -24,21 +34,26 @@ class NSStringSimplenoteTests: XCTestCase {
...
-        XCTAssertEqual(hashB, hashC)
+extension NSStringSimplenoteTests {
+    private func testNonEqualStringsCreateSameHash(_ samples: [String]) {
+        let sampleToTest = samples[0]
+        
+        for i in 1..<samples.count {
+            XCTAssertNotEqual((sampleToTest as NSString), (samples[i] as NSString))
+            print("sample to test: \(sampleToTest) sample to comare: \(samples[i])")
+            XCTAssertEqual(sampleToTest.byEncodingAsTagHash, samples[i].byEncodingAsTagHash)
+            print("sample to test: \(sampleToTest.byEncodingAsTagHash) sample to comare: \(samples[i].byEncodingAsTagHash)")
```
eshurakov: Let's remove `print` :-)

eshurakov (approved):
Left a few nitpick comments, but other than that looks great!

Let's :shipit:

SimplenoteTests/NSStringSimplenoteTests.swift
charliescheer: oops.... ;P

PR #1097, opened 2021-01-19
Issue 19 rtl issues - process guidance

### LINKED ISSUE #19: RTL support? (OS X, too)
From a user:

`I use simplenote on Mac and iOS. Both systems have native support for rtl (right to left) languages, In my case it's Arabic. Simplenote on both systems doesn't keep the text on the right as it should be for a rtl text. I could set the direction to rtl from the menu (in iOS) but it gets back to the left side after leaving the app. On Mac it’s not possible to set the note to RTL direction. I hope it’s a bug that could fixed. I really like the app, this is the only reason i didn’t fully switch to Simplenote.` 

Opened by @roundhill

charliescheer:
In this PR, I am looking for a bit of guidance to see if I am on the right path with fixing this issue or if there might be another way to do this.

This branch is intended to improve the RTL support for Simplenote iOS, which will fix #19 Currently Simplenote works fine with RTL languages as long as the app and the keyboard language are in agreement with the text orientation.  For example, if the app/system is set to English (LTR) and the keyboard being used is in Arabic (RTL) the alignment and text direction will not be correct.

The approach that I have started taking in so far to try and fix this is to setup a Notification Center observer for "UITextInputMode.currentInputModeDidChangeNotification" to be notified when ever the input mode changes.  From that notification the keyboard language can be derived and the language direction can be changed accordingly.  

I have a few issues with how this works though, which is why I am unsure about spending more time going in this direction.  
1. Currently the easiest way to make a change to the language direction is to change it for the entire text field, so all text changes alignment, so you could not have multiple languages in the same note if the languages used a different orientation
2. The notification seems to be returning the previous language not the language that has been changed to, which is fine for only two languages but will likely create issues if there are more
3. The system it's self seems to be holding onto it's one orientation values, so when I have been able to change the alignment for RTL languages, as soon as I start typing, it switches back to left alignment.  I haven't sorted that one out yet, but makes me think that I could be on the wrong track.

What do you think of this current approach, do you think I am on the right track here, or do you have additional suggestions of things I could look into?

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
@@ -709,17 +709,22 @@ extension SPNoteEditorViewController {
         tagView.keyboardAppearance = .simplenoteKeyboardAppearance
     }
 
-    private func refreshTextStorage() {
+    private func refreshTextStorage(with textDirection: NSWritingDirection = .natural) {
         let headlineFont = UIFont.preferredFont(for: .title1, weight: .bold)
         let defaultFont = UIFont.preferredFont(forTextStyle: .body)
         let textColor = UIColor.simplenoteNoteHeadlineColor
-        let lineSpacing = defaultFont.lineHeight * Metrics.lineSpacingMultipler
+        
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.lineSpacing = defaultFont.lineHeight * Metrics.lineSpacingMultipler
+        paragraphStyle.baseWritingDirection = textDirection
+        
         let textStorage = noteEditorTextView.interactiveTextStorage
+        
 
         textStorage.defaultStyle = [
             .font               : defaultFont,
             .foregroundColor    : textColor,
-            .paragraphStyle     : NSMutableParagraphStyle(lineSpacing: lineSpacing)
+            .paragraphStyle     : paragraphStyle
         ]
 
         textStorage.headlineStyle = [
--- a/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
@@ -108,6 +108,17 @@ extension SPNoteEditorViewController {
         interlinkProcessor.delegate = self
         interlinkProcessor.contextProvider = self
     }
+    
+    @objc
+    func configureKeyboardObserver() {
+        let nc = NotificationCenter.default
+        nc.addObserver(self, selector: #selector(inputModeDidChange), name: UITextInputMode.currentInputModeDidChangeNotification, object: nil)
+    }
+    
+    @objc
+    func inputModeDidChange(_ notification: Notification) {
+
+    }
 }
 
 
--- a/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
@@ -117,9 +117,8 @@ extension SPNoteEditorViewController {
     
     @objc
     func inputModeDidChange(_ notification: Notification) {
-        guard let userInfo = notification.userInfo,
-              let inputMode = userInfo["UITextInputFromInputModeKey"] as? UITextInputMode,
-              let lang = inputMode.primaryLanguage else {
+        guard let input = noteEditorTextView.textInputMode,
+              let lang = input.primaryLanguage else {
             return
         }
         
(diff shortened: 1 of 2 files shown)


eshurakov:
Hey @charliescheer! 

Found [this document](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/SupportingRight-To-LeftLanguages/SupportingRight-To-LeftLanguages.html) that mentions that standard iOS components automatically handle bidirectional text. So it could be that we incorrectly set up our text view or did some customisation that affects RTL support.

What we can do:
* Check Apple Notes app and see if described behaviour works there.
* To isolate an issue we can create a simple screen with only `UITextView`, present it somewhere in the app and check if it works there. If it works, we can dig into our custom text view to see what prevents it from working as expected.

charliescheer:
>Check Apple Notes app and see if described behaviour works there.
To isolate an issue we can create a simple screen with only UITextView, present it somewhere in the app and check if it works there. If it works, we can dig into our custom text view to see what prevents it from working as expected.

Yup, did that test and with a plain UITextView the LTR/RTL changes with the keyboard no problem.  Will look into our logic to see if I can find what is overriding the default

charliescheer:
Interestingly, the header of a note does align it's self correctly, just not the body.  Not sure why yet...:

eshurakov:
`SPInteractiveTextStorage` looks like it could be a culprit because it is handling storage of the text.

jleandroperez:
Interesting. Maybe this is related (method `refreshTextStorage`):

```
        textStorage.defaultStyle = [
            .font               : defaultFont,
            .foregroundColor    : textColor,
            .paragraphStyle     : NSMutableParagraphStyle(lineSpacing: lineSpacing)
        ]

        textStorage.headlineStyle = [
            .font               : headlineFont,
            .foregroundColor    : textColor,
        ]
```

LTR can be [set via an AttributedString property](https://stackoverflow.com/questions/47280999/nsattributedstring-add-writing-direction-for-arabic-language). I'm not sure, though, why Headline would not get such style. But it's a good starting point!

charliescheer:
Looks like you are correct @jleandroperez Removed the active bits of refreshTextStorage and the issue goes away

charliescheer:
Okay y'all, I made some improvements on this, but I am missing a few things and would like to do another check in before I go too much further with this.

So improvement, a note can now be in both LTR and RTL at the same time:



### Issues:
Issue 1. it can not do the different orientations on the same line (base apple textView can)


Issue 2. if you leave the note and return to it the text direction gets lost and it defaults to the natural text direction state of the device


For issue 1. I did some printing of what is actually going on when a basic textView saves the attributed text, and it seems that it may be saving the settings per word..?  which is how it is able to change direction in each line.  

For issue 2. doing that same test I see that after we reload the text the attributes change to natural, so somewhere. along the process of saving the strings we are losing the attributes

**before save:**
}في{
    NSColor = "<UIDynamicProviderColor: 0x60000325e460; provider = <__NSMallocBlock__: 0x600003cb3f00>>";
    NSFont = "<UICTFont: 0x7ff7b966e4e0> font-family: \".ArabicUIText-Regular\"; font-weight: normal; font-style: normal; font-size: 17.00pt";
    NSParagraphStyle = "Alignment 4, LineSpacing 0, ParagraphSpacing 0, ParagraphSpacingBefore 0, HeadIndent 0, TailIndent 0, FirstLineHeadIndent 0, LineHeight 0/0, LineHeightMultiple 0, LineBreakMode 0, Tabs (\n    28L,\n    56L,\n    84L,\n    112L,\n    140L,\n    168L,\n    196L,\n    224L,\n    252L,\n    280L,\n    308L,\n    336L\n), DefaultTabInterval 0, Blocks (\n), Lists (\n), **BaseWritingDirection 1**, HyphenationFactor 0, TighteningForTruncation NO, HeaderLevel 0 LineBreakStrategy 0";
}
**after save:**
}في{
    NSColor = "<UIDynamicProviderColor: 0x60000275dd00; provider = <__NSMallocBlock__: 0x600002942c40>>";
    NSFont = "<UICTFont: 0x7fa391659a50> font-family: \".ArabicUIText-Regular\"; font-weight: normal; font-style: normal; font-size: 17.00pt";
    NSParagraphStyle = "Alignment 4, LineSpacing 4.55711, ParagraphSpacing 0, ParagraphSpacingBefore 0, HeadIndent 0, TailIndent 0, FirstLineHeadIndent 0, LineHeight 0/0, LineHeightMultiple 0, LineBreakMode 0, Tabs (\n    28L,\n    56L,\n    84L,\n    112L,\n    140L,\n    168L,\n    196L,\n    224L,\n    252L,\n    280L,\n    308L,\n    336L\n), DefaultTabInterval 0, Blocks (\n), Lists (\n), **BaseWritingDirection -1**, HyphenationFactor 0, TighteningForTruncation NO, HeaderLevel 0 LineBreakStrategy 0";
}

So my first question is, are we saving notes in such a way that we can persist the BaseWritingDirection attribute?

And my second question is would it be possible to set refresh this property for each word instead of each line?

charliescheer:
I am going to close this PR for now.  There is more research into what we should/could support before this one can be continued.

Issue #1086, opened 2021-01-13
Email Verification


jleandroperez:
### Details:
- [x] Simperium's JSONStorage support #1091
- [x] Simperium Support to sync current Verification State #1091
- [x] Create “Review You Account” screen #1089
- [x] Create “Verify Your Email” screen #1089
- [x] Show “Review You Account” on app launch if user is not verified and verification was not requested
- [x] Show “Verify Your Email” if verification was requested
- [x] Wire up the call to verification endpoint from “Review You Account” and “Verify Your Email” screens
- [x] Open `app.simplenote.com/settings` in a browser if user clicks on “Change Email”
- [x] Analytics

### Mockups:
<img width="300" src="https://user-images.githubusercontent.com/1195260/104479597-60153e80-55a2-11eb-8446-5aedc7cbdf8e.png"> <img width="300" src="https://user-images.githubusercontent.com/1195260/104479603-62779880-55a2-11eb-905b-4da5d83f9c7e.png">

pachlava:
Verified all mentioned PRs that are UI-testable. iPhone XR | iOS 14.3 | SN 4.31

PR #1101, opened 2021-01-20
Unit tests for account verification

eshurakov:
### Details
In this PR we're adding unit tests for `AccountVerificationController`. HTTP call from `AccountVerificationController` is extracted to a separate class `AccountVerificationRemote`.

Ref. #1086

### Test
- [ ] Unit tests are green

### Review

### DIFF
--- a/Simplenote/Verification/AccountVerificationRemote.swift
+++ b/Simplenote/Verification/AccountVerificationRemote.swift
@@ -1,0 +1,55 @@
+import Foundation
+
+// MARK: - AccountVerificationRemote
+//
+class AccountVerificationRemote {
+    private let urlSession: URLSession
+
+    init(urlSession: URLSession = URLSession.shared) {
+        self.urlSession = urlSession
+    }
+
+    /// Send verification request for specified email address
+    ///
+    func verify(email: String, completion: @escaping (_ success: Bool) -> Void) {
+        guard let request = verificationURLRequest(with: email) else {
+            completion(false)
+            return
+        }
+
+        let dataTask = urlSession.dataTask(with: request) { (data, response, error) in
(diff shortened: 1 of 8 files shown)


jleandroperez (approved):
Thank you for doing this Evgeny!!

cc @charliescheer Please take a peek at this PR, it's a Dependency Injection crash course 😄 

:shipit:

Simplenote/Verification/AccountVerificationRemote.swift
```diff
@@ -0,0 +1,55 @@
...
+            completion(false)
+            return
+        }
+
+        let dataTask = urlSession.dataTask(with: request) { (data, response, error) in
+            DispatchQueue.main.async {
+                // Check for 2xx status code
+                guard let response = response as? HTTPURLResponse, response.statusCode / 100 == 2 else {
+                    completion(false)
+                    return
```
dmsnell, line 25: just a note that I've designed the API to _always_ send a 200 regardless of what happens. that means the only time we shouldn't get a 200 is when an issue on the server sends a 5xx.

in those cases it might be hard to figure out what's wrong or do anything with the knowledge.

could be easier to "fire and forget"

did you have a response based on the status code you wanted to do?

eshurakov: Oh I see. I believe when I first checked the code some time ago I saw that that endpoint could return 4xx. 

When status code is not 2xx or there is no status code at all in case of network errors we show a generic error alert.

Screenshot 2021-01-21 at 09 16 17

eshurakov:
Thank you @jleandroperez! 
Will merge this now and we can tackle possible issues later in a separate PR.

Issue #1061, opened 2020-12-18
Update Designs for Information Sheet


SylvesterWilmott:
This is an issue to request updated designs for iOS information sheet. See below for illustration.

<img width="377" alt="Screenshot 2020-12-18 at 09 54 37" src="https://user-images.githubusercontent.com/36532468/102600716-178b7200-4117-11eb-8183-d95b0c211c7a.png">

### Changes
- Title bar removed on iPhone
- Title replaced with "Document" on iPad
- Italic text removed from references titles
- Styling changed for "Referenced In" heading
- Separator removed for last child in document and referenced in lists
- Reference count added.

charliescheer:
@SylvesterWilmott I have been putting in some work on these changes and am almost finished with them.

Question though, for the updating of the styling on Referenced In.  I see that we removed the uppercasing for that text, but are there any other changes you were looking for.  It seems like the text may have gotten a bit larger, but I wasn't sure.

Also for removing the title bar, I see you added a dismiss swipe bar at the top, is this an image that we are already using in Simplenote.  I dug around a bit to see if this might be built into iOS already and didn't see a way to activate it, but I may have missed something.

Thanks, will let you know if I have any more questions

PR #1105, opened 2021-01-24
Information Sheet Style Updates

charliescheer:
### Fix
This PR fixes most of the style updates in issue #1061.  

The style updates that are covered in this PR:

1. Removed title bar from iPhone view
2. Replaced title with Document
3. Removed italics from references titles
4. Updated stying for Referenced In
5. Removed separator for last child in document and reference lists
6. Added reference count

What is not covered in this PR:
-the addition of a drag bar at the top of the modal when viewed on an iPhone.  Will create a separate PR/branch to add that feature.

### Test
To test, please make sure the following elements look correct per the changes in #1061 

- [ ] Title bar removed for iphone
- [ ] Title on iPad changed to Document
- [ ] Reference titles no longer in italics
- [ ] Styling for "Referenced In" is correct
- [ ] Separator below table view cells does not appear after last document setting or reference
- [ ] reference count appears in link references


Simplenote/Information/NoteInformationController.swift
```diff
@@ -142,13 +142,18 @@ private extension NoteInformationController {
         }
 
         let referenceRows = references.map { (note) -> Row in
+            let date = DateFormatter.dateFormatter.string(from: note.modificationDate)
+            let value = String(references.count) +
+                " \(references.count > 1 ? "References" : "Reference")" +
+                ", Last Modified \(date)"
```
eshurakov: This string needs to be localized, please check the following example: https://github.com/Automattic/simplenote-ios/blob/733fede9b6d5ebfc833384bbf775ff746e933be4/Simplenote/Classes/PinLock/PinLockBaseController.swift  

Using `> 1` condition is also not ideal because it doesn't work in all languages. That's something we need to discuss and improve later as a team. For now let's continue using `> 1` :-)

Simplenote/Information/NoteInformationViewController.swift
```diff
@@ -154,6 +155,15 @@ private extension NoteInformationViewController {
     func configureAccessibility() {
         dismissButton.accessibilityLabel = Localization.dismissAccessibilityLabel
     }
+    
+    func configureScreenTitleLabel() {
+        screenTitleLabel.text = Localization.document
+        
+        if UIDevice.current.userInterfaceIdiom == .phone {
```
eshurakov: Nitpick: we have an extension for UIDevice to check if device is iPad `UIDevice.isPad`, maybe we can use it here.

eshurakov: It looks like with these changes we don't use screenTitleLabel and dismissButton at all, so I think we can safely remove the whole `headerStackView`. It wasn't used on iPad and now we're not using it on iPhone as well.

Simplenote/Information/NoteInformationViewController.swift
```diff
@@ -319,6 +336,7 @@ private struct Localization {
     static let information = NSLocalizedString("Information", comment: "Card title showing information about the note (metrics, references)")
```
eshurakov: I believe this string is no longer used and can be removed.

eshurakov (changes requested):
Looks good! Sending you a few comments.

Simplenote/Information/NoteInformationViewController.swift
charliescheer: Makes sense, updated

Simplenote/Information/NoteInformationViewController.swift
charliescheer: I wasn't sure if we wanted to remove that or not.  Removed it

Simplenote/Information/NoteInformationController.swift
charliescheer: That is a good way to do it!  Thanks for the suggestion.  Updated

SylvesterWilmott:
Design-wise looks good so far.

There's one minor spacing issue. The distance between the top and bottom tables in the sheet appear to be too large. This space should be 10pt.

Screenshot 2021-01-26 at 16 45 00

Simplenote/Information/NoteInformationViewController.swift
eshurakov: Sorry I wasn't very clear. I meant that we can completely delete `headerStackView` from the code and interface file because I believe it is not used anymore.

Simplenote/Information/NoteInformationController.swift
```diff
@@ -142,13 +142,18 @@ private extension NoteInformationController {
         }
 
         let referenceRows = references.map { (note) -> Row in
+            let date = DateFormatter.dateFormatter.string(from: note.modificationDate)
+            let value = "\(Localization.interlinkReferences(references.count)), "
+                + "\(NSLocalizedString("Last Modified", comment: "Date of note last modified")) "
+                + date
```
eshurakov: Lets extract `NSLocalizedString("Last Modified", comment: "Date of note last modified")` to `Localization` struct, so that all localization is in one place. Lets also add `date` as a parameter so that if needed the order of words can be changed during localization.

It can look like this: `NSLocalizedString("Last Modified %1$@", comment: "Date of note last modified. Parameter: %1$@ - formatted date")`

charliescheer:
Alright @eshurakov  and @SylvesterWilmott 
Those changes are updated.  Let me know if you see anything else!

eshurakov (approved):
Looks good to me. Thank you @charliescheer.

SylvesterWilmott:
I think there are some issues with the "Referenced In" title area. It looks as though the title area for "Referenced In" is about 40pt. This area should ideally be 36pt with a 10pt space between the information list and the references list. Illustrated below:

Screenshot 2021-01-28 at 09 27 30

charliescheer:
Okay... will need some feedback on this one.  

@SylvesterWilmott  can you take a look and see if the sizing looks right to you now?  I was able to add the additional 10pt space above the referenced in section of the information sheet.

@eshurakov Can you also take a look at it?  I found adding spacing to a cell but only on the top (or bottom) of the cell was quite difficult.  I couldn't find a reliable way to using constraints or heights to add a space above referenced in, while also keeping it centered to the actual cell it was in.  

So what I did was added a case in Spacer.Row called .spacer, inserted one at the beginning of the referenced in section, and then set that to be 10pt.  There may be a much easier way to do this that I didn't find.  What do you think?

eshurakov:
Hey @charliescheer! 
The app crashes when you open Info sheet that has references.
`unable to dequeue a cell with identifier UITableViewCell - must register a nib or a class for the identifier or connect a prototype cell in a storyboard`

I think the approach with having another cell to create a space is valid, but what you can also do is to increase top margin of the "Referenced In" cell. 

Please take a look here: https://github.com/Automattic/simplenote-ios/pull/1105/files#diff-f546a56eb2c42f5fa7c11c8bdf1cc44375b2d8fb0c6987d7d497001f6dd73a53L59
Top inset was set to 24 before, if you set it to 10 it will be exactly the same as using a spacer with the height of 10. If you want it to be more configurable, you can keep a reference to the top constraint in `TableHeaderViewCell` and update it to appropriate value when configuring a cell. 
Right now we use this cell only in Info sheet, so I think hardcoding top inset to be 10 is fine.

charliescheer:
Thanks for taking a look @eshurakov  I could have sworn that the first thing I did was use the edge insets to create that spacing, but I guess not...

Anyways, I removed the spacers I had created (which were now causing a crash for me as well, but not yesterday :shrug:) and set up the edge insets to the correct 10pts.

Should be all set now.  @SylvesterWilmott does that sizing look right to you?

SylvesterWilmott:
Thanks @charliescheer.

One final thing, can we change the section title area to be 36pt tall? Right now it looks like 40pt.

Screenshot 2021-02-02 at 12 03 08

pachlava:
iPhone XR | iOS 14.3 | Simplenote 4.32.0.0
and
iPad Pro Sim | iOS 14.4 | Simplenote 4.32.0.0

PR #1106, opened 2021-01-24
Add dismiss handle to Information Sheet

charliescheer:
### Fix
This PR continues with the design changes in issue #1061. Specifically this PR adds in a visual element to alert the user that they can dismiss the information sheet by dragging down on the top of the information sheet element.

The drag bar element will not appear on iPad and should look a bit like this:

### Test

1. Open up Simplenote on an iPhone and go to a Note > Information sheet, confirm that the drag bar appears
2. Open up Simplenote again on an iPad and go to a note > Information sheet, confirm there is no drag bar

### Review
One developer and one designer are required to review these changes, but anyone can perform the review.

### DIFF
--- a/Simplenote/Information/NoteInformationViewController.swift
+++ b/Simplenote/Information/NoteInformationViewController.swift
@@ -5,7 +5,7 @@ import SimplenoteFoundation
 //
 final class NoteInformationViewController: UIViewController {
     @IBOutlet private weak var tableView: UITableView!
-    @IBOutlet weak var dragBar: SPDragBar!
+    @IBOutlet private weak var dragBar: SPDragBar!
 
     private var transitioningManager: UIViewControllerTransitioningDelegate?
 
@@ -122,7 +122,7 @@ private extension NoteInformationViewController {
         preferredContentSize = tableView.contentSize
     }
     
-    func configureDragBar() {
+    private func configureDragBar() {
         dragBar.isHidden = navigationController != nil
     }
 }
--- a/Simplenote/Information/NoteInformationViewController.swift
+++ b/Simplenote/Information/NoteInformationViewController.swift
@@ -6,7 +6,6 @@ import SimplenoteFoundation
 final class NoteInformationViewController: UIViewController {
     @IBOutlet private weak var tableView: UITableView!
     @IBOutlet private weak var screenTitleLabel: UILabel!
-    @IBOutlet private weak var dismissButton: UIButton!
     @IBOutlet private weak var headerStackView: UIStackView!
     @IBOutlet weak var dragBar: SPDragBar!
     private lazy var blurEffectView = SPBlurEffectView()
@@ -155,7 +154,7 @@ private extension NoteInformationViewController {
     }
 
     func configureAccessibility() {
-        dismissButton.accessibilityLabel = Localization.dismissAccessibilityLabel
+        //TO DO: Should we do something for accessibility for dismissing the info sheet now that the dismiss button is gone?
     }
     
     func configureScreenTitleLabel() {
@@ -163,7 +162,6 @@ private extension NoteInformationViewController {
         
         if UIDevice.current.userInterfaceIdiom == .phone {
             screenTitleLabel.isHidden = true
-            dismissButton.isHidden = true
         }
     }
 
--- a/Simplenote/Information/NoteInformationViewController.swift
+++ b/Simplenote/Information/NoteInformationViewController.swift
@@ -5,7 +5,6 @@ import SimplenoteFoundation
 //
 final class NoteInformationViewController: UIViewController {
     @IBOutlet private weak var tableView: UITableView!
-    @IBOutlet private weak var headerStackView: UIStackView!
     @IBOutlet weak var dragBar: SPDragBar!
 
     private var transitioningManager: UIViewControllerTransitioningDelegate?
@@ -58,13 +57,12 @@ final class NoteInformationViewController: UIViewController {
 
     override func viewWillLayoutSubviews() {
         super.viewWillLayoutSubviews()
-        configureHeaderLayoutMargins()
         refreshPreferredSize()
     }
 
     override func viewDidLayoutSubviews() {
         super.viewDidLayoutSubviews()
-        updateAdditionalSafeAreaInsets()
+        additionalSafeAreaInsets = Consts.tableViewSafeAreaInsets
     }
 }
 
@@ -116,32 +114,12 @@ private extension NoteInformationViewController {
         }
     }
 
-    func configureHeaderLayoutMargins() {
-        headerStackView.isLayoutMarginsRelativeArrangement = true
-
-        var layoutMargins = Consts.headerExtraLayoutMargins
-        layoutMargins.left += tableView.layoutMargins.left
-        layoutMargins.right += tableView.layoutMargins.right
-
-        // Sync layout margins with table view so labels are aligned
-        headerStackView.layoutMargins = layoutMargins
-    }
-
     func configureAccessibility() {
         //TO DO: Should we do something for accessibility for dismissing the info sheet now that the dismiss button is gone?
     }
 
     func refreshPreferredSize() {
         preferredContentSize = tableView.contentSize
-    }
-
-    func updateAdditionalSafeAreaInsets() {
-        guard !headerStackView.isHidden else {
-            additionalSafeAreaInsets = .zero
-            return
-        }
-
-        additionalSafeAreaInsets = UIEdgeInsets(top: headerStackView.frame.maxY, left: 0, bottom: 0, right: 0)
     }
     
     func configureDragBar() {
@@ -301,5 +279,5 @@ private struct Localization {
 }
 
 private struct Consts {
-    static let headerExtraLayoutMargins = UIEdgeInsets(top: 13.0, left: 0.0, bottom: 13.0, right: 0.0)
+    static let tableViewSafeAreaInsets = UIEdgeInsets(top: 26, left: 0, bottom: 0, right: 0)
 }
--- a/Simplenote/Information/NoteInformationViewController.swift
+++ b/Simplenote/Information/NoteInformationViewController.swift
@@ -182,13 +182,6 @@ private extension NoteInformationViewController {
     
     func configureDragBar() {
         if navigationController == nil {
-            dragBar.translatesAutoresizingMaskIntoConstraints = false
-            NSLayoutConstraint.activate([
-                dragBar.heightAnchor.constraint(equalToConstant: 5),
-                dragBar.widthAnchor.constraint(equalToConstant: 36),
-                dragBar.topAnchor.constraint(equalTo: view.topAnchor, constant: 5),
-                dragBar.centerXAnchor.constraint(equalTo: view.centerXAnchor)
-            ])
         } else {
             dragBar.isHidden = true
         }
--- a/Simplenote/Information/NoteInformationViewController.swift
+++ b/Simplenote/Information/NoteInformationViewController.swift
@@ -119,10 +119,6 @@ private extension NoteInformationViewController {
         }
     }
 
-    func configureAccessibility() {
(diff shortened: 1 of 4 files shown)


charliescheer:
I pulled this PR out of the other one I made for the information sheet style updates because I wanted to ask if this is something that we want to extract into it's own class.  It is rather simple, just a small UIView with a background color, but if it turned out to be something that we wanted to insert onto other screens, I could see how it could be setup to simply be added to any modal and save on repeat code later. 

Thoughts?

mokagio:
Pushing this PR to the next milestone, 4.31, because I'm running the code-freeze today and this is labelled "enhancements, spring cleaning": it can safely wait till the next release.

eshurakov:
Hi @charliescheer, I believe you created this branch from `issue/1061-information-sheet-updates`. I updated base branch in this PR to `issue/1061-information-sheet-updates` so that we only see changes that you made on top of that branch.

Simplenote/Information/NoteInformationViewController.swift
```diff
@@ -177,6 +181,24 @@ private extension NoteInformationViewController {
...
+        dragBar.alpha = 0.2
+        dragBar.layer.cornerRadius = 2.5
+        dragBar.layer.masksToBounds = true
+        
+        view.addSubview(dragBar)
+        
+        dragBar.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            dragBar.topAnchor.constraint(equalTo: view.topAnchor, constant: 5),
+            dragBar.bottomAnchor.constraint(equalTo: view.topAnchor, constant: 10),
```
eshurakov: I think it's better to use `height`. Easier to understand dimensions :-)

eshurakov (changes requested):
Hey Charlie,

I think it's good to extract it in a separate class, not necessarily because we will reuse it later, but to keep view controller more tidy. 

Another thing is, we're using xib file to construct information screen, so let's move this handle to the xib and hide it in case we're presenting it in a popover. I just realised that checking for `UIDevice.isPad` won't always work, because in split view on iPad information screen is presented as a card if width class is compact or in a popover if width class is regular. What we can do is to check if there is a navigation controller attached to this view controller. When we present an information controller in a popover, we wrap it in a navigation controller, so if there is a navigation controller, we don't need to show a handle.

Simplenote/Information/NoteInformationViewController.swift
charliescheer: I have made those changes.  Check it out!  Let me know if you see anything else that needs to be updated!

Simplenote/Information/NoteInformationViewController.swift
```diff
@@ -177,6 +179,20 @@ private extension NoteInformationViewController {
...
+            dragBar.translatesAutoresizingMaskIntoConstraints = false
+            NSLayoutConstraint.activate([
+                dragBar.heightAnchor.constraint(equalToConstant: 5),
+                dragBar.widthAnchor.constraint(equalToConstant: 36),
+                dragBar.topAnchor.constraint(equalTo: view.topAnchor, constant: 5),
+                dragBar.centerXAnchor.constraint(equalTo: view.centerXAnchor)
+            ])
+        } else {
+            dragBar.isHidden = true
+        }
```
jleandroperez: Since the initialization is already done in IB (`NoteInformationViewController`), would it be possible to move the Autolayout constraints there?

---

On a second note, we'd typically wanna avoid inline constants. Instead, you would probably wanna do something like this:

```
   dragBar.heightAnchor.constraint(equalToConstant: Metrics.dragBarSize.height),
   dragBar.widthAnchor.constraint(equalToConstant: Metrics.dragBarSize.width),
   dragBar.topAnchor.constraint(equalTo: view.topAnchor, constant: Metrics.dragBarTopPadding),
   dragBar.centerXAnchor.constraint(equalTo: view.centerXAnchor)
```

And at the bottom of the source file you could have:

```
private enum Metrics {
    static let dragBarSize = CGSize(width: 36, height: 5)
    static let dragBarTopPadding = CGFloat(5)
}
```

### Why:
Numbers used inline in the code are best avoided, for few reasons. It's harder to maintain, it's impossible to reuse. Suppose there were two components onscreen that needed the same **Top Padding**, without a constant, you'd have a `5` in two places.

**Plus:** if you had to change that `5`, without a proper constant you'd need to read the whole code, and patch two spots.

### And another note:
Why `enum Metrics` instead of, say, `struct Metrics`?  Because if you used `struct Metrics` this statement would be valid:

```
let metrics = Metrics()
```

It's a "workaround" we've been using everywhere, and allows us to namespace a bunch of constants. Other programming languages (such as C#) have a way to define a namespace that doesn't really belong to anything that can be instantiated. Swift... really doesn't, for now!

SylvesterWilmott:
Looks good!

Spacing and position is perfect. It appear in dark mode the handle remains dark, can we update the colour of the handle in dark mode to #ffffff with 0.2% opacity?

charliescheer:
Thanks for the suggestions @jleandroperez  and @SylvesterWilmott 

I have added dark mode for the drag bar
and moved the constraints to the IB.

Let me know if there are any more changes that need to be made

Simplenote/Information/NoteInformationViewController.swift
```diff
@@ -177,6 +179,12 @@ private extension NoteInformationViewController {
 
         additionalSafeAreaInsets = UIEdgeInsets(top: headerStackView.frame.maxY, left: 0, bottom: 0, right: 0)
     }
+    
+    func configureDragBar() {
+        if navigationController != nil {
+            dragBar.isHidden = true
```
jleandroperez: Nipticky (cosmetic, feel free to disregard!!). You can simplify this one a bit:

```
dragBar.isHidden = navigationController != nil
```

Simplenote/SPDragBar.swift
```diff
@@ -0,0 +1,24 @@
...
+    
+    func configure() {
+        if SPUserInterface.isDark {
+            self.backgroundColor = UIColor.white
+        } else {
+            self.backgroundColor = UIColor.black
+        }
+        self.alpha = 0.2
+        self.layer.cornerRadius = 2.5
+        self.layer.masksToBounds = true
```
jleandroperez: No need for `self` (unless you're inside a closure)

Simplenote/SPDragBar.swift
```diff
@@ -0,0 +1,24 @@
...
+        super.init(frame: frame)
+        configure()
+    }
+    
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        configure()
+    }
+    
+    func configure() {
```
jleandroperez: This one should probably be `private`

Simplenote/SPDragBar.swift
```diff
@@ -0,0 +1,24 @@
...
+        if SPUserInterface.isDark {
+            self.backgroundColor = UIColor.white
+        } else {
+            self.backgroundColor = UIColor.black
+        }
```
jleandroperez: This snippet can be simplified a bit:

```
backgroundColor = SPUserInterface.isDark ? .white : .black
```

Simplenote/Information/NoteInformationViewController.swift
charliescheer: I like it, much cleaner

Simplenote/SPDragBar.swift
charliescheer: Much simpler!

charliescheer:
Thanks @jleandroperez  I made those changes

Simplenote/Information/NoteInformationViewController.swift
```diff
@@ -177,6 +179,10 @@ private extension NoteInformationViewController {
 
         additionalSafeAreaInsets = UIEdgeInsets(top: headerStackView.frame.maxY, left: 0, bottom: 0, right: 0)
     }
+    
+    func configureDragBar() {
```
eshurakov: This can also be private :-)

Simplenote/Information/NoteInformationViewController.swift
```diff
@@ -8,6 +8,7 @@ final class NoteInformationViewController: UIViewController {
     @IBOutlet private weak var screenTitleLabel: UILabel!
     @IBOutlet private weak var dismissButton: UIButton!
     @IBOutlet private weak var headerStackView: UIStackView!
+    @IBOutlet weak var dragBar: SPDragBar!
```
eshurakov: This can be private like other `@IBOutlet` properties

Simplenote/SPDragBar.swift
```diff
@@ -0,0 +1,20 @@
...
+        configure()
+    }
+    
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        configure()
+    }
+    
+    private func configure() {
+        backgroundColor = SPUserInterface.isDark ? .white : .black
```
eshurakov: Please take a look at `UIColor+Studio.swift` extension. You can define a pair of light / dark colors and use it here as a single UIColor. The benefit is that it will automatically update when theme changes. 
For example right now if you open an information sheet and change a theme (Simulator -> Features -> Toggle Appearance), the color of the handle won't update.

eshurakov (commented):
Hey Charlie, now that we show a handle instead of title and dismiss button, I believe we can safely delete the following things from the code and interface file:
* screenTitleLabel
* dismissButton
* headerStackView
* blurEffectView

SylvesterWilmott (approved):
Design looks good. Thanks @charliescheer

charliescheer:
> now that we show a handle instead of title and dismiss button, I believe we can safely delete the following things from the code and interface file

Good call @eshurakov I have removed those extra things, and I don't see any issues with it having been removed.

I have also updated the dragbar to use the color studio colors and marked the methods as private.

Question for you though.  By removing the dismiss button we are also removing the accessibility label for dismissing.  Is there anything we can, or should do for accessibility labeling on the drag bar? maybe:
`    func configureAccessibility() {
        dragBar.accessibilityLabel = Localization.dismissAccessibilityLabel
    }`

@SylvesterWilmott can you take a quick look at the info sheet one more time and see if the spacing between the top of the info sheet and the first text item looks right to you?  It should have the correct spacing, but wanted to double check

eshurakov:
If we add a label to the drag bar, we also should handle tap on it so that it can be activated though accessibility. 
I think we can start by implementing [accessibilityPerformEscape](https://developer.apple.com/documentation/objectivec/nsobject/1615091-accessibilityperformescape) method to dismiss Info sheet. It is activated when you draw a “Z” with two fingers anywhere on the screen and should act like `ESC` key.

charliescheer:
>If we add a label to the drag bar, we also should handle tap on it so that it can be activated though accessibility.
I think we can start by implementing accessibilityPerformEscape method to dismiss Info sheet. It is activated when you draw a “Z” with two fingers anywhere on the screen and should act like ESC key.

Great idea, I have added both of those things to the drag bar.  Both the Z to dismiss and the tap to dismiss should work with VoiceOver on

Simplenote/Information/NoteInformationViewController.swift
```diff
@@ -121,76 +119,33 @@ private extension NoteInformationViewController {
...
-    }
-    
-    func configureScreenTitleLabel() {
-        screenTitleLabel.text = Localization.document
-        
-        if UIDevice.current.userInterfaceIdiom == .phone {
-            screenTitleLabel.isHidden = true
-            dismissButton.isHidden = true
-        }
+        //TO DO: Should we do something for accessibility for dismissing the info sheet now that the dismiss button is gone?
```
eshurakov: I think can remove this one :-)

eshurakov (approved):
Looks good to me! 👍 

:shipit:

Simplenote/Information/NoteInformationViewController.swift
charliescheer: Ah, missed that note.  Thanks!  Donoe

pachlava:
[testing] fix/feature verified on iPhone XR | iOS 14.3 | Simplenote 4.32.0.0

Issue #1036, opened 2020-11-19
Sharing: Text hidden behind the keyboard when large text is selected.


rezzap:
When using the option to share from one app to Simplenote and a large portion of text is added, the text becomes trapped behind the keyboard and there is no way to edit the section that overflows. 

### Expected

I'd expect to be able to scroll the full way down to edit the full text.

### Observed

The text is hidden by the keyboard and can't be editable. 

### Reproduced

> 1. Go to a different app, like Instapaper or News.
> 2. Tap on the share icon and select the option to Share to Simplenote.
> 3. Make sure the text box has large amounts of text, or paste a large portion of text in the box. 
> 4. You will see the error if you try to scroll to the bottom of the screen. 

![Screen Recording 2020-11-19 at 01 30 30 pm](https://user-images.githubusercontent.com/10325627/99672736-bc108a80-2a6b-11eb-86f7-0a12f87ce8a1.gif)

***(Optional)*** If you reproduced the bug on a specific device, please replace DEVICE_MAKE, DEVICE_MODEL, IOS_VERSION, and APP_VERSION below with the details for that device.  If you reproduced the bug on multiple devices, please add a row for each device.

Make|Model|iOS Version|App Version
-|-|-|-
iPhone|X|iOS 14.2|4.27

User reported in 3499081-zen

PR #1109, opened 2021-01-25
Fix issue where text could be hidden behind keyboard when sharing to SN

charliescheer:
### Fix
This PR fixes issue number #1036 

The issue described in #1036  is that when sharing to Simplenote from another application (like safari) if the text was too long, the text could hide behind the keyboard.  With this change the edge insets for the text view adjust with the keyboard appearing.

### Test
To test:
1. After loading Simplenote onto device or simulator, go to another app that has sharing like safari
2. Click on the share button and choose to share to Simplenote
3. paste in more text than can fit on the screen
4. Scroll down and make sure you can reach the bottom of the pasted in text

### DIFF
--- a/SimplenoteShare/Presentation/ExtensionPresentationController.swift
+++ b/SimplenoteShare/Presentation/ExtensionPresentationController.swift
@@ -71,7 +71,7 @@ final class ExtensionPresentationController: UIPresentationController {
     }
 
     override func containerViewWillLayoutSubviews() {
-        presentedView?.frame = frameOfPresentedViewInContainerView
+//        presentedView?.frame = frameOfPresentedViewInContainerView
         presentedView?.layer.cornerRadius = Appearance.cornerRadius
         presentedView?.clipsToBounds = true
     }
--- a/SimplenoteShare/ViewControllers/ShareViewController.swift
+++ b/SimplenoteShare/ViewControllers/ShareViewController.swift
@@ -75,7 +75,7 @@ class ShareViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         setupNavigationBar()
-        registerKeyboardNotifications()
+//        registerKeyboardNotifications()
         textView.textContainerInset = Constants.textViewInsets
         loadContent()
     }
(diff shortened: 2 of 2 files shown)


jleandroperez (changes requested):
Two notes on this one! The Share Sheet appears to have different behaviors:

- iOS 14 / device: its taking the full available height
- iOS 13 / simulator: Acquires only the available height above the keyboard

In iOS 13 this has a side effect (you end up with extra unneded insets, see video below).

In order to attack this one you may wanna:

A. Switch to `UIKeyboardWillChangeFrameNotification`: because it'll get fired anytime the keyboard's location changes. This happens on iPad when you use the floating keyboard, or Show / Dismiss the keyboard.

B. Adapt the notification handler so that it only applies an Inset for the intersected area.

[Sample code here!](https://github.com/Automattic/simplenote-ios/blob/develop/Simplenote/Classes/SPNoteListViewController+Extensions.swift#L824)

Thank yoou!

https://user-images.githubusercontent.com/1195260/105849494-3ce58880-5fbf-11eb-887b-d78916b9c189.mov

charliescheer:
Hey @jleandroperez  Questions for you.

>A. Switch to UIKeyboardWillChangeFrameNotification: because it'll get fired anytime the keyboard's location changes. This happens on iPad when you use the floating keyboard, or Show / Dismiss the keyboard.

I am running up against an issue using UIKeyboardWillChangeFrameNotification.  Well two issues, when I started out using keyboard WILL change frame, the value of the size of the keyboard was coming in as where it was, not where it will be.  so the math to change the insets always came out wrong.  If I used keyboard DID change frame then I could get the correct value.  Am I missing something about how to get the correct value from the will change notification?

Second problem, when I am using Did change frame, I am coming up against an issue where after I start interacting with the text view, the notification fires again, but this time the keyboard registers as a different size so it breaks the insets.  It doesn't happen every time, but often enough to be an issue.  I updated the branch to show my latest changes, and here is a quick [video of what is happening](https://d.pr/i/p9hLuT). You can see in the console that initially the calculated value is around 250, which is what we want, but when I click in the text field (sometimes but not all the time) the keyboard frame notification fires again and now I get -8 or sometimes -330




Probably a math error, but I am not sure what is causing it.  Any suggestions?

jleandroperez:
Hi there @charliescheer!

May I ask you what's the state of this one?

charliescheer:
Hey @jleandroperez  I put some more work into this today, and something strange is going on.  With the way that this is supposed to be setup/working, the class ExtensionPresentationController should be resizing the ShareViewController textview so that it is just about 8 pts off the keyboard when it appears, and for iPhone it does this correctly, however on iPad (simulator) it does not.

What happens on the iPad is that the sizing calculations end up being about 24 pixels off of where they should be, but what is strange is the values should be correct.  For example on iPhone 12 the screen height is 844, and when the keyboardWillChangeFrame method is called the keyboard height is 336, we are using a padding of 8, so the bottom of the text view should land at 500, and when calculated that is where it lands
but using iPad Air 4, the screen size is 1180, keyboard size is 313 with a padding of 8, so the bottom of the text view should be 836.   When calculated or hard coded that is still too long, and I am not sure why the math works for iPhones (I tried a few others and they seemed to work) but not iPads.

I figure we have two options.  We can continue to look into the why the sizing is not working correctly on iPad (if we go with this one, I would like your help hunting down the issue if possible) OR we could use insets instead of resizing the view.

I personally prefer the second option.  I think that it looks better with the view going longer passed the keyboard and then the edge insets stopping the text from going passed the keyboard instead of there being a little gap and the view being much smaller.  What do you think?

charliescheer:
closing in favor of #1139

PR #1110, opened 2021-01-25
Bugfix: Configure popover when presenting tag delete sheet on iPad

eshurakov:
### Fix
This PR fixes a crash which happens while trying to delete a tag from a sidebar on iPad. The app crashes because of missing popover configuration for action sheet. 

### Test
1. On iPad: open sidebar, make sure to have some tags
2. Enter editing mode
3. Tap on Trash icon next to one of the tags

- [x] Check that popover appears next to the row of the tag that you're trying to delete
- [ ] App doesn't crash

### Review


jleandroperez (approved):
Thanks for fixing this one!!

:shipit:

eshurakov:
Thank you!

Issue #1092, opened 2021-01-15
Project: Keyboard Support


eshurakov:
### What
Support keyboard navigation and shortcuts.

- [x] Save scroll position
- [x] While in editor, `Cmd + Enter` brings you back to the notes list
- [x] Add shortcuts
- [x] Support RTL
- [x] Add tracks

Action | Shortcut
-|-
Focus search field | Cmd + Shift + F
Jump to next match in note | Cmd + G
Jump to previous match in note | Cmd + Shift + G
Toggle note list | Cmd + Shift + L
Create new note |  Cmd + N
Toggle markdown preview | Cmd + Shift + P
Insert checklist | Cmd + Shift + C

Similar issues: #291, #8
Reference: p2XJRt-2Ao

PR #1111, opened 2021-01-25
Keyboard support: MK1

eshurakov:
### Details
In this PR we're adding keyboard navigation in Note List and Sidebar.
What was added:
* Navigation in the list using up / down arrow
* Navigation between list and sidebar using left / right arrow
* Navigation back through the view controllers stack using cmd + return. Helpful when editor is focused and we can't use left arrow for that
* Search using `cmd + shift + f` (when search bar is focused ESC ends the search)
* Create new note using `cmd + n`

Coming next:
* Shortcuts in the editor
* Improve creating new note from the editor (now it goes back to list)
* Preserving selection in the note list
* Save scroll position
* RTL
* Tracks

Ref: #1092

### Test (iPhone / iPad with external keyboard)

1. Open the app

- [x] It is possible to navigate between sidebar and list using arrows
- [x] It is possible to navigate inside the list using arrows
- [x] `cmd + shift + f` opens search
- [x] `cmd + n` creates new note
- [ ] Hold `cmd` to see available shortcuts

### Review

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -543,6 +543,8 @@
     self.searchBar.userInteractionEnabled = YES;
 
     self.navigationController.navigationBar.userInteractionEnabled = YES;
+
+    [self becomeFirstResponder];
 }
 
 
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -538,6 +538,9 @@
 
     self.navigationController.navigationBar.userInteractionEnabled = YES;
 
+    // We want this VC to be first responder to support keyboard shortcuts.
+    // We don't want to steal first responder from the search bar in case it's already active.
+    // It Can happen if we open the app directly into search mode from the home screen quick action
     if (!self.isSearchActive) {
         [self becomeFirstResponder];
     }
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -538,7 +538,9 @@
 
     self.navigationController.navigationBar.userInteractionEnabled = YES;
 
-    [self becomeFirstResponder];
+    if (!self.isSearchActive) {
+        [self becomeFirstResponder];
+    }
 }
 
 
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -302,13 +302,7 @@
 #pragma mark - BarButtonActions
 
 - (void)addButtonAction:(id)sender {
-    
-    [SPTracker trackListNoteCreated];
-    
-    // the editor view will create a note. Passing no note ensures that an emty note isn't added
-    // to the FRC before the animation occurs
-    [self.tableView setEditing:NO];   
-    [self openNote:nil animated:YES];
+    [self createNewNote];
 }
 
 - (void)sidebarButtonAction:(id)sender {
--- a/Simplenote/Classes/UITableView+Simplenote.swift
+++ b/Simplenote/Classes/UITableView+Simplenote.swift
@@ -46,3 +46,120 @@ extension UITableView {
         return dequeueReusableHeaderFooterView(withIdentifier: T.reuseIdentifier) as? T
     }
 }
+
+// MARK: - Navigation
+//
+extension UITableView {
+
+    /// Returns first index path from the first non-empty section
+    ///
+    var firstIndexPath: IndexPath? {
+        let indexPath = IndexPath(row: 0, section: 0)
+        if numberOfRows(inSection: 0) == 0 {
+            return nextIndexPath(after: indexPath)
+        }
+
+        return indexPath
+    }
+
+    /// Returns index path after currently selected index path
+    ///
+    var nextIndexPath: IndexPath? {
+        guard let indexPath = indexPathForSelectedRow else {
+            return nil
+        }
+
+        return nextIndexPath(after: indexPath)
+    }
+
+    /// Returns index path before currently selected index path
+    ///
+    var prevIndexPath: IndexPath? {
+        guard let indexPath = indexPathForSelectedRow else {
+            return nil
+        }
+
+        var row = indexPath.row
+        var section = indexPath.section
+
+        while true {
+            if row == 0 {
+                if section == 0 {
+                    return nil
+                }
+
+                section -= 1
+                row = numberOfRows(inSection: section)
+            } else {
+                return IndexPath(row: row - 1, section: section)
+            }
+        }
+    }
+
+    /// Selects row after currently selected row or first row if no row is currently selected
+    ///
+    func selectNextRow() {
+        guard let indexPath = (indexPathForSelectedRow != nil ? nextIndexPath : firstIndexPath) else {
+            return
+        }
+        deselectSelectedRow()
+        selectRow(at: indexPath, animated: false, scrollPosition: .none)
+        scrollRectToVisible(rectForRow(at: indexPath), animated: false)
+    }
+
+    /// Selects row before currently selected row or first row if no row is currently selected
+    ///
+    func selectPrevRow() {
+        guard let indexPath = prevIndexPath ?? firstIndexPath else {
+            return
+        }
+        deselectSelectedRow()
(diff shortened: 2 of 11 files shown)


Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -543,6 +537,10 @@ - (void)sidebarContainerDidHideSidebar:(SPSidebarContainerViewController *)sideb
     self.searchBar.userInteractionEnabled = YES;
 
     self.navigationController.navigationBar.userInteractionEnabled = YES;
+
+    if (!self.isSearchActive) {
```
jleandroperez, line 544: Nipticky: Mind adding a comment on this one? (I had to look it up why we'd set the VC as the first responder).

Just in case, to prevent someone in the future from deleting this and breaking shortcuts! 😄

Simplenote/Classes/UITableView+Simplenote.swift
```diff
@@ -46,3 +46,120 @@ extension UITableView {
...
+    }
+
+    /// Returns index path before currently selected index path
+    ///
+    var prevIndexPath: IndexPath? {
+        guard let indexPath = indexPathForSelectedRow else {
+            return nil
+        }
+
+        var row = indexPath.row
```
jleandroperez: Nipticky: Maybe extract this, for symmetry?

```
    func indexPath(before indexPath: IndexPath) -> IndexPath?
```

Simplenote/Classes/UITableView+Simplenote.swift
```diff
@@ -46,3 +46,120 @@ extension UITableView {
...
+        }
+
+        deselectRow(at: indexPath, animated: false)
+    }
+}
+
+// MARK: - Navigation (Private)
+//
+private extension UITableView {
+    func nextIndexPath(after prevIndexPath: IndexPath) -> IndexPath? {
```
jleandroperez: Nipticky: Hooow about:

```
    func indexPath(after indexPath: IndexPath) -> IndexPath? {
Write
```

Since the direction is implied by the `after` ?

Simplenote/TagListViewController.swift
```diff
@@ -376,89 +372,104 @@ private extension TagListViewController {
...
-            return false
+// MARK: - Selection
+//
+private extension TagListViewController {
+    func updateCurrentSelection() {
+        tableView.selectRow(at: currentSelection, animated: false, scrollPosition: .none)
+    }
+
+    var currentSelection: IndexPath? {
+        let filter = NotesListFilter(selectedTag: SPAppDelegate.shared().selectedTag)
```
jleandroperez, line 385: I'm wondering if the TagListViewController should actually own the Selected Tag.

(And down the road, change its type from String to NotesListFilter?)

Simplenote/TagListViewController.swift
```diff
@@ -376,89 +372,104 @@ private extension TagListViewController {
...
+        case .tag(let tagName):
+            guard let index = resultsController.fetchedObjects.firstIndex(where: {
+                $0.name == tagName
+            }) else {
+                return nil
+            }
+
+            return IndexPath(row: index,
+                             section: Section.tags.rawValue)
         }
```
jleandroperez, line 405: Nipticky: (please feel free to disregard, just brainstorming!!)

```
private extension NotesListFilter {
    var indexPath: IndexPath
}
```

Simplenote/TagListViewController.swift
```diff
@@ -331,6 +322,11 @@ extension TagListViewController: UITableViewDataSource, UITableViewDelegate {
     func tableView(_ tableView: UITableView, editingStyleForRowAt indexPath: IndexPath) -> UITableViewCell.EditingStyle {
         return .none
     }
+
+    func reloadTableView() {
+        tableView.reloadData()
+        updateCurrentSelection()
```
jleandroperez, line 328: Why do we need to "recalculate" the selected row? is it to deal with selectedRowIndexPath changes applied via the new UITableView extension?

jleandroperez (approved):
Works like magic!! Added few cosmetic comments, feel free to disregard!

Beautiful work @eshurakov !!

:shipit:

SylvesterWilmott:
Everything looks great!

Can we use cmd + shift + f to toggle the search instead of just to focus? So using the shortcut again will cancel the search as well as using ESC.

Simplenote/Classes/SPNoteListViewController.m
eshurakov: Good point!

Simplenote/Classes/UITableView+Simplenote.swift
eshurakov: I like it, done!

Simplenote/Classes/UITableView+Simplenote.swift
eshurakov: Good point, done!

Simplenote/TagListViewController.swift
eshurakov: We can't do that, because we use `resultsController` to get correct index for a tag

Simplenote/TagListViewController.swift
eshurakov: We update selected index path in case the currently selected row changed its index path after the reload. Can happen if we add, delete or reorder tags.

Simplenote/TagListViewController.swift
eshurakov: That's a good point. I think there are several options, we can also have a separate object holding the current state of the app and inject it when needed. 
Note list for example can listen for changes in that object and update the list.

eshurakov:
@SylvesterWilmott thank you! I've added support to cancel the search via the same shortcut in the follow up PR: https://github.com/Automattic/simplenote-ios/pull/1113

SylvesterWilmott (approved):

Simplenote/TagListViewController.swift
jleandroperez: Ahhhh totally missed that, apologies about the noise!!

Simplenote/TagListViewController.swift
jleandroperez: Thanks for the clarification!

jleandroperez (approved):
@eshurakov Thank you for the latest changes, love this PR!!

:shipit:

eshurakov:
Thank you!

pachlava:
I experience shortcuts not working often while using the MBP keyboard with simulator.
Will return to verification in next days when I get wireless keyboard and will test on a real device.

pachlava:
[testing] fix/feature verified on iPhone XR | iOS 14.4 | Simplenote 4.32.0.0 + Magic Keyboard

pachlava:
@eshurakov all works great with real device and wireless keyboard. There's one thing that bugs me: when the user has something in search, and then navigates to the notes list, enters and leaves the note, they might want to return to search and correct/clarify something in the existing search string. What happens now is that the existing search string will be cleared when Cmd+Shift+f is used.

IMO it makes sense not to clear the text. Let me know if I should create an enhancement for that.

UPD: after reading [MK2](https://github.com/Automattic/simplenote-ios/pull/1113) in detail, I see that it's expected:
> On the note list: when search bar is active, pressing Cmd + Shift + F cancels search

For the user this might look like not the best perspective. This means that every search should be typed in from scratch (if the user uses only the external keyboard). While it's OK for occasional searches, it's not the best experience when the user wants to search a lot, especially entering tags, and adjusting the search. Maybe this area can be improved.

eshurakov:
Thanks @pachlava!
Let's ask @SylvesterWilmott about ⬆️

pachlava:
@eshurakov 
> Hold cmd to see available shortcuts

^ Should this work on iPhone? I saw this on iPad sim, but never on iPhone (both real and sim)

SylvesterWilmott:
>There's one thing that bugs me: when the user has something in search, and then navigates to the notes list, enters and leaves the note, they might want to return to search and correct/clarify something in the existing search string. What happens now is that the existing search string will be cleared when Cmd+Shift+f is used.

Oh, well spotted. We could add some logic that if a search string already exists in the search bar then the shortcut to focus search should not clear it.

eshurakov:
> > Hold cmd to see available shortcuts
> 
> ^ Should this work on iPhone? I saw this on iPad sim, but never on iPhone (both real and sim)

@pachlava that window is automatically shown by the system and we can't control it. I guess it works only on iPad.

PR #1112, opened 2021-01-25
Merge release 4.31 beta 1 into develop

mokagio:


jkmassel (approved):

jkmassel:
Approved but not merged because I just wanted to double check – we don't do editorial review for Simplenote release notes?

If not, good to merge, but if we do we shouldn't update the `.pot` file here – otherwise the translations will land in `develop` twice and may be translated twice, which we don't want (because we'd be throwing away the first set).

loremattei:
Right, we don't do editorial reviews for Simplenote @jkmassel 👍

PR #1113, opened 2021-01-27
Keyboard Support: MK2

### LINKED ISSUE #1092: Project: Keyboard Support

eshurakov:
### Details
In this PR we're adding the following features:
* shortcuts support in the note editor
* in-memory caching of scroll position for each note
* RTL support for keyboard navigation
* Tracks events for keyboard shortcuts

Closes #1092

### Test (iPhone / iPad with external keyboard)
1. Open a note

- [x] If markdown is enabled, preview can be toggled with `Cmd + Shift + P`
- [x] Toggle checklist with `Cmd + Shift + C`
- [x] If editor is not active, `left` arrow brings you back to the list
- [x] If editor is not active, `Cmd + Return` brings you back to the list

---

1. Open a note from the search

- [x] It is possible to navigate through matches with `Cmd + G` and `Cmd + Shift + G`

---

- [x] Notes remember its scroll position as you switch between different notes
- [x] When apps interface is RTL, left and right arrows work as expected
- [x] When using a shortcut there is a `Tracked: shortcut_used` printed in the console
- [x] On the note list: when search bar is active, pressing `Cmd + Shift + F` cancels search

### Review

### Release
`RELEASE-NOTES.txt` was updated in 1033eff0 with:

> Added support for keyboard navigation and shortcuts #1092


SylvesterWilmott:
Looks good and works as expected.

One minor thing I noticed is that there is a slightly inconsistent behaviour between using the Cmd + G shortcut to cycle through search matches and using the on screen chevrons to do the same:

1. Perform a search
2. Go into the note
3. Use Cmd + G to go to the last match
4. Use Cmd + G one more time
5. Use Cmd + Shift + G
6. Notice that the last match remains selected

Using the on screen chevrons to perform the same task doesn't have the same effect.

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -648,6 +649,32 @@ private extension SPNoteListViewController {
...
+            tableView.selectRow(at: indexPath, animated: false, scrollPosition: .none)
+        }
+    }
+
+    private func updateSelectedNoteBasedOnSelectedIndexPath() {
+        if let indexPath = tableView.indexPathForSelectedRow, let note = notesListController.object(at: indexPath) as? Note {
+            selectedNote = note
+        } else {
+            selectedNote = nil
+        }
```
jleandroperez, line 672:
```
        selectedNote = tableView.indexPathForSelectedRow.flatMap { indexPath in
            notesListController.object(at: indexPath) as? Note
        }
```

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -57,20 +57,6 @@ extension SPAppDelegate {
...
-    ///         In both scenarios, there's already an Editor onScreen, and the standard SDK behavior is to "pop".
-    ///
-    /// - Note: Furthermore, when pressing an Interlink from the Editor, the "Link Text Interaction" injects a floating view, so that the link appears to "float".
-    ///         Refreshing the Editor's contents right there (with any kind of animation) ends up interacting with this "Floating Link" behavior.
-    ///         For such reason, we're opting for simply pushing another VC.
-    ///
-    private func replaceNoteEditor(_ editorViewController: SPNoteEditorViewController) {
-        navigationController.presentedViewController?.dismiss(animated: true, completion: nil)
-        navigationController.setViewControllers([noteListViewController, editorViewController], animated: true)
-    }
```
jleandroperez, line 73: Thank you 😄

eshurakov:
Thank you @SylvesterWilmott, good catch! This has been fixed now.

SylvesterWilmott (approved):

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -580,32 +589,32 @@ - (void)updateWithSearchQuery:(SearchQuery *)searchQuery
...
-    [self highlightSearchResultAtIndex:MAX(0, self.highlightedSearchResultIndex - 1) animated:YES];
+    [self highlightSearchResultAtIndex:(self.highlightedSearchResultIndex - 1) animated:YES];
 }
 
 - (void)highlightSearchResultAtIndex:(NSInteger)index animated:(BOOL)animated
 {
+    NSInteger searchResultCount = self.searchResultRanges.count;
+
+    index = MIN(index, searchResultCount - 1);
+    index = MAX(index, 0);
```
jleandroperez, line 562: AAhhhh I was gonna as why were we dropping the bounds checks 😄

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -431,6 +444,9 @@ - (void)openNote:(Note *)note ignoringSearchQuery:(BOOL)ignoringSearchQuery anim
 {
     [SPTracker trackListNoteOpened];
 
+    self.selectedNote = note;
```
jleandroperez, line 437: Q: do you think we could, maybe, avoid keeping a reference here?

Maybe something in the lines of...

```
let editor = navigationController.viewControllers.first { $0 is NoteEditorViewController }
return editor?.note
```

Maybe not worthy? (just trying to avoid the need to "maintain" yet another property)

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -443,15 +459,7 @@ - (void)openNote:(Note *)note ignoringSearchQuery:(BOOL)ignoringSearchQuery anim
...
-    // Failsafe:
-    // We were getting (a whole lot!) of crash reports with the exception
-    // 'Pushing the same view controller instance more than once is not supported'. This is intended to act
-    // as a safety net. Ref. Issue #345
-    if (self.navigationController.visibleViewController != self) {
-        [self.navigationController popToRootViewControllerAnimated:NO];
-    }
-
-    [self.navigationController pushViewController:editor animated:animated];
+    [self.navigationController setViewControllers:@[self, editor] animated:animated];
```
jleandroperez, line 451: Much better!

jleandroperez (commented):
@eshurakov *Awesome PR!!* Just found one minor glitch:

### Issue: Multiple Markdown Preview(s)

1. Open a note with Markdown enabled
2. Press CMD + Shift + P (and leave it pressed)

As a result, multiple **MarkdownPreviewViewController** instances will be pushed

### Question:
Is **Tab** expected to toggle Edition mode while in the editor? CC @SylvesterWilmott ?

Works super well, but couldn't find it in the original issue / spreadsheet.

SylvesterWilmott:
> Is **Tab** expected to toggle Edition mode while in the editor?

We don't want to block the Tab key in the editor or assign it to any navigation shortcut so the editor uses the Cmd + Enter shortcut to remove focus.

jleandroperez:
Oh!! The `Tab` key only toggles the Editor as the first responder (that is, "initiates edition").
(Otherwise it should work as usual!)

Simplenote/Classes/SPNoteListViewController+Extensions.swift
eshurakov: Nice kung fu 🥋

Simplenote/Classes/SPNoteListViewController.m
eshurakov: Yeah, this part is ugly. But right now I don't have any better ideas :-(

At first I also thought about checking which note is now in the editor and use this information to restore selected note when we come back to the list. If I do this in `viewWillAppear` method of the list view controller, there is no editor because it was already removed from the navigation controller stack.

eshurakov:
@jleandroperez Thank you for the review, great catch with multiple markdown previews. It's fixed now :-)

pachlava:
I experience shortcuts not working often while using the MBP keyboard with simulator.
Will return to verification in next days when I get wireless keyboard and will test on a real device.

pachlava:
[testing] fix/feature verified on iPhone XR | iOS 14.4 | Simplenote 4.32.0.0 + Magic Keyboard

PR #1118, opened 2021-01-31
Localize keyword for tag search

### LINKED ISSUE #860: "tag:" prefix gets localized and is harder to type in some languages
This is an interesting one!

A user reports (#3261645-zen):

> I like the idea of tag search very much, like “tag:food” for example. The problem is that I use Swedish language setting on my iphone. That means that I’d have to type “etikett:food” instead. That makes the feature so much less useful because it takes much longer to type. Could you please make it so that the use of “tag” works in any language so that I could use “tag:food” on my phone?

I'm not sure what the best course of action would be for this, since we probably don't want to remove the localized tag name prefix. I wonder if we can support both the English "tag:" and whatever the localized word for "tag:" is?

cc @SylvesterWilmott

charliescheer:
### Fix
This issue fixes #860 

When doing a tag search in Simplenote iOS, currently you can do so with the keyword tag: but not a localized version of the word tag.  Originally in issue 860, only the localized version worked, but through some changes to SNiOS the problem reversed, only "tag:" would work and not the localized version.  With this PR both tag: and localTag: will work.

### Test
1. After loading the branch, check and make sure that the correct [Simplenote Search branch](https://github.com/Automattic/SimplenoteSearch-Swift/tree/issue/860-localize-tag-search) has also been loaded
2. In the localizedString files, choose a language and add a key pair for "tag:" = _the localized version of tag: in the chosen language_ I used French which would be `"tag:" = "étiquette";`
3. Load up the simulator, with the simulator still set to English.  Go to a search and look for a tag by name It should appear with tag: as the prefix
4. While still in English type in tag: and then search for a tag by name, the prefix should be tag:
5. Switch the simulator to the language you added the translation for
6. Return to search and repeat steps 3 and 4 to make sure that none localized tag: still works
7. In the search field type in the localized version of the tag `étiquette:` or your chosen translation.  With the translated key complete in the search field the results will show the word tag translated

One thing to consider here, is it worth translating the result if a user enters étiquette: ?  Given how specific the entry has to be or that event to trigger, I am not sure that it is.  I had another possible option, where the key tag: does not appear in the search results.  Currently there is an icon that appears next to the tags and there is a header which is localized declaring the tag search results section.  We could remove tag:/étiquette: from the result and simply rely on the header and icon. 

One more thing to consider that I am not 100% sure on here.  One of the things that we had been looking for is to make sure that NSLocalizedString gets placed in such a way that the scripts can pick up the strings and then change them.  Because the SearchQuerySettings struct is in SimplenoteSearch, I am not sure that this will happen.  Can you take a look to see if the Strings should get picked up by the localizing scripts, and if not, can you advise on changes to make to fix it?


Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -630,7 +630,13 @@ private extension SPNoteListViewController {
         let cell = tableView.dequeueReusableCell(ofType: SPTagTableViewCell.self, for: indexPath)
         cell.leftImage = .image(name: .tag)
         cell.leftImageTintColor = .simplenoteNoteShareStatusImageColor
-        cell.titleText = String.searchOperatorForTags + tag.name
+        cell.titleText = String.searchOperatorForTags.tagsKeyword + tag.name
+        
+        //Display localized tag search
+        if searchPrefixIsLocalizedTag() {
+            cell.titleText = String.searchOperatorForTags.localizedTagKeyword + tag.name
+        }
```
jleandroperez: By the way!! I'm not sure if we need to change the way this is presented to the user. As far as I know we just need to support `Localized("tag:", ...)` and `tag:` (because users mentioned that, in some languages, just typing `tag` is easier).

Can we check with @SylvesterWilmott on this one?

charliescheer: When I was setting it up I thought it was odd that I would type étiquette and I would see tag, except in the header, so I thought translating the key would be less confusing for the user.

I think I prefer removing the word tag: or translated-tag: from the results all together, The user would see the header that says "search by tag" and the tag icon next to it.  I think adding the word tag to the results is overkill.

But yes, @SylvesterWilmott WDYT

SylvesterWilmott: Localising the key looks like the ideal scenario IMO. I would run it by the team as well since we will want to align with this.

jleandroperez:
@charliescheer please LMK whenever this one is ready for another look! not sure we have a decision on the `tag:` prefix detail!

mokagio:
Pushing this PR to the next milestone, 4.33, because I'm running the code-freeze today and this is an enhancement PR: it can safely wait till the next release.

charliescheer:
@jleandroperez Okay, this one got pushed to the back burner for a few days, but I have updated the PR to reflect the changes you asked for.  When you get a chance can you take another look? Thank you!!

jleandroperez (approved):
@charliescheer Good to go sir!!

:shipit:

pachlava:
Since the `tag:` string was localized only for EN, I tested this by adding a translation for French locally, and it worked:

Screenshot 2021-03-09 at 15 00 35
Screenshot 2021-03-09 at 14 56 15

I wonder if this should also work with diacritics removed (or considered as a future improvement), but in any case, the feature works.

[testing] fix/feature verified on iPhone 12 Sim / iPad Pro 11" Sim | iOS 14.4 | Simplenote 4.33.0.0

PR #1123, opened 2021-02-02
Issue/517 share to books

### LINKED ISSUE #517: Share as PDF to iBooks Available But Does Not Work
The option to share a note to Books is available, but does not work

### Expected
After clicking on the Send > Books option, a modal appears saying that a PDF is being created -- at that point, I'd expect to be sent to be sent to Books

### Observed
The modal appears, but I stay in Simplenote, and no new PDF is added to Books.

### Reproduced
1. Create a new note and add content to the note
2. Click on the (i) icon, and choose Send
3. From the app options, choose Books
4. Wait for Creating PDF modal to close
5. Go to Books app, verify no new PDF has been created

-|-|-|-
Apple|iPhone 11 Pro|13.2|4.12
Apple|iPhone Pad|13.2|4.11

Reported in #2471529-zen

charliescheer:
### Fix
This is a PR to fix #517 

The issue in #517  is that sharing from Simplenote to Apple Books is an option in the share sheet, but it doesn't work.  A message appears saying that it is converting the note to a PDF but nothing ever happens.

In this PR we are creating the PDF for Books instead and then supplying the ActivityView a link to where the file can be found so books can open the link.

This is a proof of concept about how we could potentially fix this problem, looking for some feedback on the approach

### Test
To test:
1. Open up a note and go to Options > Share > Books (you may have to look for it in the more options)
2. books should open and a PDF copy of your note should appear

A few things to keep in mind.  
-This is not styled in any way at this time, and we may need to more carefully consider what the PDF output should be (markup?  text styles?  etc)
-This should be able to continue creating new versions of PDFs if the note has changes to it, if there are no changes it should open the existing file
-This should only trigger the creation of the PDF once, at the time that share to books is tapped, but would be good to confirm.
-Making the PDF takes a bit of time, 2-3 seconds, but this happens while you are looking at books, so while it feels a bit laggy, it doesn't feel like it is laggy on SNs end

What do you think about how this is setup?

### Release


Simplenote/Classes/Note+Properties.swift
```diff
@@ -105,6 +105,17 @@ extension Note {
     }
 }
 
+// MARK: Export Identifier
+///
```
jleandroperez, line 109: Nipticky: Inconsistent `//` - `///`

Simplenote/Classes/Note+Properties.swift
```diff
@@ -105,6 +105,17 @@ extension Note {
...
+extension Note {
+    
+    ///Returns a unique identifier for a note that can be used to create a file name
```
jleandroperez: Nipticky: Missing space

Simplenote/PDFLinkActivityItemProvider.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+public class PDFLinkActivityItemProvider : UIActivityItemProvider {
+    public let targetURL: URL
+    public let filename: String
+    public let note: Note
+    
+    public init(note: Note) {
+        self.note = note
+        self.filename = String(format: "%@.pdf", "Thisisastring")
```
jleandroperez: Perhaps this one should be `note.exportFilename + ".pdf"` ?

Simplenote/PDFLinkActivityItemProvider.swift
```diff
@@ -0,0 +1,24 @@
...
+        self.targetURL = FileManager.documentsURL.appendingPathComponent(filename)
```
jleandroperez: I was thinking that **PERHAPS** we got the WP exporter wrong. Can we switch to [temporaryDirectory](https://www.google.com/search?client=safari&rls=en&q=NSFileManager+temporaryDirectory&ie=UTF-8&oe=UTF-8) ? 

[Ref. here](https://stackoverflow.com/questions/16666812/whats-the-difference-between-saving-content-at-document-directory-or-temp-folder). Files stored in `temporary` aren't meant to persist thru launch OPs

Simplenote/PDFLinkActivityItemProvider.swift
```diff
@@ -0,0 +1,24 @@
...
+        
+        super.init(placeholderItem: targetURL)
+    }
+    
+    public override var item: Any {
+        guard activityType?.rawValue != "com.apple.UIKit.activity.RemoteOpenInApplication",
```
jleandroperez: Let's move the inline string to a constant 😄 

You can do something like...

```
private enum Settings {
    static let supportedActivityType = "..."
}
```

(Or a simple `let` constant on top of the file with a comment would just be ferpect!)

Simplenote/PDFLinkActivityItemProvider.swift
```diff
@@ -0,0 +1,24 @@
...
+        let url = SimplenotePDFExporter().exportNoteToFiles(note: note, filename: filename) else {
```
jleandroperez: Multiple invocations over this getter would cause multiple PDF Export OPs.

One way to solve that is by doing something like this:

```
private lazy var documentURL = SimplenotePDFExporter().exportNoteToFiles(note: note, filename: filename)

override var item: Any {
    documentURL ?? note.content
}
```

Simplenote/PDFLinkActivityItemProvider.swift
jleandroperez: **PLUS** are we 100% sure we need to check on the Activity Type?

Since we're exporting a PDF and passing along a URL that contains the `.pdf` extension, perhaps we can drop the `activityType` check?

Furthermore, probably checking on the activity type on a parameter-less getter is probably not good. In theory, getters should behave the same, regardless of the caller. Otherwise, the internal behavior would be hard to tell, from outside the API.

Simplenote/SimplenotePDFExporter.swift
```diff
@@ -0,0 +1,39 @@
...
+            return nil
+        }
+        return targetURL
+    }
+    
+    private func createPDFDataFromNote(note: Note) -> Data {
+        let format = UIGraphicsPDFRendererFormat()
+        
+        let pageWidth = 8.5 * 72.0
+        let pageHeight = 11 * 72.0
```
jleandroperez: Can we move these to constants?

```
private struct Metrics {
   static let pageSize = CGSize(...)
}
```

Plus: what's the page size of these constants? (please add a comment, or name the variable afterwards)

Simplenote/SimplenotePDFExporter.swift
```diff
@@ -0,0 +1,39 @@
...
+        let pageRect = CGRect(x: 0, y: 0, width: pageWidth, height: pageHeight)
+        let renderer = UIGraphicsPDFRenderer(bounds: pageRect, format: format)
+        
+        let data = renderer.pdfData { (context) in
+          context.beginPage()
+          let attributes = [
+            NSAttributedString.Key.font: UIFont.boldSystemFont(ofSize: 16)
+          ]
+            if let text = note.content {
+                text.draw(at: CGPoint(x: 0, y: 0), withAttributes: attributes)
```
jleandroperez:
```suggestion
                text.draw(at: .zero, withAttributes: attributes)
```

Simplenote/PDFLinkActivityItemProvider.swift
jleandroperez: Second note: the `Target URL` is also calculated by `SimplenotePDFExporter`. It'd be neat to converge everything in a single spot 🔥

jleandroperez (commented):
Few nipticky comments!! general approach is solid!!

🔥

Simplenote/PDFLinkActivityItemProvider.swift
charliescheer: I didn't think that we did, but if it is not there then the share for things like mail and apple notes were getting the note content and a PDF.  By adding this check in before creating the PDF we can stop that

charliescheer:
@jleandroperez I have made a bunch of changes to the PDF exporter.  When you get a chance, please take another look and  let me know what you think.

Simplenote/Classes/FileManager+Simplenote.swift
```diff
@@ -34,4 +34,17 @@ extension FileManager {
 
         return targetURL
     }
+    
+    /// Writes a data file to the Documents Folder
+    ///
+    class func writeDataToDocuments(data: Data, to targetURL: URL) -> URL? {
```
jleandroperez, line 40: - Since this API doesn't really save `to documents` exclusively (the URL could point anywhere), perhaps we could drop the `DataToDocuments` suffix?

- Since the code isn't related to FileManager itself (and the log mentions Note Exporter), maybe the best spot for it is `PDFLinkActivityItemProvider` itself?

Simplenote/PDFLinkActivityItemProvider.swift
```diff
@@ -0,0 +1,30 @@
...
+        documentURL ?? content
+    }
+    
+    private func shareToPDF() -> URL? {
+        guard activityType?.rawValue == Consts.remoteOpenInAppActivityType else {
+            return nil
+        }
+        let data = SimplenotePDFExporter.exportStringToPDFData(content)
+        return FileManager.writeDataToDocuments(data: data, to: targetURL)
+    }
```
jleandroperez, line 25: How about if instead of coupling the output of this API to the `activityType` instead we implement this API?
(And drop the `item: Any` getter)

```
    public override func activityViewController(_ activityViewController: UIActivityViewController, itemForActivityType activityType: UIActivity.ActivityType?) -> Any? {
        guard activityType?.rawValue == Consts.remoteOpenInAppActivityType else {
            return nil
        }

        return documentURL ?? content
    }
```

Simplenote/PDFLinkActivityItemProvider.swift
```diff
@@ -0,0 +1,30 @@
...
+}
+
+private struct Consts {
```
jleandroperez, line 28:
```suggestion
private struct SupportedActivityType {
```

Simplenote/PDFLinkActivityItemProvider.swift
```diff
@@ -0,0 +1,30 @@
...
+    static let remoteOpenInAppActivityType = "com.apple.UIKit.activity.RemoteOpenInApplication-ByCopy"
```
jleandroperez, line 29:
```suggestion
    static let remoteOpenInApp = "com.apple.UIKit.activity.RemoteOpenInApplication-ByCopy"
```

Simplenote/UIActivityViewController+Simplenote.swift
```diff
@@ -16,7 +16,8 @@ extension UIActivityViewController {
 
         let print = SPSimpleTextPrintFormatter(text: content)
         let source = SimplenoteActivityItemSource(note: note)
+        let link = PDFLinkActivityItemProvider(content: content, filename: String(format: "%@.pdf", note.exportFilename()))
```
jleandroperez, line 19: Perhaps we can send over just the `filename` parameter, and append `+ ".pdf"` in the PDF ItemProvider? (since it'll always be PDF)

Simplenote/SimplenotePDFExporter.swift
```diff
@@ -0,0 +1,25 @@
...
+                NSAttributedString.Key.font: UIFont.boldSystemFont(ofSize: 16)
+            ]
+            string.draw(at: .zero, withAttributes: attributes)
+        }
+        
+        return data
+    }
+}
+
+private struct Consts {
```
jleandroperez: Mind if we use `Metrics` or `Constants` for consistency? (SORRY Nipticky!!!)

Simplenote/SimplenotePDFExporter.swift
```diff
@@ -0,0 +1,25 @@
...
+    public static func exportStringToPDFData(_ string: String) -> Data {
+        let format = UIGraphicsPDFRendererFormat()
+        
+        let pageRect = CGRect(x: 0, y: 0, width: Consts.pageWidth, height: Consts.pageHeight)
+        let renderer = UIGraphicsPDFRenderer(bounds: pageRect, format: format)
+        
+        let data = renderer.pdfData { (context) in
+            context.beginPage()
+            let attributes = [
+                NSAttributedString.Key.font: UIFont.boldSystemFont(ofSize: 16)
```
jleandroperez: Are we sure all of the text should be bold?

Simplenote/SimplenotePDFExporter.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+struct SimplenotePDFExporter {
+    public static func exportStringToPDFData(_ string: String) -> Data {
+        let format = UIGraphicsPDFRendererFormat()
+        
+        let pageRect = CGRect(x: 0, y: 0, width: Consts.pageWidth, height: Consts.pageHeight)
```
jleandroperez: Perhaps move this to the structure below, since it's always the same?

```
static let pageBounds = ...
```

jleandroperez (commented):
Sending you few nipticky comments!!

Regarding the actual PDF generation:

1. The resulting document has no hading nor margin. Looks like a square on my end
2. The text itself gets horizontally clipped (overflows the readable area)
3. Do we support multiple pages?

charliescheer:
closing in favor of #1135

PR #1124, opened 2021-02-03
Added automated UI tests to the workspace

pachlava:
### Fix
During my trial I automated 21 UI test, and kept them in my fork. Now created a branch here. Having these tests in develop will help running automated tests against next betas.

I understand this is a huge one. I wrote this for two weeks only in my fork. But committing this in parts in current state will make tests execution impossible.

### Test
Works best on iPhone 12. Certain tests will fail on previous versions, because of different ways to locate controls. I'm planning to add support of multiple versions gradually.

1. Make sure that 'Use the Same Keyboard Language as macOS' and 'Connect Hardware Keyboard' are disabled in Simulator properties ('I/O' → 'Keyboard'). 
2. All tests are located in SimplenoteUITests.swift
3. Run them in similar way as existing 'SimplenoteTests' or 'SimplenoteScreenshots'

### Review
Besides new files with tests themselves, please check if I did not break anything 'unrelated':
- .bundle/config
- Simplenote.xcodeproj/project.pbxproj
- Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme
- Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteScreenshots.xcscheme (have no idea why this was affected - maybe because I created one more 'UI Testing Bundle' target, and this interfered with existing SimplenoteScreenshots.

It's the first PR in my life :-)

### Release


.bundle/config
```diff
@@ -1,2 +1,4 @@
 ---
 BUNDLE_PATH: "vendor/bundle"
+BUNDLE_RETRY: "3"
+BUNDLE_JOBS: "3"
```
mokagio, line 4: This change is due to how certain versions of [Bundler](https://bundler.io/) update the `.bundle/config` file when called with options. In particular, this is the result of:

https://github.com/Automattic/simplenote-ios/blob/883bd582bfc659bddf8a989735023e3f9088b6c6/Rakefile#L49

I think it's fine to keep this change here. As a followup, we should remove the options from the callsite in the `Rakefile` since Bundler will already use them now that they're tracked here.

Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme
```diff
@@ -47,6 +47,16 @@
                ReferencedContainer = "container:Simplenote.xcodeproj">
             </BuildableReference>
          </TestableReference>
+         <TestableReference
+            skipped = "NO">
```
mokagio: Xcode automatically adds a new test target to the test action of the main one which is handy to get started with it right away, but undesirable in the case of UI tests.

UI tests are slower and brittle. Making them run together with the unit tests would decrease the speed at which developers can get feedback from them.

Could you please skip the UI tests in the test action of this scheme?

image

We can always create a dedicated scheme to run the UI tests only, similar to the screenshots tests one.

Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteScreenshots.xcscheme
```diff
@@ -40,6 +40,16 @@
...
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "D864B15325CAE25000F9B73E"
+               BuildableName = "SimplenoteUITests.xctest"
+               BlueprintName = "SimplenoteUITests"
+               ReferencedContainer = "container:Simplenote.xcodeproj">
+            </BuildableReference>
+         </TestableReference>
```
mokagio: From the PR description:

> have no idea why this was affected - maybe because I created one more 'UI Testing Bundle' target, and this interfered with existing SimplenoteScreenshots.

Yeah, I'm guessing this is just what Xcode does: when you add a new test target it adds it to the existing schemes that are using the same host app.

Same comment as above: could you please undo this change?

In this case, running the UI tests together with those that take the screenshots would result in a greater execution time for the screenshot automation.

Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteScreenshots.xcscheme
```diff
@@ -69,15 +79,6 @@
...
       debugDocumentVersioning = "YES">
-      <MacroExpansion>
-         <BuildableReference
-            BuildableIdentifier = "primary"
-            BlueprintIdentifier = "46A3C96017DFA81A002865AE"
-            BuildableName = "Simplenote.app"
-            BlueprintName = "Simplenote"
-            ReferencedContainer = "container:Simplenote.xcodeproj">
-         </BuildableReference>
-      </MacroExpansion>
```
mokagio, line 80: I wonder why this disappeared 🤔 Maybe a difference in how Xcode manages its schemes since this was created? 🤷‍♂️ 

I run the screenshots tests locally and they still work 👍

SimplenoteUITests/Playground.swift
```diff
@@ -0,0 +1,59 @@
...
+/*
+let allElsCount = app.descendants(matching: .any).count
+
+for index in 0...allElsCount - 1
+{
+    let el = app.descendants(matching: .any).element(boundBy: index)
+    print(el.description)
+    print(el.label)
+}
+*/
```
mokagio: Do we need to keep this file?

TESTING-CHECKLIST.md
```diff
@@ -2,9 +2,9 @@
 
 ### Login
 
-- [ ] Logout
-- [ ] Login with wrong password fails
-- [ ] Login with correct password succeeds
+- [ ] Logout (A)
```
mokagio, line 6: What do you think of documenting that `(A)` means the test is automated at the start of this file?

Pretty cool to see how many you automated! 🙌

SimplenoteUITests/AllNotesClass.swift
```diff
@@ -0,0 +1,93 @@
+//
+//  AllNotesClass.swift
+//  SimplenoteUITests
+//
+//  Created by Sergiy Fedosov on 03.02.2021.
+//  Copyright © 2021 Automattic. All rights reserved.
+//
+
```
mokagio: Nitpick: I'm not sure if it's explicitly documented anywhere, but I think as a general convention we don't keep these comments that Xcode generates in the file.

The reason is that they only add clutter while all the information is already available between Xcode itself and Git. I'm not 100% sure about the copyright, but I think that because of the open-source licensing that's not necessary either.

Feel free to address removing all these headers in a dedicated commit as part of this PR or as a followup PR. No stress 👍

SimplenoteUITests/SimplenoteUITests.swift
```diff
@@ -0,0 +1,531 @@
...
+
+class SimplenoteUISmokeTestsLogin: XCTestCase {
+
+    let testDataInvalidEmail = "user@gmail."
+    let testDataNotExistingEmail = "nevergonnagiveyouup@gmail.com"
+    let testDataInvalidPassword = "ABC"
+    let testDataNotExistingPassword = "ABCD"
+
+    override func setUpWithError() throws {
+        app.launchArguments = ["enable-testing"]
```
mokagio: I can't see this value ever being used? A search for "enable-testing" only results in these files, there's no production code that looks into `CommandLine.launchArguments` for them and then triggers a change in the app's state as a result.

Is there some production code that goes along with these changes that didn't make it in the PR, or is this genuinely unused?

If it's unused, I'd suggest removing this.

SimplenoteUITests/SimplenoteUITests.swift
```diff
@@ -0,0 +1,531 @@
...
+        continueAfterFailure = true
+        app.launch()
+        let _ = attemptLogOut()
+    }
+
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
```
mokagio:
```suggestion
```

Along the same line as the previous comment. If you don't plan on adding code to this method from the test template in the PR, let's just remove it.

SimplenoteUITests/SimplenoteUITests.swift
```diff
@@ -0,0 +1,531 @@
...
+    // Test below is automatically generated one, disabled for now to save time
+    /*
+    func testLaunchPerformance() throws {
+        if #available(macOS 10.15, iOS 13.0, tvOS 13.0, *) {
+            // This measures how long it takes to launch your application.
+            measure(metrics: [XCTApplicationLaunchMetric()]) {
+                XCUIApplication().launch()
+            }
+        }
+    }*/
```
mokagio: Feel free to remove this. 

I think Xcode adds this to the template of a new UI tests bundle to show new developers how to write performance tests.

If we'll ever want to write performance tests in the UI test suite, we can easily find the [documentation online](https://developer.apple.com/documentation/xctest/xctestcase/1496290-measure).

SimplenoteUITests/AllNotesClass.swift
```diff
@@ -0,0 +1,93 @@
...
+    
+    class func isOpen() -> Bool {
+        return app.navigationBars[uidNavBar_AllNotes].exists
+    }
+
+    class func open() {
+        if !isOpen() {
+            app.navigationBars.element.buttons[uidButton_Menu].tap()
+            app.tables.staticTexts[uidButton_AllNotes].tap()
+        }
```
mokagio: Nitpick: Swift has a `guard` statement that we could use here to make the code (in my opinion) easier to read and (objectively) free from an extra level of indendation:

```suggestion
        guard isOpen() else { return }

        app.navigationBars.element.buttons[uidButton_Menu].tap()
        app.tables.staticTexts[uidButton_AllNotes].tap()
```

SimplenoteUITests/AllNotesClass.swift
```diff
@@ -0,0 +1,93 @@
...
+            cell.swipeLeft()
+            cell.buttons[uidButton_NoteCell_Trash].tap()
+        }
+    }
+
+    class func waitForLoad() {
+        let allNotesNavBar = app.navigationBars[uidNavBar_AllNotes]
+        sleep(4)
+        while allNotesNavBar.staticTexts[uidText_AllNotes_InProgress].exists {
+        }
```
mokagio: Do we need the `sleep` call given there's code that waits already?

Also, I wonder if there's a way to wait for the "in progress" text do disappear that leverages the test framework instead 🤔 This code has the potential of turning into an infinite loop if something breaks in the loading progress.

For example:

```swift
class func waitForLoadedExpectation() -> XCTestExpectation {
    let allNotesNavBar = app.navigationBars[uidNavBar_AllNotes]
    let predicate = NSPredicate { _, _ in
        allNotesNavBar.staticTexts[uidText_AllNotes_InProgress].exists
    }
    return XCTNSPredicateExpectation(predicate: predicate, object: .none)
}
```

Which can then be used like

```swift
wait(for: [AllNotes.waitForLoadedExpectation()], timeout: 30)
```

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -391,6 +391,17 @@
 		B5F3FFFF23F44679007A7C59 /* SPSortBar.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5F3FFFE23F44679007A7C59 /* SPSortBar.swift */; };
 		BA5C1C0725BF9D6C006E3820 /* SPDragBar.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA5C1C0625BF9D6C006E3820 /* SPDragBar.swift */; };
 		D435D38FCA5863E8447D5C2C /* Pods_Automattic_SimplenoteShare.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 183BAB896957F07FDCAB6DF5 /* Pods_Automattic_SimplenoteShare.framework */; };
+		D864B16925CAE47C00F9B73E /* Generic.swift in Sources */ = {isa = PBXBuildFile; fileRef = D864B16825CAE47C00F9B73E /* Generic.swift */; };
+		D864B18225CAE4BE00F9B73E /* SimplenoteUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D864B18125CAE4BE00F9B73E /* SimplenoteUITests.swift */; };
+		D864B19225CAE54300F9B73E /* AllNotesClass.swift in Sources */ = {isa = PBXBuildFile; fileRef = D864B19125CAE54300F9B73E /* AllNotesClass.swift */; };
```
mokagio: Nitpick on the file names: I find the "Class" prefix unnecessary. The main type in the file is `AllNotes` and if a reader opens it, they'll see it's a `class`.

Xcode also clearly shows that it's a class in the Open Quickly (<kbd>Shift Cmd O</kbd>) view:

image

SimplenoteUITests/SimplenoteUITests.swift
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    override func setUpWithError() throws {
+        continueAfterFailure = true
+        app.launch()
+
+        AllNotes.clearAllNotes()
+        Trash.empty()
+        AllNotes.open()
+    }
```
mokagio: There are two setup methods here. The first is the `class` setup, it will run only once, before the tests start; the second will run before every test.

Do we need them both? If so, I _think_ `continueAfterFailure` ought to be set in `class setUp` instead of being reset before every test. I also _think_ there's no need to run `app.launch()` again either. I say I _think_ because I haven't verified either, so correct me if I'm wrong, maybe I'm missing something 👌

SimplenoteUITests/AllNotesClass.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+class AllNotesAssert {
+
+    class func noteExists(noteName: String) {
+        XCTAssertTrue(app.tables.cells[noteName].exists, "\"" + noteName + noteNotFoundInAllNotes)
+    }
```
mokagio: Out of curiosity, why did you decide to split the assertions in a dedicated class as opposed to keep them in `AllNotes` and prefix the function name with "assert", e.g.:

```swift
AllNotes.assertNoteExists(noteName: ...)
```

Just out of curiosity though, I don't see any issue in this.

mokagio (changes requested):
I run the tests on the iPod Touch and iPhone 12 Simulators and they went well, as far as UI tests go: only one failure in each 😅 

#### iPod Touch

Screen Shot 2021-02-09 at 2 18 11 pm

#### iPhone 12

Screen Shot 2021-02-09 at 2 29 59 pm

The failure was:

Screen Shot 2021-02-09 at 2 32 28 pm

I like how you used the page object pattern, with the dedicated classes for each screen. The tests are quite readable as a result of that 👏 

I left a few suggestions and nitpicks, but overall I'm keen in getting this merged soon and refine later as you keep working on it because, as you say:

> Having these tests in develop will help running automated tests against next betas.

I'm requesting changes just to make sure we don't merge this with the new UI tests running in the test action of the other schemes.

> I'm planning to add support of multiple versions gradually.

That's the way to go 👍 

In fact, I'm excited to have you on board working on these. There's been different ad hoc efforts to write test automation but it seems to me like the time is ripe to get serious about it and define some good standards.

I'm sure there's possibilities to share code or helpers between this and the screenshot automation. I'm working on other stuff at the moment, but I'll try to set aside some time soon to look into it. It's exciting because I feel it would end up improve _both_ test suites.

Also cc @jkmassel who's done some great work on the UI tests / screenshots automation.

SimplenoteUITests/AllNotesClass.swift
pachlava: I think this happened because I assumed that a class with actions and assertion can grow enough that it should be split into separate files, and also because it's easier to have an overview of what's included into the class, like this:

Screenshot 2021-02-15 at 14 26 11

I agree that it makes a very little difference, and just creates another class. I'd tend to keep it this way meanwhile, in order not to overload the review with even more changes.

SimplenoteUITests/SimplenoteUITests.swift
pachlava: I'm using both methods because there's stuff that should be done only once for a pack of tests (e.g. log in under a certain account), and stuff that should be done before each test (like remove all notes and empty trash).

I tried moving `continueAfterFailure` to `setUp()` earlier, but this caused the error:

`Instance member 'continueAfterFailure' cannot be used on type 'SimplenoteUISmokeTestsNoteEditor'`

I did not look deep into it before, but decided to google a bit after your review, and [found](https://stackoverflow.com/a/62015300) that the error will not be shown if I change `override class func setUp()` to `override func setUp()`. But this change will actually cause all tests to fail...

As for removing `app.launch()` from `setUpWithError()`: I'd really like to do this, and tried this before, but about half of tests start failing after this. I need to look in detail why this happens, because this will save lots of execution time.

SimplenoteUITests/AllNotesClass.swift
pachlava: Thank you! I never used Swift before, and was not aware of `guard` statement. Applied it there and in a few other places.

It looks like `!isOpen()` still should be used in guard statement.

.bundle/config
pachlava: Thank you for the explanation :-)

Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme
pachlava: Done. Thanks!

Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteScreenshots.xcscheme
pachlava: Also done, thank you! :)

SimplenoteUITests/Playground.swift
pachlava: Removed this file 🗑️

SimplenoteUITests/AllNotesClass.swift
pachlava: Removed them, feels better without them 😄

SimplenoteUITests/SimplenoteUITests.swift
pachlava: You're right, this currently serves no purpose, I kept it in order not to forget that this can be used for the future. Removed.

SimplenoteUITests/SimplenoteUITests.swift
pachlava: Agree, removed, thanks!

SimplenoteUITests/SimplenoteUITests.swift
pachlava: Went to 🗑️

TESTING-CHECKLIST.md
pachlava: Done, thank you :)

Simplenote.xcodeproj/project.pbxproj
pachlava: Renamed the files, thanks!

SimplenoteUITests/AllNotesClass.swift
pachlava: Fixed sleeps and infinite waits are definitely not the way to go. Again, this is something I planned to improve, but kept like that during trial, because of hurry, and forgot about this. Your solution is much better than the one I would probably create: just a finite loop that will stop once the statement is fulfilled or timeout is reached. For the moment I merged the code you proposed into one function, because it seems to me that it will be the only scenario of its usage:

```
    class func waitForLoad() {
        let allNotesNavBar = app.navigationBars[uidNavBar_AllNotes]
        let predicate = NSPredicate { _, _ in
            allNotesNavBar.staticTexts[uidText_AllNotes_InProgress].exists
        }
        
        let expectation = XCTNSPredicateExpectation(predicate: predicate, object: .none)
        XCTWaiter().wait(for: [expectation], timeout: 30)
    }
```

Thank you!

pachlava:
@mokagio Huuuge thanks for your review, I've implemented everything you proposed, apart from two last points, with explanations why. Thanks for helping to make my code cleaner, it still has many space for improvement.

For some reason the failing tests that you mention never failed for me, but instead another one might fail - the new `testUsingInsertChecklistInsertsChecklist`, usually when it's executed as a part of the whole suite. But it doesn't fail in separate 😅 .

My latest changes include the ones coming from your review, but as well some new code for another automated test - sorry for the size. Hopefully next times my changes will be more atomic.

SimplenoteUITests/Generic.swift
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func logOut() -> Bool {
+    app.navigationBars[uidNavBar_AllNotes].buttons[uidButton_Menu].tap()
+    app.tables.staticTexts[uidCell_Settings].tap()
+    app.tables.staticTexts[uidButton_Settings_LogOut].tap()
+    return app.buttons[uidButton_LogIn].waitForExistence(timeout: maxLoadTimeout)
+}
+
+public func trackTest(_ function: String = #function) {
```
mokagio: Does this need to be `public`?

```suggestion
func trackTest(_ function: String = #function) {
```

SimplenoteUITests/Generic.swift
```diff
@@ -0,0 +1,80 @@
+import XCTest
+
+var stepIndex = 0
```
mokagio: Nitpick: this could be `private` so the only way for consumer code to increment the counter is to call `logStep()`

```suggestion
private var stepIndex = 0
```

SimplenoteUITests/AllNotes.swift
```diff
@@ -0,0 +1,88 @@
...
+    }
+    
+    class func waitForLoad() {
+        let allNotesNavBar = app.navigationBars[uidNavBar_AllNotes]
+        let predicate = NSPredicate { _, _ in
+            allNotesNavBar.staticTexts[uidText_AllNotes_InProgress].exists == false
+        }
+        
+        let expectation = XCTNSPredicateExpectation(predicate: predicate, object: .none)
+        XCTWaiter().wait(for: [expectation], timeout: 10)
```
mokagio, line 67: Nice touch using `XCTWaiter` here 👍 I hadn't considered this option in my original suggestion.

SimplenoteUITests/SimplenoteUITests.swift
```diff
@@ -0,0 +1,579 @@
...
+        let noteTextInitial = "Inserting checkbox with a button",
+            noteNameInitial = noteTextInitial,
+            noteTextWithCheckbox = " " + noteTextInitial,
+            noteNameWithCheckbox = "- [ ]" + noteTextWithCheckbox
+        
+        trackTest()
+
+        logStep()
+        AllNotes.addNoteTap()
+        NoteEditorAssert.editorShown()
```
mokagio: Looking at this and the series of `logStep()` calls below, it might be that my suggestion above regarding [grouping tests into sub-steps with activities](https://developer.apple.com/documentation/xctest/activities_and_attachments/grouping_tests_into_substeps_with_activities) might actually become applicable sooner rather than later 😄 

I wouldn't worry about it in the context of this PR, though. Better to get these big changes in first, and then improve upon them later.

SimplenoteUITests/Generic.swift
```diff
@@ -0,0 +1,80 @@
...
+    app.tables.staticTexts[uidButton_Settings_LogOut].tap()
+    return app.buttons[uidButton_LogIn].waitForExistence(timeout: maxLoadTimeout)
+}
+
+public func trackTest(_ function: String = #function) {
+    print("> Test: \(function)")
+    stepIndex = 1
+}
+
+func logStep(){
```
mokagio: Nitpick for consistent naming with `trackTest` above:

```suggestion
func trackStep() {
```

SimplenoteUITests/Generic.swift
```diff
@@ -0,0 +1,80 @@
...
+    print(">> Step " + String(stepIndex))
+    stepIndex += 1
+}
```
mokagio, line 33: Semi-related to these two functions to log the start of a test and its steps in the console: I recently learned about [`XCTActivity` and `XCTGroup`](https://developer.apple.com/documentation/xctest/activities_and_attachments/grouping_tests_into_substeps_with_activities) which sounds like it might be useful to us:

> Use activities to break longer test methods, such as UI tests and integration tests, into smaller named substeps. For example, break a complex UI login test into three substeps named "Open login window", "Enter password", and "Close login window", each of which contains the checks and assertions appropriate to that substep.

mokagio (approved):
Thank you for taking my suggestions into account. 

I run the tests and they mostly passed. I was pleased by the fact that even when one test fails, the rest of the suite keeps working. That's something that doesn't always happen, so kudos for designing the tests in a resilient way 👏 .

Speaking of resilience. By chance, I run the tests when the instance of the app on the Simulator was logged in with an account that hadn't confirmed the email and they all failed because of the verification screen being in the way. If I recall correctly, the flow is still a bit in flux and there are plans to make the screen non-dismissible, so it might not be worth adding code to take that into account. Just wanted to call that to your attention.

I also noticed there is no shared scheme for running the UI tests. I took the liberty of adding one and pushed the commit b8e3fd01.

If you want, you can add that change to your branch via

```
git fetch && git cherry-pick b8e3fd01
```

For reference, this is how I added the scheme. Notice the _Shared_ checkbox in the scheme settings at the end.


https://user-images.githubusercontent.com/1218433/108304634-8ad76000-71fc-11eb-90e8-533089e41ed3.mov

Let's merge this in 💪 Looking forward to new granular PRs with one new test each 🤖 🤖 🤖

SimplenoteUITests/Generic.swift
pachlava: Thank you! Applied

SimplenoteUITests/Generic.swift
pachlava: Thanks! Commited suggestion

SimplenoteUITests/SimplenoteUITests.swift
pachlava: Thanks! Will read this

pachlava:
@mokagio 👋 Thanks again for your feedback, especially for the detailed video: I cherry-picked your commit, and it's in this branch now. Also, I applied your suggestions on access control for methods and variables and renaming of methods.

The issue with verification screen is something I remember about, but did not work on it yet. There's still a lot of space for improvement, even with this amount of automated tests...

PR #1125, opened 2021-02-03
Refactor SimplenoteActivityItemSource

charliescheer:
### Fix
This is a quick PR to refactor the setup for SimplenoteActivityItemSource.  Previously the SimplenoteActivityItemSource  required a note to be passed in at initialization, this was only required so that the FileManager could access the simperiumID when writing the note content to a file.

In this PR I have changed it so that SimplenoteActivityItemSource accepts text and a filename for the destination file, decoupling SimplenoteActivityItemSource from FileaManager and Note

### Test
Should work the same as before, but just to make sure
1. run unit tests
2. open up a note and go to Info > Share and try a few different options other than Books and print
3. Make sure to share to Wordpress, there is custom logic in SimplenoteActivityItemSource for that

### Review
> Only one developer s required to review these changes, but anyone can perform the review.

### Release

### DIFF
--- a/Simplenote/Classes/FileManager+Simplenote.swift
+++ b/Simplenote/Classes/FileManager+Simplenote.swift
@@ -14,19 +14,12 @@ extension FileManager {
 
         return url
     }
-
-    /// Writes a given Note to the Documents folder
+    
+    /// Writes a given String to the documents folder
     ///
-    class func writeToDocuments(note: Note) -> URL? {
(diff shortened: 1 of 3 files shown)


charliescheer:
Note that one additional change we may want to make here is change from saving the .txt file to the documents folder to saving to FileManager.default.temporaryDirectory instead

charliescheer:
closing in favor of #1136

PR #1127, opened 2021-02-05
Add Fastlane API Key

jkmassel:
### Fix
Fixes code signing when shipping 4.31 to the App Store.

### Test
- Update the `.user.env` file, removing the `FASTLANE_USER` and `DELIVER_USER` keys, and replacing them with `PILOT_API_KEY_PATH=".configure-files/app_store_connect_fastlane_api_key.json"`. Try building for App Store Connect locally.
- Note that this change will need to be made in CircleCI environment variables prior to cutting a build, otherwise it will fail.


mokagio (approved):
The change looks good.

I'm struggling to test this locally. I'm not sure whether the Simplenote iOS setup is different, out of date, or if maybe there's something messed up in my local machine.

Anyways, I'll merge these changes and verify if they work from CI. Any fix would have to do with the Simplenote iOS setup rather than the addition of this file, so I feel okay at making it in a possible dedicated PR later on.

PR #1133, opened 2021-02-05
Moving tags editors from text view to view controller

eshurakov:
### Details
In this PR we're moving tags editor from text view to view controller in order to simplify interaction between different components and leverage auto layout.

There should be no changes in functionality.

### Test
1. Open a note

- [x] Tags are visible on screen if the content size is smaller than screen size
- [x] Tags are not visible if content size is bigger than screen size
- [x] As you scroll and content comes closer to the tags, tags go offscreen
- [x] It is possible to add and delete tags

### Test (Voiceover)

- [x] When VoiceOver is running tags are always on screen

### Review


Simplenote/Classes/SPEditorTextView.h
```diff
@@ -16,15 +15,14 @@
...
-@property (nonatomic) BOOL lockTagEditorPosition;
-
-@property (nonatomic, strong) SPTagView *tagView;
 @property (nonatomic, strong) UIFont *checklistsFont;
 @property (nonatomic, strong) UIColor *checklistsTintColor;
 
 @property (nonatomic, readonly) BOOL isInserting;
 @property (nonatomic, readonly) BOOL isDeletingBackward;
 
+@property (nonatomic, strong) void (^onContentPositionChange)(void);
```
jleandroperez, line 24: Trying to remember, I think this should be `copy` rather than `strong` ? In old ObjC jargon?

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -52,22 +52,12 @@ extension SPNoteEditorViewController {
...
     }
 
-    /// Sets up the Bottom View:
-    /// - Note: This helper view covers the area between the bottom edge of the screen, and the safeArea's bottom
-    ///
-    @objc
-    func configureBottomView() {
-        bottomView = UIView()
-        bottomView.isHidden = true
-    }
```
jleandroperez, line 62: Thank you so much 😄

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -705,14 +703,72 @@ extension SPNoteEditorViewController: InterlinkProcessorDelegate {
...
+
+        NSLayoutConstraint.activate([
+            tagView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            tagView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
+        ])
+
+        tagListBottomConstraint = tagView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+        tagListBottomConstraint.isActive = true
+
+        tagListViewController.didMove(toParent: self)
```
jleandroperez, line 725: Not sure if it's 100% needed, but missing call to `willMove(toParent:)`

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -705,14 +703,72 @@ extension SPNoteEditorViewController: InterlinkProcessorDelegate {
...
+    private func updateTagListPosition() {
+        guard !isVoiceOverEnabled else {
+            tagListViewController.view.transform = .identity
+            return
+        }
+
+        let contentHeight = noteEditorTextView.contentSize.height - noteEditorTextView.textContainerInset.bottom
+        let maxContentY = noteEditorTextView.convert(CGPoint(x: 0, y: contentHeight), to: view).y
+
+        tagListViewController.view.transform = .identity
```
jleandroperez: Can we get rid of this line?

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -705,14 +703,72 @@ extension SPNoteEditorViewController: InterlinkProcessorDelegate {
...
 
 
+// MARK: - Tags
+//
+extension SPNoteEditorViewController {
+    @objc
+    func configureTagListViewController() {
+        tagListViewController = NoteEditorTagListViewController(note: note)
+        addChild(tagListViewController)
+        let tagView = tagListViewController.view!
```
jleandroperez: I've noticed `tagListViewController.view` is used in this API + `updateTagListPosition` + `refreshTagsEditor`.

Maybe something like this would be neat?

```
private var tagView: UIView {
    tagListViewController.view!
}
```

Simplenote/Tags/NoteEditorTagListViewController.swift
```diff
@@ -0,0 +1,143 @@
...
+        didSet {
+            tagView.tagDelegate = self
+            tagView.backgroundColor = .clear
+            tagView.keyboardAppearance = .simplenoteKeyboardAppearance
+        }
+    }
+    @IBOutlet private weak var tagViewTopConstraint: NSLayoutConstraint!
+
+    private let note: Note
+    private let objectManager = SPObjectManager.shared()
```
jleandroperez, line 25: Eventually we'll 🔥 this!!

jleandroperez (commented):
@eshurakov I LOVE this refactor!!

I found a tiny glitch:

1. Add a new Note
2. Press anywhere to reveal the keyboard
3. Swipe downwards to dismiss (Interactive Keyboard Dismiss)

As a result, the Tags List will remain in place, until you lift your thumb.

jleandroperez (approved):
My bad, I've missed the fact that the Tags Editor + Interactive Dismiss exhibits the same glitch in `develop`.

:shipit: when ready sir!!

Simplenote/Classes/SPEditorTextView.h
eshurakov: At first blocks had to be declared as `copy`, but with ARC and modern `clang` `strong` is just fine :-)

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: 🔥

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: When you add a child view controller, `willMove(toParent:)` is not needed. 
Ref: https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: We can not. We need to reset transform to get original frame without any modifications so that we can correctly calculate intersection with editor content on the next line :-)

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
eshurakov: Updated in https://github.com/Automattic/simplenote-ios/pull/1133/commits/5b14794fc11281504bf5f2161eb176ffcb58c11e
Thanks 👍

eshurakov:
Thank you for the review!

pachlava:

PR #1135, opened 2021-02-08
Exclude share to books from UIActivityViewController

charliescheer:
### Fix
This PR responds to issue #517 where sharing from Simplenote using iOS sharing there is an option to share to Apple Books, when selected a message appears saying a PDF is being generated, but it fails silently and no PDF is generated

I spent some time on this, ultimately decided the best way forward is to exclude sharing to books as an option in the share sheet.  This PR removes sharing to books as an option

### Test
To Test:
1. Open a note and go to Options > Share
2. look for books, it should not be one of the options available under apps

### Release

### DIFF
--- a/Simplenote/UIActivityViewController+Simplenote.swift
+++ b/Simplenote/UIActivityViewController+Simplenote.swift
@@ -18,5 +18,9 @@ extension UIActivityViewController {
         let source = SimplenoteActivityItemSource(note: note)
 
         self.init(activityItems: [print, source], applicationActivities: nil)
+        
+        excludedActivityTypes = [
+            UIActivity.ActivityType.openInIBooks
+        ]
     }
 }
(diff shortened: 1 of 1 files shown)


Simplenote/UIActivityViewController+Simplenote.swift
```diff
@@ -18,5 +18,9 @@ extension UIActivityViewController {
         let source = SimplenoteActivityItemSource(note: note)
 
         self.init(activityItems: [print, source], applicationActivities: nil)
+        
+        excludedActivityTypes = [
```
jleandroperez, line 24: Mind adding a comment detailing why this is done? (For future reference!!)

jleandroperez (approved):
:shipit:

pachlava:

PR #1136, opened 2021-02-09
Refactor SimplenoteActivityItemSource

charliescheer:
### Fix
This is a quick PR to refactor the setup for SimplenoteActivityItemSource.  Previously the SimplenoteActivityItemSource  required a note to be passed in at initialization, this was only required so that the FileManager could access the simperiumID when writing the note content to a file.

In this PR I have changed it so that SimplenoteActivityItemSource accepts text and a filename for the destination file, decoupling SimplenoteActivityItemSource from FileManager and Note.  

Also, to WPiOS, FileManager will use a temporary directory instead of the uses documents directory

### Test
Should work the same as before, but just to make sure
1. run unit tests
2. open up a note and go to Info > Share and try a few different options other than Books and print
3. Make sure to share to Wordpress and save draft to Wordpress, there is custom logic in SimplenoteActivityItemSource for that

### Review
> Only one developer s required to review these changes, but anyone can perform the review.

### Release


Simplenote/Classes/FileManager+Simplenote.swift
```diff
@@ -14,19 +14,18 @@ extension FileManager {
...
-        guard let payload = note.content else {
-            return nil
-        }
-
-        let filename = String(format: "%@.txt", note.simperiumKey)
-        let targetURL = FileManager.documentsURL.appendingPathComponent(filename)
-
+    class var temporaryDirectoryURL: URL {
+        return FileManager.default.temporaryDirectory
+    }
```
jleandroperez: No need to wrap access to the `temporaryDirectory` (since it's accessible from the SDK singleton)

Simplenote/Classes/FileManager+Simplenote.swift
```diff
@@ -14,19 +14,18 @@ extension FileManager {
...
+    
+    /// Writes a given String to the documents folder
+    ///
+    class func writeStringToURL(string: String, to targetURL: URL) -> URL? {
```
jleandroperez: This API doesn't use `FileManager` (and it was originally misplaced).

It does interact with a String + URL, how about if we move this one to the spot where it's actually used? (`SimplenoteActivityItemSource`)

Simplenote/UIActivityViewController+Simplenote.swift
```diff
@@ -13,9 +13,10 @@ extension UIActivityViewController {
         guard let content = note.content else {
             return nil
         }
+        let shareFilename = note.exportFilename()
 
         let print = SPSimpleTextPrintFormatter(text: content)
-        let source = SimplenoteActivityItemSource(note: note)
+        let source = SimplenoteActivityItemSource(content: content, filename: String(format: "%@.txt", shareFilename))
```
jleandroperez: (From the outside, as a reader) it's not clear why an `ActivityItemSource` would need a filename, nor why it would need a `.txt` file extension.

How about if, instead, we pass along the document Unique ID, and deal with filenames / extensions in `SimplenoteActivityItemSource` itself?

```
SimplenoteActivityItemSource(content: content, identifier: note.simperiumKey)
```

(Otherwise the API is "leaking" internal details).

jleandroperez (commented):
Looking good! sending you few nipticky comments, thanks in advance!!

charliescheer:
Okay @jleandroperez I have updated the branch to reflect your changes.  Let me know what you think!

jleandroperez (approved):
Thank you for doing this @charliescheer !!

:shipit:

pachlava:
Tested sharing to WP, Notes and Messages

PR #1137, opened 2021-02-09
Persist scroll positions to a local file

eshurakov:
### Details
In this PR we're persisting scroll position of notes to a local file in documents folder. To keep the file size small we clean it up by removing deleted notes when the app goes to background.

Ref: #1092

### Test
1. Open any note and scroll, remember where you scrolled.
2. Repeat first step a few times
3. Kill the app and open it again

- [x] Check that notes that you scrolled before open exactly at the same position

### Review

### DIFF
--- a/Simplenote/Classes/EditorFactory.swift
+++ b/Simplenote/Classes/EditorFactory.swift
@@ -16,7 +16,10 @@ class EditorFactory: NSObject {
 
     /// Scroll position cache
     ///
-    private let scrollPositionCache = NoteScrollPositionCache()
+    let scrollPositionCache: NoteScrollPositionCache = {
+        let fileURL = FileManager.documentsURL.appendingPathComponent(Constants.filename)
+        return NoteScrollPositionCache(storage: FileStorage(fileURL: fileURL))
+    }()
 
     /// You shall not pass!
     ///
@@ -43,4 +46,11 @@ class EditorFactory: NSObject {
         }
         return note
     }
+}
+
+
+// MARK: - Constants
+//
+private struct Constants {
+    static let filename = ".scroll-cache"
 }
--- a/Simplenote/Classes/FileStorage.swift
+++ b/Simplenote/Classes/FileStorage.swift
@@ -1,0 +1,28 @@
+import Foundation
+
+// MARK: - FileStorage
+//
+class FileStorage<T: Codable> {
+    private let fileURL: URL
+
+    private lazy var decoder = JSONDecoder()
+    private lazy var encoder = JSONEncoder()
+
+    init(fileURL: URL) {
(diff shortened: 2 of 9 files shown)


Simplenote/Classes/EditorFactory.swift
```diff
@@ -44,3 +47,10 @@ class EditorFactory: NSObject {
         return note
     }
 }
+
+
+// MARK: - Constants
+//
+private struct Constants {
+    static let filename = ".scroll-cache"
```
eshurakov, line 55: WDYT about filename? 🤔

jleandroperez: I think looks good!!

jleandroperez: (Plus I love the `.` prefix detail!!)

Simplenote/Classes/FileStorage.swift
```diff
@@ -0,0 +1,28 @@
...
+        return try decoder.decode(T.self, from: data)
+    }
+
+    /// Save an object
+    ///
+    func save(object: T) throws {
+        let data = try encoder.encode(object)
+        try data.write(to: fileURL)
+    }
+}
```
jleandroperez, line 28: Hey, this is neat!

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -355,3 +355,21 @@ private extension SPAppDelegate {
...
+    @objc
+    func cleanupScrollPositionCache() {
+        let allNotes = SPObjectManager.shared().notes()
+        let allIdentifiers: [String] = allNotes.compactMap {
+            guard let note = $0 as? Note, !note.deleted else {
+                return nil
+            }
+
+            return note.simperiumKey
+        }
```
jleandroperez, line 368: Suggestion!!

1. Edit `SPObjectManager.h` and set the collection types:
```
- (NSArray<Note *> *)notes;
- (NSArray<Tag *> *)tags;
```

2. Then you can do something like this!
```
let allIdentifiers: [String] = allNotes.compactMap { note in
   note.deleted ? nil : note.simperiumKey
}
```

jleandroperez (approved):
Solid!! Sent you a tiny suggestion, other than that :shipit: !!

charliescheer (commented):
Looks good to me.

Simplenote/SPAppDelegate+Extensions.swift
eshurakov: Great suggestion. Love it.

eshurakov:
Thank you!

pachlava:

PR #1138, opened 2021-02-09
Remove Trailing Comma on Tag

### LINKED ISSUE #1120: Tags: Use comma as a delimiter when adding tags to a note
Currently, Web/Electron and native MacOS use the comma character as a delimiter when adding tags to a note. 
Based on some internal discussion p2XJRt-2Fi-p2 iOS should work the same way.

### What
When adding tags to a note inputting the comma character should submit the tag and be used as a delimiter instead of being part of the tag.

### Why
This would help be consistent across the other platforms and follow other conventions.

illusaen:
Fixes [#1120](https://github.com/Automattic/simplenote-ios/issues/1120)
### Fix
When entering tags and separating via `, `, tags added had a trailing comma. Removing that comma to be on par with other platforms.

### Test
1. Go to note.
2. Tap on tag field.
3. Enter `tag, `.
4. See that tag is entered as `tag` instead of `tag,`.
5. Check that tags list has tag as `tag`, not `tag,`.

### Screenshots
Feb-17-2021 14-37-55

<details>
<summary>Old screenshots</summary>

#### Trailing comma removed:
tagbugged

#### Mid-word comma still present:
tagbug-mid
</details>

### Review

### Release
`RELEASE-NOTES.txt` was updated in 63c887e5 with: 
`Removing trailing comma from tags when used as delimiter #1120`


Simplenote/Classes/SPTagView.m
```diff
@@ -357,8 +356,9 @@ - (void)processTextInFieldToTag {
 
     if (name.length > 0 && [self.tagDelegate tagView:self shouldCreateTagName:name] && !containsEmailAddress) {
         
-        [self newTagPillWithString:name];
-        [self.tagDelegate tagView:self didCreateTagName:name];
+        NSString *cleanedName = [name stringByTrimmingCharactersInSet:[NSCharacterSet punctuationCharacterSet]];
```
dmsnell: as far as I know this is overly constrictive and will remove tags with legitimate characters. so far we have only disallowed spaces and the comma

eshurakov (changes requested):
Hi @illusaen, sending you a few comments.

We want to use comma in a similar way that space works now. When you press space or return the tag is created, so pressing comma should do the same.

The best way to achieve this is to update `TagTextFieldInputValidator` and include comma in the character set of disallowed characters. 

Also note that it is also possible to rename a tag from the side bar. The same logic should be applied there as well.

illusaen:
> Hi @illusaen, sending you a few comments.
> 
> We want to use comma in a similar way that space works now. When you press space or return the tag is created, so pressing comma should do the same.
> 
> The best way to achieve this is to update `TagTextFieldInputValidator` and include comma in the character set of disallowed characters.
> 
> Also note that it is also possible to rename a tag from the side bar. The same logic should be applied there as well.

Hi @eshurakov! Thanks for the tips! I see what you mean, I didn't realize that that was how it should work.

By the way, also replying to @dmsnell in this: I used `punctuationCharacterSet` here because I didn't really think that people would/should? use things like `?/)/etc` in their tags.

illusaen:
@eshurakov Okay, updated `TagTextFieldInputValidator` and removed the previous change :)

eshurakov (commented):
Thank you!

Do you mind adding a few tests to `TagTextFieldInputValidatorTests`?

Sorry I didn't mention it earlier, but we also need to fix pasting text that contains comma. There is a `preprocessForPasting` method in `TagTextFieldInputValidator`. I think it makes sense to extract disallowed character set to a variable and use it from both `validateInput` and `preprocessForPasting`.

Another note is now that we extended our disallowed set with comma it would make sense to update methods and variables that use `whitespace` in their name to better reflect what's going on. For example we can rename `Result.endingWithWhitespace` to maybe `Result.endingWithDisallowedCharacter`. Suggestions are welcome :-)

illusaen:
> Thank you!
> 
> Do you mind adding a few tests to `TagTextFieldInputValidatorTests`?
> 
> Sorry I didn't mention it earlier, but we also need to fix pasting text that contains comma. There is a `preprocessForPasting` method in `TagTextFieldInputValidator`. I think it makes sense to extract disallowed character set to a variable and use it from both `validateInput` and `preprocessForPasting`.
> 
> Another note is now that we extended our disallowed set with comma it would make sense to update methods and variables that use `whitespace` in their name to better reflect what's going on. For example we can rename `Result.endingWithWhitespace` to maybe `Result.endingWithDisallowedCharacter`. Suggestions are welcome :-)

Good suggestions all! Done, done, and done :)

eshurakov (approved):
Works great!

:shipit:

pachlava:

PR #1139, opened 2021-02-09
Issue/1036 text hidden by share keyboard

charliescheer:
### Fix
This PR fixes issue #1036 where it is possible while sharing to Simplenote from another app, the text can get hidden behind the keyboard if the text is too long.

In this PR, I have made a few changes:
changed the styling of the share sheet to look more like a modal (removed the side padding) 
changed the the reaction to the keyboard so that insets are added to the textView instead of making the textView smaller
Removed the keyboard observing from ExtensionPresentationController in favor of ShareViewController
set logic for the insets to be added correctly on iPad and iPhone

### Test
Please make sure to test on an an iPad and an iPhone as they are handled slightly differently.

1. Before running, choose the ShareSheet target
2. In the app that is used for sharing, select Share > Simplenote
3. Once the share sheet loads paste in enough content to have more text than can fit on the page
4. swipe down to make sure that the bottom of the text can be view with the keyboard open or closed

### Review

### Release


SimplenoteShare/ViewControllers/ShareViewController.swift
```diff
@@ -90,6 +99,51 @@ class ShareViewController: UIViewController {
     }
 }
 
+extension ShareViewController: KeyboardObservable {
+    func keyboardWillChangeFrame(beginFrame: CGRect?, endFrame: CGRect?, animationDuration: TimeInterval?, animationCurve: UInt?) {
+        guard let beginFrame = beginFrame,
+              let endFrame = endFrame else {
+            return
+        }
+        //BeginFrame y will be larger IF the keyboard was hidden
```
jleandroperez: Nipticky: Space between `//` and the text (for readability purposes!)

SimplenoteShare/ViewControllers/ShareViewController.swift
```diff
@@ -90,6 +99,51 @@ class ShareViewController: UIViewController {
...
+        return UIEdgeInsets(top: 0, left: 0, bottom: keyboardFrame.height, right: 0)
+    }
+    
+    func keyboardDidChangeFrame(beginFrame: CGRect?, endFrame: CGRect?, animationDuration: TimeInterval?, animationCurve: UInt?) {
+        guard let beginFrame = beginFrame,
+              let endFrame = endFrame else {
+            return
+        }
+        
+        //BeginFrame y will be larger IF the keyboard was hidden
```
jleandroperez: Nipticky: Space between `//` + Text (for readability purposes!)

SimplenoteShare/ViewControllers/ShareViewController.swift
```diff
@@ -90,6 +99,51 @@ class ShareViewController: UIViewController {
...
+        if beginFrame.origin.y < endFrame.origin.y {
+            print(beginFrame)
+            print(endFrame)
+            UIView.animate(withDuration: Constants.animationTimeInterval) {
+                var insets = UIEdgeInsets()
```
jleandroperez: This assignment is never really read (always overwritten below)

SimplenoteShare/ViewControllers/ShareViewController.swift
```diff
@@ -90,6 +99,51 @@ class ShareViewController: UIViewController {
...
+    func keyboardDidChangeFrame(beginFrame: CGRect?, endFrame: CGRect?, animationDuration: TimeInterval?, animationCurve: UInt?) {
+        guard let beginFrame = beginFrame,
+              let endFrame = endFrame else {
+            return
+        }
+        
+        //BeginFrame y will be larger IF the keyboard was hidden
+        if beginFrame.origin.y < endFrame.origin.y {
+            print(beginFrame)
+            print(endFrame)
```
jleandroperez: Let's drop the print statements 🔥

SimplenoteShare/ViewControllers/ShareViewController.swift
```diff
@@ -90,6 +99,51 @@ class ShareViewController: UIViewController {
...
 }
 
+extension ShareViewController: KeyboardObservable {
+    func keyboardWillChangeFrame(beginFrame: CGRect?, endFrame: CGRect?, animationDuration: TimeInterval?, animationCurve: UInt?) {
+        guard let beginFrame = beginFrame,
+              let endFrame = endFrame else {
+            return
+        }
+        //BeginFrame y will be larger IF the keyboard was hidden
+        if beginFrame.origin.y > endFrame.origin.y {
```
jleandroperez: You **probably** always want to update the insets. Otherwise OPs such as hiding the keyboard won't update the Insets

SimplenoteShare/ViewControllers/ShareViewController.swift
```diff
@@ -90,6 +99,51 @@ class ShareViewController: UIViewController {
...
+                var insets = UIEdgeInsets()
+                
+                if UIDevice.isPad {
+                    insets = self.makeEdgeInsets(from: endFrame)
+                } else {
+                    insets = Constants.textViewInsets
+                }
+                
+                self.textView.contentInset = insets
+                self.textView.scrollIndicatorInsets = insets
```
jleandroperez: You only really really need to modify the **.bottom** insets, always (since that's the overlapping surface with the keyboard). That means that:

1. No need for an `insets` helper var
2. You can also drop the `makeEdgeInsets` API

```
let bottomInset = UIDevice.isPad ? endFrame.height : Constants.textViewInsets.bottom
self.textView.contentInset.bottom = bottomInset
self.textView.scrollIndicatorInsets.bottom = bottomInset
```

(Same note for `keyboardWillChangeFrame`, you can drop the makeEdgeInsets API)

jleandroperez (changes requested):
Few Style-Y notes, plus:

### iPad: Text covered by Autocomplete Bar
On iPad Devices (9.7 Inch / iOS 14.4) the last line of text may get covered by the keyboard:



### Top Location

The Top Location appears to be a bit off. Specially noticeable when in landscape. This is how it looks like, when compared to WordPress iOS's Share Ext:

Simplenote|WordPress iOS
-|-
|

charliescheer:
@jleandroperez can you take another look at this one? I made a change to how the insets are being calculated that seems to be more stable, but does rely on getting the screen size from UIScreen.  In the back of my mind I remember you saying that wasn't the best practice, but can't find where that comment happened to see more on why.

Is there a better way to get the screen size other than UIScreen.main.bounds that we could use instead?

I think this new calculation takes care of the cases of the insets not appearing correctly on portrait mode.  But let me know what you find.  ~Be aware that to compile this new version you will need to change the version of Simplenote Search to 1.2.0~

charliescheer:
@jleandroperez just a quick update, per our chat earlier, I merged develop back into this branch to deal with the conflict I was having previously with the update to SimplenoteSearch.  You don't need to downgrade the version of the package now.

charliescheer:
@jleandroperez Another update that deals with the floating keyboard issue.  Tested and it works, thanks for the suggestion on a potential solution!  Worked great.

When you get a chance this is ready to be reviewed again

Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -6,8 +6,8 @@
         "repositoryURL": "git@github.com:Automattic/SimplenoteFoundation-Swift.git",
         "state": {
           "branch": null,
-          "revision": "2731e4d28c42d394ddc62cd864d9f7ff96759228",
-          "version": "1.3.0"
+          "revision": "b21479caeffbff6d544d0aeeec8e627acd8e350c",
+          "version": "1.2.0"
```
jleandroperez: This one should remain 1.3.0, can we revert this change?

SimplenoteShare/ViewControllers/ShareViewController.swift
```diff
@@ -90,6 +99,26 @@ class ShareViewController: UIViewController {
...
+        let newEdgeInset = newKeyboardFloats ? .zero : offset
+
+        UIView.animate(withDuration: Constants.animationTimeInterval) {
+            self.textView.contentInset.bottom = newEdgeInset
+            self.textView.scrollIndicatorInsets.bottom = newEdgeInset
+        }
+    }
+
+    func keyboardDidChangeFrame(beginFrame: CGRect?, endFrame: CGRect?, animationDuration: TimeInterval?, animationCurve: UInt?) {
+        //currently not used
```
jleandroperez: Nipticky: mind adding a spacing `//` and the text, for readability?

SimplenoteShare/ViewControllers/ShareViewController.swift
```diff
@@ -195,9 +224,17 @@ private extension ShareViewController {
     }
 }
 
+extension UIDevice {
+    static var isPad: Bool {
+        current.userInterfaceIdiom == .pad
+    }
+}
```
jleandroperez: Can we move this one to its own file?

jleandroperez (commented):
Hey there @charliescheer !!

Found a few issues with this approach, I'm afraid!!


### Issue 1: White View
1. Enable Dark Mode
2. Launch the Share Extension from Safari
3. Switch to any other app
4. Open / dismiss the Keyboard
5. Switch back to Safari

As a result, the Simplenote Share Extension will be onscreen, with the keyboard dismissed.
It would appear there's a view without the proper background set, and it's noticeable with the new layout.



### Issue 2: Top Padding
If you compare Simplenote Share with Reminders (both fullscreen), you'll notice that the gap at the top is different.

**I think** the proper way of fixing that is in these lines (`ExtensionPresentationController`):

```
    override func size(forChildContentContainer container: UIContentContainer, withParentContainerSize parentSize: CGSize) -> CGSize {
        if let target = container as? ExtensionPresentationTarget,
            target.shouldFillContentContainer == true {
            return parentSize
        }

        let safeInsets = containerView?.safeAreaInsets ?? .zero
        let extraPadding = CGFloat(10)  // << move this to a proper constant elsewhere!

        return CGSize(width: parentSize.width, height: parentSize.height - safeInsets.top - extraPadding)
    }
```



### Issue 3: Scroll Gap
The gap between the Editor and the Cursor feels a bit odd (we're not doing this in the main app).

Do you think we could drop it?

SimplenoteShare/ViewControllers/ShareViewController.swift
charliescheer: Question, we already have this isPad check built into Simplenote in UIDevice+Simplenote.swift.  I added it here because that file's targets don't include Simplenote Share.
Could we simply add SimplenoteShare to the targets for that file?

charliescheer: Actually....  I went back to try that/test it and this update isn't using the isPad bool anymore, so I can just remove it

Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
charliescheer: Yup, thought I had fixed that already.

jleandroperez:
@charliescheer please LMK whenever this one is ready for another peek! thank you!!

mokagio:
This PR is milestoned 4.33, which I'm code freezing today. It's a bug fix, so I think it should make it into the build, but it looks like there's more work to do here.

To keep the release train moving, I'll go ahead and start the code freeze without it. 

If you want to include this bug fix into a new beta, please update the base branch to `release/4.33`. Otherwise, please change the milestone to 4.34.

Let me know, thanks 🙇‍♂️

charliescheer:
@mokagio I went ahead and set the milestone to future.  I think that it is close to being fixed, but there is still some weirdness going on and I'm not sure when I will be able to circle back to it.

mokagio:
Thank you for letting me know @charliescheer 👍

PR #1140, opened 2021-02-09
Fixes an issue where changing keyboard input causes inset sizing issue

charliescheer:
### Fix
During testing for #1139 I noticed a bit of an edge case issue where a keyboard frame change notification can get triggered, and under circumstances it seems that the height returned from the notification is not correct.  This is an attempt to fix that. 

I feel like the situation is a bit of an edge case, and my sense is that this solution may introduce potential inconsistencies in how the insets in #1139 get set, so wanted to get some feedback on the option.

Setting up the edge case:
So to make this issue happen, on a physical device, choose a language that allows for the predictive text bar at the top of the keyboard. Set that as the current keyboard for your device
Go to settings > General > Keyboard and make sure predictive text is enabled
Now add a keyboard that does NOT have the predictive text keyboard (the only one that I know of is Hebrew)

First it may be easier to understand the edge case if you try the problem without this fix.
If you load #1139 follow these steps:
1. From the SimplenoteShare target, run the code and select the app to share from
2. Tap on share, select Simplenote
3. Paste into the textView that appears enough text that it runs off screen
4. Scroll down to see that the insets work correctly
5. change keyboards till you land on the non predictive text keyboard
6. Check the insets again and the insets will be incorrect, the text will run behind the keyboard

Now you are ready to test:

### Test
1. update the code to using this branch
2. From the SimplenoteShare target, run the code and select the app to share from
3. Tap on share, select Simplenote
4. Paste into the textView that appears enough text that it runs off screen
5. Scroll down to see that the insets work correctly
6. change keyboards till you land on the non predictive text keyboard
7. Check the insets again and the text should still appear correctly.

### Review

### Release

### DIFF
--- a/SimplenoteShare/ViewControllers/ShareViewController.swift
+++ b/SimplenoteShare/ViewControllers/ShareViewController.swift
@@ -59,6 +59,7 @@ class ShareViewController: UIViewController {
     ///
     private var keyboardObserverTokens: [Any] = []
 
+    private var keyboardInputChanging: Bool = false
 
     // MARK: Initialization
 
@@ -68,6 +69,7 @@ class ShareViewController: UIViewController {
         self.context = context
         super.init(nibName: type(of: self).nibName, bundle: nil)
         keyboardObserverTokens = addKeyboardObservers()
+        setInputModeObserver()
     }
 
     required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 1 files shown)


charliescheer:
Closing this PR as the changes are very edge case and the solution has the potential to not work from iOS version to iOS version

PR #1152, opened 2021-02-15
Make Interlink Popover UI reusable in tags editor

eshurakov:
### Details
In this PR I'm extracting presentation and geometry related code from `InterlinkProcessor` and `InterlinkViewController` into separate classes. It is needed so that it can be later used from the tags editor to show tags suggestions.

**How does it work now?**
`PopoverPresenter` is an object that knows how to present any view controller as a popover. It has methods to show or hide a popover. It also handles all geometry calculations and sets correct constraints for the popover.

`PopoverViewController` provides UI components, rounded corners, clipping and blur. It is a wrapper for our view controller.

`InterlinkViewController` is now a simple list view and not aware how it is presented.

Again apologies about the size of this PR 🙇 

### Test
Do a quick smoke test around interlink popover.

### Review


Simplenote/HuggableTableView.swift
```diff
@@ -5,6 +5,20 @@ import UIKit
...
+        return constraint
+    }()
+
+    lazy var minHeightConstraint: NSLayoutConstraint = {
+        let constraint = heightAnchor.constraint(greaterThanOrEqualToConstant: 0)
+        constraint.priority = UILayoutPriority(rawValue: 999)
+        constraint.isActive = true
+        return constraint
+    }()
+
```
jleandroperez, line 21: TBH I'm not 100% sure this is the **right** place for these constraints (specially because they're not initialized right here).

But considering that it's analog to `heightAnchor`, so to speak, sounds reasonable?

jleandroperez (approved):
Thank you for doing this @eshurakov !!

:shipit:

Simplenote/HuggableTableView.swift
eshurakov: I thought to keep them lazy because they're not always needed, but if a user of this class wants to set min / max height, they provide an easy way to do it.

eshurakov:
Thank you 👍

pachlava:

model | iOS | Simplenote
-|-|-
iPhone XR | 14.4 | 4.33.0.0
iPad Pro 11” Sim | 14.4 | 4.33.0.0

PR #1153, opened 2021-02-15
Fix keyboard navigation in search bar

### LINKED ISSUE #1151: [Ext. Kbd.] Right arrow can't be used for navigation in Searchbar.
### Expected
It's possible to use both left/right arrow keys to move the cursor inside a searchbar

### Observed
Right arrow key does nothing. (For RTL it's Left arrow key)

### Reproduced
1. Use external keyboard
2. Open note list
3. Focus Searchbar with Cmd+Shift+f
4. Enter any text
5. Move the cursor to the left with left arrow key ~> OK
6. Move the cursor to the right with right arrow key ~> nothing happens

-|-|-|-
iPhone + Magic Keyboard|XR|14.4|4.32.0.0

eshurakov:
### Fix
This PR:
* fixes keyboard navigation in search bar as described in #1151 
* adjusts `cmd + shift + f` so that it dismisses the search only when search bar is the first responder

Closes #1151 
Ref #1111

### Test
1. Activate search bar

- [x] Check that both left and right arrows can be used to navigate entered text
- [x] Open one of the notes and come back to the note list
- [ ] Check that pressing `cmd + shift + f` will bring focus to search bar

### Review

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -941,10 +941,11 @@ extension SPNoteListViewController {
     open override var keyCommands: [UIKeyCommand]? {
         var commands = tableCommands
         if isSearchActive {
-            commands.append(contentsOf: [
-                UIKeyCommand(input: "f", modifierFlags: [.command, .shift], action: #selector(keyboardStopSearching)),
-                UIKeyCommand(input: UIKeyCommand.inputEscape, modifierFlags: [], action: #selector(keyboardStopSearching))
-            ])
+            commands.append(UIKeyCommand(input: UIKeyCommand.inputEscape, modifierFlags: [], action: #selector(keyboardStopSearching)))
+
+            if searchBar.isFirstResponder {
+                commands.append(UIKeyCommand(input: "f", modifierFlags: [.command, .shift], action: #selector(keyboardStopSearching)))
+            }
         }
         return commands
     }
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -943,6 +943,8 @@ extension SPNoteListViewController {
         if isSearchActive {
             commands.append(UIKeyCommand(input: UIKeyCommand.inputEscape, modifierFlags: [], action: #selector(keyboardStopSearching)))
 
+            // We add this shortcut only when search bar is first responder because when it's not we don't want to clear the search.
+            // The shortcut that actually focuses on the searchbar is located in `SPSidebarContainerViewController`. This is done to make shortcut work from multiple screens
             if searchBar.isFirstResponder {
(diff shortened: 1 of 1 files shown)


jleandroperez (approved):
Works as advertised!!

:shipit:

eshurakov:
Thank you @jleandroperez

eshurakov:
👋 @mokagio, we've discovered a small issue in release 4.32 and fixed it with this PR. Could we get a new build please? Thank you!

pachlava:
@eshurakov Got my hands on 4.32.0.1 only today. It looks like the fix introduced another issue: when leaving the searchbar with Down Arrow, the focus stays in the searchbar, blocking the user from the possibility to use Left and Right arrows to open note / show sidebar:

https://user-images.githubusercontent.com/73365754/108364901-44cfcb80-71ff-11eb-9e67-112b88aa4fd3.MP4

pachlava:
[testing] fix/feature verified on iPhone XR | iOS 14.3 | Simplenote 4.32.0.1 + Magic Keyboard.
✅  Can use both arrows to navigate searchbar
✅  `cmd + shift + f ` dismisses the search only when search bar is the first responder

Related issue mentioned in comment above might be handled separately.

eshurakov:
@pachlava Thank you for testing it. 
Right now when you press down arrow search bar stays focused and active. That's why it is possible to use left and right arrow keys navigate the search bar and also to edit the search query. It works similar to autocomplete on google. It is still possible to open a note using `return` key.

Another possibility is to remove focus from the search bar once you press down arrow. This will enable the usage of arrow keys to navigate between screens. To regain focus you will have to press `cmd+shift+f` again.

Let's ask @SylvesterWilmott :-)

pachlava:
@eshurakov

> Another possibility is to remove focus from the search bar once you press down arrow. This will enable the usage of arrow keys to navigate between screens. To regain focus you will have to press cmd+shift+f again.

This is exactly how I imagined this to work, in a way similar to using no keyboard: when user taps something outside searchbar - searchbar looses focus. But let's discuss

SylvesterWilmott:
The implementation looks correct to me, but I understand why it feels incorrect now.

The keyboard shortcuts are being optimised for iPad. The issue here is that we don't yet have the split view, unfortunately. So, in the case of having the split view:

- Highlighting a note list item has the same behaviour as a tap. So whichever note is highlighted is displayed in the editor.
- If I physically **tap** a search result, focus should not be removed from the search bar because there is no particular reason for us to move the cursor elsewhere or hide the on-screen keyboard from this interaction.
- If we consider the highlighted note to be equal with a tap, then highlighting search results should not remove focus from the search bar.

The "right arrow" key does not equate to a tap, instead it changes the "active panel". That is technically true in the current state of the app but will be much a lot visually apparent when we move over to the split view.

pachlava:
Thank makes a lot of sense now, when thinking of it in the context of iPad split view. Thanks @SylvesterWilmott!

Issue #1067, opened 2021-01-04
Project: Spring Cleaning


eshurakov:
### Spring Cleaning

- [x] [Update Designs for Information Sheet](https://github.com/Automattic/simplenote-ios/issues/1061), https://github.com/Automattic/simplenote-ios/issues/1061
- [x] [Condensed Mode: Minimum Height](https://github.com/Automattic/simplenote-ios/issues/1045), https://github.com/Automattic/simplenote-ios/issues/1045
- [x] [Nü Tags Edition Mode](https://github.com/Automattic/simplenote-ios/issues/1042), https://github.com/Automattic/simplenote-ios/issues/1042
- [x] [Refresh: Tags Editor](https://github.com/Automattic/simplenote-ios/issues/990), https://github.com/Automattic/simplenote-ios/issues/990
- [x] [Nü Passcode](https://github.com/Automattic/simplenote-ios/issues/761), https://github.com/Automattic/simplenote-ios/issues/761
- [x] [Refresh: Editor Metrics](https://github.com/Automattic/simplenote-ios/issues/480), https://github.com/Automattic/simplenote-ios/issues/480
- [x] List: Update Sort Bar #1159

The following should be shipped as one feature:
- [x] Move New Note Button to the bottom toolbar in the list and editor: #1044
- [x] [List: Multiple Selection Support](https://github.com/Automattic/simplenote-ios/issues/1043), https://github.com/Automattic/simplenote-ios/issues/1043

Issues:
- [x] Tags Editor: RTL support https://github.com/Automattic/simplenote-ios/pull/1149#issuecomment-779874151 #1160
- [x] Tags Editor: Add tags field should stay on screen https://github.com/Automattic/simplenote- ios/pull/1149#issuecomment-779874151 #1160

Design project thread: p2XJRt-1dQ

charliescheer:
All of the tasks on this are finished, so I am going to close it

PR #1159, opened 2021-02-17
Update search Sort Bar

eshurakov:
### Details
In this PR we're updating search sort bar by replacing system action sheet with our custom popover. We're also showing all options combined from mode and order.

Ref: #1067 

### Screenshots
|||
|-|-|
|Simulator Screen Shot - iPhone 12 - 2021-02-17 at 13 21 38|Simulator Screen Shot - iPhone 12 - 2021-02-17 at 13 21 40|

### Test
1. Perform a search

- [ ] Popover is presented as you tap on sort mode
- [ ] Tapping outside the popover hides it
- [ ] Tapping on one of the options dismisses the popover, updates sort bar and note list
- [ ] Sort bar is visible only when there is a search query
- [ ] On big screen (iPad or iPhone in landscape) the popover is centred 

### Review

### Release
`RELEASE-NOTES.txt` was updated in ba62cea1 with:

> Updating sort bar


Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -902,32 +900,34 @@ extension SPNoteListViewController {
...
 
-        let popoverPresentationController = alertController.popoverPresentationController
-        popoverPresentationController?.sourceRect = sortBar.dividerView.frame
-        popoverPresentationController?.sourceView = sortBar.dividerView
-        popoverPresentationController?.permittedArrowDirections = .any
-        self.popoverController = popoverPresentationController
+        let viewController = SortModePickerViewController(currentSelection: Options.shared.searchSortMode)
+        viewController.onSelectionCallback = { sortMode in
+            Options.shared.searchSortMode = sortMode
+            presenter.dismiss()
```
jleandroperez, line 927: Maybe `[weak presenter]` just in case?

**EDIT:** Okay I got lost. There is **no retain cycle here** and it's all good. 

Question is who is retaining the Presenter 😂 (because I seriously can't find that reference?)

Simplenote/Classes/SortMode.swift
```diff
@@ -35,9 +35,9 @@ enum SortMode: Int, CaseIterable {
     var description: String {
         switch self {
         case .alphabeticallyAscending:
-            return NSLocalizedString("Alphabetically: A-Z", comment: "Sort Mode: Alphabetically, ascending")
+            return NSLocalizedString("Name: A-Z", comment: "Sort Mode: Alphabetically, ascending")
         case .alphabeticallyDescending:
-            return NSLocalizedString("Alphabetically: Z-A", comment: "Sort Mode: Alphabetically, descending")
+            return NSLocalizedString("Name: Z-A", comment: "Sort Mode: Alphabetically, descending")
```
jleandroperez, line 40: YES!

Simplenote/Classes/SortModePickerViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    private let data: [SortMode] = [
+        .alphabeticallyAscending,
+        .alphabeticallyDescending,
+        .createdNewest,
+        .createdOldest,
+        .modifiedNewest,
+        .modifiedOldest
+    ]
```
jleandroperez, line 23: I **think** you can just use `SortMode.allCases` (since it conforms to CaseIterable!)

jleandroperez (approved):
Looks **SUPER**!!! One question sent your way. Other than that :shipit: !!

jleandroperez:
Okay, cracked it. Answer is: **nobody** retains the Popover presenter 😂

All good sir, thank you for implementing this!!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
eshurakov: Yeah, it is tricky. But the benefit is that we don't need another property on `SPNoteListViewController` to hold on to the presenter.
`SortModePickerViewController` is retaining the presenter via the closure. This creates a retain loop presenter -> presenter view controller -> `SortModePickerViewController` -> presenter.
But once `SortModePickerViewController` is dismissed, the connection between presenter -> presenter view controller is cut and there is no longer a retain loop :-)

SylvesterWilmott:
Look great.

This component is pretty versatile!

My only feedback is that in this case the popup is needlessly wide since all the labels are quite short (in English at least). Would it be worth having a version of this component that can adjust to the content it contains?

SylvesterWilmott (approved):
After discussing, the above feedback will be addressed at a later stage.

eshurakov:
> My only feedback is that in this case the popup is needlessly wide since all the labels are quite short (in English at least). Would it be worth having a version of this component that can adjust to the content it contains?

Great suggestion @SylvesterWilmott, I created a task to track it: https://github.com/Automattic/simplenote-ios/issues/1164

Simplenote/Classes/SortModePickerViewController.swift
eshurakov: That's true, but we want to show them in this specific order. By changing order in `SortMode` the order in this picker would change as well. It is easy to forget about that, so I think keeping a separate list in `SortModePickerViewController` is safer.

pachlava:
@eshurakov only one question, and probably not directly related to this PR. When on iPad, there might be this toolbar displayed at the bottom (I have only a Simulator, with setting `I/O` -> `Keyboard` -> `Connect Hardware Keyboard` set ON, which I guess is similar to a situation of real iPad with external keyboard connected). Should we care about sort bar be covered with it?

Screenshot_2021-03-09 12 45 54_SZqAR8

pachlava:

eshurakov:
> @eshurakov only one question, and probably not directly related to this PR. When on iPad, there might be this toolbar displayed at the bottom (I have only a Simulator, with setting `I/O` -> `Keyboard` -> `Connect Hardware Keyboard` set ON, which I guess is similar to a situation of real iPad with external keyboard connected). Should we care about sort bar be covered with it?

@pachlava I think it is expected for the sort bar to be covered as this toolbar is considered to be part of the keyboard (even if you're using and external one). It can still be accessed once you press "return" in the search field and dismiss the keyboard.

pachlava:
@eshurakov  Thank you!

PR #1149, opened 2021-02-12
Tags editor, bubblified: MK2

eshurakov:
### Details
In this PR:
* Tags in the editor are shown as bubbles
* SPTagView.m  is gone and replaced with TagView.swift
* Use auto layout for tags editor
* Autocomplete is temporarily gone and will be restored in one of the next PRs.

Apologies about the size of this PR 🙇 

Ref: #990

### Screenshots

|||
|-|-|
|Simulator Screen Shot - iPhone 12 - 2021-02-15 at 14 36 11|Simulator Screen Shot - iPhone 12 - 2021-02-15 at 14 36 15|

### Test
Run a quick smoke around tags in the editor. Except the change in design and missing autocomplete the behaviour should be the same.

### Review
Only one developer and one designer are required to review these changes, but anyone can perform the review.

PR #1160, opened 2021-02-17
Tags editor, usability improvements: MK5

eshurakov:
### Fix
In this PR we're adding the following to the tags editor:
* RTL support
* Add tag field stays visible when active

Both issues were mentioned in https://github.com/Automattic/simplenote-ios/pull/1149#issuecomment-779874151

Ref: #1067

### Test
- [ ] Add tag field is always visible when you enter new tags
- [ ] When device's locale is RTL, tags are pinned to the right side and ordered right-to-left

### Review


Simplenote/Classes/TagView.swift
```diff
@@ -128,7 +128,10 @@ class TagView: UIView {
             return
         }
 
-        scrollView.scrollRectToVisible(addTagField.frame.insetBy(dx: -Constants.wrappingStackViewMargins.right, dy: 0),
+        layoutIfNeeded()
```
jleandroperez, line 131: This invocation here _feels_ off. But TBH I'm not sure where it could be otherwise!!

jleandroperez (approved):
@eshurakov thank you for fixing this!!

:shipit:

Simplenote/Classes/TagView.swift
eshurakov: Totally agree on that 🤔

eshurakov:
Thank you!

pachlava:

model | iOS | Simplenote
-|-|-
iPhone XR | 14.4 | 4.33.0.0
iPad Pro 11” Sim | 14.4 | 4.33.0.0

PR #1165, opened 2021-02-21
UIAutomation: Adjusted setUp() and setUpWithError() methods to speed up execution

pachlava:
UI Tests execution sped up by 110 seconds (~466 seconds vs ~570 seconds)

### Fix
1. Adjusted `setUp()` and `setUpWithError()` methods to avoid app launch before every test
2. For this created method that attempts returning to 'base' screen for every test: `getToAllNotes()` (for Note Editor tests)

Also (not related to speed):
3. Created several methods like `AllNotesAssert.notesExist() `on top of `AllNotesAssert.noteExists()` to allow passing an array of note names instead of calling the method N times
4. Added guard statements to several old methods

### Test
1. Running all tests and seeing no new fails introduced, with execution time being within 460 sec range.

### Review

### Release

### DIFF
--- a/Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteUITests.xcscheme
+++ b/Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteUITests.xcscheme
@@ -21,6 +21,77 @@
                BlueprintName = "SimplenoteUITests"
                ReferencedContainer = "container:Simplenote.xcodeproj">
             </BuildableReference>
+            <SelectedTests>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsLogin/testLogInWithCorrectCredentials()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsLogin/testLogInWithExistingEmailIncorrectPassword()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsLogin/testLogInWithInvalidEmail()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsLogin/testLogInWithNoEmail()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsLogin/testLogInWithNoEmailNoPassword()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsLogin/testLogInWithNoPassword()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsLogin/testLogInWithTooShortPassword()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsLogin/testLogOut()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsNoteEditor/testAddedURLIsLinkified()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsNoteEditor/testBulletedLists()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsNoteEditor/testCanFlipToEditMode()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsNoteEditor/testCanPreviewMarkdownBySwiping()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsNoteEditor/testCreateCheckedItem()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsNoteEditor/testCreateUncheckedItem()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsNoteEditor/testLongTappingOnLinkOpensLinkInNewWindow()">
+               </Test>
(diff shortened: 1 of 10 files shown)


Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteUITests.xcscheme
```diff
@@ -21,6 +21,77 @@
...
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsTrash/testCanRestoreNote()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsTrash/testCanTrashNote()">
+               </Test>
+               <Test
+                  Identifier = "SimplenoteUISmokeTestsTrash/testCanViewTrashedNotes()">
+               </Test>
+            </SelectedTests>
```
mokagio, line 94: Is this list of selected tests intentional?

SimplenoteUITests/AllNotes.swift
```diff
@@ -5,29 +5,35 @@ class AllNotes {
     class func openNote(noteName: String) {
         app.tables.cells[noteName].tap()
     }
-    
+
```
mokagio, line 8: 🎉 for removing trailing whitespaces. 👍 

Totally useless for how the code work, but sticklers like me really appreciate 🤣 🙇‍♂️

SimplenoteUITests/AllNotes.swift
```diff
@@ -5,29 +5,35 @@ class AllNotes {
...
 
-    class func createNoteAndLeaveEditor (noteName: String) {
+    class func createNoteAndLeaveEditor(noteName: String) {
+        print(">>> Creating note: " + noteName)
         AllNotes.addNoteTap()
         NoteEditor.clearAndEnterText(enteredValue: noteName)
         NoteEditor.leaveEditor()
     }
 
+    class func createNotes(noteNamesArray: Array<String>) {
```
mokagio: Nitpick. No need to address this, just wanted to point out that, while `Array<String>` is correct, `[String]` is the most common way of writing this.

`[String]` is syntax sugar for `Array<String>`.

mokagio: Another nitpick.

The [Swift API Guidelines](https://swift.org/documentation/api-design-guidelines/#weak-type-information) recommend to precede each weakly typed parameter, such as `Any`, `Int`, or `String, with a noun describing its role:

screen shot 2019-02-06 at 1 43 58 pm

In this case, `Array<String>` is clear enough that appending "array" to the name is note necessary.

What do you think of this name?

```suggestion
    class func createNotes(names: Array<String>) {
```

SimplenoteUITests/UIDs.swift
```diff
@@ -2,6 +2,7 @@
 let uidPicture_AppLogo = "simplenote-logo"
 
 let uidNavBar_AllNotes = "All Notes"
+let uidNavBar_LogIn = "Log In"
```
mokagio: I should have noticed this in the previous review but I didn't, so sorry for the out of scope suggestion. No need to address it now, just a nice-to-have tidiness idea for later.

Swift uses camelcase, there are no instances that I know of where we use `_` to separate parts of the name of a value or method. The only exception is that some testing practitioners do use `_` to separate test components in the test method name, but that's an exception and is not even a standard practice.

Anyways... What I'm trying to say is that `uidNavBar_LogIn` is not "swifty". One option could be to simply remove it and use `uidNavBarLogIn`. Another suggestion, more interesting in my opinion, would be to leverage Swift nested types feature and do this:

```swift
enum UID {

    enum NavBar {
        static let allNotes = "All Notes"
        static let logIn = "Log In"
        // ...
    }

    enum Buttons {
        static let allNotes = "All Notes"
        static let trash = "Trash"
        // ...
    }
}
```

Using an `enum` without any case means that it won't be able to instantiated (you can't do `_ = UID()`).

An extra benefit is that Xcode will help with its autocompletion:

mokagio (approved):
Pending input on the change in the scheme, this is good to go 👍 

```
Executed 23 tests, with 0 failures (0 unexpected) in 443.029 (480.719) seconds
```

image

Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteUITests.xcscheme
pachlava: This change appeared as soon as I opened the scheme and viewed the 'Tests' section. I did not change anything in the selection of tests. I guess it was initialized this way.

SimplenoteUITests/AllNotes.swift
pachlava: Thanks! Will update this during next set of changes :-)

pachlava: Thanks! Applied

SimplenoteUITests/UIDs.swift
pachlava: Thanks for bringing this up! I realized this will be discussed sooner or later. Changed this to nested type as you suggested. The UIDs file now looks much tidier 🎉

pachlava:
@mokagio thanks! I applied everything you have proposed. Actually every comment of yours made the code shorter, this I something I really like!

PR #1168, opened 2021-02-23
Merge finalized release 4.32 into develop

mokagio:


fastlane/Fastfile
```diff
@@ -638,7 +638,7 @@ end
       readonly: options[:readonly] || is_ci,
       app_identifier: ["com.codality.NotationalFlow",
                        "com.codality.NotationalFlow.Share"],
-      template_name: "NotationalFlow Keychain Access",
+      template_name: "NotationalFlow Keychain Access (Distribution)",
```
mokagio, line 641: This is something that popped up while working on the automation to regenerate the provisioning profiles. With the previous value fail, the call to `match` failed.

On the Developer Portal, the Entitlement has "(Distribution)" in the name.

fastlane/metadata/ar-SA/keywords.txt
```diff
@@ -1 +1 @@
-برنامج simplenote، والملحوظة، والتدوين، والمُدوِّن، والملحوظات، والمزامنة، والمشاركة، وبسيط، والضبط، والوسم، والمهام، والقائمة، والتأمين، والسحابة، والكتابة
\ No newline at end of file
+برنامج simplenote، والملحوظة، والتدوين، والمُدوِّن، والملحوظات، والمزامنة، والمشاركة، وبسيط، والضبط
```
mokagio, line 1: This is the usual diff due to ASC having a limit of 100 characters.

It would be good to prevent this at the root in GlotPress, but in the meantime we could build an extra step in the translation management pipeline in Fastlane to trim this for us after pulling but before committing the file?

AliSoftware (approved):
LGTM!

PR #1171, opened 2021-02-23
"Tags&Search" tests automation MK1

pachlava:
This PR starts the automation of "Tags&Search" section. PR looks huge due to splitting of file with tests into 3 smaller files, and also because of applying camelCase to files with UIDs. The number of 'real' changes is low though.

### Fix
1. Created new file `SimplenoteUITestsSearch.swift` that contains new automated `Clearing the search field immediately updates filtered notes ` and `Can search by keyword` from [smoke test](https://github.com/Automattic/simplenote-ios/blob/develop/TESTING-CHECKLIST.md) + another test for excerpts.
2. Original `SimplenoteUITests.swift` with all tests was renamed to `SimplenoteUITestsNoteEditor.swift`, with 'Trash' and 'Login' tests moved from it to new `SimplenoteUITestsTrash.swift` and `SimplenoteUITestsLogin.swift`
3. Changes in `SimplenoteUITests.xcscheme` were caused by me unknowingly deselecting `Automatically include new tests` option in the scheme. I reverted this back here.
4. Created a few new search-related methods: `AllNotes.searchForText(text)`, `AllNotes.searchCancel()`, `AllNotesAssert.noteContentIsShownInSearch(noteName, expectedContent)`
5. Applying camelCase to UIDs, which caused changes all over

### Test
1. Running the `SimplenoteUISmokeTestsSearch` tests
2. Optionally, also running all the rest, making sure new tests did not introduce issues with existing ones. 

### Review
I really appreciate your comments - thanks @mokagio!

### Release


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -402,8 +402,11 @@
...
 		D435D38FCA5863E8447D5C2C /* Pods_Automattic_SimplenoteShare.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 183BAB896957F07FDCAB6DF5 /* Pods_Automattic_SimplenoteShare.framework */; };
+		D843AD6F25E3E633007E5B15 /* SimplenoteUITestsLogin.swift in Sources */ = {isa = PBXBuildFile; fileRef = D843AD6E25E3E633007E5B15 /* SimplenoteUITestsLogin.swift */; };
+		D843AD7825E3E7F9007E5B15 /* SimplenoteUITestsTrash.swift in Sources */ = {isa = PBXBuildFile; fileRef = D843AD7725E3E7F9007E5B15 /* SimplenoteUITestsTrash.swift */; };
+		D843AD8F25E3EB16007E5B15 /* SimplenoteUITestsSearch.swift in Sources */ = {isa = PBXBuildFile; fileRef = D843AD8E25E3EB16007E5B15 /* SimplenoteUITestsSearch.swift */; };
 		D864B16925CAE47C00F9B73E /* Generic.swift in Sources */ = {isa = PBXBuildFile; fileRef = D864B16825CAE47C00F9B73E /* Generic.swift */; };
-		D864B18225CAE4BE00F9B73E /* SimplenoteUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D864B18125CAE4BE00F9B73E /* SimplenoteUITests.swift */; };
+		D864B18225CAE4BE00F9B73E /* SimplenoteUITestsNoteEditor.swift in Sources */ = {isa = PBXBuildFile; fileRef = D864B18125CAE4BE00F9B73E /* SimplenoteUITestsNoteEditor.swift */; };
```
mokagio, line 409: Good idea renaming this file and splitting out the other tests clusters in dedicated files. Makes it clearer what it's about when looking at the files list 👍

SimplenoteUITests/AllNotes.swift
```diff
@@ -61,26 +61,49 @@ class AllNotes {
...
+		guard searchField.exists else { return }
+		searchField.tap()
+
+		let clearTextButton = searchField.buttons[UID.Button.clearText]
+		if clearTextButton.exists {
+			clearTextButton.tap()
+		}
+
+		searchField.typeText(text)
+		sleep(2)
```
mokagio, line 90: Not a blocker, but it would be great if we had some kind of wait condition here rather than an arbitrary sleep. 

2 seconds might be good on a fast machine but not enough on a slow one. I'm pretty sure Simplenote searches the notes locally, but if it was talking with a backend, it can happen that the connection has an hiccup and take 3 seconds instead of 2.

Worth pointing out that sleeping for longer is a slippery slope. Once could make the same argument for most time intervals lower than the [default 60 seconds `URLSessionConfiguration` timeout](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408259-timeoutintervalforrequest). Plus, the longer we sleep the slower the tests gets.

SimplenoteUITests/AllNotes.swift
```diff
@@ -61,26 +61,49 @@ class AllNotes {
...
 }
 
 class AllNotesAssert {
 
     class func noteExists(noteName: String) {
-        print(">>> Asserting note existsence: " + noteName)
-        XCTAssertTrue(app.tables.cells[noteName].exists, "\"" + noteName + noteNotFoundInAllNotes)
+        print(">>> Asserting that note is shown once: " + noteName)
+		let matchesCount = Table.getCellsWithExactLabelCount(label: noteName)
+		XCTAssertEqual(1, matchesCount)
```
mokagio:
```suggestion
		XCTAssertEqual(matchesCount, 1)
```

The XCTest convention is to have the actual value as the first parameter and the expected one as the second.

See the [example in the `XCTAssertEqual` documentation](https://developer.apple.com/documentation/xctest/2142776-xctassertequal) or the one in the [XCTest introduction](https://developer.apple.com/documentation/xctest/defining_test_cases_and_test_methods).

<table>
<tr>
	<td>

image

</td>
	<td>

image

</td>

</table>

TESTING-CHECKLIST.md
```diff
@@ -42,9 +42,9 @@ Note: (A) means Automated Test. If automated UI tests were executed for the app,
 - [ ] Can filter by tag when clicking on tag in tag drawer
 - [ ] Searching in the search field highlights matches in note list
 - [ ] Searching in the search field highlights matches in the note editor
-- [ ] Clearing the search field immediately updates filtered notes
+- [ ] Clearing the search field immediately updates filtered notes (A)
 - [ ] Clicking on different tags or All Notes or Trash immediately updates filtered notes
-- [ ] Can search by keyword
+- [ ] Can search by keyword (A)
```
mokagio, line 47: Imagine how good it's going to be when most of the tests will be automated and the list here will only contain the few that are too hard or fiddly to automate.

SimplenoteUITests/SimplenoteUITestsLogin.swift
```diff
@@ -0,0 +1,82 @@
...
+	let testDataNotExistingEmail = "nevergonnagiveyouup@gmail.com"
+	let testDataInvalidPassword = "ABC"
+	let testDataNotExistingPassword = "ABCD"
+
+	override class func setUp() {
+		app.launch()
+	}
+
+	override func setUpWithError() throws {
+		continueAfterFailure = true
```
mokagio: I was going to suggest to move this into the `class` `setUp()` because we only need to set it once (unless we want to change it to `false` at some point during the test run).

I looked [at the docs](https://developer.apple.com/documentation/xctest/xctestcase/1496260-continueafterfailure) to make sure my assumption was true and discovered that `true` is the default value. Really, I should have known that because that's what my tests do all the time 🤣 .

Given it's the default value, what do you think about removing this explicit setting?

If you'd like to keep it because it might be handy to toggle it while debugging a test failure, then I'd suggest to add it as a commented code to the `class` setup.

```swift
// If you need to debug a failing test and don't want to waste time with the test suite 
// continuing after it fails, uncomment the following line.
//continueAfterFailure = false
```

mokagio (approved):
None of my comments is a blocker, so I'm approving the PR and leaving you the choice of which and when to address.

I run the tests on the iPhone X Simulator and they all failed... 🤔 🙃 

I run them on the iPhone 12 Pro Simulator, twice. Most tests passed. `testUsingInsertChecklistInsertsChecklist` failed on both runs, `testCanDeleteNoteForeverIndividually` failed only on the second run.

It's annoying to have unreliable tests, but we can live with them for now. We'll keep working on making the tests stable, and I have some ideas about it, so _stay tuned_.

:shipit: 

---

On tip regarding the PR being big but most changes being the file split and the camel case adoption.

I always recommend making atomic commits. Atomic commits are commits that:

- Do only one thing
- Build and have tests passing (_okay, in the case of UI tests maybe we can close an eye on the tests passing_ 😉 )

The value of doing only one thing per commit is that it makes it easier to understand (and possibly revert) the changes.

When opening a PR like this one, reviewers can look at commits individually, skimming through the ones that split the files and adopted camel case, then focus on the genuine code changes of adding the new test.

Even better! You can open a refactor PR first where you do all the high noise low signal changes, like going from `SomeName` to `someName`, and follow it up with a small PR that adds the new tests.

Personally, I find small focused PRs much easier to review. It's in everyone's interest to make small PRs. If it's easier for the reviewer to tackle them, they'll get them done sooner, so the author will be able to keep working on the next step of the feature.

SimplenoteUITests/AllNotes.swift
pachlava: Thanks Gio! Definitely my miss, it was the other way around in the framework I used previously. Changed this everywhere.

SimplenoteUITests/SimplenoteUITestsLogin.swift
pachlava: I was mistakenly thinking that having this set to `false` will actually cancel all following tests is a class. I really had no reason to have a failing test itself to continue the execution. Removed it from everywhere. Thanks!

SimplenoteUITests/AllNotes.swift
pachlava: Fully agree here. I will work on this in a separate upcoming 'refactor PR'.

PR #1173, opened 2021-02-25
"Tags&Search" tests automation MK2

pachlava:
This PR introduces no new tests, but some basic operations with tags
- Add a tag to note
- Delete all tags from Sidebar
- Select a tag in Sidebar
- Service functions for all above ^

I still feel that this PR is bigger than it should be, and most probably I needed to create a separate PR for a new file. Please let me know if this is a case.

### Fix
1. Added several new elements to `UIDs.swift`
2. Added `NoteEditor.dismissKeyboard()`
3. Added `NoteEditor.addTag(tagName)`
4. Added an optional `tagsOptional` parameter to `AllNotes.createNoteAndLeaveEditor(noteName, tagsOptional)` which allows assigning tags to a note during note creation
5. Created new file `Sidebar.swift ` with some low-level functions, and two that are likely to be used in tests/setUp methods directly: `Sidebar.tagSelect(tagName)` and `Sidebar.tagsDeleteAll()`.
6. Two 1-second sleeps in `Sidebar.tagsDeleteAll()` are planned to be fixed in the next 'refactor PR'

### Test
1. Running tests from `SimplenoteUISmokeTestsSearch` and seeing that notes are created with tags when `setUp()` is called
2. Seeing that a demo-test `testTagsTapping` passes, and you can see tags being selected and corresponding notes shown.

### Review

### Release

### DIFF
--- a/SimplenoteUITests/AllNotes.swift
+++ b/SimplenoteUITests/AllNotes.swift
@@ -7,24 +7,31 @@ class AllNotes {
     }
 
     class func isOpen() -> Bool {
+		//Warning for future: this works only for 'All Notes' obviously
+		//Will not work when a tag is selected.
         return app.navigationBars[UID.NavBar.allNotes].exists
     }
 
     class func open() {
         guard !isOpen() else { return }
 
-        app.navigationBars.element.buttons[UID.Button.menu].tap()
+		Sidebar.open()
         app.tables.staticTexts[UID.Button.allNotes].tap()
     }
 
     class func addNoteTap() {
-        app.navigationBars[UID.NavBar.allNotes].buttons[UID.Button.newNote].tap()
+		app.navigationBars.buttons[UID.Button.newNote].tap()
     }
 
-    class func createNoteAndLeaveEditor(noteName: String) {
-        print(">>> Creating note: " + noteName)
+	class func createNoteAndLeaveEditor(noteName: String, tagsOptional: [String] = []) {
+        print(">>> Creating a note: " + noteName)
         AllNotes.addNoteTap()
         NoteEditor.clearAndEnterText(enteredValue: noteName)
+
+		for tag in tagsOptional {
+			NoteEditor.addTag(tagName: tag)
+		}
+
         NoteEditor.leaveEditor()
     }
 
@@ -61,7 +68,7 @@ class AllNotes {
         for _ in 0..<notesNumber {
             let cell = app.tables.cells.element(boundBy: startingIndex)
             cell.swipeLeft()
-            cell.buttons[UID.Button.noteCellTrash].tap()
+            cell.buttons[UID.Button.itemTrash].tap()
         }
     }
 
@@ -118,7 +125,7 @@ class AllNotesAssert {
 
     class func notesNumber(expectedNotesNumber: Int) {
         let actualNotesNumber = AllNotes.getNotesNumber()
-        XCTAssertEqual(expectedNotesNumber, actualNotesNumber, numberOfNotesInAllNotesNotExpected)
+        XCTAssertEqual(actualNotesNumber, expectedNotesNumber, numberOfNotesInAllNotesNotExpected)
     }
 
     class func screenShown() {
--- a/SimplenoteUITests/AllNotes.swift
+++ b/SimplenoteUITests/AllNotes.swift
@@ -7,30 +7,30 @@ class AllNotes {
     }
 
     class func isOpen() -> Bool {
-		//Warning for future: this works only for 'All Notes' obviously
-		//Will not work when a tag is selected.
+        //Warning for future: this works only for 'All Notes' obviously
+        //Will not work when a tag is selected.
         return app.navigationBars[UID.NavBar.allNotes].exists
     }
 
     class func open() {
         guard !isOpen() else { return }
 
-		Sidebar.open()
+        Sidebar.open()
         app.tables.staticTexts[UID.Button.allNotes].tap()
     }
 
     class func addNoteTap() {
-		app.navigationBars.buttons[UID.Button.newNote].tap()
+        app.navigationBars.buttons[UID.Button.newNote].tap()
     }
 
-	class func createNoteAndLeaveEditor(noteName: String, tagsOptional: [String] = []) {
+    class func createNoteAndLeaveEditor(noteName: String, tags: [String] = []) {
         print(">>> Creating a note: " + noteName)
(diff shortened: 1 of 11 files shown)


SimplenoteUITests/AllNotes.swift
```diff
@@ -7,24 +7,31 @@ class AllNotes {
...
+		//Warning for future: this works only for 'All Notes' obviously
+		//Will not work when a tag is selected.
         return app.navigationBars[UID.NavBar.allNotes].exists
     }
 
     class func open() {
         guard !isOpen() else { return }
 
-        app.navigationBars.element.buttons[UID.Button.menu].tap()
+		Sidebar.open()
```
mokagio: From the diff on the indentation, it looks like something happened to your Xcode settings and it went from using spaces to tabs for indentation.

Either before or after merging this PR, could you please fix that and reformat these files (really, we should have automation that does this for us, sorry you have to do it manually).

SimplenoteUITests/Sidebar.swift
```diff
@@ -0,0 +1,69 @@
...
+
+class Sidebar {
+
+	class func getButtonSettings() -> XCUIElement {
+		let predicate = NSPredicate(format: "identifier == '\(UID.Cell.settings)'")
+		return app.tables.cells.element(matching: predicate)
+	}
+
+	class func getButtonTagsEdit() -> XCUIElement {
+		// Returning first match, because there are two 'Done' buttons
```
mokagio:
```suggestion
		// Returning first match, because there are two 'Edit' buttons
```

SimplenoteUITests/Sidebar.swift
```diff
@@ -0,0 +1,69 @@
...
+
+	class func getButtonTagsDone() -> XCUIElement {
+		// Returning first match, because there are two 'Done' buttons
+		// for some reason, having same coordinates...
+		return app.tables.buttons[UID.Button.done].firstMatch
+	}
+
+	class func isOpen() -> Bool {
+		// Checking for 'isHittable' of Settings button, because Settings button
+		// is actually shown even when Sidebar is closed, but has the negative coords.
```
mokagio:
```suggestion
		// is actually shown even when Sidebar is closed, but has negative coordinates.
```

SimplenoteUITests/AllNotes.swift
mokagio: 🤔 looks like this is something that happened earlier than this PR, too. I went back to #1171 and I can see them there, too –not sure how I missed them...

Definitely something to address in a dedicated PR, then.

mokagio: I think this is the setting you should tweak

Screen Shot 2021-02-26 at 3 03 59 pm

SimplenoteUITests/AllNotes.swift
```diff
@@ -7,24 +7,31 @@ class AllNotes {
...
     }
 
     class func addNoteTap() {
-        app.navigationBars[UID.NavBar.allNotes].buttons[UID.Button.newNote].tap()
+		app.navigationBars.buttons[UID.Button.newNote].tap()
     }
 
-    class func createNoteAndLeaveEditor(noteName: String) {
-        print(">>> Creating note: " + noteName)
+	class func createNoteAndLeaveEditor(noteName: String, tagsOptional: [String] = []) {
```
mokagio: I'm assuming the name `tagsOptional` means that the tags may not be there?

If that's the case, then you can use the simpler `tags`. Having the default value, `= []` already tells the reader and the Xcode autocompletion that it's not necessary to provide a value for this parameter unless you want to override the default.

```suggestion
	class func createNoteAndLeaveEditor(noteName: String, tags: [String] = []) {
```

SimplenoteUITests/AllNotes.swift
```diff
@@ -7,24 +7,31 @@ class AllNotes {
...
+        print(">>> Creating a note: " + noteName)
         AllNotes.addNoteTap()
         NoteEditor.clearAndEnterText(enteredValue: noteName)
+
+		for tag in tagsOptional {
```
mokagio: If you choose to adopt the suggestion above:

```suggestion
		for tag in tags {
```

SimplenoteUITests/UIDs.swift
```diff
@@ -32,11 +33,13 @@ enum UID {
 		static let noteEditorChecklist = "Inserts a new Checklist Item"
 		static let noteEditorInformation = "Information"
 		static let noteEditorMenu = "note-menu"
-		static let noteCellTrash = "icon trash"
+		static let itemTrash = "icon trash"
```
mokagio: You mention this PR being too long, still. I don't think that's the case it's a decent size 👍 

One thing I'd recommend, though, would be making changes like renaming a value in their own commit.

I also noticed you listed 3 operations that the tests verify in the commit message. Those, too, could have been dedicated commits.

- Commit 1: Rename noteCellTrash to itemTrash in UI tests UIDs
- Commit 2: Add UI test to add a tag to a note
- Commit 3: Add UI test to delete all tags from the sidebar
- Commit 4: Add UI test to select a tag in the sidebar

The value of having small commits is that it's easier to look at the changes in isolation. Also, if one small commit introduces an error, it's easier to revert that without loosing the rest of the work done in the PR.

mokagio (commented):
Run on the iPhone 12 Simulator and the tests passed 👍

mokagio (approved):
_I meant to approve the PR..._

SimplenoteUITests/Sidebar.swift
pachlava: Thanks for the notice!

SimplenoteUITests/Sidebar.swift
pachlava: Fixed

SimplenoteUITests/AllNotes.swift
pachlava: Fixed, thank you!

SimplenoteUITests/AllNotes.swift
pachlava: Done!

SimplenoteUITests/AllNotes.swift
pachlava: This was my fault: I intentionally changed spaces to tabs in the settings above some time ago, hoping it will help with trailing whitepaces. But I had to use another setting from 'Editing' tab for this. Changed. Thanks!

PR #1175, opened 2021-02-27
Issue/1407 reference count incorrect

### LINKED ISSUE #1147: [Info Sheet] References count is not counting right
### Expected
References count should show the count of usage in every note separately

### Observed
When the note is referenced in other notes, the ref. count will show the same number for each note where it's referenced (i.e. total number of references VS individual number of references):

Screenshot_2021-02-12 11 44 50_ooWsfL

### Reproduced
1. Copy an internal like of Note1
2. Insert the link into Note2 once and Note3 also once
3. Open Note1
4. Open Info Sheet and check the references number ~> 

**Actual**: it shows '2' for Note2 and '2' for Note3.
**Expected**: should show '1' for both

**Additionally**: it also looks like references inside note are not counted correctly. For example, referencing Note1 in Note2 twice will still show 1 reference.

-|-|-|-
iPhone|XR|14.3|4.32.0.0
iPad|Pro (Sim)|14.4|4.32.0.0

charliescheer:
### Fix
This PR updates the reference count logic in the Note Information sheet to calculate the number of internal references are contained in a linked note.  This fixes #1147 

Previous:

New:
 

### Test
1. Create a note to be the note that is referenced to (NoteA).
2. In the new note, go to the note menu > Get internal link
3. go to another note (NoteB) and paste in the internal link one time into the note
4. Go to a third note (NoteC) and paste the internal link two different times in the content of the note
5. Return to NoteA and open the information sheet

There you should see both NoteB and NoteC listed as references, noteB should say 1 reference noteC should say 2

### Review

### DIFF
--- a/Simplenote/Classes/String+Simplenote.swift
+++ b/Simplenote/Classes/String+Simplenote.swift
@@ -171,3 +171,10 @@ extension String {
         return ContentSlice(content: self, range: lowerBound..<upperBound, matches: matchingWordsRange)
     }
 }
+
+// MARK: Substring Instance Count
+extension String {
+    func occurancesOf(_ string: String) -> Int {
+        return components(separatedBy: string).count - 1
+    }
+}
--- a/Simplenote/Classes/String+Simplenote.swift
+++ b/Simplenote/Classes/String+Simplenote.swift
@@ -174,7 +174,7 @@ extension String {
 
 // MARK: Substring Instance Count
 extension String {
-    func occurancesOf(_ string: String) -> Int {
+    func occurrences(of string: String) -> Int {
         return components(separatedBy: string).count - 1
     }
 }
--- a/Simplenote/Information/NoteInformationController.swift
+++ b/Simplenote/Information/NoteInformationController.swift
@@ -143,9 +143,9 @@ private extension NoteInformationController {
 
         let referenceRows = references.map { (referenceNote) -> Row in
             let date = DateFormatter.dateFormatter.string(from: referenceNote.modificationDate)
-            let value = Localization.interlinkReferences(Note.instancesOfReferenceToNoteWith(simperiumKey: note.simperiumKey, in: referenceNote.content)) +
-                        ", " +
-                        Localization.lastModified(date)
+            let instancesOfReference = referenceNote.instancesOfReference(to: note)
+
+            let value = "\(Localization.interlinkReferences(instancesOfReference)), \(Localization.lastModified(date))"
 
             return .reference(interLink: referenceNote.plainInternalLink,
                               title: referenceNote.titlePreview,
--- a/Simplenote/Information/NoteInformationController.swift
+++ b/Simplenote/Information/NoteInformationController.swift
@@ -141,12 +141,12 @@ private extension NoteInformationController {
             return []
         }
 
-        let referenceRows = references.map { (note) -> Row in
-            let date = DateFormatter.dateFormatter.string(from: note.modificationDate)
-            let value = "\(Localization.interlinkReferences(references.count)), \(Localization.lastModified(date))"
+        let referenceRows = references.map { (referenceNote) -> Row in
+            let date = DateFormatter.dateFormatter.string(from: referenceNote.modificationDate)
+            let value = "\(Localization.interlinkReferences(note.instancesOfInternalLinkIn(referenceNote.content))), \(Localization.lastModified(date))"
 
-            return .reference(interLink: note.plainInternalLink,
-                              title: note.titlePreview,
+            return .reference(interLink: referenceNote.plainInternalLink,
+                              title: referenceNote.titlePreview,
                               reference: value)
         }
 
--- a/Simplenote/Information/NoteInformationController.swift
+++ b/Simplenote/Information/NoteInformationController.swift
@@ -143,7 +143,9 @@ private extension NoteInformationController {
 
         let referenceRows = references.map { (referenceNote) -> Row in
             let date = DateFormatter.dateFormatter.string(from: referenceNote.modificationDate)
-            let value = "\(Localization.interlinkReferences(note.instancesOfInternalLinkIn(referenceNote.content))), \(Localization.lastModified(date))"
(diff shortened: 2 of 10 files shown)


Simplenote/Classes/String+Simplenote.swift
```diff
@@ -171,3 +171,10 @@ extension String {
         return ContentSlice(content: self, range: lowerBound..<upperBound, matches: matchingWordsRange)
     }
 }
+
+// MARK: Substring Instance Count
+extension String {
+    func occurancesOf(_ string: String) -> Int {
```
eshurakov: Should it be `occurrences`? 🤔 

Interesting approach, too bad swift doesn't have a built in method for that 😞

Simplenote/Note+Links.swift
```diff
@@ -36,4 +36,9 @@ extension Note {
 
         return SimplenoteConstants.simplenotePublishedBaseURL + targetURL
     }
+
+    static func instancesOfReferenceToNoteWith(simperiumKey: String, in content: String) -> Int {
+        let internalLink = SimplenoteConstants.simplenoteScheme + "://" + SimplenoteConstants.simplenoteInterlinkHost + "/" + simperiumKey
+        return content.occurancesOf(internalLink)
+    }
```
eshurakov: How about `func instancesOfReference(to note: Note) -> Int`? Seems cleaner and we can put all the logic here.
We can also reuse `plainInternalLink` here, which will remove code duplication.

SimplenoteTests/StringSimplenoteTests.swift
```diff
@@ -104,4 +104,20 @@ extension StringSimplenoteTests {
...
     }
 
+    /// Verifies the correct return from occurancesOf
+    ///
+    func testOccurancesOfReturnsCorrectValue() {
+        let sampleA = "test x value, x test"
+        let sampleB = "test x value x"
+        let sampleC = "x test x value"
+        let sampleD = "X test X value"
+        let sampleE = "x test xxx valxue"
```
eshurakov, line 114: It's good to test and empty case here as well.

SimplenoteTests/InterlinkResultsControllerTests.swift
```diff
@@ -105,6 +105,20 @@ class InterlinkResultsControllerTests: XCTestCase {
...
+        let noteContentWithOneReference = "This note refers to \(plainInternalLink) one time"
+        let noteContentWithTwoReferences = "This note refers to \(plainInternalLink) two times: \(plainInternalLink)"
+        let noteContentWithOutRefernces = "No references"
+        let contentWithKeyNotLink = "this note hase the \(simperiumKey) but not a link"
+
+        XCTAssertEqual(Note.instancesOfReferenceToNoteWith(simperiumKey: simperiumKey, in: noteContentWithOneReference), 1)
+        XCTAssertEqual(Note.instancesOfReferenceToNoteWith(simperiumKey: simperiumKey, in: noteContentWithTwoReferences), 2)
+        XCTAssertEqual(Note.instancesOfReferenceToNoteWith(simperiumKey: simperiumKey, in: noteContentWithOutRefernces), 0)
+        XCTAssertEqual(Note.instancesOfReferenceToNoteWith(simperiumKey: simperiumKey, in: contentWithKeyNotLink), 0)
+    }
```
eshurakov: I think it's better to put it in a separate test class, because it doesn't test InterlinkResultsController, but Note extension.

eshurakov (commented):
Hey Charlie, sending you a few comments!

Simplenote/Information/NoteInformationController.swift
```diff
@@ -141,12 +141,14 @@ private extension NoteInformationController {
...
 
-        let referenceRows = references.map { (note) -> Row in
-            let date = DateFormatter.dateFormatter.string(from: note.modificationDate)
-            let value = "\(Localization.interlinkReferences(references.count)), \(Localization.lastModified(date))"
-
-            return .reference(interLink: note.plainInternalLink,
-                              title: note.titlePreview,
+        let referenceRows = references.map { (referenceNote) -> Row in
+            let date = DateFormatter.dateFormatter.string(from: referenceNote.modificationDate)
+            let value = Localization.interlinkReferences(Note.instancesOfReferenceToNoteWith(simperiumKey: note.simperiumKey, in: referenceNote.content)) +
```
jleandroperez: For readability purposes, can we move the `Note.instancesOfReferenceToNoteWith()` invocation up, and assign that to a constant?

charliescheer: Yes we can.

Simplenote/Classes/String+Simplenote.swift
charliescheer: yes it should...  Spelling is not my greatest skill... fixed

Simplenote/Note+Links.swift
charliescheer: Thanks for the suggestions on how I can make that work with the unit tests.  Was able to get that working with `func instancesOfReference(to note: Note) -> Int`

charliescheer:
Thanks for the suggestions @eshurakov  and @jleandroperez  I have updated the PR.  When you have a moment, can y'all take another look.  I think it is good to go

SimplenoteTests/NoteExtensionTests.swift
```diff
@@ -0,0 +1,48 @@
+import XCTest
+@testable import Simplenote
+
+
+// MARK: - Note+Extension Tests
+//
+class NoteExtensionTests: XCTestCase {
```
jleandroperez: Maybe `NoteLinkTests` ? (since there's another Extension per se)

Simplenote/Note+Links.swift
```diff
@@ -36,4 +36,11 @@ extension Note {
 
         return SimplenoteConstants.simplenotePublishedBaseURL + targetURL
     }
+
+    func instancesOfReference(to note: Note) -> Int {
+        guard let internalLink = note.plainInternalLink else {
+            return 0
```
jleandroperez: `return .zero` looks sooo much cooler 😂

jleandroperez (approved):
Thank you for fixing this @charliescheer !!

:shipit:

pachlava:

PR #1176, opened 2021-02-28
Automated two more Search tests

pachlava:
Automated two more tests:
`Can filter by tag when clicking on tag in tag drawer`
`Clicking on different tags or All Notes or Trash immediately updates filtered notes`

### Fix
In order to automate tests from above in a better way, I added a few methods related to navbars:

1. `NoteList.getNavBarIdentifier()`
2. `isNoteListForTagOpen(tag)`
3. `NoteListAssert.trashShown()`
4. `NoteListAssert.noteListForTagOpen()`

Also, minor improvements to logging.

### Test
1. Running two new tests from `SimplenoteUISmokeTestsSearch`:
 - `testCanFilterByTagWhenClickingOnTagInTagDrawer`
 - `testClickingOnDifferentTagsOrAllNotesOrTrashImmediatelyUpdatesFilteredNotes`
2. All other tests are not affected

### Review

### Release

### DIFF
--- a/SimplenoteUITests/NotesList.swift
+++ b/SimplenoteUITests/NotesList.swift
@@ -6,12 +6,24 @@ class NoteList {
         app.tables.cells[noteName].tap()
     }
 
+    class func getNavBarIdentifier() -> String {
+        let navBar = app.navigationBars.element
+        guard navBar.exists else { return "" }
+        print(">>> Currently active navigation bar: " + navBar.identifier)
+        return navBar.identifier
+    }
+
     class func isAllNotesListOpen() -> Bool {
         return app.navigationBars[UID.NavBar.allNotes].exists
     }
 
+    class func isNoteListForTagOpen(tag: String) -> Bool {
+        return app.navigationBars[tag].exists
+    }
+
     class func openAllNotes() {
         guard !NoteList.isAllNotesListOpen() else { return }
+        print(">>> Opening \"All Notes\"")
         Sidebar.open()
         app.tables.staticTexts[UID.Button.allNotes].tap()
     }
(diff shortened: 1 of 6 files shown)


SimplenoteUITests/NotesList.swift
```diff
@@ -6,12 +6,24 @@ class NoteList {
         app.tables.cells[noteName].tap()
     }
 
+    class func getNavBarIdentifier() -> String {
+        let navBar = app.navigationBars.element
+        guard navBar.exists else { return "" }
```
mokagio, line 11: Returning an empty string to signal that the nav bar wasn't found does the job, but could spiral into weird test failures downstream if a test doesn't check for the value not to be `""` and tries to access a nav bar with it.

A different take that leverages Swift's refined type system would be to use `Optional<String>`, `String?` for short. With this type, we can signal that no match was found by returning `.none`. The Swift compiler forces the consumer code to deal with the optionality, this is a great way to let future developers know that the nav bar may not have been found, without writing a single line of documentation.

The only changes required to the function are:

```suggestion
    class func getNavBarIdentifier() -> String? {
        let navBar = app.navigationBars.element
        guard navBar.exists else { return .none }
```

I haven't tried it, but I'm pretty sure the existing assertions won't need to change either.

SimplenoteUITests/NotesList.swift
```diff
@@ -6,12 +6,24 @@ class NoteList {
...
+        print(">>> Currently active navigation bar: " + navBar.identifier)
+        return navBar.identifier
+    }
+
     class func isAllNotesListOpen() -> Bool {
         return app.navigationBars[UID.NavBar.allNotes].exists
     }
 
+    class func isNoteListForTagOpen(tag: String) -> Bool {
```
mokagio: Nitpick. Swift allows us to have two different argument labels, one for the call site and to use internally. This helps write code that reads more like English:

```suggestion
    class func isNoteListOpen(forTag tag: String) -> Bool {
```

This code still uses `tag` internally, but is called like `isNoteListOpen(forTag: "a-tag")` which reads more like a sentence, imho.

mokagio (approved):
Only left non blocking nitpicks.

All the tests passed on the iPhone 12 Simulator 🎉

SimplenoteUITests/NotesList.swift
pachlava: Thank you! Used your suggestion, and currently the impact (changes required I mean) was 0, because the method was never called. It's more for the future.

SimplenoteUITests/NotesList.swift
pachlava: Thanks A LOT for this @mokagio! I will create a separate PR for this 🥳

PR #1181, opened 2021-03-01
UI Automation Refactor: Using optionals in return values

pachlava:
- Using optional strings in return values instead of returning empty strings
- Renamed two methods

### Fix
1. `Table.getContentOfCell()` and `NoteList.getNavBarIdentifier()` now return optional strings
2. Two methods: `NoteListAssert.noteListShown()` and `NoteListAssert.noteContentIsShownInSearch()` that are using the methods from 1 now include unwrapping of optional strings
3. Renamed `allNotesScreenShown()` to `allNotesShown()`
4. Since there were three painfully similar methods: `allNotesShown()` , `trashShown()` and `noteListForTagOpen()` which do the same, but with different inputs, I created `noteListShown(forSelection selection: String)`, which now replaces 'noteListForTagOpen()' and also now called by `allNotesShown()` and `trashShown()`

### Test
1. Running all tests

### Review

### Release

### DIFF
--- a/SimplenoteUITests/NotesList.swift
+++ b/SimplenoteUITests/NotesList.swift
@@ -144,7 +144,7 @@ class NoteListAssert {
         if let navBarID = NoteList.getNavBarIdentifier() {
             XCTAssertEqual(navBarID, selection, selection + navBarNotFound)
         } else {
-            XCTAssertTrue(false, foundNoNavBar)
+            XCTFail(foundNoNavBar)
         }
     }
 
@@ -163,7 +163,7 @@ class NoteListAssert {
         if let noteContent = Table.getContentOfCell(noteName: noteName) {
             XCTAssertTrue(noteContent.contains(expectedContent), "Content NOT found")
         } else {
-            XCTAssertTrue(false, "Could not find note")
+            XCTFail("Could not find note")
         }
     }
 }
--- a/SimplenoteUITests/NotesList.swift
+++ b/SimplenoteUITests/NotesList.swift
@@ -6,9 +6,9 @@ class NoteList {
         app.tables.cells[noteName].tap()
     }
 
-    class func getNavBarIdentifier() -> String {
+    class func getNavBarIdentifier() -> String? {
         let navBar = app.navigationBars.element
-        guard navBar.exists else { return "" }
+        guard navBar.exists else { return .none }
         print(">>> Currently active navigation bar: " + navBar.identifier)
         return navBar.identifier
     }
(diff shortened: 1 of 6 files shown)


SimplenoteUITests/NotesList.swift
```diff
@@ -137,27 +137,33 @@ class NoteListAssert {
...
-        XCTAssertEqual(NoteList.getNavBarIdentifier(), UID.NavBar.allNotes, UID.NavBar.allNotes + navBarNotFound)
+    class func noteListShown(forSelection selection: String ) {
+        print(assertNavBarIdentifier + selection)
+        XCTAssertTrue(app.navigationBars[selection].waitForExistence(timeout: maxLoadTimeout))
+
+        if let navBarID = NoteList.getNavBarIdentifier() {
+            XCTAssertEqual(navBarID, selection, selection + navBarNotFound)
+        } else {
+            XCTAssertTrue(false, foundNoNavBar)
+        }
```
mokagio: `XCTAssertTrue(false)` is a neat trick to force a test failure. There's a dedicated API that you can use for that, though:

```suggestion
        if let navBarID = NoteList.getNavBarIdentifier() {
            XCTAssertEqual(navBarID, selection, selection + navBarNotFound)
        } else {
            XCTAssertFail(foundNoNavBar)
        }
```

Another cool XCTest API is `XCTUnwrap`; you can use it to unwrap optional values and fail the test if the value is `.none`.

I haven't tried the code in the IDE, but it should work out to something like:

```suggestion
		let navBarId = try XCTUnwrap(NoteList.getNavBarIdentifier(), foundNoNavBar)
        XCTAssertEqual(navBarID, selection, selection + navBarNotFound)
```

mokagio (approved):
Nice. Glad you liked the suggestion and found it useful.

All tests passing on iPhone 12 Simulator.

SimplenoteUITests/NotesList.swift
pachlava: Thank you @mokagio, it definitely makes sense for me to explore a bit more for alternatives before writing certain code.

I noticed that using `try XCTUnwrap...` results in a chain of changes: marking all subsequent methods with `throws`, and also prefixing all calls with `try`, including the tests code. For now, I went for your first suggestion of using `XCTFail()`.

mokagio: Good point. I didn't take into account the fact that this is not a test method, that is, the place where I'd usually call `XCTUnwrap`.

pachlava: I generally have some amount of questions/places in my code I don't like that much, but I'd like to firstly discuss how to proceed with it. I thought of maybe having a short call at some point?

mokagio: Sure, happy to give my two cents.

What do you think of writing down all of them (or the subset that can be easily written done) in a P2 (it's a shame to not do it in the open, but P2 has got a better threading support than GitHub) so I can start thinking about them first?

PR #1182, opened 2021-03-01
SignUp flow update

eshurakov:
### Details
In this PR we're updating SignUp flow.

### Review


Simplenote/Classes/SignupRemote.swift
```diff
@@ -0,0 +1,57 @@
...
+              let url = URL(string: SimplenoteConstants.signupURL) else {
+            return nil
+        }
+
+        var request = URLRequest(url: url,
+                                 cachePolicy: .reloadIgnoringLocalAndRemoteCacheData,
+                                 timeoutInterval: Constants.timeoutInterval)
+        request.httpMethod = Constants.httpMethod
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        request.httpBody = try? JSONEncoder().encode(["username": base64EncodedEmail])
```
dmsnell: as noted in Slack, we did determine that sending the username inside a JSON body would be sufficient since JSON properly handles Unicode data without encoding/decoding issues. we can pass a JSON encoded object with a `username` field containing the the plain/raw email address.

Simplenote/Classes/SignupRemote.swift
```diff
@@ -0,0 +1,57 @@
...
+    private func request(with email: String) -> URLRequest? {
+        guard let base64EncodedEmail = email.data(using: .utf8)?.base64EncodedString(),
+              let url = URL(string: SimplenoteConstants.signupURL) else {
+            return nil
+        }
+
+        var request = URLRequest(url: url,
+                                 cachePolicy: .reloadIgnoringLocalAndRemoteCacheData,
+                                 timeoutInterval: Constants.timeoutInterval)
+        request.httpMethod = Constants.httpMethod
```
dmsnell: feel free to ignore me, but this seems more confusing than helpful. these are somewhat pervasive string constants, would it not be clearer to use `"POST"` here?

if not, if we require a constant, it would definitely take less work to hunt around figuring out which HTTP method `httpMethod` uses if we had `Constants.POST` or `Constants.HttpPost`, but again, if the literal `POST` ever changes the whole internet will break, so maybe it's fine to inline it?

eshurakov: Thank you for bringing this up. We try to extract literals and inline numbers to constants to make more clear what they mean. Of course here it would be clear without a constant that it is an http method :-) But it's also good to follow the code style of the project.
The name that is currently used for a constant indeed doesn't tell us what the method is. I've updated the name to fix it. 👍

Simplenote/Classes/SPAuthHandler.swift
```diff
@@ -65,16 +65,19 @@ class SPAuthHandler {
...
-            onCompletion(nil)
-        }, failure: { (statusCode, response, error) in
-            let error = SPAuthError(signupErrorCode: statusCode, response: response, error: error)
+    func signupWithCredentials(username: String, onCompletion: @escaping (SPAuthError?) -> Void) {
+        SignupRemote().signup(with: username) { (success) in
+            let error: SPAuthError? = {
+                guard !success else {
+                    return nil
+                }
+                return SPAuthError(signupErrorCode: 0, response: nil, error: nil)
```
jleandroperez: We **may** need to update this a bit, so that the remote StatusCode is relayed back perhaps.

Simplenote/Classes/SignupRemote.swift
```diff
@@ -0,0 +1,49 @@
...
+class SignupRemote {
+    private let urlSession: URLSession
+
+    init(urlSession: URLSession = URLSession.shared) {
+        self.urlSession = urlSession
+    }
+
+    /// Send signup request for specified email address
+    ///
+    func signup(with email: String, completion: @escaping (_ success: Bool) -> Void) {
```
jleandroperez: Yup, for troubleshooting purposes we mayyyyy wanna send back the status code?

Simplenote/Classes/SignupRemote.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+
+// MARK: - SignupRemote
+//
+class SignupRemote {
```
jleandroperez, line 5: Since now we've got two Remotes (AccountVerificationRemote / SignupRemote), do you think we should move them both to a (project) Folder?

(Not filesystem, but I mean, group them in the project itself)

Simplenote/SPAppDelegate.m
```diff
@@ -578,6 +578,10 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
 {
     if (!self.simperium.user.authenticated) {
         [self performDotcomAuthenticationWithURL:url];
+
+        if (!self.simperium.user.authenticated && url) {
+            [self performMagicLinkAuthenticationWith:url];
+        }
```
jleandroperez, line 584: I *think* the plan was to display an alert asking if you wanna log out otherwise.

(But no need to do anything in this PR, I also need to do the same in the Mac!)

jleandroperez (approved):
@eshurakov Sorry I crashed your PR, sending a CI build to Sly.

Looks **SUPER SOLID**, we'll need to:

1. Ship a new Simperium version (doing that shortly). 
2. Drop the testing URL

Other than that :shipit:

Simplenote/MagicLinkAuthenticator.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
```
jleandroperez: Oh!! I've just noticed this file isn't in the `Classes/` folder, would it be possible to move it?

jleandroperez: Ummmm I'm thinking we may need to go even further. Because... well, this is not a class, isn't it!😂

eshurakov: Argh, happens all the time to me :-(

Simplenote/Classes/SignupRemote.swift
eshurakov: Right. I think it need to be discussed, because there are many ways to organise files. Right now those files are in the groups that contain most of the files for that feature. So when you work on onboarding or verification feature you can find all the files in one place :-)

Simplenote/Classes/SignupRemote.swift
eshurakov: Added it in [this commit](https://github.com/Automattic/simplenote-ios/pull/1182/commits/098461f21fe3c7ba7062143e41086e8e11c206a6), could you take a look pls? :-)

jleandroperez: Looks super, thank you!! (I did something similar for the mac!)

Simplenote/Classes/SignupRemote.swift
jleandroperez: Agreed!

jleandroperez (approved):
✅  Verified Signup Flow with **Backgrounded App** during the Magic Link Handling step
✅  Verified Signup Flow with **Dead App** during the Magic Link Handling step

:shipit:

eshurakov:
Thank you!

pachlava:

Issue #1177, opened 2021-02-28
Project: In App Notifications - Implement NoticeView UI


charliescheer:
### Details:

- [x] Create view xib and classes (NoticeView, NoticeAction) #1185 
- [x] add support for dark mode #1185
- [x] add support for dynamic type #1185 
- [x] add support for VoiceOver #1185 

Screenshot
<img src="https://user-images.githubusercontent.com/1195260/96787162-af930680-13c7-11eb-84e4-5c1fb6344b5d.gif" >

PR #1185, opened 2021-03-02
Project: In App Notifications - View Implementation and Styling

charliescheer:
### Fix
This PR begins the process of implementing in app notifications to Simplenote iOS.  It deals with the first two to do items on  issue #1177 Implement Notice View and enable dark mode styling.

Note that this current setup does not animate the appearance of the notices and is only intended to display them in the correct location and styled correctly.

light mode:

dark mode:

### Test
There is currently no implementation to display notes, so to test, first go to this [bin](https://mc.a8c.com/pb/28971/) and copy the code there.  Paste it into SPNoteViewController+Extensions.swift starting at line 37 replacing configureNavigationBarItems( ) and adding setup/display functions for the notices

1. once setup with the testing code, open up Simplenote and go to any note.
2. tap on the action button to see a notice without an action, long press on the notification and check the log for a response then wait for it to disappear
3. tap on the create note button to see a notice with an action appear, long press on the notification and check the log for a response, then wait for it to disappear
4. tap on the create note again, and then tap on the "retry" button you should see in the log a message appear after tapping the button
5. Set device to dark mode and repeat steps 2-4 again
6. Test device rotation to make sure that the notice stay centered and just above bottom of the screen
7. This PR also covers accessibility, so make sure to test that the notices appear correctly when dynamic text is enabled

Note that these do not currently account for keyboards, so if the keyboard is present the notification will not be visible. 

### Release


charliescheer:
Note: I updated the PR per our conversation about constraints

charliescheer:
New note.  I made some more changes today that added accessibility to the notices.  I had planned on making this on another branch/in another PR, but turns out Voiceover was already there by default.  So there were only some IB changes that added dynamic type.  Didn't seem to make sense in it's own PR at that point.

Updated the test instructions

Simplenote/NoticeView.swift
```diff
@@ -0,0 +1,93 @@
...
+
+class NoticeView: UIView {
+
+    // MARK: Properties
+    //
+    @IBOutlet weak var stackView: UIStackView!
+    @IBOutlet weak var noticeLabel: UILabel!
+    @IBOutlet weak var noticeButton: UIButton!
+
+    var delegate: NoticePresentingDelegate?
```
jleandroperez: This one should probably be `weak`, to avoid retain cycles

Simplenote/NoticeView.swift
```diff
@@ -0,0 +1,93 @@
...
+    func noticePressEnded()
+}
+
+class NoticeView: UIView {
+
+    // MARK: Properties
+    //
+    @IBOutlet weak var stackView: UIStackView!
+    @IBOutlet weak var noticeLabel: UILabel!
+    @IBOutlet weak var noticeButton: UIButton!
```
jleandroperez: Nipticky: Missing visibility modifier 👀

Simplenote/NoticeView.swift
```diff
@@ -0,0 +1,93 @@
...
+        addGestureRecognizer(longPressGesture)
+    }
+
+    private func setupViewStyles() {
+        stackView.backgroundColor = UIColor(lightColor: .spGray3, darkColor: .darkGray2)
+        stackView.layer.cornerRadius = Constants.cornerRadius
+        stackView.clipsToBounds = true
+
+        noticeLabel.textColor = UIColor(lightColor: .gray100, darkColor: .white)
+        noticeButton.titleLabel?.textColor = UIColor(lightColor: .spBlue50, darkColor: .spBlue30)
```
jleandroperez:
```suggestion
        noticeButton.titleLabel?.textColor = .simplenoteInteractiveTextColor
```

Simplenote/NoticeView.swift
```diff
@@ -0,0 +1,93 @@
...
+        let longPressGesture = UILongPressGestureRecognizer(target: self, action: #selector(viewWasLongPressed(_:)))
+        addGestureRecognizer(longPressGesture)
+    }
+
+    private func setupViewStyles() {
+        stackView.backgroundColor = UIColor(lightColor: .spGray3, darkColor: .darkGray2)
+        stackView.layer.cornerRadius = Constants.cornerRadius
+        stackView.clipsToBounds = true
+
+        noticeLabel.textColor = UIColor(lightColor: .gray100, darkColor: .white)
```
jleandroperez:
```suggestion
        noticeLabel.textColor = .simplenoteTextColor
```

Simplenote/NoticeView.swift
```diff
@@ -0,0 +1,93 @@
...
+        layoutIfNeeded()
+    }
+
+    private func setupLongPress() {
+        let longPressGesture = UILongPressGestureRecognizer(target: self, action: #selector(viewWasLongPressed(_:)))
+        addGestureRecognizer(longPressGesture)
+    }
+
+    private func setupViewStyles() {
+        stackView.backgroundColor = UIColor(lightColor: .spGray3, darkColor: .darkGray2)
```
jleandroperez: Nipticky: can we add a color in `UIColor+Studio`?  Maaaybe **simplenoteNoticeBackgroundColor** (or so)

Simplenote/NoticeView.swift
```diff
@@ -0,0 +1,93 @@
...
+            longPressBegan()
+        case .ended:
+            longPressEnded()
+        default:
+            return
+        }
+    }
+
+    private func longPressBegan() {
+        print("long press began")
```
jleandroperez: Nipticky: log statement 🔥

Simplenote/NoticeView.swift
```diff
@@ -0,0 +1,93 @@
...
+        delegate?.noticePressBegan()
+    }
+
+    private func longPressEnded() {
+        print("long press finished")
```
jleandroperez: 🔥

Simplenote/UntouchableView.swift
```diff
@@ -0,0 +1,16 @@
+import UIKit
+
+class UntouchableView: UIView {
+    override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {
+        let visibleViews = subviews.filter { view -> Bool in
+            return view.alpha >= 0.01 && !view.isHidden && view.isUserInteractionEnabled
```
jleandroperez: Q: would `> .zero` work?

Simplenote/UntouchableView.swift
```diff
@@ -0,0 +1,16 @@
+import UIKit
+
+class UntouchableView: UIView {
+    override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {
```
jleandroperez: I **suspect** you may wanna override hitTest instead maybe?

jleandroperez (commented):
Code looks good @charliescheer !!
Sending few notes your way, 

Thank you sir!!


### iOS 12 / 13
The Notice isn't getting its background color properly applied (the bubble's BG appears to be clear color!)

### Animation
There's something odd going on with the animation. Please note that the `Retry` button's text appears to be clipped for a split second, then, corrected.

To fix this one, please try switching the Button Type to **Custom** (that's how we usually set it!).

https://user-images.githubusercontent.com/1195260/110128635-cb5ede00-7da5-11eb-99c1-a69f47c9e7e7.mov

Simplenote/NoticeView.swift
charliescheer: I'd left these prints in because the tapping didn't do anything currently otherwise.  They will be doing something once the controllers get brought in.  Removed them

Simplenote/UntouchableView.swift
charliescheer: Yes it does

Simplenote/NoticeView.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    // MARK: Action
+    //
+    @IBAction func noticeButtonWasTapped(_ sender: Any) {
+        action?()
+    }
+}
+
+// NOTE: long press recognizing has not been connected to anything yet
+// Currently just prints to log that a press event happened.
```
eshurakov, line 87: Just curious, what will be the purpose of long press?

charliescheer: It interrupts the automatic dismissal of the notice.  So if the user wants the notice to not disappear yet, as long as they press on the notice it will stay

Simplenote/UntouchableView.swift
charliescheer: As discussed in slack, I checked out PassthruView and that works to do what we were looking for here.  Have removed the UntouchableView file.

charliescheer:
>The Notice isn't getting its background color properly applied (the bubble's BG appears to be clear color!)
Well would you look at that, pre iOS 14 StackViews had no background color...  TIL

@jleandroperez this should be ready to take a look at again.  Note I have updated the [code bin ](https://mc.a8c.com/pb/28a5e/)a little bit to use a passthrough view and instantiate the notice view from nib.

Let me know if there is anything else!

Simplenote/NoticeView.swift
eshurakov: The notice is not very big, so I wonder if you press it with finger, can you still read it? 🤔

Simplenote/NoticeView.swift
```diff
@@ -0,0 +1,125 @@
...
+    var message: String? {
+        get {
+            noticeLabel.text
+        }
+        set {
+            noticeLabel.text = newValue
+        }
+    }
+
+    var action: NoticeAction? {
```
jleandroperez: Can we drop this one?

It's coupling the View + Model, which is something you'd generally wanna avoid.

You could, instead, do something like (and maintain the Model <> View details elsewhere)


```
var onAction: (() -> Void)?
var actionTitle: String? {
    get {
        noticeButton.title(for: .normal)
    }
    set {
        noticeButton.setTitle(newValue.title, for: .normal)
    }
}
```

jleandroperez (commented):
Just one minor comment, other than that, looks great!

Simplenote/NoticeView.swift
charliescheer: Yup, like that much better

jleandroperez (approved):
Looks good!! Nice work Charlie!!

:shipit:

PR #1189, opened 2021-03-05
Don't create a welcome note

eshurakov:
### Fix
In this PR we're removing Welcome Note creation logic from the client. Welcome Note will be created by the backend.

### Test
1. Make sure it's a clean install (delete and install the app)
2. Sign up with a new account

- [ ] Check that there is only one welcome note

### Review

### DIFF
--- a/Simplenote/SPAppDelegate.m
+++ b/Simplenote/SPAppDelegate.m
@@ -161,9 +161,9 @@
     [self loadSelectedTheme];
     
     // Check to see if first time user
-    if ([self isFirstLaunch]) {        
+    if ([self isFirstLaunch]) {
+        _noteListViewController.firstLaunch = YES;
         [[SPPinLockManager shared] removePin];
-        [self createWelcomeNoteAfterDelay];
         [self markFirstLaunch];
     } else {
         [self showPasscodeLockIfNecessary];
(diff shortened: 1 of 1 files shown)


Simplenote/SPAppDelegate.m
```diff
@@ -263,30 +263,6 @@ - (void)markFirstLaunch
...
-    Note *welcomeNote = [noteBucket objectForKey:welcomeKey];
-    
-    if (welcomeNote) {
-        return;
-	}
-    
-    welcomeNote = [noteBucket insertNewObjectForKey:welcomeKey];
-    welcomeNote.modificationDate = [NSDate date];
-    welcomeNote.creationDate = [NSDate date];
-    welcomeNote.content = NSLocalizedString(@"welcomeNote-iOS", @"A welcome note for new iOS users");
```
jleandroperez, line 284: We may also wanna (manually) drop `welcomeNote-iOS` from the .strings files (since that particular string was manually put there I believe!)

jleandroperez (approved):
LOVE this PR!!!

:shipit:

Simplenote/SPAppDelegate.m
eshurakov: Hmm, I think it will be gone when release is created. We will see :-)

eshurakov:
Thank you!

pachlava:

PR #1190, opened 2021-03-05
Adding `XCUIElementQuery: Sequence` extension

pachlava:
The only functional change: adding the `XCUIElementQuery: Sequence` [extension](https://github.com/onmyway133/blog/issues/628) to `XCTest+Extensions.swift`, which allows iterating through query with:

```
for cell in cells {
    cell.doStuff...
```
instead of
```
for index in 0...cellsNum - 1 {
    let cell = app.tables.cells.element(boundBy: index)
    cell.doStuff...
```

### Fix
1. Added extension that allows iterating through XCUIElementQuery
2. Renamed `matchesNum`, `matchesCount` etc to `matches` in various places
3. Applying String Interpolation in several methods
4. Added `Table.getAllCells()` method

### Test
1. Running the tests, no unexpected fails should occur.

### Review

### Release

### DIFF
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -29,7 +29,7 @@ func trackTest(_ function: String = #function) {
 }
 
 func trackStep() {
-    print(">> Step " + String(stepIndex))
+    print(">> Step \(stepIndex)")
     stepIndex += 1
 }
 
@@ -50,34 +50,18 @@ class Table {
         // otherwise we will include invisible elements from Sidebar pane, when Notes List is open
         // or the elements from Notes List when Settings are open
         // or the visible tags search suggestions
-        let cells = Table.getAllCells()
-        var matches = 0
-        guard cells.count > 0 else { return matches }
-
-        for cell in cells {
-            if cell.frame.minX == 0.0 && cell.label.count > 0 {
-                matches += 1
-            }
-        }
-
-        return matches
+        return Table.getAllCells()
+            .filter { $0.frame.minX == 0.0 && $0.label.isEmpty == false }
+            .count
     }
 
     class func getVisibleNonLabelledCellsNumber() -> Int {
         // We need only the table cells that have X = 0 and an empty label
         // Currently (besides using object dimentions) this is the way to
         // locate tags search suggestions
-        let cells = Table.getAllCells()
-        var matches = 0
-        guard cells.count > 0 else { return matches }
-
-        for cell in cells {
-            if cell.frame.minX == 0.0 && cell.label.count == 0 {
-                matches += 1
-            }
-        }
-
-        return matches
+        return Table.getAllCells()
+            .filter { $0.frame.minX == 0.0 && $0.label.isEmpty == true }
+            .count
     }
 
     class func trashCell(noteName: String) {
@@ -138,11 +122,17 @@ class WebView {
 
 class WebViewAssert {
 
-    class func textShownOnScreen(textToFind: String) {
-        let textPredicate = NSPredicate(format: "label MATCHES '" + textToFind + "'")
+    class func textShownOnScreen(text: String) {
+        let textPredicate = NSPredicate(format: "label MATCHES '" + text + "'")
         let staticText = app.staticTexts.element(matching: textPredicate)
 
-        XCTAssertTrue(staticText.exists, "\"" + textToFind + textNotFoundInWebView)
+        XCTAssertTrue(staticText.exists, "\"" + text + textNotFoundInWebView)
+    }
+
+    class func textsShownOnScreen(texts: [String]) {
+        for text in texts {
+            WebViewAssert.textShownOnScreen(text: text)
+        }
     }
 }
 
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -41,41 +41,43 @@ func getToAllNotes() {
 
 class Table {
 
+    class func getAllCells() -> XCUIElementQuery {
+        return app.tables.element.children(matching: .cell)
+    }
+
     class func getVisibleLabelledCellsNumber() -> Int {
         // We need only the table cells that have X = 0 and a non-empty label
         // otherwise we will include invisible elements from Sidebar pane, when Notes List is open
         // or the elements from Notes List when Settings are open
         // or the visible tags search suggestions
-        let cellsNum = app.tables.element.children(matching: .cell).count
-        var matchesNum: Int = 0
-        guard cellsNum > 0 else { return matchesNum }
+        let cells = Table.getAllCells()
+        var matches = 0
+        guard cells.count > 0 else { return matches }
 
-        for index in 0...cellsNum - 1 {
-            let cell = app.tables.cells.element(boundBy: index)
+        for cell in cells {
             if cell.frame.minX == 0.0 && cell.label.count > 0 {
-                matchesNum += 1
+                matches += 1
             }
         }
 
-        return matchesNum
+        return matches
     }
 
     class func getVisibleNonLabelledCellsNumber() -> Int {
         // We need only the table cells that have X = 0 and an empty label
         // Currently (besides using object dimentions) this is the way to
         // locate tags search suggestions
-        let cellsNum = app.tables.element.children(matching: .cell).count
-        var matchesNum: Int = 0
-        guard cellsNum > 0 else { return matchesNum }
+        let cells = Table.getAllCells()
+        var matches = 0
+        guard cells.count > 0 else { return matches }
 
-        for index in 0...cellsNum - 1 {
-            let cell = app.tables.cells.element(boundBy: index)
+        for cell in cells {
             if cell.frame.minX == 0.0 && cell.label.count == 0 {
-                matchesNum += 1
+                matches += 1
             }
         }
 
-        return matchesNum
+        return matches
     }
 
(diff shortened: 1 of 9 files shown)


SimplenoteUITests/Generic.swift
```diff
@@ -41,41 +41,43 @@ func getToAllNotes() {
...
         // We need only the table cells that have X = 0 and a non-empty label
         // otherwise we will include invisible elements from Sidebar pane, when Notes List is open
         // or the elements from Notes List when Settings are open
         // or the visible tags search suggestions
-        let cellsNum = app.tables.element.children(matching: .cell).count
-        var matchesNum: Int = 0
-        guard cellsNum > 0 else { return matchesNum }
+        let cells = Table.getAllCells()
+        var matches = 0
+        guard cells.count > 0 else { return matches }
```
mokagio: Nitpick

```suggestion
        guard cells.isEmpty == false else { return matches }
```

From a performance point of view, using `isEmpty` is better than counting all the elements when the only thing we care about is whether the array is empty or not.

SimplenoteUITests/Generic.swift
```diff
@@ -41,41 +41,43 @@ func getToAllNotes() {
...
-            let cell = app.tables.cells.element(boundBy: index)
+        for cell in cells {
             if cell.frame.minX == 0.0 && cell.label.count > 0 {
-                matchesNum += 1
+                matches += 1
             }
         }
 
-        return matchesNum
+        return matches
```
mokagio: Nitpick. I'm fond of using collections manipulations instead of for loops.

Here's an alternative way of writing this code:

```swift
return Table.getAllCells()
    .filter { $0.frame.minX == 0.0 && $0.label.isEmpty == false }
    .count
```

I did not run the tests with it, so it might need some tweaking, but the idea is to use filter to get only the cells that match the criteria defined in your comment and return the count. If no cell matches, the count will be 0.

The advantage of this approach, in my opinion, is that there are no variables in the code, so less chances for mutable state to creep in and result in bugs. There's also less code to read, which is, _usually_, a good thing.

SimplenoteUITests/Generic.swift
```diff
@@ -41,41 +41,43 @@ func getToAllNotes() {
...
-        var matchesNum: Int = 0
-        guard cellsNum > 0 else { return matchesNum }
+        let cells = Table.getAllCells()
+        var matches = 0
+        guard cells.count > 0 else { return matches }
 
-        for index in 0...cellsNum - 1 {
-            let cell = app.tables.cells.element(boundBy: index)
+        for cell in cells {
             if cell.frame.minX == 0.0 && cell.label.count > 0 {
```
mokagio: Same as above.

```suggestion
            if cell.frame.minX == 0.0 && cell.label.isEmpty == false {
```

SimplenoteUITests/Generic.swift
```diff
@@ -41,41 +41,43 @@ func getToAllNotes() {
...
-        var matchesNum: Int = 0
-        guard cellsNum > 0 else { return matchesNum }
+        let cells = Table.getAllCells()
+        var matches = 0
+        guard cells.count > 0 else { return matches }
 
-        for index in 0...cellsNum - 1 {
-            let cell = app.tables.cells.element(boundBy: index)
+        for cell in cells {
             if cell.frame.minX == 0.0 && cell.label.count == 0 {
```
mokagio:
```suggestion
            if cell.frame.minX == 0.0 && cell.label.isEmpty {
```

Using `isEmpty` is a better option even when the comparison is `== 0` instead of `> 0`.

SimplenoteUITests/Generic.swift
```diff
@@ -97,19 +99,17 @@ class Table {
     class func getCellsWithExactLabelCount(label: String) -> Int {
         let predicate = NSPredicate(format: "label == '" + label + "'")
         let matchingCells = app.cells.matching(predicate)
-        let matchesCount = matchingCells.count
-        print(">>> Found " + String(matchesCount) + " Cell(s) with '" + label + "' label")
-        return matchesCount
+        let matches = matchingCells.count
+        print(">>> Found \(matches) Cell(s) with '\(label)' label")
```
mokagio, line 87: +1 for using string interpolation here.

SimplenoteUITests/NoteEditor.swift
```diff
@@ -96,35 +96,35 @@ class NoteEditor {
     }
 
     class func getCheckboxesForTextCount(text: String) -> Int {
-        let matchesCount = NoteEditor.getTextViewsWithExactLabelCount(label: text)
-        print(">>> ^ Found " + String(matchesCount) + " Checkboxe(s) for '" + text + "'")
-        return matchesCount
+        let matches = NoteEditor.getTextViewsWithExactLabelCount(label: text)
+        print(">>> ^ Found " + String(matches) + " Checkbox(es) for '" + text + "'")
```
mokagio:
```suggestion
        print(">>> ^ Found \(matches) Checkbox(es) for '\(text)'")
```

SimplenoteUITests/NoteEditor.swift
```diff
@@ -96,35 +96,35 @@ class NoteEditor {
...
-            let currentValue = textViews.element(boundBy: index).value as! String
+        for textView in textViews {
+            let currentValue = textView.value as! String
             let currentValueStripped = currentValue.replacingOccurrences(of: "\u{fffc}", with: "")
 
             if currentValueStripped == value {
-                matchesCounter += 1
+                matches += 1
             }
         }
```
mokagio: Collection manipulations can be used here, too.

```swift
let matches = textViews
    .compactMap { ($0.value as? String)?.replacingOccurrences(of: "\u{fffc}", with: "") }
    .filter { $0 == value }
```

Once again, no variables and less code to read.

mokagio: By the way, speaking of reading code. I had to look up what the "\u{fffc}" was.

It might help future readers to encapsulate the value in a constant, e.g.:

```swift
let unicodeReplacementCharacter = "\u{fffc}"
```

Or, one could even define a `String` extension to document what the code above does

```swift
extension String {

    func strippingUnicodeReplacementCharacter() -> String {
        replacingOccurrences(of: "\u{fffc}", with: "")
    }
}
```

Which would make the code here clearer, IMHO:

```swift
let currentValueStripped = currentValue.strippingUnicodeReplacementCharacter()
```

SimplenoteUITests/XCTest+Extensions.swift
```diff
@@ -19,3 +19,17 @@ extension XCUIElement {
         self.typeText(text)
     }
 }
+
+extension XCUIElementQuery: Sequence {
```
mokagio: Nitpick. Since this is a port of open source code, let's give credit in the source, too.

This is also useful because, if something breaks here, we can check if the source has been updated or got there to suggest updates.

```suggestion
// Credits to https://github.com/onmyway133/blog/issues/628
extension XCUIElementQuery: Sequence {
```

mokagio (approved):
All tests passed on iPhone 12 Simulator.

None of my comments are blocking, but I'd like the `isEmpty` usages to be implemented at some point. Could be done here or in a dedicated PR auditing the test suite for usages of `count == 0` and `count > 0`.

SimplenoteUITests/XCTest+Extensions.swift
pachlava: Of course, fully agree. Added. 🙇

SimplenoteUITests/Generic.swift
pachlava: This is a beautiful improvement @mokagio ! Thank you!

SimplenoteUITests/NoteEditor.swift
pachlava: Thank you!

SimplenoteUITests/NoteEditor.swift
pachlava: @mokagio Again, this is a great improvement. 🙇 I've applied `filter` and `compactMap` everywhere where it was possible, instead of loops. In the next PR I will try to remove even more loops...

SimplenoteUITests/Generic.swift
pachlava: @mokagio After replacing loops with collections manipulations, as you suggested 🔥, the number of `count == 0` and `count > 0` usages went down a lot. But two cases of usage still remain, and in both cases the `count` was used on `XCUIElementQuery`, which is not an array, but a query, as I understood, and has no `isEmpty` value. 

I've started looking more into this, and saw that it becomes possible once I convert the `XCUIElementQuery` to array with `compactMap { $0 }`, for example. But, I guess, this operation involves traversing through all elements too, which makes the whole thing meaningless?

Also, I've read more, and [found](https://betterprogramming.pub/check-swift-array-is-empty-f15e63bf1026) that `isEmpty` is faster than `count` only for String, but for Array the performance is comparable, because Array conforms to [RandomAccessCollection](https://developer.apple.com/documentation/swift/randomaccesscollection).

I agree that `isEmpty` reads better, and can think of adding an `isEmpty()` extension `XCUIElementQuery`, which returns `count == 0`, but this looks not very honest, and only improves the look, but not the performance.

I'm not sure which way to go with this issue, and would appreciate your thoughts, as always.

I will leave this change for the future and meanwhile will merge this PR, with all your other suggestions 🚀  applied.

PR #1191, opened 2021-03-06
Improve Note Contextual Action VoiceOver Titles

### LINKED ISSUE #1018: Screen Reading Voice Over Improvements.
On IOS with Voiceover:

When in the list of all notes and swiping down for actions, the current
labels are ïcon trash", "icon pin", "icon link" and "icon share". That's
understandable, of course, but "delete", "pin', "copy link" and "share"
would be better.

Thank you. Also I just downloaded the IOS Beta and if there's something
in particular you would like me to test with Voiceover, I'm happy to.

Reported in 3487018-zen

I'm asking the user for more specific device information.

charliescheer:
### Fix
This is a small PR that fixes the issue #1018 

When you are in Simplenote iOS using VoiceOver, there are not currently titles or accessibility labels set for the actions, so Simplenote defaults to the title of the icon being used on the image. In this PR I have added titles to the UIContextualActions so that they will have a title when accessed via VoiceOver.

### Test
1. Activate Voiceover on your device
2. Go to the Note list
3. Tap on a note once to select it
4. Make repeated swipe down actions to hear the different options in the contextual menu

The old titles were:
"icon_trash"
"icon_pin" or "icon_unpin"
"icon_share"
"icon_link"

I updated them to the existing action titles so now they should be:
"Move to Trash"
"Pin to Top" or "Unpin"
"Share"
"Copy Internal Link"

In the ticket that alerted us to this issue, the user asked for very basic descriptions trash, pin, share, link.  I put in more verbose descriptions, partly because they were already there and partly because I thought they might be more helpful if they were more descriptive. @SylvesterWilmott can you take a look and see if you have any suggestions for updating these accessibility labels?

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -700,24 +700,25 @@ private extension SPNoteListViewController {
 
     func regularContextActions(for note: Note) -> [UIContextualAction] {
         let pinImageName: UIImageName = note.pinned ? .unpin : .pin
+        let pinActionTitle: String = note.pinned ? ActionTitle.unpin : ActionTitle.pin
 
         return [
-            UIContextualAction(style: .destructive, image: .image(name: .trash), backgroundColor: .simplenoteDestructiveActionColor) { [weak self] (_, _, completion) in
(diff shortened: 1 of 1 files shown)


eshurakov (approved):
Looks good to me, thank you!

:shipit:

pachlava:
Using iPhone XR + iOS 14.4.1 + Simplenote 4.34.0.0:

✔️ VoiceOver description for context actions in note list are `Share`, `Copy Internal link`,  `Pin to Top` (`Unpin`) and `Move to Trash`

Would it be nice to implement the same for Trash view?

PR #1192, opened 2021-03-08
Merge version 4.33 beta 1 into develop

mokagio:
This is the first step of the release cycle.

This PR is important because, once merged, a bot will pick up the new localized strings and upload them to [GlotPress](https://translate.wordpress.com/) for translation.


Scripts/extract_release_notes.rb
```diff
@@ -0,0 +1,59 @@
+# Parses the release notes file to extract the current version information and
+# puts it to STDOUT.
+#
+# To update the release notes for localization:
+#
+# ruby ./this_script >| Simplenote/Resources/release_notes.txt
+#
+# To generate the GitHub release copy:
+#
+# ruby ./this_script -k | pbcopy
```
mokagio: I'm sure there are many improvements we could make on this script. Eventually, it would be great for the release toolkit to actually do it, but until each project has a different convention for the release notes formatting, that might not be feasible.

The current iteration does the job for what I need to do right now.

Scripts/extract_release_notes.rb
```diff
@@ -0,0 +1,81 @@
...
+#
+# ruby ./this_script >| Simplenote/Resources/release_notes.txt
+#
+# To generate the App Store Connect release message:
+#
+# ruby ./this_script | pbcopy
+#
+# To generate the GitHub and App Center release message:
+#
+# ruby ./this_script -k | pbcopy
```
mokagio, line 14: I'm sure there are many improvements we could make on this script. Eventually, it would be great for the release toolkit to actually do it, but until each project has a different convention for the release notes formatting, that might not be feasible.

The current iteration does the job for what I need to do right now.

loremattei (approved):
LGTM!

PR #1196, opened 2021-03-08
Accessibility Color updates

### LINKED ISSUE #760: Accessibility: New color scheme mismatches with accessibility contrast setting enabled
### Expected

I expect the new Simplenote blue introduced in https://github.com/Automattic/simplenote-ios/pull/748 to be consistent throughout the app.

### Observed

When I have the "Increase Contrast" accessibility setting enabled on my device, there are some places where the blue doesn't match. It looks like the increased contrast makes it a darker blue in most places, but in a few places the non-contrast (lighter) blue appears.

### Reproduced

1. Open the Simplenote app and notice the new, consistent blue throughout.
2. Go to Settings app > Accessibility > Display & Test Size and enable Increase Contrast.
3. Open the Simplenote app again.
4. Search for a note and notice that the highlighted text matching the search is a lighter blue than the Cancel button in the top right.
5. Open a note in the editor and add a tag. Notice that the tag is a lighter blue than the text/icons in the top navigation bar and lighter than any linkified URLs in the note.
6. Preview a markdown note. Notice that the Back button text is a lighter blue than the back arrow.

-|-|-|-
Apple|iPhone XS|13.4.1|4.19.0.0 TestFlight beta

charliescheer:
### Fix
mostly fixes #760 
This is a small PR that fixes some issues with colors.

In this PR I have updated the color that markdown preview uses for links, which was out of sync with the current design colors. 

This also adjusts the markdown preview link colors when accessibility contrast is enabled so that the link colors there also update

Note that these changes will only work on iOS 13 and later

Note that this PR does not fix one of the issues referenced in the original issue where in high contrast, the search results highlighting and the search cancel button don't have the same color.  The high contrast settings is done automatically by iOS, and because the size and boldness of the cancel button it will always change to a higher contrast color where the search results won't.  To interrupt that was gonna be a pretty big lift, after some discussion we decided to leave it as is.

### Test
On a device with iOS 13 or later:
1. create a note with a link in it and activate markdown
2. enter the markdown preview and confirm that the color for the back button and the link are the same
3. Enable dark mode on check the colors match
4. Enable high contrast and make sure the colors still match
5. disable dark mode (so you are now in high contrast/light mode) and check the colors again

### Release

### DIFF
--- a/Simplenote/Classes/SPMarkdownParser.m
+++ b/Simplenote/Classes/SPMarkdownParser.m
@@ -50,13 +50,13 @@
             "<style media=\"screen\" type=\"text/css\">\n";
     NSString *headerEnd = @"</style></head><body><div class=\"note-detail-markdown\">";
 
-    NSString *css = [SPMarkdownParser loadCSSAtPath:self.cssPath];
+    NSString *css = [self loadCSSAtPath:self.cssPath];
 
     // Check if increase contrast is enabled.  If enabled amend CSS to include high contrast colors
     //
     if (@available(iOS 13.0, *)) {
         if ([[UITraitCollection currentTraitCollection] accessibilityContrast] == UIAccessibilityContrastHigh) {
-            NSString *contrast = [SPMarkdownParser loadCSSAtPath:self.contrastCssPath];
+            NSString *contrast = [self loadCSSAtPath:self.contrastCssPath];
             css = [css stringByAppendingString:contrast];
         }
     }
--- a/Simplenote/Classes/SPMarkdownParser.m
+++ b/Simplenote/Classes/SPMarkdownParser.m
@@ -50,19 +50,24 @@
             "<style media=\"screen\" type=\"text/css\">\n";
     NSString *headerEnd = @"</style></head><body><div class=\"note-detail-markdown\">";
 
-    NSString *css = [NSString stringWithContentsOfURL:[[NSBundle mainBundle] URLForResource:self.cssPath withExtension:nil]
-                                             encoding:NSUTF8StringEncoding error:nil];
+    NSString *css = [SPMarkdownParser loadCSSAtPath:self.cssPath];
 
     // Check if increase contrast is enabled.  If enabled amend CSS to include high contrast colors
     //
     if (@available(iOS 13.0, *)) {
         if ([[UITraitCollection currentTraitCollection] accessibilityContrast] == UIAccessibilityContrastHigh) {
-            NSString *contrast = [NSString stringWithContentsOfURL:[[NSBundle mainBundle] URLForResource:self.contrastCssPath withExtension:nil]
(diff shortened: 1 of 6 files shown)


Simplenote/Classes/SPMarkdownParser.m
```diff
@@ -52,6 +52,15 @@ + (NSString *)htmlHeader
 
     NSString *css = [NSString stringWithContentsOfURL:[[NSBundle mainBundle] URLForResource:self.cssPath withExtension:nil]
                                              encoding:NSUTF8StringEncoding error:nil];
+
+    // Check if increase contrast is enabled.  If enabled amend CSS to include high contrast colors
+    //
+    if (@available(iOS 13.0, *)) {
+        if ([[UITraitCollection currentTraitCollection] accessibilityContrast] == UIAccessibilityContrastHigh) {
+            NSString *contrast = [NSString stringWithContentsOfURL:[[NSBundle mainBundle] URLForResource:self.contrastCssPath withExtension:nil]
+                                                         encoding:NSUTF8StringEncoding error:nil];
```
jleandroperez: Perhaps it's worth extracting this to an API? (duplicates the mechanism in L53).

```
- (NSString *)loadCSSAtPath:(NSString *)path;
```

Simplenote/Resources/markdown-dark-contrast.css
```diff
@@ -0,0 +1,5 @@
+
+/* Accessibility high contrast blue */
+.note-detail-markdown a {
```
jleandroperez, line 3: Question: I've noticed the `.note-detail-markdown a` style ends up duplicate (with the second entry being the one we want).

Do we need to use `!important` to signal this one should superceed the first one?

charliescheer: I had it in before, but removed it cause it would work without it.  Back in support, the CSS guild would always shy away from !important for... reasons, but yeah I think it is probably better to have it just in case since we can't do more specific targeting

Simplenote/Classes/SPMarkdownParser.m
```diff
@@ -50,21 +50,26 @@ + (NSString *)htmlHeader
             "<style media=\"screen\" type=\"text/css\">\n";
     NSString *headerEnd = @"</style></head><body><div class=\"note-detail-markdown\">";
 
-    NSString *css = [NSString stringWithContentsOfURL:[[NSBundle mainBundle] URLForResource:self.cssPath withExtension:nil]
-                                             encoding:NSUTF8StringEncoding error:nil];
+    NSString *css = [SPMarkdownParser loadCSSAtPath:self.cssPath];
```
jleandroperez: Since this method is static, you don't really need to write the classname: 

`[self loadCSSAtPath:...]` should work as well 🔥

jleandroperez (approved):
Looks super, thank you for fixing this!!

:shipit:

PR #1200, opened 2021-03-09
Added support for accessibility three finger navigation

### LINKED ISSUE #1172: Cannot switch between Markdown editor and preview if accessibility VoiceOver feature is active
Reported on 3757989-zen

> Using the IOS built in screen-reader Voiceover, I want to be able to access the preview function for notes which are written in Markdown. On Windows the command is control+shift.p. I read your blog post on Markdown which indicated you could somehow swipe to trigger a preview on the IOS app, but if there's a way to make that work with Voiceover I haven't found it yet. So the easiest I think would to add a "Preview" button.

Details on accessibility options can be found at https://www.apple.com/accessibility/vision/

### Expected
To be able to swipe between the Markdown editor and preview using either two-fingers (like default) or three-fingers (specific when VoiceOver is active).

### Observed
We cannot switch between editor and preview modes.

### Reproduced

1. In your device, go to Settings > Accessibility > VoiceOver, and enable it. Note that this will change the way your phone works.
2. Open the Simplenote app by double tapping (if is on a different sub-screen, you can use three-finger swipe to move through) 
3. Try to switch between the editor and the preview mode in any of your notes, either by two-finger or three-finger swiping, this is not responsive.

The user suggest we could add a simple button if VoiceOver is active in order to switch between modes.

Reproduced on Apple | iPhone7 | iOS 14.2 | Simplenote 4.31

charliescheer:
### Fix
This PR fixes #1172 

This is a small PR that deals with a problem in SNiOS where if you are using Voiceover there is no way to open markdown preview mode.

With accessibility voiceover enabled, single finger swipe does not work like a standard swipe.  Generally this is replaced with a 3 finger swipe.  I have added actions in the note editor where 3 finger swipe left will open markdown preview and three finger swipe right will dismiss the note and return to the note list.

### Test
1. enable voiceover on the device that you are testing with
2. navigate to a note that has markdown enabled
3. in voiceover select some text in the note (if navigation is selected the swipe actions don't work)
4. do a three finger swipe from right to left, this will should open markdown
5. tap on the back button to return to the note, now three finger swipe from left to right, this should dismiss the note

### Review

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
@@ -244,19 +244,36 @@ extension SPNoteEditorViewController: KeyboardObservable {
 
 // MARK: - Voiceover Support
 //
-private extension SPNoteEditorViewController {
+extension SPNoteEditorViewController {
 
     /// Indicates if VoiceOver is running
     ///
-    var isVoiceOverEnabled: Bool {
+    private var isVoiceOverEnabled: Bool {
         UIAccessibility.isVoiceOverRunning
     }
 
     /// Whenver VoiceOver is enabled, this API will lock the Tags List in position
     ///
     @objc
-    func refreshVoiceOverSupport() {
+    private func refreshVoiceOverSupport() {
         updateTagListPosition()
+    }
+
+    /// Sets behavior for accessibility three finger scroll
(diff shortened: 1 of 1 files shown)


jleandroperez (approved):
Looks great @charliescheer !!!

:shipit:

PR #1201, opened 2021-03-10
Merge finalized release 4.33 into develop

mokagio:
Doing this earlier to re-enable the signup flow. Internal ref. paaHJt-1Sn-p2.


fastlane/Fastfile
```diff
@@ -106,6 +106,44 @@ end
...
+      end
+
+      until keywords.length <= app_store_connect_keywords_length_limit do
+        keywords = keywords.split(separator)[0...-1].join(separator)
+        puts keywords
+      end
+
+      File.write(keywords_path, keywords)
+    end
+  end
```
mokagio, line 145: I got tired of manually tweaking the `sa-AR` keywords every submission, so I wrote this automation to do it.

Then, I realized that the best place to run this is after downloading the translations from GlotPress, so that the problem doesn't arise. That will require updating the Swift download script, and since I already invested time on this today, I decided to leave it for another time.

AliSoftware: I'm not sure I'm a fan of this.

1. I think this kind of fix should be fixed at the root of its issue – i.e. on GlotPress (and/or in ADC) once and for all, where those keywords get pulled from.
2. I'm confused why the `ar-SA` locale use an arabic comma in its keywords; I thought that AppStore Connect would always expect an ASCII comma U+002C as the separator for keywords – whatever the typographic comma is in the locale in the context of a regular sentence – because that U+002C comma is used to separate keywords (i.e. to treat the long string as if it were an array of words), not to use an actual semantic comma within a regular sentence…
   * I'm not even sure that using the arabic comma in the keywords field make them be considered be separate keywords for ASO / AppStore search, it might even be that they are considered like a giant / very long single keyword…
   * Probably translators of the keywords don't now about this and use arabic commas like they'd do when using semantic commas in regular arabic sentences. In that case I think the right think to do is, again, to fix that at its source, i.e. in GlotPress

---

I'm all for having our lane generate a **warning** with the keywords we pull from GlotPress exceed the character limit… or use the wrong type of commas. And when that happens I'm ok if we do a quick one-time fix for this time around so that it doesn't block the release for this time.
But then I think the next step is to ping #i18n team to ask them to fix the keywords in GlotPress for the problematic locale(s), so that it would not happen anymore at all during the future releases. Especially since maybe it would make more sense (for #i18n folks and marketing / data folks) to trim different keywords to make it fit rather than always removing the last one(s), so automating the removal of the last keywords like you did here might not always be the right move.

AliSoftware (commented):
Overall LGTM, but I'm not a fan of the new lane to fix keywords (and its consequences) as I feel like that's not the right solution for the problem 🤔

jleandroperez (approved):
:shipit: on my end sir!

fastlane/Fastfile
mokagio: @AliSoftware thanks for looking into this. You made realized I was working with the wrong assumption. 

When I wrote that code, I had in my mind that the keywords changed on every release. _I'm not sure why, as it doesn't make much sense in hindsight. Maybe because I got the error from running `deliver` which I associate to updating release-specific metadata, even though it does more than that, of course._ 🤔 🤷‍♂️ 

If they were ephemeral, so to speak, trimming them seems to me like a reasonable compromise to get through the release process smoothly. 

But ☝️, that's not the case for the keywords. In such case it makes total sense to fix the problem at the root. _It_ always _makes sense to fix the problem at the root, of course, but the fact that the keywords don't change that often means that trimming them every time on the client is indeed "not the right solution for the problem"._

I'm going to merge this PR and follow up with one that removes the lane 👍

PR #1202, opened 2021-03-10
Merge release 4.33 into trunk

mokagio:
Doing this earlier to re-enable the signup flow. Internal ref. paaHJt-1Sn-p2.

Leaving this as a draft till we actually release it. Still, I'd appreciate a review anyway.

Thanks 🙇‍♂️ .


Scripts/extract_release_notes.rb
```diff
@@ -0,0 +1,81 @@
...
+# ruby ./this_script -k | pbcopy
+
+GITHUB_URL = 'https://github.com/Automattic/simplenote-ios'
+
+RELEASE_NOTES_FILE = 'RELEASE-NOTES.txt'
+NOTES = File.read(RELEASE_NOTES_FILE)
+lines = NOTES.lines
+
+def replace_pr_number_with_markdown_link(string)
+  string.gsub(/\#\d*$/) do |pr_number|
```
AliSoftware, line 23: You probably wanted `\d+` not `\d*`, otherwise this will match singled-out `#` too 😉

Scripts/extract_release_notes.rb
```diff
@@ -0,0 +1,81 @@
...
+    "[#{GITHUB_URL}/pull/#{pr_number.gsub('#', '')}](#{pr_number})"
+  end
+end
+
+# This is a very bare bone option parsing. It does the job for this simple use
+# case, but it should not be built upon.
+#
+# If you plan to add more options, please consider using a gem to manage them
+# properly.
```
AliSoftware, line 32: note: actually `OptionParser` is part of ruby's stdlib, so no need to use an external gem if you wanted to do it properly 😛 – https://ruby-doc.org/stdlib-2.7.2/libdoc/optparse/rdoc/OptionParser.html

Scripts/extract_release_notes.rb
```diff
@@ -0,0 +1,81 @@
...
+# ruby ./this_script | pbcopy
+#
+# To generate the GitHub and App Center release message:
+#
+# ruby ./this_script -k | pbcopy
+
+GITHUB_URL = 'https://github.com/Automattic/simplenote-ios'
+
+RELEASE_NOTES_FILE = 'RELEASE-NOTES.txt'
+NOTES = File.read(RELEASE_NOTES_FILE)
```
AliSoftware, line 19: Don't we have another `.txt` file that have the release notes just for the current version (already extracted and isolated) by one of our release-toolkit lanes/actions already (somewhere in `fastlane/metadata/`)?

Parsing that other file instead of the `RELEASE-NOTES.txt` would allow you to simplify that script a lot by avoid to have to do the whole "ignore the first empty section for next version" + "ignore the current version header" + "parse up until next empty line" (as that would have already been done by our release-toolkit step), and make this script only have to do the `gsub(/^- */,'- ')` part + the `replace_pr_number_with_markdown_link` call?

Scripts/swiftlint.sh
```diff
@@ -0,0 +1,54 @@
...
+  echo "warning: skipping SwiftLint build phase because running on CI."
+  exit 0
+fi
+
+# Get the directory of this file.
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+# Temporarily move to the root directory so that SwiftLint can correctly
+# find the paths returned from the `git` commands below.
+pushd $DIR/../ > /dev/null
```
AliSoftware, line 21: You probably want quotes here, in case `$DIR` contains spaces

fastlane/Fastfile
```diff
@@ -106,6 +106,44 @@ end
...
+
+      until keywords.length <= app_store_connect_keywords_length_limit do
+        keywords = keywords.split(separator)[0...-1].join(separator)
+        puts keywords
+      end
+
+      File.write(keywords_path, keywords)
+    end
+  end
+
```
AliSoftware, line 146: Same q as in https://github.com/Automattic/simplenote-ios/pull/1201#discussion_r591356015 here 🤷

fastlane/Fastfile
```diff
@@ -493,6 +531,13 @@ end
       # But fail completely after those 3 retries
       stop_after_first_error: true,
 
+      # Run one Simulator at a time. One of the screenshots requires adding
+      # text to a note but because we're using the same account across all the
+      # Simulators, editing the note on one will result in the changes
+      # appearing on the other making for unexpected and inconsistent
+      # screnshots
```
AliSoftware, line 538: Great to have documented/commented about this 👍 👏

AliSoftware (approved):
Overall LGTM.

Left some comments on a couple of your scripts – since that's the first time I have the occasion to see/read them, took the occasion to give some pointers – but since they are not really related to the "merge release to trunk" part of this PR and were already merged upstream, it's just food for thoughts that you might want to fix in a separate PR one day, so those should probably not block the merge of this PR.

Scripts/extract_release_notes.rb
mokagio: 🎉

mokagio:
Thanks for the input on the script @AliSoftware! As you suggest, I'll address it in a dedicated PR 🙇‍♂️ 

This has now been published on the Store: Marking it "Ready for Review" and merging it.

PR #1203, opened 2021-03-10
Add Notice Controller and Presenter to In App Notifications

### LINKED ISSUE #1178: Project In App Notifications - Presenter/Controller Implementation
Details:

- [x] Implement Controller class
- [x] Implement Presenter class
- [x] connect the two classes
- [x] connect classes together with View classes to test implementation

charliescheer:
### Fix
This PR Implements the controller class and Presenter class for in app notifications.  Implements #1178 

With the addition of the controller and presenter classes inserting Notice views is much easier, the notices respond to keyboard location, handle changing orientation of device, animate in, and animate to disappear

The notices will look much as they did in the Notice View PRs, as the appearance has not changed much.

### Test
To test, first you will need to go to the bin and get the code to modify the the note option buttons to display notices.
You can find the presentation code [here](https://mc.a8c.com/pb/289a8/)

Once you have added that code to the SPNoteEditorViewController+extensions.swift file test out the following things:
QUEUEING
1. go to a note and tap on the information button to see a notice without an action
2. tap on the new note button to see a notice with an action, tap on the action to make sure that works (will print to the console)
3. Tap on the information button again, and while the notice is presenting tap on the new note button.  This should queue the notice w/action to present when the first notice is finished
4. Tap on the information button, after the notice appears tap on it again.  A duplicate notice should not get queued and then presented

KEYBOARD AND ORIENTATION
5. Tap on the information button again and while the notice is there, change the screen orientation, the notice should recenter it's self
6. Display a notice and then make the keyboard appear and disappear, then notice should track to the keyboard
7. Test on an iPad with floating keyboard.  If the keyboard is floating the notice will go to the bottom position (same as if the keyboard was hidden)

### Review

Note there is some weirdness with the commits, some of them are duplicated and some have been reverted because of an update to the package.resolved file that wouldn't go away.  That change was committed but has been reverted.


Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,89 @@
...
+    private var current: Notice?
+    private let noticePresenter = NoticePresenter()
+
+    private var activeViewIsBeingTouched: Bool = false
+
+    private init() { }
+
+    func setupNoticeController() {
+        noticePresenter.startListeningToKeyboardNotifications()
+    }
```
eshurakov, line 31: Why not to call `startListeningToKeyboardNotifications` in the init of `NoticePresenter`?

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
+import UIKit
+
+class NoticePresenter: KeyboardObservable {
+
+    // MARK: Properties
+    //
+    var containerView = PassthruView(frame: .zero)
```
eshurakov: Is there a benefit of having a container view vs directly adding a notice to the window?

Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,34 +0,0 @@
...
-        "state": {
-          "branch": null,
-          "revision": "499d2809d169fcbeb9ff75568d9f1f937f290ffc",
-          "version": "1.3.0"
-        }
-      }
-    ]
-  },
-  "version": 1
-}
```
eshurakov: We should not delete this file because it has pinned versions of libraries.

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Confirms if a notice is already contained in notices queue. Appends to queue if new
+    ///
+    private func appendToQueueIfNew(_ notice: Notice) {
+        if notices.contains(notice) {
+            return
+        }
+
+        if let current = current {
```
eshurakov: There is no need to unwrap optional values when you compare, so you can just write `if notice == current `

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
...
+            return .zero
+        }
+        return window.frame
+    }
+}
+
+private struct Times {
+    static let waitShort = 1.5
+    static let waitLong = 2.75
+    static let animationTime = TimeInterval(0.5)
```
eshurakov: There are some animation constants defined in `UIKitConstants.swift`

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
+import UIKit
+
+class NoticePresenter: KeyboardObservable {
+
+    // MARK: Properties
+    //
+    var containerView = PassthruView(frame: .zero)
+    var noticeView: NoticeView?
+    var noticeBottomConstraint: NSLayoutConstraint?
```
eshurakov: I think some of the vars don't need to be public?

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
...
+    func presentNoticeView(_ noticeView: NoticeView, completion: @escaping () -> Void) {
+        guard let keyWindow = getKeyWindow() else {
+            return
+        }
+        self.noticeView = noticeView
+
+        containerView.addSubview(noticeView)
+        keyWindow.addFillingSubview(containerView)
+
+        noticeBottomConstraint = noticeView.bottomAnchor.constraint(equalTo: containerView.bottomAnchor, constant: Constants.notificationStartingPosition)
```
eshurakov: Relying on a fixed value to hide the view is not the best practice as the height can vary. The best is to have initial constraint from the top of notice view to the bottom of container view. Then switch it to "bottom to bottom".

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
...
+        UIView.animate(withDuration: Times.animationTime) {
+            self.containerView.layoutIfNeeded()
+        } completion: { (_) in
+            DispatchQueue.main.asyncAfter(deadline: .now() + delay) {
+                completion()
+            }
+        }
+    }
+
+    private func configureOnScreenConstraint() -> CGFloat {
```
eshurakov: From the method name I would expect that this methods updates the constraint and doesn't return anything.

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
...
+        guard let endFrame = endFrame,
+              let animationCurve = animationCurve,
+              let animationDuration = animationDuration else {
+            return
+        }
+        setStoredStates(frame: endFrame)
+        animateNoticeToNewKeyboardLocation(frame: endFrame, curve: animationCurve, animationDuration: animationDuration)
+    }
+
+    private func setStoredStates(frame: CGRect) {
```
eshurakov: We don't usually use `set*` and `get*` in method names in Swift. 
Maybe something like `updateKeyboardHeight(with frame: CGRect)`

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
...
+        let windowFrame = getWindowFrame()
+        keyboardVisible = frame.height != .zero
```
eshurakov: Do we need to have `keyboardVisible`? If the keyboard is hidden, the intersection with window frame should be zero and won't affect our constraint.

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
...
+    //
+    private func getKeyWindow() -> UIWindow? {
+        return UIApplication.shared.windows.first(where: { $0.isKeyWindow })
+    }
+
+    private func getWindowFrame() -> CGRect {
+        guard let window = getKeyWindow() else {
+            return .zero
+        }
+        return window.frame
```
eshurakov: Can be written as `getKeyWindow()?.frame ?? .zero`

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
...
+
+private struct Times {
+    static let waitShort = 1.5
+    static let waitLong = 2.75
+    static let animationTime = TimeInterval(0.5)
+    static let tapWait = TimeInterval(3)
+}
+
+private struct Constants {
+    static let notificationBottomMargin = CGFloat(-150)
```
eshurakov: Not used?

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,172 @@
...
+    static let notificationStartingPosition = CGFloat(100)
+    static let notificationOnScreenPosition = CGFloat(-50)
```
eshurakov: It's hard to immediately understand what `position` means in this case. Maybe `bottomMargin` would be a better fit?

Simplenote/NoticeView.swift
```diff
@@ -9,6 +9,7 @@ class NoticeView: UIView {
 
     // MARK: Properties
     //
+    @IBOutlet weak var backgroundView: UIView!
```
eshurakov: Can be private

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,89 @@
...
+
+        noticePresenter.presentNoticeView(noticeView) {
+            if !self.activeViewIsBeingTouched {
+                self.dismiss(noticeView)
+            }
+        }
+    }
+
+    private func dismiss(_ noticeView: NoticeView) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + Times.tapDelay) {
```
eshurakov: Why do we need a delay here? 🤔

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,89 @@
...
+    private var activeViewIsBeingTouched: Bool = false
+
+    private init() { }
+
+    func setupNoticeController() {
+        noticePresenter.startListeningToKeyboardNotifications()
+    }
+
+    func present(_ notice: Notice) {
+        if noticePresenter.isPresenting {
```
eshurakov: Do you think we can use `current != nil` here?

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,89 @@
...
+
+    func noticePressEnded() {
+        activeViewIsBeingTouched = false
+
+        guard let noticeView = noticePresenter.noticeView else {
+            return
+        }
+        dismiss(noticeView)
+    }
+}
```
eshurakov: Wonder if this logic should be part of presenter along with the timer?

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,89 @@
...
+                }
+            }
+        }
+    }
+
+    private func makeNoticeView(from notice: Notice) -> NoticeView {
+        let noticeView: NoticeView = NoticeView.instantiateFromNib()
+        noticeView.message = notice.message
+        noticeView.actionTitle = notice.action?.title
+        noticeView.handler = notice.action?.handler
```
eshurakov, line 70: Should it also be dismissed when the user taps on the action?

eshurakov (commented):
Sending you some comments 💬 

Also thinking out loud:
Right now presenter can work only with one notice at the time because it stores notice view and other notice related vars. What if we create a notice specific presenter each time we need to present it. That way we can have less optionals and less state changes inside of presenter.

Simplenote/NoticePresenter.swift
charliescheer: The container view allows me to call layoutIfNeeded on something other than the window.  I think I was running into issues if I called it on the notice view it's self, and I had been advised against layout out the entire window just to move one object

Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
charliescheer: There has been something really odd going on with this branch, it is continuously creating and deleting this file.  There are several commits I have since reverted where that was deleted or added.  I will make sure it is there and correct

Simplenote/NoticeController.swift
charliescheer: Originally I did have them in the init, but I was having an issue with timing.  The Presenter wasn't being initialized by the controller until the first notice was created.  
IF the keyboard already was displayed when that happened, the presenter wouldn't know the current keyboard state so it would present the notices as if the keyboard wasn't there.
By moving start listening to it's own function, I could call it when the app was starting so it began listening right away.

Simplenote/NoticePresenter.swift
charliescheer: Good to know.  Moved these times to the UIKit constants

Simplenote/NoticePresenter.swift
charliescheer: nice super simple

Simplenote/NoticePresenter.swift
charliescheer: oops :)

Simplenote/NoticeController.swift
charliescheer: nope

Simplenote/NoticeController.swift
charliescheer: My thought here was that the controller acts as the triggering mechanism for changes to the notices, and the presenter handles the animations.  So the animation should certainly be in the presenter, I moved that.

I have left the tap responding in the controller, mostly because I am retriggering the dismiss from the controller once the press is finished.  If I trigger the animation dismissal from the presenter, I don't have a way of letting the controller know that the dismissal is complete currently.  By having the controller handle the long press, it can re trigger it's dismissal method so the completion handler can be resolved to the controller when animation is complete

Simplenote/NoticeController.swift
charliescheer: That wasn't in the original design description, but I have wondered if there should be some sort of swipe to dismiss or tap to dismiss the notices.

@SylvesterWilmott WDYT?

charliescheer:
@eshurakov thanks for the suggestions!  I have updated the PR code to reflect many of the suggestions.  If you get a chance let me know what you think of the changes!

Simplenote/NoticeController.swift
jleandroperez: I think this can now be addressed?

Since the Presenter is **let** and `startListeningToKeyboardNotifications` only hooks up the keyboard notifications, (unless I'm missing something) you could now simplify things up by moving this initialization inside the `init()`

Simplenote/NoticeController.swift
jleandroperez: +1 it'd probably be best to guard against the `current` ivar (since that's the actual property that determines the current state)

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        noticePresenter.presentNoticeView(noticeView) {
+            if !self.activeViewIsBeingTouched {
+                self.dismiss(noticeView)
+            }
+        }
+    }
+
+    private func dismiss(_ noticeView: NoticeView) {
+        self.noticePresenter.dismissNotification {
```
jleandroperez:
```suggestion
        noticePresenter.dismissNotification {
```

Simplenote/NoticeController.swift
jleandroperez: +1 otherwise the action could be executed N times

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,86 @@
...
+        noticeView.handler = notice.action?.handler
+        noticeView.delegate = self
+
+        return noticeView
+    }
+
+    /// Confirms if a notice is already contained in notices queue. Appends to queue if new
+    ///
+    private func appendToQueueIfNew(_ notice: Notice) {
+        if notices.contains(notice) {
```
jleandroperez: We don't really need to guard against this. If the `present` API was invoked multiple times, it would be a bug in the caller side (and shouldn't be handled by the Controller itself).

Same logic applies when using Arrays. When we invoke `array[index]`, if the `index` isn't valid, the system will crash (because the problem lies in the caller's code)

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,86 @@
...
+extension NoticeController: NoticePresentingDelegate {
+    func noticePressBegan() {
+        if !noticePresenter.isPresenting {
+            return
+        }
+        guard let timer = noticePresenter.timer else {
+            return
+        }
+
+        timer.invalidate()
```
jleandroperez: Perhaps this snippet should live inside the Presenter itself?

```
presenter.invalidateTimerIfNeeded()
```

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    func noticePressEnded() {
+        guard let noticeView = noticePresenter.noticeView else {
```
jleandroperez: This makes me think that perhaps the `NoticePresentingDelegate` needs an update:

```
protocol NoticePresentingDelegate: class {
    func noticePressBegan(sender: NoticeView)
    func noticePressEnded(sender: NoticeView)
}
```

This would allow you to drop the guard, and, in theory, would make it possible to deal with multiple Notices onscreen without a glitch 🔥

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,170 @@
+import UIKit
+
+class NoticePresenter: KeyboardObservable {
+
+    // MARK: Properties
+    //
+    private var containerView = PassthruView(frame: .zero)
+    var noticeView: NoticeView?
+    private var noticeVariableConstraint: NSLayoutConstraint?
```
jleandroperez, line 9:
```suggestion
    private var noticeTopConstraint: NSLayoutConstraint?
```

Nipticky: Would improve a biiit readability (all of the constraints are, technically, variable!)

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,170 @@
...
+
+        containerView.addSubview(noticeView)
+        keyWindow.addFillingSubview(containerView)
+
+        noticeVariableConstraint = noticeView.topAnchor.constraint(equalTo: containerView.bottomAnchor)
+        noticeVariableConstraint?.isActive = true
+        noticeView.centerXAnchor.constraint(equalTo: containerView.centerXAnchor).isActive = true
+        containerView.layoutIfNeeded()
+
+        displayNotificationView() {
```
jleandroperez: `noticeVariableConstraint` is set in L38 and immediately after, it's replaced in L71.

You can probably drop L37 + L38 + L40

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    private func displayNotificationView(completion: @escaping () -> Void) {
+        prepareConstraintToDisplayNotice()
+
+        UIView.animate(withDuration: UIKitConstants.animationLongDuration) {
+            self.containerView.layoutIfNeeded()
+        } completion: { (_) in
+            completion()
+        }
```
jleandroperez, line 55: Tip: If you match the UIView.animate's completion signature, you can do this:


```
        UIView.animate(withDuration: ..., animations: {
// ...
        }, completion: onCompletion)
```

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    private func makeBottomConstraintConstant() -> CGFloat {
```
jleandroperez:
```suggestion
    private var noticeBottomPadding: CGFloat {
```

Typically all of the `make*` APIs would return objects (but in this particular case it's a calculation)

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,170 @@
...
+        noticeVariableConstraint = noticeView?.bottomAnchor.constraint(equalTo: containerView.bottomAnchor, constant: constant)
+        noticeVariableConstraint?.isActive = true
+    }
+
+    func dismissNotification(completion: @escaping () -> Void) {
+        guard let noticeView = noticeView else {
+            return
+        }
+
+        let delay = noticeView.handler == nil ? UIKitConstants.animationDelayExtraLong : UIKitConstants.animationDelayExtraExtraLong
```
jleandroperez: Perhaps it'd be clear to move those constants as a local `NoticePresenter` thing? (since it's not really an animation delay, but a "Dismissal timeout" or so?

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,170 @@
...
+
+    @objc
+    private func dismissAnimation(noticeView: NoticeView, completion: @escaping () -> Void) {
+        UIView.animate(withDuration: UIKitConstants.animationLongDuration) {
+            noticeView.alpha = .zero
+        } completion: { (_) in
+            self.noticeView?.removeFromSuperview()
+            self.containerView.removeFromSuperview()
+            self.noticeView = nil
+            self.timer = nil
```
jleandroperez: I'm not 100% sure but I **think** you need to invalidate the timer first perhaps?

jleandroperez (commented):
Great progress Chrarlie!!! 👏 

Sending you few nipticky comments. Love how this is shaping!

Simplenote/NoticeController.swift
eshurakov: @jleandroperez `static let shared = NoticeController()` is lazy, so we will still need to somehow access / instantiate it (and presenter along the way) when the app starts

Simplenote/NoticeController.swift
eshurakov: I think in this scenario it is easier and less code to handle this here once rather than on the caller's side. 

Let's say we show a notice when you try to log in and forgot to enter a password. You can trigger notice multiple times and there is no way for caller to know that the app is currently showing a notice and which one it is showing.

Simplenote/NoticePresenter.swift
eshurakov: It is `variable` not because it's a var, but because at first it'a a top constraint and later it becomes bottom constraint

Simplenote/NoticePresenter.swift
eshurakov: Lines `L37 + L38 + L40` are here to set the initial offscreen position with top side of notice attached to bottom of container.

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,170 @@
...
+extension NoticePresenter {
+    // Convenience method to fetch current key window
+    //
+    private func getKeyWindow() -> UIWindow? {
+        return UIApplication.shared.windows.first(where: { $0.isKeyWindow })
+    }
+
+    private func getWindowFrame() -> CGRect {
+        return getKeyWindow()?.frame ?? .zero
+    }
```
eshurakov: If possible method names should not have `get*` in them. 
`keyWindow` and `windowFrame` would match current naming conventions better.

Simplenote/NoticePresenter.swift
eshurakov: Right, you need to layout superview to position a view because that's where position constraints are stored.

eshurakov:
Some thoughts about architecture. Writing it as one comment to group everything together :-)

I think that right now responsibilities of controller and presenter are mixed and they both have a state. 

**Controller** stores current notice and handle touches of the notice.
**Presenter** stores current notice view, container view, constraints, dismiss timer.

Because of that I become confused when reading the code. For example we immediately ask presenter to dismiss a notice right after it is being shown. Why? 🤔 From the method name I wouldn't expect that dismiss actually starts a timer.

```
noticePresenter.presentNoticeView(noticeView) {
            if !self.activeViewIsBeingTouched {
                self.dismiss(noticeView)
            }
        }
```

---

Another part that confuses me is that we handle touches in controller and dismiss timer in presenter. In my opinion this type of logic should be handled by the same object. So either controller or presenter would handle both timer and touches.

---

Looking at the API of presenter I would may assume that I can present multiple notices at the same time. I believe right now calling `presentNoticeView` several times with different `noticeView` will result in undefined behaviour, probably old notices will never be removed from the screen. 
Of course controller guards agains this behaviour, but still would be nice to make it more explicit.

Possible solution here can be to instantiate a new presenter for each notice, so that it's always 1:1 mapping between presenter and notice. Presenter can show only one notice at a time and we store contextual information for presented notice, so I think it can work nicely. It will also allow us to make multiple fields non-optional in the presenter.

---

I think puzzle pieces would fit better together in one of the following options:

1. Controller has all the logic.
Controller would set up a timer after asking presenter to show a notice and also handles action tap and long press.
Presenter would only be responsible for displaying and dismissing the view. 

2. Presenter has all the logic.
In this case presenter handles timer and touches. 
Presenter API to show a notice can look like this: `func present(noticeView: NoticeView, @escaping onDismiss: () -> Void)`
In this case the logic about how long the notice stays on screen and when it's dismissed is hidden in the presenter. Controller will get notified when it's dismissed so it can attempt to show the next one.

Simplenote/NoticeController.swift
jleandroperez: Please disregard my comment, I definitely missed the scenario in which the Keyboard would become visible, and the `NoticeController` would not start the Keyboard Notification listeners.

Sorry about the noise!! (and thank you @eshurakov !!)

Simplenote/NoticePresenter.swift
jleandroperez: Discussed over Slack with Evgeny, and indeed, the solution I was proposing to animate things relies on the `intrinsicContentSize` calculation, which is not 100% bullet proof.

Please disregard my comment Charlie!! (and thank you @eshurakov !!)

Simplenote/NoticePresenter.swift
charliescheer: A non repeating timer should invalidate it's self I think?  Did a quick search and that is what I am seeing

charliescheer:
Thanks for all of the suggestions and insights @eshurakov  and @jleandroperez 

I updated the PR to reflect the suggestions.  I decided to go with the first of @eshurakov 's options putting all of the logic inside the controller and having the presenter just display the noticeViews.  Making these changes ended up resolving many of the other questions that had been posed before.

Now the Controller handles: maintaining the queue of notices, creating the noticeView, telling the presenter to present/dismiss, handles taps/long presses, timer/delay
the Presenter handles creation of the container view, setting constraints, animating in the notice, fading out the notice, observing the keyboard

I think this makes it a lot cleaner and easier to understand.  When you get a chance please take a look and let me know what you think.

charliescheer:
One additional note.  Today I was looking into unit tests for the Controller class.  I am always somewhat unsure about what in the production code I should or should not change to make testing possible, and how much of the class information should be exposed to make testing possible.

That being said, one thing that had been bothering me and became really apparent when I went to write tests was how tied together the present and dismiss methods were. For testing, and to decouple the methods a bit, I thought it would be good to make it possible to choose when presenting if the controller should continue all the way on to dismissal or not.

The way that I did this was by adding an optional timer value which can be included when presenting.  If not used, then the default timer would be setup and triggered.  The default timer will eventually fire the dismiss method.  By doing this I can inject a timer during the tests which will take no action, or another action, if need be.

I pulled the unit tests and other changes I made for them into a new branch which can be seen [here](https://github.com/Automattic/simplenote-ios/pull/1208)

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,118 @@
...
+class NoticeController {
+    // MARK: Proerties
+    //
+    static let shared = NoticeController()
+
+    private var notices: [Notice] = []
+    private var current: Notice?
+    private let noticePresenter = NoticePresenter()
+
+    private var activeViewIsBeingTouched: Bool = false
```
eshurakov: `activeViewIsBeingTouched` is not used anymore :-)

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,118 @@
...
+extension NoticeController: NoticePresentingDelegate {
+    func noticePressBegan() {
+        if !isPresenting {
+            return
+        }
+        guard let timer = timer else {
+            return
+        }
+
+        timer.invalidate()
```
eshurakov: If you add `oldValue?.invalidate()` to `didSet` of `timer`, you can just set a timer to `nil` here

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,168 @@
+import UIKit
+
+class NoticePresenter {
+
+    // MARK: Properties
+    //
+    var noticeView: NoticeView?
```
eshurakov: Does this one need to be public? 🤔

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,168 @@
...
+
+        let containerView = PassthruView(frame: .zero)
+        self.containerView = containerView
+
+        keyWindow.addFillingSubview(containerView)
+
+        return containerView
+    }
+
+    private func prepareNoticeView(_ noticeView: NoticeView, containerView: PassthruView) -> NoticeView {
```
eshurakov: Should this method return `noticeView`? It is the same view you're passing in.

eshurakov: Maybe a better name is `add(view: UIView, into containerView: UIView)`?

Simplenote/NoticePresenter.swift
eshurakov: Also we don't really care if noticeView is `NoticeView` or any other `UIView` subclass, so I would just change this to be `UIView`.

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,168 @@
...
+
+        noticeVariableConstraint = noticeView.topAnchor.constraint(equalTo: containerView.bottomAnchor)
+        noticeVariableConstraint?.isActive = true
+        noticeView.centerXAnchor.constraint(equalTo: containerView.centerXAnchor).isActive = true
+        containerView.layoutIfNeeded()
+
+        return noticeView
+    }
+
+    private func displayNotificationView(containerView: PassthruView, noticeView: NoticeView, completion: @escaping (Bool) -> Void) {
```
eshurakov: Maybe `display(view: UIView, in containerView: UIView, completion: @escaping (Bool) -> Void)`?

Simplenote/NoticePresenter.swift
```diff
@@ -0,0 +1,168 @@
...
+
+    // MARK: Lifecycle
+    //
+    deinit {
+        stopListeningToKeyboardNotifications()
+    }
+
+    // MARK: Presenting Methods
+    //
+    func presentNoticeView(_ noticeView: NoticeView, completion: @escaping (Bool) -> Void) {
```
eshurakov: I think having `Bool` in completion here is confusing :-) If it's false, does it mean it wasn't presented?

Simplenote/NoticeView.swift
```diff
@@ -47,20 +49,23 @@ class NoticeView: UIView {
...
-        setupLongPress()
+        setupGestureRecognizers()
 
         layoutIfNeeded()
     }
 
-    private func setupLongPress() {
+    private func setupGestureRecognizers() {
         let longPressGesture = UILongPressGestureRecognizer(target: self, action: #selector(viewWasLongPressed(_:)))
+        let tapGesture = UITapGestureRecognizer(target: self, action: #selector(viewWasTapped(_:)))
```
eshurakov: Does this work as well when you tap on an action button? I didn't check, but I have an impression that UIButton would intercept touches. 🤔

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,118 @@
...
+    //
+    private init() { }
+
+    func setupNoticeController() {
+        noticePresenter.startListeningToKeyboardNotifications()
+    }
+
+    // MARK: Presenting
+    //
+    func present(_ notice: Notice, withTimer timer: Timer? = nil) {
```
eshurakov: What you probably want to do here is to have a `TimerFactory` or `TimerProvider` object injected to the controller (in the init method). This will let you replace it with a mocked version in tests.

`TimerFactory` is an object that would create and return Timers :-) 

The interface can look like this:
```
class TimerFactory {
    func scheduledTimer(with timeInterval: TimeInterval, completion: () -> Void) -> Timer
}
```

eshurakov (commented):
Looks better! Thank you for the changes 👍 

Sending you some more comments.

Simplenote/NoticeView.swift
charliescheer: I tested it.  The button intercepts the touches, so the notice does not get dismissed and the button action fires

Simplenote/NoticePresenter.swift
charliescheer: I like it.

Simplenote/NoticePresenter.swift
charliescheer: I had done that so that I could use the same completion handler signature as UIView.animate... but I agree it is confusing, the bool isn't really needed there.  I removed it

Simplenote/NoticeView.swift
charliescheer: I would think that is expected behavior though, correct?  That the button can be tapped on without dismissing the notice, and tapping anywhere else but on the button will cause a dismiss

Simplenote/NoticeController.swift
charliescheer: Makes sense!

charliescheer:
@eshurakov I have updated the PR per your suggestions.  Can you take another look?

Simplenote/NoticeController.swift
```diff
@@ -0,0 +1,114 @@
...
+        noticeView.handler = notice.action?.handler
+        noticeView.delegate = self
+
+        return noticeView
+    }
+
+    // MARK: Dismissing
+    //
+    @objc
+    func dismiss() {
```
eshurakov, line 79: This can be private I think. We don't want to dismiss from outside? 🤔

eshurakov (commented):
Looks good @charliescheer!

---

I've checked with @SylvesterWilmott about interaction with notices:

* tap on the notice -> do nothing
* tap on the action -> immediate dismiss
* tap and hold -> keep the notice. Once lifted -> dismiss after the duration the same as the duration we show each notice for

Simplenote/NoticeController.swift
charliescheer: I had left this as internal so that it could be tested. I noticed in your comments on the testing PR, there were some suggestions on how we can test this and leave it private.  I will handle that there

charliescheer:
tap on the notice -> do nothing
tap on the action -> immediate dismiss
tap and hold -> keep the notice. Once lifted -> dismiss after the duration the same as the duration we show each notice for

Gotcha, I misunderstood the intention behind the tap to dismiss.  I have updated it so that the action being tapped on will dismiss the notice, but tapping on the notice it's self does nothing.

eshurakov (approved):
Thank you @charliescheer for all the changes 👍 
Let's :shipit:

PR #1208, opened 2021-03-15
NoticeController tests

charliescheer:
### Fix
This PR is a beginning attempt at adding unit tests for the NoticeController class.  This makes a few changes to access in the NoticeController class and writes some basic unit tests for the functionality of the controller methods.

One thing that I particularly could use some advice on is the internal init that I created so I can override the NoticePresenter with a mock.  Since the NoticeController is a singleton, I feel like it is a bad idea to have an initializer that is exposed, but I couldn't find another way around that.  Would be happy to find a way to be able to override the presenter without exposing an initializer to the controller.

### Test
1. look over the access changes and make sure that  I am not exposing too much
2. run the tests to ensure they all pass

### Review
> Only one develope is required to review these changes, but anyone can perform the review.

### DIFF
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -7,7 +7,7 @@ class NoticeController {
 
     private var notices: [Notice] = []
     private var current: Notice?
-    private let noticePresenter = NoticePresenter()
+    private var noticePresenter: NoticePresenter
     private let timerFactory: TimerFactory
 
     private var timer: Timer? {
@@ -22,13 +22,14 @@ class NoticeController {
 
     // MARK: Life Cycle
     //
-    private init() {
-        self.timerFactory = TimerFactory()
-        self.noticePresenter = NoticePresenter()
+    private init(presenter: NoticePresenter = NoticePresenter(), timerFactory: TimerFactory = TimerFactory()) {
+        self.timerFactory = timerFactory
+        self.noticePresenter = presenter
     }
 
-    internal init(presenter: NoticePresentable) {
+    init(presenter: NoticePresenter, timerFactor: TimerFactory) {
         self.noticePresenter = presenter
+        self.timerFactory = timerFactor
     }
 
     func setupNoticeController() {
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -16,7 +16,7 @@ class NoticeController {
         }
     }
 
-    internal var isPresenting: Bool {
+    var isPresenting: Bool {
         current != nil
     }
 
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -95,7 +95,7 @@ class NoticeController {
 
 // MARK: NoticePresenting Delegate
 //
-extension NoticeController: NoticePresentingDelegate {
+extension NoticeController: NoticeInteractionDelegate {
     func noticePressBegan() {
         if !isPresenting {
             return
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -12,7 +12,7 @@ class NoticeController {
     private var activeViewIsBeingTouched: Bool = false
     private var timer: Timer?
 
-    private var isPresenting: Bool {
+    internal var isPresenting: Bool {
         current != nil
     }
 
@@ -60,6 +60,10 @@ class NoticeController {
         notices.append(notice)
     }
 
+    var pendingNotices: Int {
+        notices.count
+    }
+
     func startTimer(timer: Timer? = nil) {
         if timer != nil {
             self.timer = timer
@@ -69,6 +73,7 @@ class NoticeController {
         let delay = current?.action == nil ? Times.shortDelay : Times.longDelay
         self.timer = Timer.scheduledTimer(timeInterval: delay, target: self, selector: #selector(dismiss), userInfo: nil, repeats: false)
     }
+
     private func makeNoticeView(from notice: Notice) -> NoticeView {
         let noticeView: NoticeView = NoticeView.instantiateFromNib()
         noticeView.message = notice.message
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -7,7 +7,7 @@ class NoticeController {
 
     private var notices: [Notice] = []
     private var current: Notice?
-    private var noticePresenter: NoticePresenter
+    private let noticePresenter: NoticePresenter
     private let timerFactory: TimerFactory
 
     private var timer: Timer? {
@@ -16,7 +16,7 @@ class NoticeController {
         }
     }
 
-    var isPresenting: Bool {
+    private var isPresenting: Bool {
         current != nil
     }
 
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -64,10 +64,6 @@ class NoticeController {
         notices.append(notice)
     }
 
-    var pendingNotices: Int {
-        notices.count
-    }
-
     private func makeNoticeView(from notice: Notice) -> NoticeView {
         let noticeView: NoticeView = NoticeView.instantiateFromNib()
         noticeView.message = notice.message
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -7,7 +7,7 @@ class NoticeController {
 
     private var notices: [Notice] = []
     private var current: Notice?
-    private let noticePresenter = NoticePresenter()
+    private var noticePresenter: NoticePresentable
 
     private var activeViewIsBeingTouched: Bool = false
     private var timer: Timer?
@@ -18,7 +18,13 @@ class NoticeController {
 
     // MARK: Life Cycle
     //
-    private init() { }
+    private init() {
+        self.noticePresenter = NoticePresenter()
+    }
+
+    internal init(presenter: NoticePresentable) {
+        self.noticePresenter = presenter
+    }
 
     func setupNoticeController() {
         noticePresenter.startListeningToKeyboardNotifications()
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -22,14 +22,9 @@ class NoticeController {
 
     // MARK: Life Cycle
     //
-    private init(presenter: NoticePresenter = NoticePresenter(), timerFactory: TimerFactory = TimerFactory()) {
+    init(presenter: NoticePresenter = NoticePresenter(), timerFactory: TimerFactory = TimerFactory()) {
         self.timerFactory = timerFactory
         self.noticePresenter = presenter
-    }
-
-    init(presenter: NoticePresenter, timerFactor: TimerFactory) {
-        self.noticePresenter = presenter
-        self.timerFactory = timerFactor
     }
 
     func setupNoticeController() {
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -81,7 +81,7 @@ class NoticeController {
     // MARK: Dismissing
     //
     @objc
-    func dismiss() {
+    private func dismiss() {
         noticePresenter.dismissNotification {
             self.current = nil
 
--- a/Simplenote/NoticePresenter.swift
+++ b/Simplenote/NoticePresenter.swift
@@ -1,16 +1,10 @@
 import UIKit
 
-protocol NoticePresentable {
-    func startListeningToKeyboardNotifications()
-    func presentNoticeView(_ noticeView: NoticeView, completion: @escaping (Bool) -> Void)
-    func dismissNotification(completion: @escaping () -> Void)
-}
-
-class NoticePresenter: NoticePresentable {
+class NoticePresenter {
 
     // MARK: Properties
     //
-    private var noticeView: UIView?
+    var noticeView: UIView?
     private var containerView: PassthruView?
     private var noticeVariableConstraint: NSLayoutConstraint?
 
@@ -39,6 +33,19 @@ class NoticePresenter: NoticePresentable {
     //
     deinit {
         stopListeningToKeyboardNotifications()
+    }
+
+    func startListeningToKeyboardNotifications() {
+        keyboardNotificationTokens = addKeyboardObservers()
+    }
+
+    func stopListeningToKeyboardNotifications() {
+        guard let tokens = keyboardNotificationTokens else {
+            return
+        }
+
+        removeKeyboardObservers(with: tokens)
+        keyboardNotificationTokens = nil
     }
 
     // MARK: Presenting Methods
@@ -154,19 +161,6 @@ extension NoticePresenter: KeyboardObservable {
         UIView.animate(withDuration: animationDuration, delay: .zero, options: animationOptions) {
             containerView.layoutIfNeeded()
         }
-    }
-
-    func startListeningToKeyboardNotifications() {
-        keyboardNotificationTokens = addKeyboardObservers()
-    }
-
-    func stopListeningToKeyboardNotifications() {
-        guard let tokens = keyboardNotificationTokens else {
-            return
-        }
-
-        removeKeyboardObservers(with: tokens)
-        keyboardNotificationTokens = nil
     }
 }
 
--- a/Simplenote/NoticePresenter.swift
+++ b/Simplenote/NoticePresenter.swift
@@ -1,6 +1,12 @@
 import UIKit
 
-class NoticePresenter {
+protocol NoticePresentable {
+    func startListeningToKeyboardNotifications()
+    func presentNoticeView(_ noticeView: NoticeView, completion: @escaping (Bool) -> Void)
+    func dismissNotification(completion: @escaping () -> Void)
+}
+
+class NoticePresenter: NoticePresentable {
 
     // MARK: Properties
     //
--- a/Simplenote/NoticePresenter.swift
+++ b/Simplenote/NoticePresenter.swift
@@ -4,7 +4,7 @@ class NoticePresenter {
 
     // MARK: Properties
     //
-    var noticeView: UIView?
+    private var noticeView: UIView?
     private var containerView: PassthruView?
     private var noticeVariableConstraint: NSLayoutConstraint?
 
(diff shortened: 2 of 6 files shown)


Simplenote/NoticeController.swift
```diff
@@ -7,18 +7,24 @@ class NoticeController {
...
     private var notices: [Notice] = []
     private var current: Notice?
-    private let noticePresenter = NoticePresenter()
+    private var noticePresenter: NoticePresentable
 
     private var activeViewIsBeingTouched: Bool = false
     private var timer: Timer?
 
-    private var isPresenting: Bool {
+    internal var isPresenting: Bool {
```
eshurakov: You don't need to specify `internal` as it should be the default visibility.

Simplenote/NoticeController.swift
```diff
@@ -7,18 +7,24 @@ class NoticeController {
 
     private var notices: [Notice] = []
     private var current: Notice?
-    private let noticePresenter = NoticePresenter()
+    private var noticePresenter: NoticePresentable
```
eshurakov: This can stay as `let` :-)

Simplenote/NoticeController.swift
```diff
@@ -7,18 +7,24 @@ class NoticeController {
...
     // MARK: Life Cycle
     //
-    private init() { }
+    private init() {
+        self.noticePresenter = NoticePresenter()
+    }
+
+    internal init(presenter: NoticePresentable) {
+        self.noticePresenter = presenter
+    }
```
eshurakov: Can be `init(presenter: NoticePresentable = NoticePresenter())`

Simplenote/NoticePresenter.swift
```diff
@@ -1,6 +1,12 @@
 import UIKit
 
-class NoticePresenter {
+protocol NoticePresentable {
+    func startListeningToKeyboardNotifications()
+    func presentNoticeView(_ noticeView: NoticeView, completion: @escaping (Bool) -> Void)
+    func dismissNotification(completion: @escaping () -> Void)
+}
```
eshurakov: You don't necessarily need a protocol to mock `NoticePresenter`, you can also subclass it and override exposed methods.

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,142 @@
...
 
 class NoticeContorllerTests: XCTestCase {
+    private lazy var presenter = MockNoticePresenter()
+    private lazy var controller = NoticeController(presenter: presenter)
+
+func testSetupNoticeContoller() {
+    controller.setupNoticeController()
+
+    XCTAssertTrue(presenter.listeningToKeyboard == true)
+    }
```
eshurakov: Alignment looks off :-)

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,142 @@
...
+        let timer = Timer.scheduledTimer(withTimeInterval: TimeInterval(0.5), repeats: false) { (_) in
+            expectation.fulfill()
+        }
+
+        let notice = Notice(message: "Message", action: nil)
+        controller.present(notice, withTimer: timer)
+
+        wait(for: [expectation], timeout: TimeInterval(3))
+
     }
```
eshurakov: It is best to avoid using delays in unit tests as they should be fast. You can mock a timer and "fire" it when needed.
It's also not clear what is being tested here. That timer would fire? It would fire even if it's not passed to controller :-)

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,142 @@
...
+
+    func testPresentNotice() throws {
+        let notice = Notice(message: "Message", action: nil)
+        controller.present(notice, withTimer: NoticeContorllerTests.noActionTimer)
+
+        XCTAssertTrue(presenter.presented)
+
+        let presentedNoticeView = try XCTUnwrap(presenter.noticeView)
+        XCTAssertEqual(presentedNoticeView.message, notice.message)
+        XCTAssertTrue(presentedNoticeView.handler == nil)
```
eshurakov: This is the same as in `testMakeNoticeView`. Because they're testing the same thing, we can remove checks from this method.

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,142 @@
...
+        XCTAssertNil(presenter.noticeView?.handler)
     }
 
-    func testPerformanceExample() throws {
-        // This is an example of a performance test case.
-        self.measure {
-            // Put the code you want to measure the time of here.
+    func testMakeNoticeViewWithAction() {
+        let action = NoticeAction(title: "Title") {
+            print("Action")
```
eshurakov: Leftover debug code? :-)

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,142 @@
...
+        let action = NoticeAction(title: "") {
+            print("")
+        }
+        let notice = Notice(message: "", action: action)
+
+        controller.present(notice)
+        XCTAssertEqual(notice.message, presenter.noticeView?.message)
+        XCTAssertNotEqual(notice.action?.title, presenter.noticeView?.actionTitle)
+        XCTAssertNotNil(presenter.noticeView?.handler)
     }
```
eshurakov: `testMakeNoticeViewWithAction` and `testMakeNoticeViewEmptyNotice` are the same except the message. Maybe we can extract a common method?

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,142 @@
...
+        controller.present(noticeB, withTimer: NoticeContorllerTests.noActionTimer)
+
+        XCTAssertEqual(controller.pendingNotices, 1)
+
+        controller.dismiss()
+
+        XCTAssertTrue(presenter.dismissed)
+        XCTAssertTrue(controller.isPresenting)
+        XCTAssertEqual(controller.pendingNotices, 0)
+    }
```
eshurakov: My understanding is that we want to test here that notices are presented one after the other. The best way to do that is to have a log of events in `MockNoticePresenter`. 
For example presenting 2 notices should result in the following log: `present(noticeA), dismiss(noticeA), present(noticeB), dismiss(noticeB)`. Later in the test we can compare the log with what we expect. 
I think this way the behaviour is more deterministic.

eshurakov (commented):
Hey Charlie, good work on tests.
Sending you some comments.

SimplenoteTests/NoticeContorllerTests.swift
charliescheer: Yeah, I was testing that the timer in the controller present method fired.  I have removed that test

SimplenoteTests/NoticeContorllerTests.swift
charliescheer: Okay @eshurakov I have updated the tests as well.  I left a few questions for you in slack, so not all of these comments have been handled, but they are ready for another look.  Thanks!!!

Simplenote/NoticeController.swift
```diff
@@ -7,7 +7,7 @@ class NoticeController {
 
     private var notices: [Notice] = []
     private var current: Notice?
-    private let noticePresenter = NoticePresenter()
+    private var noticePresenter: NoticePresenter
```
eshurakov: This can stay `let` because it is assigned in the init :-)

Simplenote/NoticeController.swift
```diff
@@ -16,14 +16,20 @@ class NoticeController {
...
-        self.timerFactory = TimerFactory()
+    private init(presenter: NoticePresenter = NoticePresenter(), timerFactory: TimerFactory = TimerFactory()) {
+        self.timerFactory = timerFactory
+        self.noticePresenter = presenter
+    }
+
+    init(presenter: NoticePresenter, timerFactor: TimerFactory) {
+        self.noticePresenter = presenter
+        self.timerFactory = timerFactor
     }
```
eshurakov: There is no need to have 2 methods here :-)
Making the first one non-private would do the job. It's up to the caller whether to use default values or provide another value.
For example, this is a valid call `NoticeController(timerFactory: MockFactory())` that would use "default" presenter.

Simplenote/NoticePresenter.swift
```diff
@@ -4,7 +4,7 @@ class NoticePresenter {
 
     // MARK: Properties
     //
-    private var noticeView: UIView?
+    var noticeView: UIView?
```
eshurakov: Why do we need to expose this? 🤔

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,171 @@
...
+    override func scheduledTimer(with timeInterval: TimeInterval, completion: @escaping () -> Void) -> Timer {
+        if let timer = timer {
+            timer.fire()
+
+            return timer
+        } else {
+            return Timer.scheduledTimer(withTimeInterval: NoticeContorllerTests.timeInterval, repeats: false, block: NoticeContorllerTests.timerNoActionCompletionHandler)
+        }
+    }
 }
```
eshurakov: What you want to do here is to return a mock timer that won't fire till you tell it to. This will allow you to precisely control test scenarios. Mock timer could look like this:
```
class MockTimerFactory: TimerFactory {
    var timer: Timer?

    override func scheduledTimer(with timeInterval: TimeInterval, completion: @escaping () -> Void) -> Timer {
        let timer = MockTimer()
        timer.completion = completion
        return timer
    }
}

class MockTimer: Timer {
    var completion: (() -> Void)?

    override func fire() {
        completion?()
    }

    override func invalidate() {
        super.invalidate()
        completion = nil
    }
}
```

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,171 @@
...
+    override func dismissNotification(completion: @escaping () -> Void) {
+        dismissed = true
+        actionLog.append("dismissNotification(), ")
+        completion()
+    }
+
+    override func startListeningToKeyboardNotifications() {
+        listeningToKeyboard = true
+    }
+}
```
eshurakov: If we use an action log there is no reason to keep also `presented` and `dismissed`. And instead of a list of presented views we can keep only the last one.
Mock presenter could look like this:
```
class MockNoticePresenter: NoticePresenter {
    enum Action: Equatable {
        case present(String?)
        case dismiss(String?)
    }

    var actionLog: [Action] = []

    var listeningToKeyboard: Bool = false
    var lastNoticeView: NoticeView?

    override func presentNoticeView(_ noticeView: NoticeView, completion: @escaping () -> Void) {
        lastNoticeView = noticeView
        actionLog.append(.present(noticeView.message))
        completion()
    }

    override func dismissNotification(completion: @escaping () -> Void) {
        let noticeView = try! XCTUnwrap(lastNoticeView)
        lastNoticeView = nil
        actionLog.append(.dismiss(noticeView.message))
        completion()
    }

    override func startListeningToKeyboardNotifications() {
        listeningToKeyboard = true
    }
}
```

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,171 @@
...
+        XCTAssertTrue(controller.isPresenting)
+        XCTAssertEqual(controller.pendingNotices, 0)
+
+        controller.dismiss()
+
+        XCTAssertTrue(presenter.dismissed)
+        XCTAssertFalse(controller.isPresenting)
+        XCTAssertEqual(controller.pendingNotices, 0)
+        XCTAssertEqual(presenter.actionLog, "presentNoticeView(Message A), dismissNotification(), presentNoticeView(Message B), dismissNotification(), ")
+    }
```
eshurakov: An example how actionLog and mock timer can be used (I didn't run the actual test :-) )
```
func testDismissContinuesToNextNotice() {
    let noticeA = Notice(message: "Message A", action: nil)
    let noticeB = Notice(message: "Message B", action: nil)
    controller.present(noticeA)
    controller.present(noticeB)

    var expectedActions: [MockNoticePresenter.Action] = [
        .present(noticeA.message)
    ]
    XCTAssertEqual(presenter.actionLog, expectedActions)

    timerFactory.timer?.fire()

    expectedActions.append(.dismiss(noticeA.message))
    expectedActions.append(.present(noticeB.message))
    XCTAssertEqual(presenter.actionLog, expectedActions)

    timerFactory.timer?.fire()

    expectedActions.append(.dismiss(noticeB.message))
    XCTAssertEqual(presenter.actionLog, expectedActions)
}
```

This way we mimic how the actual flow would work as the client code wouldn't call `dismiss`.

We also don't need `controller.isPresenting` and `controller.pendingNotices`. And `controller.dismiss` can remain private.

eshurakov (commented):
Some other things we can test:
* pressing on action button dismisses the notice
* long press invalidates the timer and notice stays on screen
* when long press is released, timer is created and notice is dismissed after that

SimplenoteTests/NoticeContorllerTests.swift
```diff
@@ -1,33 +1,171 @@
...
-//  SimplenoteTests
-//
-//  Created by Charlie Scheer on 3/15/21.
-//  Copyright © 2021 Automattic. All rights reserved.
-//
-
 import XCTest
+@testable import Simplenote
 
 class NoticeContorllerTests: XCTestCase {
```
jleandroperez: Nipticky: typo in the classname (Controller / Contorller)

charliescheer:
Thanks for all of the suggestions!!  The action log is a great idea.  Will put that idea in my testing tool kit :)

I have updated the tests and added some new ones.  Let me know if there are any more notes @eshurakov !!

SimplenoteTests/NoticeControllerTests.swift
```diff
@@ -0,0 +1,238 @@
...
+
+        XCTAssertEqual(expectedActions, presenter.actionLog)
+        XCTAssertNil(timerFactory.timer?.completion)
+
+        controller.noticePressEnded()
+        timerFactory.timer?.fire()
+        expectedActions.append(.dismiss(noticeA.message))
+
+        XCTAssertEqual(expectedActions, presenter.actionLog)
+    }
```
eshurakov: Just a note: because we already tested some parts in other places, we can only assert things that we didn't check yet.
For example, because most of this method is tested in `testLongPressInvalidatesTimerAndNoticeDoesNotDismiss` it could look like:
```
func testLongPressReleasedDismissesTimer() {
    let noticeA = Notice(message: "Message A", action: nil)

    controller.present(noticeA)
    controller.noticePressBegan()
    controller.noticePressEnded()

    XCTUnwrap(timerFactory.timer).fire() <- can use XCTUnwrap to make sure we actually created a timer

    var expectedActions: [MockNoticePresenter.Action] = [
      .present(noticeA.message),
      .dismiss(noticeA.message)
    ]
    XCTAssertEqual(expectedActions, presenter.actionLog)
}
```

SimplenoteTests/NoticeControllerTests.swift
```diff
@@ -0,0 +1,238 @@
...
+
+    func testMakeNoticeViewWithEmptyNotice() {
+        let action = NoticeAction(title: "") { }
+        let notice = Notice(message: "", action: action)
+
+        controller.present(notice)
+
+        XCTAssertEqual(notice.message, presenter.lastNoticeView?.message)
+        XCTAssertNotEqual(notice.action?.title, presenter.lastNoticeView?.actionTitle)
```
eshurakov: To continue the previous conversation, we can update actionTitle getter in NoticeView to `noticeButton.title(for: .normal)`. That would make it behave the same as `testMakeNoticeViewWithAction`

SimplenoteTests/NoticeControllerTests.swift
```diff
@@ -0,0 +1,238 @@
...
+
+        controller.present(noticeA)
+        XCTAssertEqual(controller.pendingNotices, 0)
+
+        controller.present(noticeB)
+        XCTAssertEqual(controller.pendingNotices, 1)
+
+        controller.present(noticeB)
+        XCTAssertEqual(controller.pendingNotices, 1)
+    }
```
eshurakov: Do you think we can also use actionLog for this and get rid of `pendingNotices`?

Simplenote/NoticeController.swift
```diff
@@ -16,14 +16,15 @@ class NoticeController {
         }
     }
 
-    private var isPresenting: Bool {
+    var isPresenting: Bool {
```
eshurakov: This can stay private :-)

Simplenote/NoticeController.swift
eshurakov: 🤔

eshurakov (commented):
Thank you @charliescheer, looking better and better.

Sending you some more comments :-)

SimplenoteTests/NoticeControllerTests.swift
charliescheer: Had to think about that one for a little bit, but yes I think we can!

Simplenote/NoticeController.swift
charliescheer: I merged some changes from the branch that I think cause some properties to change in ways I had not intended....  Fixed

charliescheer:
@eshurakov thanks again for these additional suggestions.  I have updated the tests per your suggestions.

SimplenoteTests/NoticeControllerTests.swift
```diff
@@ -37,21 +37,31 @@ class NoticeControllerTests: XCTestCase {
...
+        try XCTUnwrap(timerFactory.timer).fire()
+        expectedActions.append(.dismiss(noticeA.message))
+        expectedActions.append(.present(noticeB.message))
+
+        XCTAssertEqual(expectedActions, presenter.actionLog)
+
+        try XCTUnwrap(timerFactory.timer).fire()
+        expectedActions.append(.dismiss(noticeB.message))
+
+        XCTAssertEqual(expectedActions, presenter.actionLog)
```
eshurakov, line 64: Awesome!

eshurakov (approved):
Thank you for adding tests!

Looks good, let's :shipit:

PR #1210, opened 2021-03-15
Fix issue where unit test log out simulator

### LINKED ISSUE #1025: Running Unit Tests logs you out from the simulator
### What
Running Unit Tests from Xcode logs out currently logged in user from the iOS simulator.

### Why
Keeping the user logged in will make development experience better.

### How
Possible solution can be to check in the AppDelegate that we're running unit tests and don't create / run the app. 
Make sure that UI Automation / Screenshots are still working.

charliescheer:
### Fix
This PR fixes #1025 
Currently when you run the unit tests in Simplenote iOS, the simulator get's logged out of the app.  This is happening because the KeychainMigratorTests are making calls to the username field, which has a setter that updates the Userdefaults username value.  In the tests they get set to nil, effectively logging the sim out.

This PR changes KeychainMigrator to hold onto a local reference of UserDefaults instead, so at testing we can create a KeychainMigrator and a test UserDefaults, leaving the simulator unaffected

### Test
1. Log into a simplenote account with any simulator
2. run the full unit test suite for simplenote ios
3. check and make sure that all tests pass
4. check and make sure that the simulator is still logged in

### Review


jleandroperez:
@eshurakov @charliescheer gentleman!! Leaving a note over here.

I believe it's okay to drop KeychainMigrator, it was introduced several years ago to deal with the TeamID switch we had. As Evgeny pointed out, it's from 2017, so it's fairly safe to assume that our active users already got migrated.

Thank you!!

charliescheer:
I went ahead and removed the KeychainMigrator and it's unit tests from this PR.  @eshurakov can you take another look when you have a moment

New testing instructions:
1. run unit tests to make sure they all still pass
2. make sure the simulator/app stays logged in after running tests
3. load the app and make sure there are no issues when the app starts up

eshurakov (approved):
Looks good, thank you!

I think we can also delete 
```
SPTracker.trackKeychainMigrationSucceeded()
SPTracker.trackKeychainMigrationFailed()
```

PR #1213, opened 2021-03-16
Fixes crash when editor loads in search mode but can not highlight search value

### LINKED ISSUE #1193: Crash when searching with English + non-English combinations
It was originally reported in 3780689-zen that when searching for notes using English + Japanese characters, the app crashes when they tap on the search result.

I was able to replicate this with the English + Chinese combination.

### Expected

I would expect the app doesn't crash when I tap on the search result after performing a search using English + non-English (Japanese/Chinese) characters.

### Observed

The app crashes when I tap on the search result after performing a search using English + non-English (Japanese/Chinese) characters.

### Reproduced

1. Create a note with a combination of English + Japanese/Chinese characters (e.g. Abc中文）
2. Search the note with this keyword: Abc中文
3. Tap the search result and the app crashes

Video recording: https://d.pr/v/bh9rOc/6fWEeWoEl4

-|-|-|-
Apple|iPhone SE|14.4|4.32
Apple|iPhone X|14.4|4.32
Apple|iPod touch 7th|14.4|4.32
Apple|iPad mini 5th|14.4|4.32

### LINKED ISSUE #1204: Crash when trying to open a Japanese note from Search, if character count is high
This user has reported that when they perform a note search in Japanese, the app crashes as long as they try to open a note if the search query was "long", for example long queries or multiple using words. If the query has not many characters, notes seem to open normally.

It seems to work with 3-5 character long words, like "ロール" but fails if search queries are longer than that.

### Expected
To be able to open the notes without crashing

### Observed
The app crashes when you try to open a note, if the word query used had more than 3 characters, or multiple words.

### Reproduced
I couldn't reproduce this. Some cases shared by the user:

- I write "ロールスクリーン" with Japanese characters in the search box, the app shows one note rightly. Then I try to open this note, app crashes.

- I reduce the number of characters of the search word from "ロールスクリーン" to "ロールスクリー", "ロールスクリ", "ロールスク", "ロールス". In every cases, when I try to open the note, the app crashes.

- I write "ロール" , the app responds rightly; showing 5 notes. This time, I can open every 5 notes without cash.

- I open the one note, it shows highlight position wrongly, although other 4 notes shows highlight position rightly.

Apple |  iPhone 11 | IOS 13.5.1 | Simplenote 4.32

charliescheer:
### Fix
This issue fixes #1193 

If you search for a note in Simplenote, and then open that note from search with values that can not be highlighted, Simplenote will crash when the SPNoteEditorViewController opens.  This crash can be replicated in two ways:
Tag search:
Create a new note and give it a tag.  Make sure that the tag value can not be found in the note text
Search for the tag, select the tag
select the note from the search results
when editor loads it will crash

Multi language character search:
Create a note with a combination of English + Japanese/Chinese characters (e.g. Abc中文）
Search the note with this keyword: Abc中文
Tap the search result and the app crashes

This PR puts in a guard against invalid search result highlight terms

### Test
1. create a new note, give it a tag of "A" and in the text body add the text "Abc中文"
2. exit the note and go to search, search for "A"
3. select the tag search from the options, then select the note you just created.
4. The note should load fine without crashing
5. Exit the note, return to the search and type in "Abc中文"
6. Select the note from the search results
7. The note should load fine without crashing

### Review

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -182,6 +182,11 @@ CGFloat const SPSelectedAreaPadding = 20;
 
         self.searchResultRanges = [self searchResultRangesIn:(searchText ?: @"")
                                                 withKeywords:self.searchQuery.keywords];
+
+        // If there are no found ranges, bail out
+        if (self.searchResultRanges.count < 1) {
+            return;
+        }
         
         dispatch_async(dispatch_get_main_queue(), ^{
 
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -183,10 +183,10 @@ CGFloat const SPSelectedAreaPadding = 20;
         self.searchResultRanges = [self searchResultRangesIn:(searchText ?: @"")
                                                 withKeywords:self.searchQuery.keywords];
 
-        // If there are no found ranges, bail out
-        if (self.searchResultRanges.count < 1) {
-            return;
-        }
+//        // If there are no found ranges, bail out
+//        if (self.searchResultRanges.count < 1) {
+//            return;
+//        }
         
         dispatch_async(dispatch_get_main_queue(), ^{
 
@@ -539,6 +539,10 @@ CGFloat const SPSelectedAreaPadding = 20;
 - (void)highlightSearchResultAtIndex:(NSInteger)index animated:(BOOL)animated
 {
     NSInteger searchResultCount = self.searchResultRanges.count;
+
+    if (searchResultCount < 1) {
+        return;
+    }
 
     index = MIN(index, searchResultCount - 1);
     index = MAX(index, 0);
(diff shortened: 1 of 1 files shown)


charliescheer:
Also fixes #1204

eshurakov (commented):
Thank you for fixing this issue Charlie!

I think it's better to add this check to `highlightSearchResultAtIndex:animated:` method because:
* we're already doing boundaries checks for index there
* this method is called from other methods. If we add a check for empty array there, potential crashes from other code paths will also be prevented

charliescheer:
Makes sense.  I moved it into that function.  Checked against the above cases and the crash is fixed with guarding against the index being  less than one.  When you get a chance, can you take another look @eshurakov

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -182,6 +182,11 @@ - (void)highlightSearchResultsIfNeeded
 
         self.searchResultRanges = [self searchResultRangesIn:(searchText ?: @"")
                                                 withKeywords:self.searchQuery.keywords];
+
+//        // If there are no found ranges, bail out
+//        if (self.searchResultRanges.count < 1) {
+//            return;
+//        }
```
eshurakov: Please remove this before merging :-)

eshurakov (approved):
Looks good!

Please remove commented out code before merging this.

Simplenote/Classes/SPNoteEditorViewController.m
charliescheer: Removed!

PR #1214, opened 2021-03-17
Use API calls to trigger release jobs

loremattei:
### Fix
This PR updates our release tools to trigger any CI release job via an API call instead of relying on the automatic trigger on tags.
This way, it will be easier and cleaner to handle any failure during the release build and, moreover, it'll be possible to move more steps of the releases to CI (for example, validation and lint steps that can fail by design).

Note that in order to use the release tools, users should configure a new token for CircleCI on their local environment, following these steps:

- Go to https://app.circleci.com/settings/user/tokens.
- Create a Personal API Token.
- Add `CIRCLE_CI_AUTH_TOKEN=<Your Token>` to your `~/.wpios-env.default` (you can check out the updated fastlane/env/user.env-example for a template).

This PR needs an updated version of `release-toolkit`, so it's now pointing to https://github.com/wordpress-mobile/release-toolkit/pull/223. After approval, I'll release a new version of `release-toolkit` and update this PR before merging.

### Test
Full testing is tricky because it's tight to the release cycle. Anyway, it's possible to test the critical part of it this way:

- checkout a new `release/test` branch from this one.
- add a dummy commit (the build will fail, so adding a dummy commit will avoid the failure to be reported to this PR). 
- `bundle install`
- Run `bundle exec fastlane trigger_beta_build branch_to_build:"release/test"`
- Verify that the `release build` job has been started on CircleCI. The job will fail during the upload to AppStore because the current version is already present. This is fine. 
- Delete the test branch.

### Review
I asked the review to @mokagio because he is currently the release manager of Simplenote iOS. All the code is already in production in WPiOS, so it _should_ be fine, but there may always be some nuances with porting that I missed, so I think that understanding the changes will help if something happens during the next release. 

### Release


.circleci/config.yml
```diff
@@ -99,13 +110,27 @@ jobs:
             brew install ghostscript
             curl -sL https://sentry.io/get-cli/ | bash
       - run:
-          name: Build
+          name: Setup notifications
```
mokagio: Nitpick:

```suggestion
          name: Setup Slack notifications
```

Even though the `when` step is clearly at the same level as this one, this CircleCI syntax made me think that it was related to the "Setup notifications" step. Maybe making it explicit that it's about Slack would help.

fastlane/Fastfile
```diff
@@ -377,6 +427,44 @@ end
...
+  # Usage:
+  # bundle exec fastlane trigger_beta_build [branch_to_build:<branch_name>]
+  #
+  #####################################################################################
+  lane :trigger_beta_build do | options |
+    circleci_trigger_job(
+      circle_ci_token: ENV["CIRCLE_CI_AUTH_TOKEN"], 
+      repository: REPOSITORY_NAME, 
+      organization: "Automattic",
+      branch: options[:branch_to_build],
```
mokagio, line 444: Not sure if that's an edge case worth covering given the use case for this workflow, but while testing I tried to run the lane before pushing the branch and the error I got in console was an unhelpful `[400] Bad Request`.

It might be worth adding a check at the release toolkit level to ensure the branch exists on the remote. I'd consider this a little improvement that can live in isolation, though.

mokagio (approved):
Worked like a charm. Thank you for doing this @loremattei ! 👏 

image

As you mention in the PR description, we should wait to merge till a new version of the toolkit shipped. In particular given the recent tag data loss issue with 0.15, meaning that if we were to prioritize getting this change in `develop` using the hash we'd end up bring the bug with it, too.

fastlane/Fastfile
loremattei: 👍

loremattei:
>  In particular given the recent tag data loss issue with 0.15, meaning that if we were to prioritize getting this change in develop using the hash we'd end up bring the bug with it, too.

Right! I won't merge this until a new `release-toolkit` with a fix for that bug will be available. 👍

loremattei:
cc @AliSoftware as this may be relevant to the lane refactoring you're working on.

fastlane/Fastfile
```diff
@@ -225,41 +227,89 @@ end
       ios_lint_localizations(input_dir: 'Simplenote', allow_retry: true)
       ios_bump_version_beta() unless ios_current_branch_is_hotfix
     end
-    ios_final_tag(options)
```
mokagio, line 228: @loremattei as I was working on the [Simplenote macOS version](https://github.com/Automattic/simplenote-macos/pull/927) of the API build trigger, I noticed this diff removes the call that makes the final tag.

Is this intentional? Sometimes new code makes it into the final build since a the latest beta build / tag.

loremattei: Yes, it's intentional. The idea is to move the final tag _after_ the build has been deployed to App Store and GH so that we won't have to move the tag if any error pops up during the final build. 
Basically, with the new flow, the CI job creates a draft GH release on the HEAD commit of the release branch. When the release manager publishes the draft, GitHub automatically adds the proper tag on the git repository.

loremattei:
@mokagio a new version of `release-toolkit` with the fix for the tag deletion issue has been released and I've updated this PR. 
I think it's ready to be merged if you don't have any additional comment.

fastlane/Fastfile
mokagio: > When the release manager publishes the draft, GitHub automatically adds the proper tag on the git repository.

Wait what?! 🤯 This is cool (a bit "magical") but cool 👍 

Thank you for the explanation 🙇

mokagio:
Thank you @loremattei! Merged 🚀 Looking forward to experience the new flow on my next code freeze.

PR #1215, opened 2021-03-21
UI automation maintenance: new locator for `Trash Note` button

pachlava:
Fixes the tests that started failing after [this accessibility improvement](https://github.com/Automattic/simplenote-ios/pull/1191):  'trash note' button has a `Move to Trash` title now, instead of the default identification by image name: `icon trash`. However, the 'delete note' button, shown for a note when in `Trash` still uses `icon trash` identification.

### Fix
1. Added `Move to Trash` item to `UIDs.swift`
2. Simplified the `NoteList.trashAllNotes()` function
3. To make ^ possible, added new `Table.getVisibleLabelledCells()`
3. Adjusted `Table.trashCell()` to use different titles for delete button, depending on the current note list (in `Trash` vs in other notes list)

### Test
1. Running all tests, especially the ones for `Trash`, seeing them not fail.

### Review

### Release

### DIFF
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -78,7 +78,6 @@ class Table {
         let noteCell = Table.getCell(label: noteName)
 
         noteCell.swipeLeft()
-        //sleep(1)
         noteCell.buttons[deleteButtonLabel].tap()
     }
 
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -72,7 +72,9 @@ class Table {
 
     class func trashCell(noteName: String) {
         // `Trash Note` and `Delete note forever` buttons have different labels
-        // since #1191. To use the correct label, we need to know where we are.
+        // since 07fcccf1039495768ecdf9909d3dbd1b255936cd
+        // (https://github.com/Automattic/simplenote-ios/pull/1191).
+        // To use the correct label, we need to know where we are.
         let deleteButtonLabel = app.navigationBars[UID.NavBar.trash].exists ?
             UID.Button.itemTrash : UID.Button.noteTrash
         let noteCell = Table.getCell(label: noteName)
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -45,14 +45,20 @@ class Table {
         return app.tables.element.children(matching: .cell)
     }
 
-    class func getVisibleLabelledCellsNumber() -> Int {
+    class func getVisibleLabelledCells() -> [XCUIElement] {
         // We need only the table cells that have X = 0 and a non-empty label
-        // otherwise we will include invisible elements from Sidebar pane, when Notes List is open
-        // or the elements from Notes List when Settings are open
-        // or the visible tags search suggestions
+        // Currently (besides using object dimentions) this is the way to
+        // locate note cells
         return Table.getAllCells()
             .filter { $0.frame.minX == 0.0 && $0.label.isEmpty == false }
-            .count
+    }
+
+    class func getVisibleLabelledCellsNames() -> [String] {
+        return Table.getVisibleLabelledCells().compactMap { $0.label }
+    }
+
+    class func getVisibleLabelledCellsNumber() -> Int {
+        return Table.getVisibleLabelledCells().count
     }
 
     class func getVisibleNonLabelledCellsNumber() -> Int {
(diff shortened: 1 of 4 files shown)


SimplenoteUITests/Generic.swift
```diff
@@ -45,14 +45,20 @@ class Table {
         return app.tables.element.children(matching: .cell)
     }
 
-    class func getVisibleLabelledCellsNumber() -> Int {
+    class func getVisibleLabelledCells() -> [XCUIElement] {
         // We need only the table cells that have X = 0 and a non-empty label
-        // otherwise we will include invisible elements from Sidebar pane, when Notes List is open
-        // or the elements from Notes List when Settings are open
-        // or the visible tags search suggestions
+        // Currently (besides using object dimentions) this is the way to
```
mokagio:
```suggestion
        // Currently (besides using object dimensions) this is the way to
```

SimplenoteUITests/Generic.swift
```diff
@@ -45,14 +45,20 @@ class Table {
...
         return Table.getAllCells()
             .filter { $0.frame.minX == 0.0 && $0.label.isEmpty == false }
-            .count
+    }
+
+    class func getVisibleLabelledCellsNames() -> [String] {
+        return Table.getVisibleLabelledCells().compactMap { $0.label }
+    }
+
+    class func getVisibleLabelledCellsNumber() -> Int {
```
mokagio: Nitpicks.

Maybe calling this "-count" would keep the wording consistent with what developers using Swift are used do (i.e. the `count` property).

```suggestion
    class func getVisibleLabelledCellsCount() -> Int {
```

Also, there might be only marginal benefit in having his method vs calling `.count` directly.

```swift
XCTAssertEqual(Table.getVisibleLabelledCellsCount(), 42)

// vs

XCTAssertEqual(Table.getVisibleLabelledCells.count, 42)
```

Not much difference really, the only advantage of using `.count` directly is that we'd have one less function to maintain. 🤷

SimplenoteUITests/Generic.swift
```diff
@@ -65,9 +71,14 @@ class Table {
     }
 
     class func trashCell(noteName: String) {
-        Table.getCell(label: noteName).swipeLeft()
-        sleep(1)
-        Table.getCell(label: noteName).buttons[UID.Button.itemTrash].tap()
+        // `Trash Note` and `Delete note forever` buttons have different labels
+        // since #1191. To use the correct label, we need to know where we are.
```
mokagio: It's cool how GitHub adds a link to the PR here 😍 

But, nitpick, the comment may be read from somewhere other than GitHub. In such case, a combination of commit and GitHub PR link might be more robust.

```suggestion
        // since 07fcccf1039495768ecdf9909d3dbd1b255936cd
        // (https://github.com/Automattic/simplenote-ios/pull/1191).
        // To use the correct label, we need to know where we are.
```

mokagio: As a side note, this scenario would be (will be) simpler in a page object world. Both the note editor and the trash screens would have their own object implementing a the method like this one, but without any conditional logic.

SimplenoteUITests/Generic.swift
```diff
@@ -65,9 +71,14 @@ class Table {
     }
 
     class func trashCell(noteName: String) {
```
mokagio, line 69: I'd suggest making the fact that this method not only finds the trash cell for the given note but also action it in the name.

```suggestion
    class func actionTrashCell(noteName: String) {
```

or 

```suggestion
    class func deleteNoteViaTrashCell(noteName: String) {
```

SimplenoteUITests/NoteList.swift
```diff
@@ -69,25 +69,9 @@ class NoteList {
...
-        }
-
-        for _ in 0..<notesNumber {
-            let cell = app.tables.cells.element(boundBy: startingIndex)
-            cell.swipeLeft()
-            cell.buttons[UID.Button.itemTrash].tap()
-        }
+        let noteNames = Table.getVisibleLabelledCellsNames()
+        guard noteNames.isEmpty == false else { return }
+        noteNames.forEach { Table.trashCell(noteName: $0) }
```
mokagio: If `noteNames` is empty, `forEach` will run on no notes. That's one of the many things I like about chaining operations on collections, you don't need to handle the case of an empty collection; if it's empty nothing will happen.

This simplifies (IMHO) the code:

```suggestion
        Table
            .getVisibleLabelledCellsNames()
            .forEach { Table.trashCell(noteName: $0) }
```

mokagio (approved):
All tests passed on the iPhone 12 Simulator 👍

SimplenoteUITests/Generic.swift
pachlava: Nice! Thank you @mokagio !

SimplenoteUITests/Generic.swift
pachlava: ✅  🙇‍♂️

SimplenoteUITests/NoteList.swift
pachlava: This is extra nice, thank you!

PR #1216, opened 2021-03-22
Merge release 4.33.1 into develop

charliescheer:
### Fix
This PR is to create a hot fix branch that fixes a crash caused by search highlighting in the note editor being out of range.  The issue was originally raised in #1193   and also mentioned in #1204 

There was also a PR to fix this in develop #1213 

In this PR, before highlighting searched values in a loaded note, the note editor confirms that it has valid ranges to highlight, if it does not, then it does not try to highlight (thus avoid the crash)

### Test
1. create a new note, give it a tag of "A" and in the text body add the text "Abc中文"
2. exit the note and go to search, search for "A"
3. select the tag search from the options, then select the note you just created.
4. The note should load fine without crashing
5. Exit the note, return to the search and type in "Abc中文"
6. Select the note from the search results
7. The note should load fine without crashing

### Review

### Release
Fixes a crash that was caused by issues with search result highlighting


mokagio (commented):
This LGTM, the diff is tiny and matches what went in #1213 (although I'm not sure why the changes to `SPNoteEditorViewController` appears, given it's already in `develop` via #1213 🤔 )

I'm going to wait for @jleandroperez or @eshurakov to give an approve just because so once that happens I can proceed with the deployment and merge this into develop in one go, as opposed to merge this first and then open another PR just for the finalization. I'd like to do this because it would streamline the process for the 4.34 code freeze.

mokagio:
FYI, I went ahead and submitted the build for review to save time in delivering this fix to our users, under the assumption everything is good – if something isn't we can always reject the build later.

mokagio:
cc @pachlava

jleandroperez (approved):
Looks great!!

:shipit:

PR #1222, opened 2021-03-23
 In app notifications - Notices!!

charliescheer:
Note: I am updating this PR to point back to the main feature branch as the difference between the integrations branch and the publish-mk2 branch was minimal.  Updating the description of this PR to cover adding the notices into the app

### Fix
This PR combines the Notice controller/presenter and the Notice UI into Simplenote to present notices to the user.  It also includes the creation of a publish controller which watches for changes of an object that has been published and then takes actions on based on those changes.

#### Links and Trash
1. Go to a note -> Options and select copy internal link.  A notice should appear saying Link Copied and then exit the options view
2. Return to the options menu and tap on the Move to trash button.  A notice will appear with an option to undo that action.  Test the undo restores the note
3. in the note list swipe to the left to see the contextual actions, choose the link button to see a link copied notice
4. open the contextual actions again and choose trash, to see the trashed notice

#### Publishing
1. Go to the options menu and flip the switch to enable publishing.  A notice that says "publishing..." should appear
2. after a few seconds, when the publishing is complete a notice confirming the note was published appears with an option to copy the publish link, test that link copy works
3. Unpublish the note and you will see "unpublishing..."
4. when unpublish is complete there will be another notice this time with an undo option, give that a test

### Review
> Only one developer is are required to review these changes, but anyone can perform the review.


charliescheer:
@jleandroperez I made some changes to the Publish controller per our chat.  Can you take a look at these changes again and see what you think?

charliescheer:
@SylvesterWilmott this PR bring the in app notifications to completed.  There will be an overall PR coming soon, but if you want to check them out now in process, they are all built into this version

Simplenote/Classes/OptionsViewController.swift
```diff
@@ -371,8 +371,23 @@ private extension OptionsViewController {
 
     @IBAction
     func publishWasPressed(_ newState: Bool) {
-        SPTracker.trackEditorNotePublishEnabled(newState)
```
jleandroperez, line 374: Nipticky: Mind moving this invocation back to OptionsViewController?

Reasoning is:
- PublishController's sole responsibility is to encapsulate all the code related to publishing
- And symmetry! all of the other actions call SPTracker right here (ie.: `copyInterlinkWasPressed`)

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,101 @@
...
+        startTimer(in: wrapper)
+    }
+
+    private func changePublishState(for note: Note, to published: Bool) {
+        note.published = published
+        note.modificationDate = Date()
+        SPAppDelegate.shared().save()
+    }
+
+    fileprivate func removeListenerCallback(for key: String) {
```
jleandroperez: Nipticky: Mind renaming to `private` ? (same for `startTimer`!)

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,101 @@
...
+    var timer: Timer
+
+    init(note: Note, block: @escaping (PublishState) -> Void) {
+        self.note = note
+        self.block = block
+        self.timer = Timer()
+    }
+}
+
+class PublishMap {
```
jleandroperez: Let's drop this type, since it's not really needed:

1. It only contains a single property, and it's set to `public`
2. Doesn't expose any methods
3. All of the code interacting with `PublishMap` just accesses the `.container`

SimplenoteTests/PublishControllerTests.swift
```diff
@@ -0,0 +1,99 @@
+import XCTest
+@testable import Simplenote
+
+private let storage = MockupStorageManager()
+private let timerFactory = MockTimerFactory()
+private let callbackMap = PublishMap()
+private let publishController = PublishController(timerFactory: timerFactory, callbackMap: callbackMap)
+private let mockAppDelegate = MockAppDelegate(publishController: publishController)
```
jleandroperez: These are global properties. Mind moving them into `PublishControllerTests` ?

SimplenoteTests/PublishControllerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    func update(note: Note, to published: Bool, with key: String, with url: String) {
+        note.publishURL = url
+        note.published = published
+        publishController.didReceiveUpdateFromSimperium(for: key as NSString)
+    }
+}
+
+class MockPublishMap: PublishMap {
```
jleandroperez: Good to drop this subclass, since it doesn't override / add anything to `PublishMap`

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,101 @@
...
+        callbackMap.container[note.simperiumKey] = wrapper
+
+        SPTracker.trackEditorNotePublishEnabled(published)
+        changePublishState(for: note, to: published)
+
+        published ? update(wrapper, to: .publishing) : update(wrapper, to: .unpublishing)
+    }
+
+    @objc(didReceiveUpdateFromSimperiumForKey:)
+    func didReceiveUpdateFromSimperium(for key: NSString) {
```
jleandroperez:
```suggestion
    func didReceiveUpdateFromSimperium(for key: String) {
```

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,101 @@
...
+        guard let wrapper = callbackMap.container[key as String] else {
```
jleandroperez:
```suggestion
        guard let wrapper = callbackMap.container[key] else {
```

jleandroperez: Swift bridges `Swift.String` as NSString, automatically. No need to explicitly request a NSString

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,101 @@
...
+        if wrapper.note.published && wrapper.note.publishURL != nil {
+            update(wrapper, to: .published)
+            return
+        }
+
+        if !wrapper.note.published {
+            update(wrapper, to: .unpublished)
+
+            return
+        }
```
jleandroperez: How about if the `wrapper` encapsulates the logic to calculate the actual `PublishedState` ?

(That would simplify this API a lot!)

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,101 @@
...
+    case publishing
+    case published
+    case unpublishing
+    case unpublished
+}
+
+class PublishListenWrapper: NSObject {
+    let note: Note
+    let block: (PublishState) -> Void
+    var timer: Timer
```
jleandroperez: Rather than maintaining a Timer per Published Note, how about iffff:

A. We keep a (single) timer in the PublishController, which gets scheduled / rescheduled after you publish a Note. (YES scheduling two notes in a row will result in a single timer).

B. We just maintain a Timestamp in the Wrapper, and check if the listener has expired

```
expiration.timeIntervalSinceNow < Constants.timeout
```

Reasoning behind this change is: a structure such as this Wrapper should not contain this timer. The `Callback` runs `self.removeListenerCallback` from **PublishController** (and doesn't really call `PublishListenWrapper`.

*PLUS* this change would allow you to switch this to a `struct`, and drop the initializer!

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,101 @@
...
+}
+
+enum PublishState {
+    case publishing
+    case published
+    case unpublishing
+    case unpublished
+}
+
+class PublishListenWrapper: NSObject {
```
jleandroperez:
```suggestion
class PublishListenWrapper {
```

jleandroperez (commented):
Looking good!! sending you few comment sir!

charliescheer:
@jleandroperez I have made some more updates to this PR.  Can you check it out when you get a chance.  

I could use some guidance of the PublishListenerFactory that I created in this round of updates.  I made that change in response to the feedback asking me to remove the callbackMap from the PublishMap class I had created.  I had set it up this way to be able to access the callback map during testing.  By setting the callbackMap back to a dictionary I couldn't access it anymore...

So similar to the TimerFactory I used in the Notice controller tests (HT to @eshurakov for the help with that one!) I created a factory to make the PublishListenerWrapper and then I could control that during tests instead.  I'm not sure about the implementation though, what do you think?

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,112 @@
...
+    private mutating func setExpiration() {
+        expiration = Date()
+    }
+}
+
+class PublishListenerFactory {
+    func publishListenerWrapper(note: Note, block: @escaping (PublishState) -> Void, expiration: Date) -> PublishListenWrapper {
+        return PublishListenWrapper(note: note, block: block, expiration: expiration)
+    }
+}
```
jleandroperez, line 66: Let's drop this one, since it's really not needed (it's wrapping access to the default initializer)

Simplenote/NoticeFactory.swift
```diff
@@ -0,0 +1,38 @@
...
+            UIPasteboard.general.copyPublicLink(to: note)
+            NoticeController.shared.present(NoticeFactory.linkCopied)
+        }
+        let notice = Notice(message: NSLocalizedString("Publish Successful.", comment: "Notice up succesful publishing"), action: action)
+        return notice
+    }
+}
+
+extension NoticeFactory {
+    static let copyLinkMessage = NSLocalizedString("Copy link.", comment: "Copy link action")
```
jleandroperez: Nipticky: For consistency, it might be good to either

A. Move all of the LocalizedStrings as a static let, right here
B. Move all of the LocalizedStrings to a private enum

Simplenote/NoticeFactory.swift
```diff
@@ -0,0 +1,38 @@
...
+        let action = NoticeAction(title: NSLocalizedString("Undo", comment: "Undo action")) {
+            SPAppDelegate.shared().publishController.updatePublishState(for: note, to: true, completion: { (_) in
+                NoticeController.shared.present(NoticeFactory.published(note))
+            })
+        }
+        let notice = Notice(message: NSLocalizedString("Unpublish successful.", comment: "Notice of publishing unsuccessful"), action: action)
+        return notice
+    }
+
+    static func published(_ note: Note) -> Notice{
```
jleandroperez:
```suggestion
    static func published(_ note: Note) -> Notice {
```

Simplenote/NoticeFactory.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+
+struct NoticeFactory {
+    static let linkCopied = Notice(message: NoticeFactory.copyLinkMessage, action: nil)
+    static let publishing = Notice(message: NSLocalizedString("Publishing...", comment: "Notice of publishing action"), action: nil)
+    static let unpublishing = Notice(message: NSLocalizedString("Unpublishing...", comment: "Notice of unpublishing action"), action: nil)
```
jleandroperez: Nipticky: For consistency's reasons, maybe all of the notices should be `static func` ?

Simplenote/Classes/OptionsViewController.swift
```diff
@@ -372,7 +372,23 @@ private extension OptionsViewController {
...
+                NoticeController.shared.present(notice)
+            case .published:
+                let notice = NoticeFactory.published(self.note)
+                NoticeController.shared.present(notice)
+            case .unpublishing:
+                let notice = NoticeFactory.unpublishing
+                NoticeController.shared.present(notice)
+            case .unpublished:
+                let notice = NoticeFactory.unpublished(self.note)
+                NoticeController.shared.present(notice)
```
jleandroperez: (No need to change anything!!!)

Another alternative to this, is to use an extension on Notice (in a standalone file), and do this:

```
extension Notice {
    static func unpublishingNotice() { ... }
    static func unpublishedNotice(note: Note) { .. }
}
```

That enables this syntax:

```
NoticeController.shared.present( .unpublishingNotice() )
```

Both approaches are 100% correct, no need to change anything!

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    func updatePublishState(for note: Note, to published: Bool, completion: @escaping (PublishState) -> Void) {
+        if note.published == published {
+            return
+        }
+
+        callbackMap[note.simperiumKey] = publishListenerFactory.publishListenerWrapper(note: note, block: completion, expiration: Date())
+
+        timer = timerFactory.scheduledTimer(with: Constants.timeOut, completion: {
```
jleandroperez: What happens if there was already a Timer scheduled?

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,112 @@
...
+            self.removeExpiredCallbacks()
+        })
+
+        changePublishState(for: note, to: published)
+
+        published ?
+            callbackMap[note.simperiumKey]?.update(to: .publishing) : callbackMap[note.simperiumKey]?.update(to: .unpublishing)
```
jleandroperez: `completion` is usually expected to be invoked **just once**, when the OP is completed (or fails). Otherwise, this API would effectively be coupling two different tasks:

A. Updating the Published State
B. Subscribing to Publishing State changes

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    private func removeExpiredCallbacks() {
+        for wrapper in callbackMap {
+            if wrapper.value.isExpired {
+                callbackMap.removeValue(forKey: wrapper.key)
+            }
+        }
+
+        if !callbackMap.isEmpty {
+            timer = timerFactory.scheduledTimer(with: Constants.timeOut, completion: {
```
jleandroperez: This one is coupling the API. It's got, effectively, two tasks:

A. Cleaning up
B. Scheduling the timer, again

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,112 @@
...
+    case publishing
+    case published
+    case unpublishing
+    case unpublished
+}
+
+struct PublishListenWrapper {
+    let note: Note
+    let block: (PublishState) -> Void
+    var expiration: Date
```
jleandroperez: This one should probably be immutable, and set just once (after creation).

The goal of this structure is to listen for either **Success** or **Failure** toggling the Pubilshing state (rather than subscribing to all of the PubilshState changes)

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,112 @@
...
+    mutating func handleListenResponse() {
+        if note.published && note.publishURL != nil {
+            update(to: .published)
+        }
+
+        if !note.published {
+            update(to: .unpublished)
+        }
+
+        setExpiration()
```
jleandroperez: This API should **not** change the Expiration (since it's expected to be handling a remote event).

Plus `setExpiration` might get called twice (L89 or L93 and L96)

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,112 @@
...
+        block(state)
+        setExpiration()
+    }
+
+    mutating func handleListenResponse() {
+        if note.published && note.publishURL != nil {
+            update(to: .published)
+        }
+
+        if !note.published {
```
jleandroperez: A simpler way is perhaps doing something like this:

```
extension Note {
    var pubilshState: PublishState { ... }
}

```

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,112 @@
...
+    let note: Note
+    let block: (PublishState) -> Void
+    var expiration: Date
+
+    var isExpired: Bool {
+        expiration.timeIntervalSinceNow < -Constants.timeOut
+    }
+
+    mutating func update(to state: PublishState) {
+        block(state)
```
jleandroperez: It would probably be cleaner if the caller deals with the actual execution (and this structure only exposes a couple primitives)

jleandroperez (commented):
Sending you few nipticky comments!! let's chat over Slack whenever you've got a chance!

Thank youu!!

Simplenote/Classes/OptionsViewController.swift
charliescheer: I thought about doing this a while back, and decided to go with the factory so that the notices were not coupled to Simplenote. I see how doing this in an extension still keeps the simplenote specific details of the notices being used private to simplenote 🤔  I think for now I like it as is, but would be open to changing to the using an extension instead

Simplenote/Controllers/PublishController.swift
charliescheer: I removed the timers in favor of having the callback listener dropped once a change is heard

charliescheer:
@jleandroperez have updated again. Only thing not changed is the  PublishListenerFactory.  Can you take another look and let me know what you think?

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,69 @@
...
+    func updatePublishState(for note: Note, to published: Bool) {
+        if note.published == published {
+            return
+        }
+
+        changePublishState(for: note, to: published)
+
+        beginListeningForChanges(to: note, timeOut: Constants.timeOut)
+
+        publishStatechanged(for: note)
```
jleandroperez:
```suggestion
        publishStateChanged(for: note)
```

Simplenote/NoticeFactory.swift
```diff
@@ -0,0 +1,45 @@
...
+    static func linkCopied() -> Notice {
+        return Notice(message: Messages.copyLink, action: nil)
+    }
+
+    static func publishing() -> Notice {
+        Notice(message: Messages.publishing, action: nil)
+    }
+
+    static func unpublishing() -> Notice {
+        return Notice(message: Messages.unpublishing, action: nil)
```
jleandroperez: Nipticky: Tiny inconsistency between this and L9 (good to drop return if it's the only statement!)

Simplenote/NoticeFactory.swift
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    static func noteTrashed(_ note: Note, onUndo: @escaping ()-> Void) -> Notice {
+        let action = NoticeAction(title: Messages.undo, handler: onUndo)
+        let notice = Notice(message: Messages.trashed, action: action)
+        return notice
```
jleandroperez: Nipticky: No need to assign `let notice = ` since we're not doing anything with the value

Simplenote/NoticeFactory.swift
```diff
@@ -0,0 +1,45 @@
+import Foundation
+
+struct NoticeFactory {
+    static func linkCopied() -> Notice {
+        return Notice(message: Messages.copyLink, action: nil)
```
jleandroperez: I think this Notice should say `Link Copied` (right now it says `Copy Link`)

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -67,6 +67,15 @@ extension SPAppDelegate {
...
         versionsController = VersionsController(bucket: simperium.notesBucket)
     }
+
+    @objc
+    func configurePublishController() {
+        publishController = PublishController()
+        publishController.onUpdate = {
+            (note) in
+            PublishNoticePresenter.presentNotice(for: note)
+        }
```
jleandroperez, line 76: Nipticky: the capture statement is usually placed right by the `{` (rather than in the first closure line)

```
     publishController.onUpdate = { note in
             PublishNoticePresenter.presentNotice(for: note)
         }
```

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    // MARK: Listener Notifications
+    func didReceiveUpdateNotification(for key: String, with memberNames: NSArray) {
+        guard memberNames.contains(Constants.observedProperty),
+              let note = observedNotes[key] else {
+            return
+        }
+
+        publishStatechanged(for: note)
```
jleandroperez:
```suggestion
        publishStateChanged(for: note)
```

Simplenote/Controllers/PublishController.swift
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private func changePublishState(for note: Note, to published: Bool) {
+        note.published = published
+        note.modificationDate = Date()
+        SPAppDelegate.shared().save()
+    }
+
+    private func publishStatechanged(for note: Note) {
```
jleandroperez:
```suggestion
    private func publishStateChanged(for note: Note) {
```

Simplenote/SPAppDelegate.m
```diff
@@ -470,6 +475,9 @@ - (void)bucket:(SPBucket *)bucket didChangeObjectForKey:(NSString *)key forChang
                 if ([key isEqualToString:self.noteEditorViewController.note.simperiumKey]) {
                     [self.noteEditorViewController didDeleteCurrentNote];
                 }
+
+                [self alertListenerOfDeleteWithKey:key];
```
jleandroperez: Nipticky: Do you think we could drop the bridging APIs and invoke directly `publishController?`

jleandroperez (approved):
Hey there @charliescheer !!

Impressive work, congratz!!! Few nipticky / cosmetic notes. Other than that, looks solid!!

:shipit:

Simplenote/NoticeFactory.swift
charliescheer: Ah nice catch, there should be a "copy link" button title and a "Link copied" alert message, I was missing the second one.  Added

PR #1225, opened 2021-03-30
Adding `UI Tests` job to CircleCI

pachlava:
This PR is hopefully a first step in integrating automated UI tests into CI. It's a rip-off from already existing code in this repo, with additions stolen from [WordPress-iOS](https://github.com/wordpress-mobile/WordPress-iOS) repo (with refs and props to both).

What it currently does:

- Runs login tests on every commit, only when the branch contains `UI-tests` in title (as for current PR), on iPhone 12, using Xcode 12.4.0.

It's not much, but it's the start. There was a lot of trial an error, since I'm a total noob in both Fastlane and CircleCI. Thus, I might not know of possible side effects caused by my changes. Login tests are not causing race conditions and can be executed in any number of instances. I still have several open questions, and some of them should be addressed in this PR (like the first one for sure):

1. The branch filter or cron job rule to run the tests (so far, we are talking only about login tests. Extension of executed tests should be a topic for a separate PR, and the filter should be changed once this happens). Options: 

 - make the job require an approval
 - every time, as for unit tests
 - running on schedule with cron
 - ...

2. The UI tests run in parallel to unit tests, and also reset the simulator. Because of this, I'm not using the `test_without_building` option.
3. The process takes about 9 minutes (with 1.5 minutes of actual tests), which in total is approx 2 minutes longer than unit tests. 2 minutes of 9 are taken by artifacts upload. I feel like they are not really needed for this kind of job. Should this be reduced to test results only?

### Fix
I needed to change three files:
1. Specify `Simplenote.app` as an executable in `SimplenoteUITests` scheme.
2. Add `run_UI_tests` lane to `Fastfile`
3. Add a `UI Tests` job to `.circleci/config.yml `

### Test
1. Change any part of the code (e.g. comment) in this PR and see that:
 i. `UI Tests` job starts running in parallel to `Tests`
 ii. It gets done in ~9 minutes
 iii. Once done, the status of 8 tests can be seen in `TEST` tab for a job
 iv. more details can be seen in `report.*` files in artifacts.
 v. no regression is introduced by my changes.

### Review
All input is very welcomed 🙇 

### Release
I don't think release notes are needed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,6 +144,7 @@ jobs:
           command: bundle exec fastlane run_ui_tests device:'<< parameters.device >>'
       - store_artifacts:
           path: output/scan/report.html
+      - store_artifacts:
           path: output/scan/report.junit
       - store_test_results:
           path: output/scan
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -130,7 +130,7 @@ jobs:
       xcode-version: "12.4.0"
     environment:
       FL_OUTPUT_DIR: output
-      FASTLANE_LANE: run_login_UI_tests
+      FASTLANE_LANE: run_UI_tests
     steps:
       - git/shallow-checkout
       - ios/install-dependencies:
@@ -157,7 +157,7 @@ workflows:
     jobs:
       - Test
       - UI Tests:
-          name: UI Tests (iPhone 12)
+          name: UI Tests ( iphone_test_device.device )
           <<: *iphone_test_device
   Installable Build:
     jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,9 +142,9 @@ jobs:
           name: Fastlane
           command: bundle exec fastlane $FASTLANE_LANE
       - store_artifacts:
-          path: output
+          path: $FL_OUTPUT_DIR
       - store_test_results:
-          path: output/scan
+          path: $FL_OUTPUT_DIR/scan
 
 workflows:
   simplenote_ios:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,15 +112,14 @@ jobs:
             failure_message: '${SLACK_FAILURE_MESSAGE}'
             success_message: '${SLACK_SUCCESS_MESSAGE}'
   UI Tests:
+    # Props to work done by others here and in
+    # https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/.circleci/config.yml
+    # this is a combined rip-off from both
     parameters:     
       device:
         description: The device (e.g. "iPhone 12") to use when running tests.
         type: string
         default: "iPhone 12"
-      ios-version:
-        description: The iOS version (e.g. "14.4") of the device used to run tests.
-        type: string
-        default: "14.4"
     executor:
       name: ios/default
       xcode-version: "12.4.0"
@@ -142,7 +141,7 @@ jobs:
       # END: Swift Package Manager Workaround
       - run:
           name: Fastlane
-          command: bundle exec fastlane $FASTLANE_LANE device:'<< parameters.device >>' ios-version:'<< parameters.ios-version >>'
+          command: bundle exec fastlane $FASTLANE_LANE device:'<< parameters.device >>'
       - store_artifacts:
           path: output
       - store_test_results:
@@ -154,8 +153,7 @@ workflows:
       - Test
       - UI Tests:
           name: UI Tests (iPhone 12)
-          device: iPhone 12
-          ios-version: "14.4"
+          device: "iPhone 12"
   Installable Build:
     jobs:
       - Hold:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,9 +142,9 @@ jobs:
           name: Fastlane
           command: bundle exec fastlane $FASTLANE_LANE
       - store_artifacts:
-          path: $FL_OUTPUT_DIR
+          path: output
       - store_test_results:
-          path: $FL_OUTPUT_DIR/scan
+          path: output/scan
 
 workflows:
   simplenote_ios:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,6 +144,7 @@ jobs:
           command: bundle exec fastlane run_ui_tests device:'<< parameters.device >>'
       - store_artifacts:
           path: output/scan/report.html
+          path: output/scan/report.junit
       - store_test_results:
           path: output/scan
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,8 +158,7 @@ workflows:
           filters:
             branches:
               only:
-                - develop
-                - /^release.*/
+                - /[UI-testss]]/
   Installable Build:
     jobs:
       - Hold:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,8 +143,9 @@ workflows:
   simplenote_ios:
     jobs:
       - Test
-      - UI Tests
+      - UI Tests:
           name: UI Tests (iPhone 12)
+          type: approval
   Installable Build:
     jobs:
       - Hold:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,8 +142,6 @@ jobs:
       - run:
           name: Fastlane
           command: bundle exec fastlane run_ui_tests device:'<< parameters.device >>'
-      #- store_artifacts:
-      #    path: output
       - store_test_results:
           path: output/scan
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,7 +126,6 @@ jobs:
       xcode-version: "12.4.0"
     environment:
       FL_OUTPUT_DIR: output
-      FASTLANE_LANE: run_UI_tests
     steps:
       - git/shallow-checkout
       - ios/install-dependencies:
@@ -142,7 +141,7 @@ jobs:
       # END: Swift Package Manager Workaround
       - run:
           name: Fastlane
-          command: bundle exec fastlane $FASTLANE_LANE device:'<< parameters.device >>'
+          command: bundle exec fastlane run_ui_tests device:'<< parameters.device >>'
       - store_artifacts:
           path: output
       - store_test_results:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,11 @@ jobs:
             failure_message: '${SLACK_FAILURE_MESSAGE}'
             success_message: '${SLACK_SUCCESS_MESSAGE}'
   UI Tests:
+    parameters:     
+      device:
+        description: The device (e.g. "iPhone 12") to use when running tests.
+        type: string
+        default: "iPhone 12"  
     executor:
       name: ios/default
       xcode-version: "12.4.0"
@@ -133,7 +138,7 @@ jobs:
       # END: Swift Package Manager Workaround
       - run:
           name: Fastlane
-          command: bundle exec fastlane $FASTLANE_LANE
+          command: bundle exec fastlane $FASTLANE_LANE device:'<< parameters.device >>'
       - store_artifacts:
           path: output
       - store_test_results:
@@ -145,7 +150,7 @@ workflows:
       - Test
       - UI Tests:
           name: UI Tests (iPhone 12)
-          type: approval
+          device: "iPhone 12"
   Installable Build:
     jobs:
       - Hold:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,6 @@ workflows:
       - Test
       - UI Tests
           name: UI Tests (iPhone 12)
-          type: approval
   Installable Build:
     jobs:
       - Hold:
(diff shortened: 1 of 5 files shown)


fastlane/Fastfile
```diff
@@ -593,6 +593,37 @@ end
...
+
+    #####################################################################################
+    # run_UI_tests
+    # -----------------------------------------------------------------------------------
+    # Usage:
+    # bundle exec fastlane run_UI_tests [device:<Name of the iOS Simulator >]
+    #
+    # Example:
+    # bundle exec fastlane run_UI_tests device:"iPhone 12"
+    #####################################################################################
```
mokagio: Nitpick

```suggestion
  # Props to work done by others in this file and in
  # https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/fastlane/Fastfile
  # this is a combined rip-off from both

  #####################################################################################
  # run_UI_tests
  # -----------------------------------------------------------------------------------
  # Usage:
  # bundle exec fastlane run_UI_tests [device:<Name of the iOS Simulator >]
  #
  # Example:
  # bundle exec fastlane run_UI_tests device:"iPhone 12"
  #####################################################################################
```

I don't think we need a different indentation for this block. I've seen it in some `Fastfile` but I think it will soon be sorted out as we work on standardizing the format.

If we were to use an IDE to auto-indent this file, it would bring the indentation back one level.

fastlane/Fastfile
```diff
@@ -593,6 +593,37 @@ end
...
+  # A lane to run a subset of automated UI tests. Since only login tests are
+  # selected for now, there will be no credentials race condition.
```
mokagio, line 610: Great idea starting with only one sub-set of tests to set this automation up 🙌

mokagio, line 610: Nitpick: this could go as part of the fenced comment block above, e.g. like in the `code_freeze` lane: https://github.com/Automattic/simplenote-ios/blob/b7390955954e75215a6935c59fb831f680bd79da/fastlane/Fastfile#L53-L58

fastlane/Fastfile
```diff
@@ -593,6 +593,37 @@ end
...
+  desc "Run UI tests"
+  lane :run_UI_tests do | options |
```
mokagio: Nitpick. I may be wrong, but I've never seen upper case acronyms used with snake cased names. 

Just because I've never seen it, though, it doesn't mean it's incorrect. It certainly is valid Ruby, otherwise CI would not have worked 😄 

```suggestion
  lane :run_ui_tests do | options |
```

fastlane/Fastfile
```diff
@@ -593,6 +593,37 @@ end
...
+  scan(
+    workspace: workspace_name,
+    scheme: ui_tests_scheme,
+    device: options[:device],
```
mokagio: TIL: if we don't pass a value when calling the lane, Fastlane will use iPhone 11 as the device value.

fastlane/Fastfile
```diff
@@ -704,6 +735,10 @@ def screenshots_scheme
   "SimplenoteScreenshots"
 end
 
+def ui_tests_scheme
+  "SimplenoteUITests"
+end
```
mokagio: Pretty suer I wrote some of these `def name "value" end` constructs 🤔 .

Since then, I learned of a better way of doing it: using Ruby constants.

```ruby
```

The value of using a constant is two-fold. First, it's appropriate because this value won't change in the context of the lane execution. Second, it's shorter than using a method (and possibly more performant, although I'm not sure what Ruby does under the hood and it really doesn't matter in our use case).

Worth pointing out that using a variable (`ui_tests_scheme = "..."`) wouldn't work because the lane wouldn't be able to find it in its local scope, while the constant, being a constant, can be found.

fastlane/Fastfile
```diff
@@ -593,6 +593,37 @@ end
...
+  desc "Run UI tests"
+  lane :run_UI_tests do | options |
+  scan(
+    workspace: workspace_name,
+    scheme: ui_tests_scheme,
+    device: options[:device],
+    prelaunch_simulator: true,
+    reset_simulator: true,
+    only_testing: "SimplenoteUITests/SimplenoteUISmokeTestsLogin"
+  )
```
mokagio: When I run the lane locally (simply with `bundle exec fastlane run_UI_tests`, it failed with 

```
Reason: Cannot test target “SimplenoteUITests” on “iPhone 11”: iPhone 11’s iOS Simulator 13.4 doesn’t match SimplenoteUITests’s iOS Simulator 14.4 deployment target.
```

Even though this lane is only meant to run on CI, really, it's necessary for it to run locally out of the box, too.

We may want to use fallback values for both device and iOS version:

```suggestion
  scan(
    workspace: workspace_name,
    scheme: ui_tests_scheme,
    device: options[:device] || 'iPhone 12',
    deployment_target_version: options[:ios_version] || '14.4'
    prelaunch_simulator: true,
    reset_simulator: true,
    only_testing: "SimplenoteUITests/SimplenoteUISmokeTestsLogin"
  )
```

(There's actually a `simulator_version` parameter at the bottom of the file that we could use instead of `14.4`)

Alternatively, we could check that the values for the device and iOS version are explicitly passed to the lane before calling `scan`. 

I haven't tested it, but something like:


```ruby
usage = "Usage bundle exec fastlane run_ui_tests device:<device> ios_version:<ios version>"

UI.user_error!(usage) if options[:device].nil? || options[:ios_version].nil?

scan(...)
```

.circleci/config.yml
```diff
@@ -112,11 +112,52 @@ jobs:
...
+          name: Copy Secrets
+          command: bundle exec fastlane run configure_apply
+      # Start: Swift Package Manager Workaround
+      # Ref. https://support.circleci.com/hc/en-us/articles/360044709573-Swift-Package-Manager-fails-to-clone-from-private-Git-repositories
+      - run: rm ~/.ssh/id_rsa
+      - run: for ip in $(dig @8.8.8.8 github.com +short); do ssh-keyscan github.com,$ip; ssh-keyscan $ip; done 2>/dev/null >> ~/.ssh/known_hosts || true
+      # END: Swift Package Manager Workaround
+      - run:
+          name: Fastlane
+          command: bundle exec fastlane $FASTLANE_LANE device:'<< parameters.device >>'
```
mokagio: Putting the name of the lane in an environment variable adds a layer of indirection here: to know the exact command we have to find the value of the env var.

Is there a particular reason for it? If not, I'd suggest using the value directly:

```suggestion
          command: bundle exec fastlane run_UI_tests device:'<< parameters.device >>'
```

.circleci/config.yml
```diff
@@ -112,11 +112,52 @@ jobs:
...
 workflows:
   simplenote_ios:
     jobs:
       - Test
+      - UI Tests:
+          name: UI Tests (iPhone 12)
+          device: "iPhone 12"
+          filters:
+            branches:
+              only: /.*UI-tests.*/
```
mokagio, line 161: You asked about branch filter vs cron schedule vs other options in the PR description.

What do you think is the best option? I think the answer also depends on how soon we want to know if the tests break: as soon as code changes? before every release? ad hoc? once a day?

mokagio (commented):
Great to see this shaping up! Thank you for putting in the time to learn this stuff @pachlava!

Left a few comments and suggestion...

Regarding your questions from the PR description:

> 2. The UI tests run in parallel to unit tests, and also reset the simulator. Because of this, I'm not using the `test_without_building` option.

This is something to sort out in a followup PR. I _think_ we should be able to have one pre-step that does the "build for testing" build and then run the unit and UI tests in parallel using "test without building" and the artifact built by the previous step.

I'm pretty sure that's what WordPress iOS does already:

image

If we could run the build only once, it would definitely save CI time.

> 3. The process takes about 9 minutes (with 1.5 minutes of actual tests), which in total is approx 2 minutes longer than unit tests. 2 minutes of 9 are taken by artifacts upload. I feel like they are not really needed for this kind of job. Should this be reduced to test results only?

I've never had the need to look into any of the artifacts from the unit or UI tests output. I'd say we can actually drop the whole upload process – one day we might found we actually need that data and regret it, but I'm struggling to see a case for keeping them ⚖️ Maybe if a failed UI test took a screenshot?

Keen to hear what the other @Automattic/platform-9 folks think on this.

fastlane/Fastfile
pachlava: Done! Thank you

.circleci/config.yml
pachlava: Yes, there was no real reason to do this. I think it sneaked from the documentation I was reading, so I did as was written there. Makes no sense, given the variable is used only once

fastlane/Fastfile
pachlava: @mokagio Added the fallback for `device`, thank you!

As for `deployment_target_version` parameter: if I understood correctly from the [doc](https://docs.fastlane.tools/actions/run_tests/), this does not specify the iOS version that will be actually used, but rather filters out the devices that do not run the specified iOS version, if the device was not specified. Given that we specify the device, it looks like it adds no value? Maybe I got if wrong 🤔

pachlava: I'll do more testing with `deployment_target_version` locally now...

pachlava: It actually confuses me even more now, because the Xcode target specifies the `iOS Deployment Target` anyway already, which I guess it the same thing as `deployment_target_version`

fastlane/Fastfile
pachlava: Done! Thank you 🙇

pachlava:
> I've never had the need to look into any of the artifacts from the unit or UI tests output. I'd say we can actually drop the whole upload process – one day we might found we actually need that data and regret it, but I'm struggling to see a case for keeping them ⚖️ Maybe if a failed UI test took a screenshot?

For the moment, I limited the upload only to `report.junit` and `report.html`

.circleci/config.yml
pachlava: @mokagio I'm thinking of keeping it as it is now, for next week 🤔 

I've applied everything you've mentioned in review, [apart](https://github.com/Automattic/simplenote-ios/pull/1225#discussion_r610572013) from adding `deployment_target_version` 🙇 Thank you! After limiting the copied artefacts to `report.junit` and `report.html` only, the job takes 2 minutes less.

Next week (Apr 12-18) I'll need to wrap up with automation of `Settings` tests, which is a final part of my project. During this I'm going to use branches that contain `UI-tests` part and will see how it's going to work.

pachlava: > What do you think is the best option?

Generally, in ideal case, I would like to see the tests run as much a possible, on every change. We have them now, and it costs us nothing but time. But the more tests will we select, the more problematic will it be. Probably, it will not be suitable for everyone.

I'm for having two options.

1. A subset of tests, the rule to run them can be discussed with the team.
and
2. Running all tests, but with approval.

I'm thinking to ask in Simpelnote P2 next week about 1 and gather opinions.

mokagio: > Generally, in ideal case, I would like to see the tests run as much a possible, on every change

😄 

> We have them now, and it costs us nothing but time

Worth pointing out that CI time does cost money, too. Ideally, money spent in CI means less money lost as a result of users experiencing bugs and us doing creative rather than rote work, but it's good to remember that too.

As you say, the more tests we'll have the longer they'll take to run. In that case, the feedback time grows and it might become more effective to run them ad hoc and wait for feedback or use a scheduler.

Keen to see what comes out of the P2 discussion. I think this project is the ideal place where to experiment with different UI tests integration in CI workflows 👍

fastlane/Fastfile
mokagio: Oh... Interesting... 🤔 

Maybe then there's no direct way to tell `scan` which simulator version to use. I did a quick Google and found [this](https://github.com/fastlane/fastlane/issues/10338) and [this](https://github.com/fastlane/fastlane/issues/14953) issues which seem to confirm that.

Let's not worry about that for now then. Using the latest version of Xcode in CI should guarantee the use of the latest version of iOS in the Simulator and that might be enough for the moment. We'll look into more details if we'll ever need to run across different version of iOS, I guess 🤷‍♂️

mokagio (approved):
@pachlava 

> Next week (Apr 12-18) I'll need to wrap up with automation of Settings tests, which is a final part of my project. During this I'm going to use branches that contain `UI-tests` part and will see how it's going to work.

Okay, I'm fine with using this convention for now. I don't think it would serve us well in the long run because it relies on us _remembering_ to add keywords to the branch name, which is easy to forget. As I mentioned, though, I think this is a great place where to start exploring, so let's start with this and see what we can learn 👍 👍 👍

mokagio:
Nice work @pachlava! Keen to see how this performs for us "in the real world" 😄

.circleci/config.yml
pachlava: > Worth pointing out that CI time does cost money, too.

TIL. I didn't know we pay for CI time. For some reason, I thought this is a subscription for a number of jobs to run in parallel.

mokagio: 🤔 now that you mention it... to be honest, I'm not totally sure what our arrangement is with CircleCI. If not a matter of pay-by-the-minute it's a matter of parallelization and memory available in the hardware we use. If we were to run so many tests to bloat the queue, we'd then end up paying for implicitly in terms of delayed feedback, or explicitly by throwing more money at the problem and getting more or faster hardware to reduce the time in the queue.

PR #1228, opened 2021-04-01
Feature/in app notifications

### LINKED ISSUE #1180: Project: In App Notices Project To Do List
This issue is a project outline for implementing In App Notifications as described in issue #938 

### What
Tasks:
- [x] Create UI for notice views #1177
- [x] Create Presenter/Controller classes for delivering notifications #1178
- [x] Connect Notice classes and Simplenote to display list of notices #1179

Issues:
[In App Notifications](https://github.com/Automattic/simplenote-ios/issues/938)

charliescheer:
### Fix
This PR is to merge the In App Notifications feature into develop!!  The different components of this change have been tested along the way but here is a chance to test them all together again before moving into develop.  

In app notifications builds a tool that allows us to send simple messages to users about the activity in the app, such as copying a link or publishing a note.  Some of these messages have actions associated with them like undoing the previous change.  They look like this:

### Test
Here are some things to test with these notifications:
1. copy a link (internal or public) you should see a notice
2. Copy a link from the options view, you should see a notice and the options view will be dismissed
3. Delete a note from the options view, a notice will be presented and the options view will be dismissed
4. delete a note from the notes list.  Tap on the undo button to make sure the note restores
5. In the options menu publish a note.  You should see two notices "Publishing..." and then "Publish Successful"  Make sure to tap the copy link button on the "Publish successful" notice
6. Unpublish a note and you should see two notices "unpublishing..." and "Unpublish Successful" with an undo action
7. Make sure that the notices stay aligned correctly if you rotate the device


charliescheer:
fixes #938

charliescheer:
@SylvesterWilmott I have created a separate issue for the dismissing behavior we were talking about the other day into a new issue #1229 and will make those changes there.

charliescheer:
Just dropping a note here that I did a smoke test on this today and did not find any issues

eshurakov (commented):
I played around with the app and did some tests:
* long press on the note in the list and copying internal link doesn't present a notice
* I feel that notices appear painfully slow 🐌 😆 
* wondering, should action be large / more noticeable than the main content of the notice? 🤔
* when switching publishing we already show "progress" in the list, so not sure if it's worth to show the same information as a notice?

Screenshot 2021-04-06 at 11 03 05

jleandroperez (commented):
✅ Code looks sharp
✅ Right to Left layout works super

Only note on my end is that, perhaps, everytime a new Notice gets enqueued we could / should **ensure current note is dismissed** (if any). That way only the most recent event is onscreen (and that would simplify PR #1230).

Plus +1'ing on the animation speed Evgeny mentioned!

@SylvesterWilmott WDYT?

Other than that looks suuuuper!!

charliescheer:
@eshurakov 
>long press on the note in the list and copying internal link doesn't present a notice

Didn't even know that menu was their :) added a link copied notice to that menu!  Thanks for finding that!

SylvesterWilmott:
Looks great @charliescheer !

@eshurakov to address your feedback:

>I feel that notices appear painfully slow

@charliescheer can we increase the transition speed at which they appear?

>wondering, should action be large / more noticeable than the main content of the notice?

Let's give the action a different weight eg. semibold or medium

>when switching publishing we already show "progress" in the list, so not sure if it's worth to show the same information as a notice?

I think it's good. Remember that soon this will be replaced by a popover style menu.

Screenshot 2021-04-07 at 11 44 24

charliescheer:
Thanks for taking a look @SylvesterWilmott 
I went ahead and changed the action button font to semi bold.  WDYT? @eshurakov @SylvesterWilmott  ?

I also sped up the notice animation in speed from half a second to a quarter second. We can still go faster if need be, but I tried .1 seconds and that felt really sharp in a way that I wasn't very fond of.  .25 seconds felt like a nice quick arrival time to me.

jleandroperez (approved):
✅ New animation metrics feel ferpect
✅ New **Undo** font size looks great as well

:shipit: on my end sir!!

SylvesterWilmott:
Looks fantastic! Thanks @charliescheer

SylvesterWilmott:
How many lines of text does the notice support? Could you post a screenshot here with 2 lines of text?

eshurakov:
Thank you @charliescheer, animation speed and action button look good to me!

While I was deleting and restoring notes I noticed that sometimes the notice would immediately disappear without staying on the screen.


https://user-images.githubusercontent.com/54751/114003383-fc777600-985d-11eb-941b-7fad948b0ad4.mov

charliescheer:
@eshurakov I think that immediate dismissal is an issue with the timers that I had addressed in #1230  what was happening was the dismiss timer wasn't getting invalidated, so when the next notice was presented the old timer was causing it to be dismissed immediately.
Changes I made in #1230 should fix that.

charliescheer:
>How many lines of text does the notice support? Could you post a screenshot here with 2 lines of text

The notices should support as many lines as necessary.  Here are some images:





Going back over this did lead me to an issue I thought I had fixed before with accessibility sizing.  Without an action the accessibility sizes work fine even with a lot of text:


But with an action we may have some choices to make.  In the case that there is not enough room, do we want to truncate the action or the notice message?  
Option A, preserve the action:

Option B, preserve the notification text:

WDYT?

charliescheer:
>While I was deleting and restoring notes I noticed that sometimes the notice would immediately disappear without staying on the screen.

@eshurakov I did some more testing on what you were seeing, and I realized what was going on.  When the action buttons were being tapped they were dismissing the current notice, but not doing anything with the timer.  So if you triggered it again inside the active timer, the old timer would call dismiss.

I fixed that by removing any existing timer as soon as dismiss is called.  That should stop this from happening.

eshurakov (approved):
Nice! Thank you @charliescheer 👍

SylvesterWilmott:
>But with an action we may have some choices to make. In the case that there is not enough room, do we want to truncate the action or the notice message?

Presumably this is happening because the system doesn't want to break the word "Trashed" in the example. Is there a way for us to scale the text if there will be a word break? Otherwise maybe we can allow word breaks here.

charliescheer:
>Presumably this is happening because the system doesn't want to break the word "Trashed" in the example. Is there a way for us to scale the text if there will be a word break? Otherwise maybe we can allow word breaks here.

It is possible, but a little bit inconsistent.  We can set it to break by character for the largest of accessibility sizes, however it won't create more lines in the notice after a certain point.  In this image I set the accessibility size to the largest possible while still word wrapping:


and here is what happens if the text is too big:
 
it can start character wrapping, but won't continue adding new lines to fit all the text.

Something that we could consider is setting a limit on the max size possible for the notices.  If you use unpublish successful as an example (which is currently the longest notice) and we could set it to max out at accessibility Extra large (36pt font)  there are only 2 larger sizes beyond that one ExtraExtraLarge and ExtraExtraExtraLarge, so most of the possible values options still work, and the text will all fit in the 2 lines In this image I have the font set to it's largest possible setting, but the notice stops at the max of 36pt for the button and the label.

jleandroperez:
@SylvesterWilmott **Thank you** for checking this one!!

@charliescheer In order to fix this glitch, you'll need to enable **Automattic Width** for the Notice's label (yeah automattic, double T!!)



End result will be like this (won't clip anymore):

charliescheer:
Nice @jleandroperez making the change to automatic width did wonders for the clipping that was happening, and it now works on all accessibility sizing!

Screen Shot 2021-04-13 at 9 16 16 AM

SylvesterWilmott:
Thanks @charliescheer! Looks great.

PR #1230, opened 2021-04-02
Dismiss notice on external tap

### LINKED ISSUE #1229: Dismiss notices when app outside of notification is interacted with
### What
Currently, the in app notifications will not queue additional notices of the same type, so if a user does the same action multiple times (like deleting notes) rapidly in succession they will only see one notice.  Instead what would be better is if a user starts interacting with the rest of the app after a notice has been displayed, the current notice should be dismissed. That way you would see a notice for each action, so there wouldn't be any confusion about what action the notice is referring to.

### Why
In the current setup, say you deleted 10 notes while the first notice is still displaying.  The undo button will only undo the first note.  As well you won't see any additional notices for those other notes you trashed.  This could appear as a bug to the user, and it would better if each action has an associated action.

### How
current:

charliescheer:
### Fix
This PR Fixes #1229 

Currently it is possible to make a series of changes in simplenote iOS that would create more notices, however because you can't queue several of the same notices at once, the subsequent actions receive no notification.  This could be problematic in the case that a user deletes multiple notes really quickly.  In that case the "restore" button on the trashed notice would refer to the first note trashed, not the last note trashed.

In this PR, I have made a change so that if the passthru view that the notice is contained in is interacted with, the notice will be dismissed.  That way you would never be able to create multiple notices at once, as they are dismissed once you interact with the rest of the app in anyway.

### Test
1. From the note list try to delete a note using the contextual actions, a note will appear
2. now delete a another note, you will see the first notice is dismissed before the second one can be trashed.

### DIFF
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -106,17 +106,16 @@ class NoticeController {
 
     // MARK: Dismissing
     //
-    func dismiss(withDuration duration: TimeInterval? = nil) {
+    func dismiss(withDuration duration: TimeInterval, completion: (() -> Void)? = nil) {
+        state = .dismissing
         timer = nil
-        isDismissing = true
 
         noticePresenter.dismissNotification(withDuration: duration) {
+            self.state = .inactive
             self.current = nil
-            self.isDismissing = false
 
-            if !self.notices.isEmpty {
-                self.present(self.notices.removeFirst())
-            }
+            completion?()
+            self.presentNextIfPossible()
         }
     }
 }
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -1,6 +1,13 @@
 import UIKit
 
 class NoticeController {
+
+    private enum State {
+        case inactive
+        case presenting
+        case dismissing
+    }
+
     // MARK: Proerties
     //
     static let shared = NoticeController()
@@ -9,18 +16,13 @@ class NoticeController {
     private var current: Notice?
     private let noticePresenter: NoticePresenter
     private let timerFactory: TimerFactory
+    private var state: State = .inactive
 
     private var timer: Timer? {
         didSet {
             oldValue?.invalidate()
         }
     }
-
-    private var isPresenting: Bool {
-        current != nil
-    }
-
-    private var isDismissing: Bool = false
 
     // MARK: Life Cycle
     //
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -56,11 +56,19 @@ class NoticeController {
         }
     }
 
-    private func dismissNoticeIfNeeded() {
-        if isDismissing {
+    private func dismissNoticeIfNeeded(completion: (() -> Void)?) {
+        switch state {
+        case .inactive:
+            presentNextIfPossible()
+        case .presenting:
+            dismiss(withDuration: UIKitConstants.animationQuickDuration, completion: completion)
+        case .dismissing:
             current = nil
             timer = nil
             noticePresenter.cancel()
+            completion?()
+        }
+    }
             return
         }
 
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -53,8 +53,7 @@ class NoticeController {
     }
 
     private func dismissNoticeIfNeeded() {
-        guard let current = current,
-              current.isDismissible else {
+        guard isPresenting else {
             return
         }
 
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -20,6 +20,8 @@ class NoticeController {
         current != nil
     }
 
+    private var isDismissing: Bool = false
+
     // MARK: Life Cycle
     //
     init(presenter: NoticePresenter = NoticePresenter(), timerFactory: TimerFactory = TimerFactory()) {
@@ -53,13 +55,18 @@ class NoticeController {
     }
 
     private func dismissNoticeIfNeeded() {
-        guard isPresenting else {
+        if isDismissing {
+            current = nil
+            timer = nil
+            noticePresenter.cancel()
             return
         }
 
-        timer?.invalidate()
-        dismiss(withDuration: UIKitConstants.animationQuickDuration)
-
+        if isPresenting && !isDismissing {
+            timer = nil
+            dismiss(withDuration: UIKitConstants.animationQuickDuration)
+            return
+        }
     }
 
     /// Confirms if a notice is already contained in notices queue. Appends to queue if new
@@ -85,8 +92,12 @@ class NoticeController {
     // MARK: Dismissing
     //
     func dismiss(withDuration duration: TimeInterval? = nil) {
+        timer = nil
+        isDismissing = true
+
         noticePresenter.dismissNotification(withDuration: duration) {
             self.current = nil
+            self.isDismissing = false
 
             if !self.notices.isEmpty {
                 self.present(self.notices.removeFirst())
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -29,6 +29,9 @@ class NoticeController {
 
     func setupNoticeController() {
         noticePresenter.startListeningToKeyboardNotifications()
+        noticePresenter.onExternalTap = { [weak self] in
+            self?.dismiss(withDuration: UIKitConstants.animationQuickDuration)
+        }
     }
 
     // MARK: Presenting
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -52,14 +52,21 @@ class NoticeController {
         }
     }
 
+    private func dismissNoticeIfNeeded() {
+        guard let current = current,
+              current.isDismissible else {
+            return
+        }
+
+        timer?.invalidate()
+        dismiss(withDuration: UIKitConstants.animationQuickDuration)
+
+    }
+
     /// Confirms if a notice is already contained in notices queue. Appends to queue if new
     ///
     private func appendToQueueIfNew(_ notice: Notice) {
         if notices.contains(notice) {
-            return
-        }
-
-        if notice == current {
             return
         }
 
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -38,22 +38,20 @@ class NoticeController {
     // MARK: Presenting
     //
     func present(_ notice: Notice) {
-        dismissNoticeIfNeeded()
+        appendToQueueIfNew(notice)
+        dismissNoticeIfNeeded {
+            self.presentNextIfPossible()
+        }
+    }
 
-        if isPresenting {
-            appendToQueueIfNew(notice)
+    /// Confirms if a notice is already contained in notices queue. Appends to queue if new
+    ///
+    private func appendToQueueIfNew(_ notice: Notice) {
+        if notices.contains(notice) {
             return
         }
 
-        current = notice
-        let noticeView = makeNoticeView(from: notice)
-
-        noticePresenter.presentNoticeView(noticeView) { () in
-            let delay = self.current?.action == nil ? Times.shortDelay : Times.longDelay
-            self.timer = self.timerFactory.scheduledTimer(with: delay, completion: {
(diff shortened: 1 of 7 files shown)


charliescheer:
UPDATE: See next comment, I decided to make a change in how this works and this comment is no longer valid (leaving for conversation)

@jleandroperez one thing that would be good to look at in this PR is how the dismiss is being called.  

I had investigated a few different ways that we could use a callback in the NoticeController, but didn't find them very satisfactory.  What I liked about it was that I could leave the dismiss method private, but it required me to capture it when defining the callback. In all cases I could come up with I had to do something like:
`onInteraction(NoticeController.shared)`
or
`noticePresenter.onInteraction = {
  self.dismiss()
}`

I thought of a few other ways that seem a bit better, like setting a passthruView delegate to handle the interaction, or using weak self when setting the callback:
`
noticePresenter.onInteraction = { [weak self] in
  self.dismiss()
`

But decided calling the shared object was cleaner in the end.  But what do you think?

charliescheer:
Update on my previous comment about where the dismiss was being setup.  I thought about it over the weekend and decided that by setting the callback in the noticePresenter was blurring the class responsibilities.  All other tap/press interactions were being handled by the NoticeController, so I felt that the natural place for this to be set/handled was there.
I updated my change to set a completion handler using [weak self] in the notice controller instead.

Also note that while testing, I noticed that if I was quick about it I could still cause actions fast enough to have the notices get stuck.  to fix that I created the ability for the NoticeController to override the default animation fadeout time for notices.  That way if the notice is being dismissed because the user has tapped outside the notice, it fades out quickly enough that it has completed before they can trigger a new one.
To test this, on a physical device, swipe to delete multiple notes in a row as quickly as you can, the notices will dismiss faster than normal, and _should_ be dismissed fully before a new notice can be triggered

Simplenote/NoticePresenter.swift
```diff
@@ -105,13 +108,17 @@ class NoticePresenter {
...
             return
         }
-        UIView.animate(withDuration: UIKitConstants.animationLongDuration) {
-            noticeView.alpha = .zero
+
+        UIView.animate(withDuration: duration ?? UIKitConstants.animationLongDuration,
+                       delay: .zero,
+                       options: [.allowUserInteraction]) {
+            noticeView.alpha = 0
+            containerView.alpha = 0
```
jleandroperez:
```suggestion
            containerView.alpha = UIKitConstants.alpha0_0
```

Simplenote/NoticePresenter.swift
```diff
@@ -105,13 +108,17 @@ class NoticePresenter {
...
               let noticeView = noticeView else {
             return
         }
-        UIView.animate(withDuration: UIKitConstants.animationLongDuration) {
-            noticeView.alpha = .zero
+
+        UIView.animate(withDuration: duration ?? UIKitConstants.animationLongDuration,
+                       delay: .zero,
+                       options: [.allowUserInteraction]) {
+            noticeView.alpha = 0
```
jleandroperez:
```suggestion
            noticeView.alpha = UIKitConstants.alpha0_0
```

Simplenote/NoticePresenter.swift
```diff
@@ -105,13 +108,17 @@ class NoticePresenter {
 
     // MARK: Dismissing Methods
     //
-    func dismissNotification(completion: @escaping () -> Void) {
+    func dismissNotification(withDuration duration: TimeInterval?, completion: @escaping () -> Void) {
```
jleandroperez: I **think**  (note the think, leaving room for error !! 😂) proper Swift would be:

```
    func dismissNotification(duration: TimeInterval?, completion: @escaping () -> Void) {
Write
```

[Ref. here!!](https://swift.org/documentation/api-design-guidelines/) (**"Argument Labels"**)

jleandroperez (approved):
Looking good code wise!! The "Quick Dismissal" behavior feels slightly odd on my end (maybe too fast?).

Other than that :shipit: sir!

SylvesterWilmott:
Thanks @charliescheer !

Conceptually, I think it works great. Now that I see this in action, I have some small changes that I think we should make:

- The whole concept is a good way to solve the issue of queued notices but I found that the experience with the other notices such as around publishing were negatively impacted by this change. I wonder if the trigger for prematurely hiding the notices should be more specific such as when a note list item is swiped. It would be great if we had a method we could call for specific actions that may generate multiple instances of the same notice type and apply it only very specifically to certain triggers to avoid impacting the rest of the notices.
- @jleandroperez  I agree that in this iteration the dismissal feels a bit abrupt but I'm not sure if it will be much of an issue if we try the point above.
- I noticed that the swipe options were odd sizes in this PR, I'm not sure if this was an issue that was introduced here or not since I don't see the same thing in the example gif. See below for illustration.

Screenshot 2021-04-06 at 10 24 34

charliescheer:
@SylvesterWilmott Thanks for taking a look!  

>The whole concept is a good way to solve the issue of queued notices but I found that the experience with the other notices such as around publishing were negatively impacted by this change. I wonder if the trigger for prematurely hiding the notices should be more specific such as when a note list item is swiped. It would be great if we had a method we could call for specific actions that may generate multiple instances of the same notice type and apply it only very specifically to certain triggers to avoid impacting the rest of the notices.

We can do something like this.  A couple of ways we could go about it. 
1. we could make it so that **some** notices are dismissed when interaction with the rest of the app, similar to how this PR is setup.  In this idea the swipe actions (copy link and trash) would be dismissed if the user continues on interacting with the rest of the app, but any notices created in the options view could be left to display the entire animation time (as it is harder from the options view to get to a place to initiate another notice before it dismisses naturally)
2. Alternatively, we could make it so that some notices can trigger a notice of the same type to dismiss quicker letting the new one take it's place.  We could still set this up to only happen so some notices (the swipe menu notices) The process would be, if an early dismissible notice is displaying, and a new notice is created the current one goes away as the new one appears.  

WDYT?

>I noticed that the swipe options were odd sizes in this PR, I'm not sure if this was an issue that was introduced here or not since I don't see the same thing in the example gif. See below for illustration.
Odd.  I checked out those size issues and wasn't able to replicate.  Were you using the sim or a device?

SylvesterWilmott:
>if an early dismissible notice is displaying, and a new notice is created the current one goes away as the new one appears.

I like that! I think that this would produce the best result. Thanks @charliescheer

charliescheer:
>I like that! I think that this would produce the best result. Thanks @charliescheer

Alright!  I have made this update.  While digging through this, what I discovered is that depending on where the notices are triggered makes them easy or note easy to stack with another notice.  Because most of the options View notices dismiss from the options or have multiple steps, those did not feel like good notices to make dismissible, but the notices that are triggered from the note list could easily cause notice stacking.

So what I have done is make the notices that are triggered from the note list dismissible (Swipe action copy link and trash note, and the long press copy note) If a new notice is created while one of these notices is being displayed, then it disappears really quick and a new notice appears right after.

Note that the acceleration of the animation in is in a different PR, so they feel a little slow comparatively, this is to be expected.

@SylvesterWilmott @jleandroperez can you both take another look at  this and let me know what you think?  Feels like a much better interaction to me

SylvesterWilmott:
Thanks @charliescheer, it feels super smooth now!

I think we should apply this behaviour to all of the notices since we won't lose anything by doing so. I think it's a good built in behaviour to have, it makes them bullet-proof for wherever we add them in the future (or if WPiOS want to snatch them eventually).

SylvesterWilmott:
BTW, were you able to investigate about the [swipe options](https://github.com/Automattic/simplenote-ios/pull/1230#issuecomment-813974403) with varying widths. Is it related to this PR or one of the notice PRs?

charliescheer:
I did look into the swipe options sizing and I wasn't able to replicate the issue.  When you saw that what sort of a device were you using?  Simulator?  What version was it?  I will do some more digging.
The changes in this PR shouldn't affect the sizing of the contextual buttons.  But happy to look into it more and see if I can find the problem!!!

charliescheer:
>I think we should apply this behaviour to all of the notices since we won't lose anything by doing so. I think it's a good built in behaviour to have, it makes them bullet-proof for wherever we add them in the future (or if WPiOS want to snatch them eventually).

That is totally do able.  I went ahead and updated this PR to reflect that change, now all notices will be dismissed if a new one appears

Check it out and let me know what you think.  I did test a few scenarios and mostly I like this change, but there are some cases where I think a user might miss out on a notice like: Publishing... -> Copy Link -> Note Published
In that case the copy link notice only appears for a quick second like this:


In this case I find the publishing... and copy link get a little stepped on as the new notices appear.  WDYT?

SylvesterWilmott:
>I did look into the swipe options sizing and I wasn't able to replicate the issue. When you saw that what sort of a device were you using? Simulator? What version was it?

The bug is also present in #1228 but appears as expected in production currently. Tested using the circleci build on iPhone XS max, iOS 14.4.2.

>Check it out and let me know what you think.

I'm finding it hard to test the scenario you pointed out. I think this change has caused the "stuck notice" bug again as notices keep getting stuck after pressing the "copy link" button in the publish success notice.

charliescheer:
I did some testing on this, and I have been able to find some ways to get the notices stuck on screen.  There seem to be two potential issues, one where the timer gets invalidated without dismissing a notice (this one is fixed) the other seems to be if you trigger notices quickly enough it is possible to end up with two going at once, and only one of those will be dismissed later.  
I'm looking at possible ways to fix this.

Simplenote/NoticePresenter.swift
charliescheer: hmmm..... that is a tricky one conventionally.

I chose to put withDuration because of this part of the guideline
>The argument label should normally begin at the preposition, e.g. x.removeBoxes(havingLength: 12).

But I could see that it could also be
`dismissNotificationWith(duration: TimeInterval?, onCompletion:.....)`
based on the exception
>An exception arises when the first two arguments represent parts of a single abstraction.
a.move(toX: b, y: c)
a.fade(fromRed: b, green: c, blue: d)

>In such cases, begin the argument label after the preposition, to keep the abstraction clear.
a.moveTo(x: b, y: c)
a.fadeFrom(red: b, green: c, blue: d)

I will change it to your suggestion.  Will have to think about this one 🤔

charliescheer: Hmmm.... more complex.  I went to make this change and noticed something:
func dismissNotification(**withDuration** duration: TimeInterval, completion: @escaping () -> Void) {

calls 
UIView.animate(**withDuration:** duration, delay: .zero, options: [.allowUserInteraction]) {

Should we follow the guidelines that apple is using in their API?

Simplenote/NoticeController.swift
```diff
@@ -57,11 +51,47 @@ class NoticeController {
...
+        case .inactive:
+            presentNextIfPossible()
+        case .presenting:
+            dismiss(withDuration: UIKitConstants.animationQuickDuration, completion: completion)
+        case .dismissing:
+            current = nil
+            timer = nil
+            noticePresenter.cancel()
+            completion?()
+        }
```
jleandroperez, line 68: Let's chat over Slack regarding the general approach!!  I believe the actual problem could be addressed with really small changes.

Regardless of whatever we do, API(s) should **always** do exactly what their signature imply. In this particular case:

1. Dismiss might `presentNext`
2. (OR) actually dismiss
3. (OR) invoke completion

That means that the Dismiss method is tightly coupled with the `state`, and has unpredictable behavior.
Implementing this in an FSM approach could look like this:

```
private var state: State = .inactive {
   didSet {
      stateWasUpdated(newState: state)
   }
}
```

```
func stateWasUpdated(state: State) {
    switch state {
        case .presenting(let notice: Notice):
            // present the actual notice
        case .dismissing:
           // dismiss
    }
}
```

----

An FSM involves **States** and **Transitions** between them. Along with these lines, you would usually see an API that determines what the next state should be.

jleandroperez (commented):
Sending you few comments Charlie!! let's chat over Slack about this one

mokagio:
I'm running the code freeze for version 4.35. This PR is on that milestone and received an approval, but after that happened new changes went in. @jleandroperez's latest comment is a "Comment" review not an approval and suggests there's more work that could be done.

@charliescheer I'll bump this to 4.36.

charliescheer:
Closing in favor of #1262

PR #1233, opened 2021-04-07
Fixed tag bar getting stuck on swipe to markdown preview

### LINKED ISSUE #1211: TagsEditor Stuck When swiping to markdown preview
### Details
While testing #1200 I've spotted a VoiceOver bug:

1. Open the Editor
2. Enable VoiceOver
3. Push the Markdown Preview via the 3 finger right swipe
4. Go back to the editor

As a result, the Keyboard will be dismissed, but the Tags Editor will remain visible in the middle of the screen.

### Demo
https://user-images.githubusercontent.com/1195260/111329169-b0198b80-864d-11eb-920b-67bbd634c131.mp4

charliescheer:
### Fix
Fixes #1211 
This PR fixes a problem with the tag bar where swiping from the note editor to markdown preview, while the keyboard was visible, would cause the tag bar to get stuck in the middle of the screen. 

The tag list location is animated in and out based on the changes to the keyboard, so with this PR, I have changed when the Note Editor stops listening to keyboard update notifications.  The keyboard was being dismissed after we stopped listening, so the tags bar had no way to know it needed to change locations.

### Test
1. go to a note that has mark down enabled.  
2. tap into the text view to make the keyboard appear
3. swipe to the left to make markdown appear
4. tap back to return to the editor
The tag bar should be at the bottom of the screen not in the middle

### Review
> Only one developer  is required to review these changes, but anyone can perform the review.

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -462,6 +462,7 @@ CGFloat const SPSelectedAreaPadding = 20;
 
 - (UIViewController *)nextViewControllerForInteractivePush
 {
+    [self endEditing];
     SPMarkdownPreviewViewController *previewViewController = [SPMarkdownPreviewViewController new];
     previewViewController.markdownText = [self.noteEditorTextView plainText];
     
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -217,6 +217,7 @@ CGFloat const SPSelectedAreaPadding = 20;
 
 - (void)viewDidDisappear:(BOOL)animated
 {
+    [super viewDidDisappear:animated];
     [self stopListeningToKeyboardNotifications];
 }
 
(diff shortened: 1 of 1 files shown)


Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -211,11 +211,15 @@ - (void)viewWillDisappear:(BOOL)animated
...
     [super viewWillDisappear:animated];
     [self.navigationController setToolbarHidden:YES animated:YES];
-    [self stopListeningToKeyboardNotifications];
 
     [self saveScrollPosition];
 }
 
+- (void)viewDidDisappear:(BOOL)animated
+{
+    [self stopListeningToKeyboardNotifications];
```
jleandroperez, line 221: Nipticky: `[super viewDidDisappear:animated]` on this one

jleandroperez (approved):
Works super! one nipticky comment, feel free to merge afterwards!

Thanks for fixing this sir!!

:shipit:

PR #1235, opened 2021-04-09
Update encrypted credentials for screenshots account

mokagio:
See https://github.com/Automattic/simplenote-android/pull/1354#issuecomment-816383580

### Test
Update the secrets in the repo then run the `SimplenoteScreenshots` tests. They should be able to login successfully.

If the tests fail after logging in because of the PIN lock screen, you may need to run them again. This is a known hiccup that I haven't been able to tackle yet.

### Review

### Release


loremattei (approved):
I tested it following the instruction and it worked as expected. 
:shipit:

PR #1238, opened 2021-04-11
Remove search sort bar bar

### LINKED ISSUE #1209: Remove Sorting Options from Search
Based on low usage, let's remove the sorting options from search.

Sorting options should now only appear in Settings and search results list should inherit the currently selected sort mode from the notes list.

charliescheer:
### Fix
Fixes #1209 

After some testing, it has been determined that the search sort bar isn't being used, so we are removing it.  Search sorting will now use the same settings as the primary sort order, set in the settings view.  

This PR removes the search sort bar and sets up for search settings to be organized by the same settings the general sorting of notes in the notes list

### Test
1. go to the settings view and choose a non date sorting options (alphabetical a-z or z-a) 
2. return to the note list, tap on the search bar, add a search query that will yield a result and tap on the enter button to dismiss the keyboard.  The sort bar will not appear and the results will be sorted by the option set in the settings window
3. Return to settings and choose one of the date sorting options
4. Preform another search and make sure that the sorting has changed to the setting you chose and that the results in the notes are now prepended with the date (modified or created depending on the option selected)

### Review

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -3,25 +3,6 @@ import CoreSpotlight
 import UIKit
 import SimplenoteSearch
 
-
-// MARK: - View Lifecycle
-//
-extension SPNoteListViewController {
-
-    open override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
-        super.viewWillTransition(to: size, with: coordinator)
-
-        coordinator.animate(alongsideTransition: { [weak self] _ in
-            guard let popoverController = self?.popoverController, let sortBar = self?.sortBar else {
-                return
-            }
-
-            popoverController.sourceRect = sortBar.dividerView.frame
-        }, completion: nil)
-    }
-}
-
-
 // MARK: - Components Initialization
 //
 extension SPNoteListViewController {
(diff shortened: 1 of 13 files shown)


Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -636,12 +590,23 @@ private extension SPNoteListViewController {
...
     func prefixText(for note: Note) -> String? {
         guard case .searching = notesListController.state,
-            let date = note.date(for: notesListController.searchSortMode)
+            let date = note.date(for: notesListController.sortMode)
             else {
                 return nil
         }
 
-        return DateFormatter.listDateFormatter.string(from: date)
+        return dateSortPrefix() + DateFormatter.listDateFormatter.string(from: date)
```
jleandroperez: Let's preserve the original behavior, and address the Timestamp changes in a follow up PR.

Ref. here `p2XJRt-2RQ-p2`

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -636,12 +590,23 @@ private extension SPNoteListViewController {
...
+    }
+
+    private func dateSortPrefix() -> String {
+        switch notesListController.sortMode {
+        case .createdNewest, .createdOldest:
+            return "Created: "
+        case .modifiedNewest, .modifiedOldest:
+            return "Modified: "
```
jleandroperez: (Although this code is going away!) Please always remember to NSLocalize all the things 😄

jleandroperez (approved):
Hey there @charliescheer !! Few notes!

1. Let's please drop the `dateSortPrefix` change, since we'll need to add a new `Timestamp` setting. More here: `p2XJRt-2RQ-p2`
2. You can also drop `SortModePickerViewController`

Other than that, looks super!

Feel free to :shipit: once the changes are in place sir!

PR #1239, opened 2021-04-12
Refactor search flow UI tests to use value type for fake data

mokagio:
As I mentioned, I'm keen to keep the discussion going on UI testing improvements and code sharing between our projects.

This was a quick win I had in my list. It replaces individual `static let` values for each note name, content, and tags with a `NoteData` [value type](https://www.objc.io/issues/16-swift/swift-classes-vs-structs/) to make the tests more readable and easier to maintain.
I think this pattern is valuable and adopting it whenever possible will help us in the long run.

### Test

Checkout this branch and run the UI tests.

### Review

### Release

### DIFF
--- a/SimplenoteUITests/SimplenoteUITestsSearch.swift
+++ b/SimplenoteUITests/SimplenoteUITestsSearch.swift
@@ -1,22 +1,43 @@
 import XCTest
 
-let godzillaNoteName = "Godzilla"
-let godzillaInfo = "Godzilla (Japanese: ゴジラ, Hepburn: Gojira, /ɡɒdˈzɪlə/; [ɡoꜜdʑiɾa] (About this soundlisten)) is a fictional monster, or kaiju, originating from a series of Japanese films. The character first appeared in the 1954 film Godzilla and became a worldwide pop culture icon, appearing in various media, including 32 films produced by Toho, four Hollywood films and numerous video games, novels, comic books and television shows. Godzilla has been dubbed the'King of the Monsters', a phrase first used in Godzilla, King of the Monsters! (1956), the Americanized version of the original film."
+// swiftlint:disable line_length
+let godzilla = NoteData(
+    name: "Godzilla",
+    content: "Godzilla (Japanese: ゴジラ, Hepburn: Gojira, /ɡɒdˈzɪlə/; [ɡoꜜdʑiɾa] (About this soundlisten)) is a fictional monster, or kaiju, originating from a series of Japanese films. The character first appeared in the 1954 film Godzilla and became a worldwide pop culture icon, appearing in various media, including 32 films produced by Toho, four Hollywood films and numerous video games, novels, comic books and television shows. Godzilla has been dubbed the'King of the Monsters', a phrase first used in Godzilla, King of the Monsters! (1956), the Americanized version of the original film.",
+    tags: ["sea-monster", "reptile", "prehistoric"]
+)
 
-let kingKongNoteName = "King Kong"
-let kingKongInfo = "King Kong is a film monster, resembling an enormous gorilla, that has appeared in various media since 1933. He has been dubbed The Eighth Wonder of the World, a phrase commonly used within the films. The character first appeared in the novelization of the 1933 film King Kong from RKO Pictures, with the film premiering a little over two months later. The film received universal acclaim upon its initial release and re-releases. A sequel quickly followed that same year with The Son of Kong, featuring Little Kong. In the 1960s, Toho produced King Kong vs. Godzilla (1962), pitting a larger Kong against Toho's own Godzilla, and King Kong Escapes (1967), based on The King Kong Show (1966–1969) from Rankin/Bass Productions. In 1976, Dino De Laurentiis produced a modern remake of the original film directed by John Guillermin."
+let kingGong = NoteData(
+    name: "King Kong",
+    content: "King Kong is a film monster, resembling an enormous gorilla, that has appeared in various media since 1933. He has been dubbed The Eighth Wonder of the World, a phrase commonly used within the films. The character first appeared in the novelization of the 1933 film King Kong from RKO Pictures, with the film premiering a little over two months later. The film received universal acclaim upon its initial release and re-releases. A sequel quickly followed that same year with The Son of Kong, featuring Little Kong. In the 1960s, Toho produced King Kong vs. Godzilla (1962), pitting a larger Kong against Toho's own Godzilla, and King Kong Escapes (1967), based on The King Kong Show (1966–1969) from Rankin/Bass Productions. In 1976, Dino De Laurentiis produced a modern remake of the original film directed by John Guillermin.",
+    tags: ["ape", "prehistoric"]
+)
 
-let mechagodzillaNoteName = "Mechagodzilla"
-let mechagodzillaInfo = "Mechagodzilla (メカゴジラ, Mekagojira) is a fictional mecha character that first appeared in the 1974 film Godzilla vs. Mechagodzilla. In its debut appearance, Mechagodzilla is depicted as an extraterrestrial villain that confronts Godzilla. In subsequent iterations, Mechagodzilla is usually depicted as a man-made weapon designed to defend Japan from Godzilla. In all incarnations, the character is portrayed as a robotic doppelgänger with a vast array of weaponry, and along with King Ghidorah, is commonly considered to be an archenemy of Godzilla."
+let mechagodzilla = NoteData(
+    name: "Mechagodzilla",
+    content: "Mechagodzilla (メカゴジラ, Mekagojira) is a fictional mecha character that first appeared in the 1974 film Godzilla vs. Mechagodzilla. In its debut appearance, Mechagodzilla is depicted as an extraterrestrial villain that confronts Godzilla. In subsequent iterations, Mechagodzilla is usually depicted as a man-made weapon designed to defend Japan from Godzilla. In all incarnations, the character is portrayed as a robotic doppelgänger with a vast array of weaponry, and along with King Ghidorah, is commonly considered to be an archenemy of Godzilla.",
+    tags: ["man-made", "robot"]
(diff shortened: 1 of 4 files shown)


SimplenoteUITests/SimplenoteUITestsSearch.swift
```diff
@@ -1,22 +1,43 @@
...
+    mechagodzilla,
+    diacritic
+]
 
 class SimplenoteUISmokeTestsSearch: XCTestCase {
 
     override class func setUp() {
         app.launch()
-        let _ = attemptLogOut()
+        _ = attemptLogOut()
```
mokagio, line 40: SwiftLint warned me about this

SimplenoteUITests/SimplenoteUITestsSearch.swift
```diff
@@ -1,22 +1,43 @@
...
+    tags: ["language", "diacritic"]
+)
+
+// swiftlint:enable line_length
+let allNotes: [NoteData] = [
+    godzilla,
+    kingGong,
+    mechagodzilla,
+    diacritic
+]
```
mokagio, line 34: Here's one advantage: because we have values distinct values for each note rather than a collection of constants, we can define an array with all the notes. This will simplify the tests that look for all the notes.

SimplenoteUITests/SimplenoteUITestsSearch.swift
```diff
@@ -25,10 +46,8 @@ class SimplenoteUISmokeTestsSearch: XCTestCase {
         Sidebar.open()
         Sidebar.tagsDeleteAll()
         NoteList.openAllNotes()
-        NoteList.createNoteAndLeaveEditor(noteName: godzillaNoteName + "\n\n" + godzillaInfo, tags: ["sea-monster", "reptile", "prehistoric"])
-        NoteList.createNoteAndLeaveEditor(noteName: kingKongNoteName + "\n\n" + kingKongInfo, tags: ["ape", "prehistoric"])
-        NoteList.createNoteAndLeaveEditor(noteName: mechagodzillaNoteName + "\n\n" + mechagodzillaInfo, tags: ["man-made", "robot"])
-        NoteList.createNoteAndLeaveEditor(noteName: diacriticNoteName + "\n\n" + diacriticInfo, tags: ["language", "diacritic"])
+
+        allNotes.forEach { NoteList.createNoteThenLeaveEditor($0) }
```
mokagio, line 50: Double advantage here.

1. Using a custom API that takes `NoteData` as the input, we can DRY the code to create a new note at the test level.
2. Since we have a type for the notes, we can use an array and call `forEach` instead of writing a call of same method for each different note.

SimplenoteUITests/SimplenoteUITestsSearch.swift
```diff
@@ -41,26 +60,26 @@ class SimplenoteUISmokeTestsSearch: XCTestCase {
 
         trackStep()
         NoteList.openAllNotes()
-        NoteListAssert.notesExist(names: [godzillaNoteName, kingKongNoteName, mechagodzillaNoteName, diacriticNoteName])
+        NoteListAssert.notesExist(allNotes)
```
mokagio, line 63: I'd argue that not only this is easier to type and read because there are less characters involved, but it's less cognitive demanding because the fact that we're verifying that all notes exist in the list is explicit here. In the previous version, a reader would have had to deduce from the context and by reading the top of the test file that those four notes are all the notes there are.

SimplenoteUITests/SimplenoteUITestsSearch.swift
```diff
@@ -178,14 +197,18 @@ class SimplenoteUISmokeTestsSearch: XCTestCase {
...
         NoteListAssert.tagsSuggestionsNumber(number: 5)
         NoteListAssert.notesSearchHeaderShown()
 
         trackStep()
         NoteList.searchForText(text: "a")
         NoteListAssert.tagsSearchHeaderShown()
-        NoteListAssert.tagSuggestionsExist(tags: ["tag:ape", "tag:diacritic", "tag:language", "tag:man-made", "tag:sea-monster"])
+        NoteListAssert.tagSuggestionsExist(
+            tags: ["tag:ape", "tag:diacritic", "tag:language", "tag:man-made", "tag:sea-monster"]
+        )
```
mokagio, line 211: This is just me addressing an easy to fix line length warning from SwiftLint. No code change.

SimplenoteUITests/SimplenoteUITestsSearch.swift
```diff
@@ -1,22 +1,43 @@
...
+// swiftlint:disable line_length
+let godzilla = NoteData(
+    name: "Godzilla",
+    content: "Godzilla (Japanese: ゴジラ, Hepburn: Gojira, /ɡɒdˈzɪlə/; [ɡoꜜdʑiɾa] (About this soundlisten)) is a fictional monster, or kaiju, originating from a series of Japanese films. The character first appeared in the 1954 film Godzilla and became a worldwide pop culture icon, appearing in various media, including 32 films produced by Toho, four Hollywood films and numerous video games, novels, comic books and television shows. Godzilla has been dubbed the'King of the Monsters', a phrase first used in Godzilla, King of the Monsters! (1956), the Americanized version of the original film.",
+    tags: ["sea-monster", "reptile", "prehistoric"]
+)
```
mokagio, line 8: I decided to define these values in the test as opposed to make them `static let` properties of `NoteData` to keep them closer to where they're used.

SimplenoteUITests/SimplenoteUITestsSearch.swift
pachlava: >SwiftLint warned me about this

This made me wonder what did I do differently to never see this warning, knowing that you are using the same configuration for targets as I do. 🤔 The only idea I have, is that the difference may come from how I have configured the project, however I'm following the [build instructions](https://github.com/Automattic/simplenote-ios#build-instructions), and also just tried alternative ways of installing / using SwiftLint:

There were a few hours of trial and error, which taught me a bit about `Run Script` phase in Xcode target, and about `swiftlint.sh` we use, but I still did not find the answer. Given that I have SwiftLint installed (and which I also tried installing via Homebew and Cocoapods), and used a 5-line script snippet from [here](https://github.com/realm/SwiftLint#xcode), to see the warning for all files, not only the added/modified, I still could not see warnings for `let _ = ...` pattern. I'm out of ideas for now...

SimplenoteUITests/SimplenoteUITestsSearch.swift
pachlava: @mokagio this is precious, thank you, and thank you for the [value type](https://www.objc.io/issues/16-swift/swift-classes-vs-structs/) read!

SimplenoteUITests/SimplenoteUITestsSearch.swift
pachlava: This is very true, thank you 🙇

SimplenoteUITests/SimplenoteUITestsSearch.swift
pachlava: Again, I wonder why I did not see this, presuming that I should have the same `.swiftlint.yml` as you have, and it works (I just tried disabling the `comma` rule and this was reflected in warning generation).

SimplenoteUITests/SimplenoteUITestsSearch.swift
pachlava: Fully agree. I initially thought of having these defined in a separate file, but it looks more natural to have them in the only file where they are used.

pachlava (approved):
@mokagio thanks _a lot_ for this improvement!

Besides that fact that it makes the code prettier, being a 'reviewer' makes me learn 🙇 

> As I mentioned, I'm keen to keep the discussion going on UI testing improvements and code sharing between our projects.

I'm both hands for this and would like to take part in this as soon as I'm done with what is outlined in my project, since it reaches its ETA. Your proposal for scaling the UI tests looks like a topic for the next project.

:shipit: !

pachlava:
Merging this now, in order to have the changes be automatically present in the new branches I'll be creating soon 🙇‍♂️

pachlava:
Forgot to mention initially: all tests passed twice with no fails on iPhone 12 👍

mokagio:
Regarding the SwiftLint issues... 🤔 🤔 🤔 That's bizarre.

Here's what I see on my Xcode:

image

I got to that point by adding an empty line (line 3) and running a build for tests with <kbd>Shift Cmd U</kbd>.

My version of SwiftLint is 0.43.1, installed via Homebrew.

pachlava:
Oh, your screenshot has helped! I suspect that we are using different `.swiftlint.yml` somehow. 

`vertical_whitesplace`  and `redundant_discardable_let` rules are not present in the [.swiftlint.yml](https://github.com/Automattic/simplenote-ios/blob/develop/.swiftlint.yml) used in the repo.

Once I removed `only_rules` from there, I started seeing this too.

mokagio:
🤔 there's something weird going on with SwiftLint I'd say.

The script that runs it as a build phase use a vendored version, which is supposed to be `0.41.0`. The vendored version on my install is `0.27.0`, I guess because I never updated it, but the script doesn't tell me about it 🤔 

> Once I removed `only_rules` from there, I started seeing this too.

That may make sense if your local version is lower than `0.41.0`, because that's the SwiftLint version in which the introduced that syntaxt.

What's the output of `./vendor/swiftlint/bin/swiftlint version` on your end?

Still, I don't understand why it works for me 🤔

pachlava:
> What's the output of ./vendor/swiftlint/bin/swiftlint version on your end?

It's `0.43.1`

mokagio:
@pachlava 🤔 what's going on here, then... I'm not sure.

Look, let's leave it for the moment and lock some dedicated time to revisit the SwiftLint setup if this arises again.

pachlava:
Agree 👍

PR #1240, opened 2021-04-12
Automated first test from `Settings`

pachlava:
Automates the first test from `Settings`: 
`Changing "Condensed Note List" mode immediately updates and reflects in note list`

### Changes:
1. New `Settings.swift` where I tried using enumerations for element locators (as seen [here](https://bitbar.com/blog/best-practices-for-organizing-locators-for-xcuitest/))
2. New `SimplenoteUISmokeTestsSettings.swift` with the test itself
3. `NoteList.swift` contains several new functions

I'll attempt to give more details by commenting on specific parts of the code below.

### Test
The way the test checks the condensed mode is: checking note cell height and presence of the note content preview in note list (should be absent under condensed mode).

1. Running the new test, along with previously existing ones.

### Review

### Release

### DIFF
--- a/SimplenoteUITests/NoteData.swift
+++ b/SimplenoteUITests/NoteData.swift
@@ -2,7 +2,7 @@
 struct NoteData {
     let name: String
     let content: String
-    let tags: [String]
+    var tags: [String] = []
 
     var formattedForAutomatedInput: String { "\(name)\n\n\(content)" }
 }
--- a/SimplenoteUITests/NoteList.swift
+++ b/SimplenoteUITests/NoteList.swift
@@ -13,6 +13,12 @@ class NoteList {
         return navBar.identifier
     }
 
+    class func isAllNotesListOpen() -> Bool {
+        let navBar = app.navigationBars[UID.NavBar.allNotes]
+        guard navBar.exists else { return false}
+        return navBar.frame.minX == 0.0
+    }
+
     class func isNoteListOpen(forTag tag: String) -> Bool {
         return app.navigationBars[tag].exists
     }
@@ -30,6 +36,7 @@ class NoteList {
     }
 
     class func openAllNotes() {
+        guard !NoteList.isAllNotesListOpen() else { return }
         print(">>> Opening \"All Notes\"")
         Sidebar.open()
         app.tables.staticTexts[UID.Button.allNotes].tap()
--- a/SimplenoteUITests/NoteList.swift
+++ b/SimplenoteUITests/NoteList.swift
@@ -202,7 +202,7 @@ class NoteListAssert {
         XCTAssertEqual(actualNotesNumber, expectedNotesNumber, numberOfNotesInAllNotesNotExpected)
     }
 
-    class func noteHeight(_ note: NoteData, _ noteHeight: CGFloat) {
+    class func note(_ note: NoteData, hasHeight height: CGFloat) {
         print(">>> Asserting that note height is \(noteHeight)")
         XCTAssertEqual(NoteList.getNoteCellHeight(note.name), noteHeight)
     }
--- a/SimplenoteUITests/NoteList.swift
+++ b/SimplenoteUITests/NoteList.swift
@@ -13,16 +13,23 @@ class NoteList {
         return navBar.identifier
     }
 
-    class func isAllNotesListOpen() -> Bool {
-        return app.navigationBars[UID.NavBar.allNotes].exists
-    }
-
     class func isNoteListOpen(forTag tag: String) -> Bool {
         return app.navigationBars[tag].exists
     }
 
+    class func getNoteCell(_ noteName: String) -> XCUIElement {
+        return Table.getCell(label: noteName)
+    }
+
+    class func isNotePresent(_ noteName: String) -> Bool {
+        return NoteList.getNoteCell(noteName).exists
+    }
+
+    class func getNoteCellHeight(_ noteName: String) -> CGFloat {
+        return NoteList.getNoteCell(noteName).frame.height
+    }
+
     class func openAllNotes() {
-        guard !NoteList.isAllNotesListOpen() else { return }
         print(">>> Opening \"All Notes\"")
         Sidebar.open()
         app.tables.staticTexts[UID.Button.allNotes].tap()
@@ -188,6 +195,11 @@ class NoteListAssert {
         XCTAssertEqual(actualNotesNumber, expectedNotesNumber, numberOfNotesInAllNotesNotExpected)
     }
 
+    class func noteHeight(_ note: NoteData, _ noteHeight: CGFloat) {
+        print(">>> Asserting that note height is \(noteHeight)")
+        XCTAssertEqual(NoteList.getNoteCellHeight(note.name), noteHeight)
+    }
+
     class func tagsSuggestionsNumber(number: Int) {
         let actualNumber = NoteList.getTagsSuggestionsNumber()
         XCTAssertEqual(actualNumber, number, numberOfTagsSuggestionsNotExpected)
@@ -212,14 +224,27 @@ class NoteListAssert {
         NoteListAssert.noteListShown(forSelection: UID.NavBar.trash)
     }
 
+    class func noteContentIsShown(_ note: NoteData) {
+        noteContentIsShownInSearch(noteName: note.name, expectedContent: note.content)
+    }
+
     class func noteContentIsShownInSearch(noteName: String, expectedContent: String) {
-        print(">>> Asserting that note '\(noteName)' shows the following content:")
-        print(">>>> " + expectedContent)
+        print(">>> Asserting that note '\(noteName)' has the following content:")
+        print(">>>> \"\(expectedContent)\"")
+
+        guard NoteList.isNotePresent(noteName) else {
+            XCTFail(">>>> Note not found")
+            return
+        }
 
         if let noteContent = Table.getContentOfCell(noteName: noteName) {
-            XCTAssertTrue(noteContent.contains(expectedContent), "Content NOT found")
+            XCTAssertTrue(noteContent.contains(expectedContent), "Content is different.")
+        } else if expectedContent.isEmpty {
+            /// If note content is nil, but we assert for empty content, we should not fail
+            XCTAssert(true, "Content is not empty.")
         } else {
-            XCTFail("Could not find note")
+            /// Otherwise, we should fail
+            XCTFail("Note has no content. Only title.")
         }
     }
 
--- a/SimplenoteUITests/NoteList.swift
+++ b/SimplenoteUITests/NoteList.swift
@@ -231,7 +231,7 @@ class NoteListAssert {
         NoteListAssert.noteListShown(forSelection: UID.NavBar.trash)
     }
 
-    class func noteContentIsShown(_ note: NoteData) {
+    class func contentIsShown(for note: NoteData) {
         noteContentIsShownInSearch(noteName: note.name, expectedContent: note.content)
     }
 
--- a/SimplenoteUITests/NoteList.swift
+++ b/SimplenoteUITests/NoteList.swift
@@ -203,8 +203,8 @@ class NoteListAssert {
     }
 
     class func note(_ note: NoteData, hasHeight height: CGFloat) {
-        print(">>> Asserting that note height is \(noteHeight)")
-        XCTAssertEqual(NoteList.getNoteCellHeight(note.name), noteHeight)
+        print(">>> Asserting that note height is \(height)")
+        XCTAssertEqual(NoteList.getNoteCellHeight(note.name), height)
     }
 
     class func tagsSuggestionsNumber(number: Int) {
--- a/SimplenoteUITests/NoteList.swift
+++ b/SimplenoteUITests/NoteList.swift
@@ -240,8 +240,7 @@ class NoteListAssert {
         print(">>>> \"\(expectedContent)\"")
 
         guard NoteList.isNotePresent(noteName) else {
-            XCTFail(">>>> Note not found")
-            return
+            return XCTFail(">>>> Note not found")
         }
 
         if let noteContent = Table.getContentOfCell(noteName: noteName) {
--- a/SimplenoteUITests/Settings.swift
+++ b/SimplenoteUITests/Settings.swift
@@ -1,0 +1,47 @@
+import XCTest
+
+let screenName = "Settings"
+
+enum SettingsScreen: String {
+
+    case doneButton = "Done"
+    case condensedModeSwitch = "Condensed Note List"
+
+    var element: XCUIElement {
+        switch self {
+        case .doneButton:
+            return app.navigationBars[screenName].buttons[self.rawValue]
+        case .condensedModeSwitch:
+            return app.tables.cells.containing(.staticText, identifier: self.rawValue).firstMatch
+        }
+    }
+}
+
+class Settings {
+    class func condensedModeEnable() {
+        let condensedModeSwitch = SettingsScreen.condensedModeSwitch.element
+        let switchValue = condensedModeSwitch.value as? String
+
+        if switchValue == "0" {
+            condensedModeSwitch.tap()
+        }
+    }
+
+    class func condensedModeDisable() {
+        let condensedModeSwitch = SettingsScreen.condensedModeSwitch.element
+        let switchValue = condensedModeSwitch.value as? String
+
+        if switchValue == "1" {
+            condensedModeSwitch.tap()
+        }
+    }
+
+    class func close() {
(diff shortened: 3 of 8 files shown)


SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
```diff
@@ -0,0 +1,56 @@
...
+        Settings.close()
+    }
+
+    func testUsingCondensedNoteList() throws {
+        trackTest()
+        let note = NoteData(
+            name: "Condensed Mode Test",
+            content: "Condensed Mode Content",
+            tags: []
+        )
```
pachlava: This makes the tests much tidier 🙇

SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
```diff
@@ -0,0 +1,56 @@
...
+        NoteList.openAllNotes()
+        NoteList.createNoteThenLeaveEditor(note)
+        NoteListAssert.noteContentIsShown(note)
+
+        trackStep()
+        Settings.open()
+        Settings.condensedModeEnable()
+        Settings.close()
+        NoteList.openAllNotes()
+        NoteListAssert.noteContentIsShownInSearch(noteName: note.name, expectedContent: "")
```
pachlava, line 45: This method name is misleading, because I initially created it to assert notes in search mode, but it appeared applicable for non-search mode too. I'm keeping it old in this PR to keep the noise down.

pachlava, line 45: This method name is misleading, because I initially created it to assert notes in search mode, but it appeared applicable for non-search mode too. I'm keeping it old in this PR to keep the noise down.

pachlava, line 46: The `noteContentIsShownInSearch` name is misleading, because I initially created it to assert notes in search mode, but it appeared applicable for non-search mode too. I'm keeping it old in this PR to have the noise down.

SimplenoteUITests/NoteList.swift
```diff
@@ -212,14 +224,27 @@ class NoteListAssert {
...
         if let noteContent = Table.getContentOfCell(noteName: noteName) {
-            XCTAssertTrue(noteContent.contains(expectedContent), "Content NOT found")
+            XCTAssertTrue(noteContent.contains(expectedContent), "Content is different.")
+        } else if expectedContent.isEmpty {
+            /// If note content is nil, but we assert for empty content, we should not fail
+            XCTAssert(true, "Content is not empty.")
         } else {
-            XCTFail("Could not find note")
+            /// Otherwise, we should fail
+            XCTFail("Note has no content. Only title.")
```
pachlava: Firstly check if the note exists, then assert the content. Should not fail in case when the note content preview is `nil` while we expect no content.

SimplenoteUITests/Settings.swift
```diff
@@ -0,0 +1,48 @@
...
+
+    case doneButton = "Done"
+    case condensedModeSwitch = "Condensed Note List"
+
+    var element: XCUIElement {
+        switch self {
+        case .doneButton:
+            return app.navigationBars[screenName].buttons[self.rawValue]
+        case .condensedModeSwitch:
+            return app.tables.cells.containing(.staticText, identifier: self.rawValue).firstMatch
```
pachlava: This is to return the parent of the `Condensed Note List` static text. The parent cell contains a useful 0 or 1 value, while the static text has no properties allowing to tell if the switch is on or off.

SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
```diff
@@ -0,0 +1,56 @@
...
+        NoteList.trashAllNotes()
+        Trash.empty()
+        NoteList.openAllNotes()
+    }
+
+    override class func tearDown() {
+        Settings.open()
+        Settings.condensedModeDisable()
+        Settings.close()
+    }
```
pachlava, line 25: Disabling the condensed mode at the end of the suite, because tests in other areas (like Search) will fail under condensed mode.

SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
```diff
@@ -0,0 +1,56 @@
...
+        NoteListAssert.noteContentIsShownInSearch(noteName: note.name, expectedContent: "")
+        NoteListAssert.noteHeight(note, 44.0)
+
+        trackStep()
+        Settings.open()
+        Settings.condensedModeDisable()
+        Settings.close()
+        NoteList.openAllNotes()
+        NoteListAssert.noteContentIsShown(note)
+        NoteListAssert.noteHeight(note, 81.0)
```
pachlava: For now using the exact cell height as one of the ways to assert the condensed mode. If it proves to be flaky, it looks like we can add an accuracy parameter.

SimplenoteUITests/NoteList.swift
```diff
@@ -14,13 +14,27 @@ class NoteList {
...
+    }
+
+    class func isNotePresent(_ noteName: String) -> Bool {
+        return NoteList.getNoteCell(noteName).exists
+    }
+
+    class func getNoteCellHeight(_ noteName: String) -> CGFloat {
+        return NoteList.getNoteCell(noteName).frame.height
+    }
+
```
pachlava, line 37: These three methods still use `noteName` String param but not `NoteData` type because one of the methods is used in older methods, that accepts the String param too.

I'll change that with the next refactor commit, otherwise there will be too much noise.

mokagio: > I'll change that with the next refactor commit, otherwise there will be too much noise.

Excellent idea 👍

SimplenoteUITests/NoteList.swift
```diff
@@ -188,6 +202,11 @@ class NoteListAssert {
         XCTAssertEqual(actualNotesNumber, expectedNotesNumber, numberOfNotesInAllNotesNotExpected)
     }
 
+    class func noteHeight(_ note: NoteData, _ noteHeight: CGFloat) {
```
mokagio: Nitpick / Readability idea.

```suggestion
    class func note(_ note: NoteData, hasHeight height: CGFloat) {
```

This would be used as `NoteListAssert.note(godzilla, hasHeight: 123.0)`, which I think reads nicely. But... it may just be my personal preference, so no big deal. Just tossing it there to play around with the language.

SimplenoteUITests/NoteList.swift
```diff
@@ -212,14 +231,27 @@ class NoteListAssert {
         NoteListAssert.noteListShown(forSelection: UID.NavBar.trash)
     }
 
+    class func noteContentIsShown(_ note: NoteData) {
```
mokagio: Nitpick. Again, thinking about readability, I wonder if this works best

```suggestion
    class func contentIsShown(for note: NoteData) {
```

Which would be used like `NoteListAssert.contentIsShown(for: godzilla)`.

"note list assert content is shown for godzilla" may read a bit better than "note list assert not content is shown godzilla".

When I say "I wonder if..." I really am wondering. The answer can be no, I'm fine with that.

SimplenoteUITests/NoteList.swift
```diff
@@ -212,14 +231,27 @@ class NoteListAssert {
...
+
     class func noteContentIsShownInSearch(noteName: String, expectedContent: String) {
-        print(">>> Asserting that note '\(noteName)' shows the following content:")
-        print(">>>> " + expectedContent)
+        print(">>> Asserting that note '\(noteName)' has the following content:")
+        print(">>>> \"\(expectedContent)\"")
+
+        guard NoteList.isNotePresent(noteName) else {
+            XCTFail(">>>> Note not found")
+            return
```
mokagio: I nice trick I like is to inline this

```suggestion
            return XCTFail(">>>> Note not found")
```

It works because the function returns `Void` and that exactly the return type of calling `XCTFail`.

SimplenoteUITests/Settings.swift
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    static func condensedModeDisable() {
+        let condensedModeSwitch = SettingsScreen.condensedModeSwitch.element
+        let switchValue = condensedModeSwitch.value as? String
+
+        if switchValue == "1" {
+            condensedModeSwitch.tap()
+        }
+    }
```
mokagio: Nitpick. You could DRY this code with something like

```swift
static func condensedModeDisable() {
  switchCondensedModeIfMatching(value: "1")
}

static private func switchCondensedModeIfMatching(value: String) {
    let condensedModeSwitch = SettingsScreen.condensedModeSwitch.element
    let switchValue = condensedModeSwitch.value as? String

    guard switchValue == value else { return }

    condensedModeSwitch.tap()
}
```

In the context of a screen object like this, having as much code that introspects the UI DRY is valuable because when know UI changes often and centralization lowers the cost of updating our tests.

SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
```diff
@@ -0,0 +1,56 @@
+import XCTest
+
+class SimplenoteUISmokeTestsSettings: XCTestCase {
+
+    override class func setUp() {
+        app.launch()
+        getToAllNotes()
+        _ = attemptLogOut()
+        EmailLogin.open()
+        EmailLogin.logIn(email: testDataExistingEmail, password: testDataExistingPassword)
```
mokagio, line 10: Side note. We could use the same value type approach from #1239 for the credentials.

```swift
struct TestUser { // or TestUserData, or UserData, or Credentials, or...
  let email: String
  let password: String
}
```

SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
mokagio: Glad to see it's useful already 😄 

We could make it even tidier by using a default value for the tags.

```swift
struct NoteData {

   let ...

   init(name: String, content: String, tags: [String] = []) {
      ...
```

This way, we'd only have to add tags when they matter and this code would become
```suggestion
        let note = NoteData(
            name: "Condensed Mode Test",
            content: "Condensed Mode Content"
        )
```

SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
mokagio: I like the accuracy parameter idea. Let's hope we don't need to use it though 😄 .

One suggestion I have is to [_Replace Magic Number_](https://refactoring.com/catalog/replaceMagicLiteral.html). The `44.0` and `81.0` are clearly expected notes height, but where do they come from? It would be good to extract them in constants to make that clearer. 

```swift
let defaultIOSCellHeight = 44.0
let customExpandedModeCellHeight = 81.0
```

I'm guessing those5 are the meaning of those values, but I'm not sure.

SimplenoteUITests/Settings.swift
```diff
@@ -0,0 +1,48 @@
...
+
+    var element: XCUIElement {
+        switch self {
+        case .doneButton:
+            return app.navigationBars[screenName].buttons[self.rawValue]
+        case .condensedModeSwitch:
+            return app.tables.cells.containing(.staticText, identifier: self.rawValue).firstMatch
+        }
+    }
+}
```
mokagio: Nitpick. When looking at the types names only, it might be hard to guess the difference between `SettingScreen` and `Settings`.

I think this is a good use case for [nested types](https://docs.swift.org/swift-book/LanguageGuide/NestedTypes.html).

```swift
class Settings {

    enum Elements {

        case doneButton
        case condensedModeSwitch
        
        ...
    }

    static func condensedModeEnable() {
        let condensedModeSwitch = Elements.condensedModeSwitch.element
        ...
```

I'm not in love with the `Elements.NAME.element` syntax, but I can't come up with a better name 😳 .

mokagio (approved):
Looking good 👍 All tests passed.

SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
pachlava: The funny part is, I initially had them as constant variables `cellHeightUsual` and `cellHeightCondensed`, but then removed them, thinking that you will propose to remove them, since they are used only once 😆 ... I'll revert to using them 🙂

SimplenoteUITests/NoteList.swift
pachlava: Oh nice! Thank you!

SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
pachlava: Thank you! I just learned that I could also define the default value for tag using by the variable instead of constant, and providing the default value:

`var tags: [String] = []`

In that case it's not needed to initialize all constants, as with `init` approach. Are there any trade-offs I did not consider here (for this case)?

SimplenoteUITests/NoteList.swift
pachlava: I'm for it, because I wanted to build other methods using `has...` (hasPin, hasContent...) Thank you!

SimplenoteUITests/NoteList.swift
pachlava: Applied your change. As you said, this is a matter of preference, so I'd like to try how it feels

SimplenoteUITests/Settings.swift
pachlava: Oh I feel bad for not doing this 🤦. Finding more opportunities for code reuse was my favourite, whenever I was reviewing the automated tests before at my previous place. And I fell into this trap myself. Thank you!

SimplenoteUITests/SimplenoteUISmokeTestsSettings.swift
pachlava: Yes, this was my observation too, yesterday. Should be a part of refactor PR. Thank you!

SimplenoteUITests/Settings.swift
pachlava: I've moved the code to another more generic `toggleSwitchIfNeeded(_ switchElement: XCUIElement, _ value: String)` which I placed in `Generic.swift` for now, because it'll be useful for the other switched too. So now it looks like this:

```
    static func condensedModeEnable() {
        switchCondensedModeIfNeeded(value: "1")
    }

    static func condensedModeDisable() {
        switchCondensedModeIfNeeded(value: "0")
    }

    static func switchCondensedModeIfNeeded(value: String) {
        toggleSwitchIfNeeded(SettingsScreen.condensedModeSwitch.element, value)
    }

...

func toggleSwitchIfNeeded(_ switchElement: XCUIElement, _ value: String) {
    let switchValue = switchElement.value as? String

    guard switchValue != value else { return }

    switchElement.tap()
}
```

SimplenoteUITests/Settings.swift
pachlava: This looks better, giving it a try 🙏

pachlava:
@mokagio thank you for all your nitpicks and proposals! They are very valuable 🔥

SimplenoteUITests/NoteData.swift
```diff
@@ -2,7 +2,7 @@
 struct NoteData {
     let name: String
     let content: String
-    let tags: [String]
+    var tags: [String] = []
```
mokagio, line 5: That's a nice way to have the compiler generate the `init` for us 🙌 

One suggestion: make it `private` so the instance remains immutable for consumers, that is, they cannot change the value of `tags` after creating it.

```swift
private var tags: [String] = []
```

SimplenoteUITests/Settings.swift
mokagio: Nice!

SimplenoteUITests/NoteData.swift
pachlava: >instance remains immutable for consumers

I did some playing around with this, and it looks like making `tags` variable `private` makes it impossible to create an instance in the first place, which _looks_ logical to me:

`'NoteData' initializer is inaccessible due to 'private' protection level`

So, it seems, that in order to achieve the instance immutability after creation, we really need to have all properties defined as constants, and have an `init()` with default value(s). (this is the approach you suggested intially)

mokagio: My bad 😞 . The right code was `private(set) var`... See #1247.

pachlava: No worries! :-) This made me learn/debug more, compared to the case if you have written `private(set)...` initially, so it's good

PR #1241, opened 2021-04-13
Feature: New Note Bar

### LINKED ISSUE #1044: Editor: New Note Bar
### Details:
- [ ] Implement the New Note Bar: Check with Sly how the **Search Results Bar**  should interact with this new UI!
- [ ] Drop the legacy New Note navigationBar Button

cc @eshurakov @SylvesterWilmott

charliescheer:
### Fix
Fixes #1044 

This PR serves a dual purpose of Adding the new note bar feature and removing the current add new note button from the top navigation bar.

This new note bar will appear on both the note list and the note editor, making it easy for a user to create a new note where ever they are in Simplenote iOS.  This is added using the navigationController bottom toolbar, same as the search navigation bar, so there is some logic added to switch between the two in the note editor.  This will also share ui with the multiple edit delete mode (the next step in this process) but that is not covered in this PR

Note list:

Note Editor:

### Test
1. Open up the note list and see the new note bar.  Tap on the button and a new note will be created and the editor will load
2. Exit that new note, select one of your notes to see the new note bar there.  Again test tapping on the button to see a new note be created
3. Return to the note list and do a search.  Select a note from your search, you will see the note editor load with the search results navigation.  Tap done to see the toolbar transition to the new note bar
4. Make sure that the new note button no longer appears in the top toolbar in the editor. Check that when the keyboard appears the dismiss keyboard button still appears
5. Go to the trash, check to see that the delete button still appears.  Return to the note list (not trash mode) to see there is no button in the top right.
6. test dark mode to make sure colors/images are correct

### Review

### DIFF
--- a/Simplenote/Classes/UIViewController+Simplenote.swift
+++ b/Simplenote/Classes/UIViewController+Simplenote.swift
@@ -91,7 +91,7 @@ extension UIViewController {
     /// Sets up a new note bar with given selector
     ///
     @objc
-    func addNewNoteBar(with selector: Selector) {
+    func configureNewNoteBar(with selector: Selector) {
         let newNoteImage = UIImage(named: UIImageName.newNote.lightAssetFilename)
         let newNoteButton = UIBarButtonItem.init(image: newNoteImage,
                                             style: .plain,
--- a/Simplenote/Classes/UIViewController+Simplenote.swift
+++ b/Simplenote/Classes/UIViewController+Simplenote.swift
@@ -87,4 +87,24 @@ extension UIViewController {
             taskBlock()
         }
     }
+
+    /// Sets up a new note bar with given selector
+    ///
+    @objc
+    func addNewNoteBar(with selector: Selector) {
+        let newNoteImage = UIImage(named: UIImageName.newNote.lightAssetFilename)
+        let newNoteButton = UIBarButtonItem.init(image: newNoteImage,
+                                            style: .plain,
+                                            target: self,
+                                            action: selector)
+        newNoteButton.accessibilityLabel = Strings.newNoteLabel
+
+        let flexibleSpace = UIBarButtonItem.init(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
+
+        setToolbarItems([flexibleSpace, newNoteButton], animated: true)
+    }
+}
+
+private struct Strings {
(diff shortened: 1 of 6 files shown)


SylvesterWilmott:
Looking good!

A couple of points:
- On iPad, the new note button should always appear in the top right side of the navigation bar. This is less important now but will become important in the future when we implement the split view. This is because for users who use the on-screen keyboard, the bottom bar would almost always be covered. For reference: zpl://screen?pid=5db058eaf5c10222cfd424b6&sid=5fca5da3acf2934fa94c02a6
- The bottom bar should not be visible in the search results. This is because creating a new note is currently not possible from the search results and in the point above, won't be possible to do on iPad.

Simplenote/Classes/UIViewController+Simplenote.swift
```diff
@@ -87,4 +87,24 @@ extension UIViewController {
             taskBlock()
         }
     }
+
+    /// Sets up a new note bar with given selector
+    ///
+    @objc
+    func configureNewNoteBar(with selector: Selector) {
+        let newNoteImage = UIImage(named: UIImageName.newNote.lightAssetFilename)
+        let newNoteButton = UIBarButtonItem.init(image: newNoteImage,
```
jleandroperez, line 96: In Swift you never really need to invoke `.init`

Simplenote/Classes/UIViewController+Simplenote.swift
```diff
@@ -87,4 +87,24 @@ extension UIViewController {
             taskBlock()
         }
     }
+
+    /// Sets up a new note bar with given selector
+    ///
+    @objc
+    func configureNewNoteBar(with selector: Selector) {
```
jleandroperez, line 94: This API is not generic enough to grant a spot in a `UIViewController` extension. There are aroooound ~18 VCs and it's used in only two spots.

How about if this was moved over as a `UIBarButtonItem` extension instead?

jleandroperez (commented):
Hey there @charliescheer !!

Sending you few notes (code wise). Found two bugs:

### iOS 12
1. Launch an iOS 12 simulator
2. Switch themes

As a result, the new Toolbar won't get adjusted



### Swipe Back
1. Launch the app
2. Press over any Note to access the editor
3. Swipe back

As a result, the toolbar will end up misplaced (reproduced in iOS 14 / device).

PR #1242, opened 2021-04-13
UI tests helper framework proof of concept

mokagio:
This PR is a first step towards the idea of having a layer of UI generic tests helper methods that we can 1) share across our projects and 2) make available to the wider community (with the caveat that general purpose, catch-all frameworks often end up being just more bloat; still, even just having all the sources in a single place would be beneficial as a reference from where to pick and choose).

In this proof of concept I created a new framework, `UITestHelpers`,  (keen on feedback on the name) in the Simplenote project, and extracted two helpers into it, one for each UI tests target.

I originally planned to make this a Swift package, but I couldn't get past a certain compilation error so I gave up, for the moment.

Another option to get started could have been to extract a page object into a shared framework. I didn't go down that route because I wanted something achievable in a few hours. Before being able to extract a page object, I'd like to define what's the pattern to follow for those, since we're doing things differently between apps.

I decided not to add a test target for the moment. When we'll extract this into a dedicated project, we shall add at least a dummy UI app to test it.

For now, since it will only contains code consumed by the test targets in this project, we can skip the tests.

### Test

Checkout this branch, then run the `SimplenoteUITests` and `SimplenoteScreenshots` target. They should both build and pass on the iPhone 12 Simulator.

_Notice I used the `*UI-tests*` branch name pattern to have CI build the main UI tests target and run the login smoke test scenario_ 👍 

### Review

### Release

### DIFF
--- a/SimplenoteScreenshots/SimplenoteScreenshots.swift
+++ b/SimplenoteScreenshots/SimplenoteScreenshots.swift
@@ -1,3 +1,4 @@
+import UITestHelpers
 import XCTest
 
 class SimplenoteScreenshots: XCTestCase {
@@ -300,20 +301,6 @@ extension XCUIElement {
 
         if let string = previousPasteboardContents {
             UIPasteboard.general.string = string
-        }
-    }
-}
-
-extension XCUIApplication {
-
-    func isDeviceIPhone8Plus(_ device: XCUIDevice = .shared) -> Bool {
-        let iPhone8PlusScreenHeight = CGFloat(736)
-
-        let frame = windows.element(boundBy: 0).frame
-
-        switch device.orientation {
-        case .landscapeLeft, .landscapeRight: return frame.width == iPhone8PlusScreenHeight
-        case _: return frame.height == iPhone8PlusScreenHeight
         }
     }
 }
--- a/SimplenoteUITests/Preview.swift
+++ b/SimplenoteUITests/Preview.swift
@@ -1,3 +1,4 @@
+import UITestHelpers
 import XCTest
 
 class Preview {
--- a/SimplenoteUITests/Preview.swift
+++ b/SimplenoteUITests/Preview.swift
@@ -3,6 +3,7 @@ import XCTest
 class Preview {
 
     class func getText() -> String {
+        // swiftlint:disable:next force_cast
         return app.webViews.descendants(matching: .staticText).element.value as! String
     }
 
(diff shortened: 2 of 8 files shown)


SimplenoteScreenshots/SimplenoteScreenshots.swift
```diff
@@ -303,17 +304,3 @@ extension XCUIElement {
...
-        let iPhone8PlusScreenHeight = CGFloat(736)
-
-        let frame = windows.element(boundBy: 0).frame
-
-        switch device.orientation {
-        case .landscapeLeft, .landscapeRight: return frame.width == iPhone8PlusScreenHeight
-        case _: return frame.height == iPhone8PlusScreenHeight
-        }
-    }
-}
```
mokagio, line 319: This logic is now part of the framework. The only changed needed in this file was to `import` it.

SimplenoteUITests/Preview.swift
```diff
@@ -1,8 +1,10 @@
+import UITestHelpers
```
mokagio, line 1: This `import` is to use the functionality enabled by the `XCUIElementQuery` extension that's been moved into the framework.

UITestHelpers/Info.plist
```diff
@@ -0,0 +1,22 @@
...
+	<key>CFBundleName</key>
+	<string>$(PRODUCT_NAME)</string>
+	<key>CFBundlePackageType</key>
+	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleVersion</key>
+	<string>$(CURRENT_PROJECT_VERSION)</string>
+</dict>
+</plist>
```
mokagio, line 22: As I mentioned, I tried to make this a Swift package first which would have avoided some of this cruft, but stumped into a build failure and decided to take a faster route.

Migrating to a package later down the track should be relatively straightforward.

pachlava:
> Notice I used the *UI-tests* branch name pattern to have CI build the main UI tests target and run the login smoke test scenario 👍

❤️

pachlava (approved):
I think this is a great start of implementing your proposal for scaling the UI tests, and to me this looks definitely useful. Thank you for spending your time on this!

> In this proof of concept I created a new framework, UITestHelpers, (keen on feedback on the name)

The name serves its purpose perfectly, IMO, and I can't come up with any better name.

> Checkout this branch, then run the SimplenoteUITests and SimplenoteScreenshots target. They should both build and pass on the iPhone 12 Simulator.

Tests from both targets are passing, but I had to work around the issue I had with `SimplenoteScreenshots`. I did not spend _too_ much time to find the root cause yet, since I'm about to head out, and maybe you could help me to figure this out. In any case, the issue I had is not a blocker, and I think we should merge asap 👍 

What happened when I ran the tests (using `SimplenoteScreenshots` scheme) is that `Credentials/ScreenshotsCredentials.swift` content was constantly rewritten with default values. I did not understand yet why this happened. I worked this around by directly editing the entered email and password values in `testScreenshots()` with another test account. I created "Lemon Cake & Blueberry", "Blueberry Recipes" and "Color" notes, and the test passed after that. 6 screenshots were created.

I suspect that maybe I had to use fastlane to run the screenshot tests? (because I saw the `tools.fastlane` folder created only after I tried using fastlane)? Also, maybe I had to set up any additional repositories? Thank you!

mokagio:
Thank you for taking the time @pachlava 

> What happened when I ran the tests (using `SimplenoteScreenshots` scheme) is that `Credentials/ScreenshotsCredentials.swift` content was constantly rewritten with default values

I think that might be because you don't have the original decrypted file in you local version?

If you run `bundle install && bundle exec fastlane run configure_apply`, the file should appear. If you try, let me know how it goes.

mokagio:
I'm going to merge this to keep iterating fast, but keen to hear your thoughts @thehenrybyrd and @startuptester. 

If there's something that stands out as odd, I'll correct it as I go 👍

PR #1243, opened 2021-04-14
Test notes for search are now created via paste

pachlava:
Since the notes used for search tests are huge, and they were created by typing the text, the time it took to create them was inappropriately long. This PR changes this and now the notes for search are created via pasting. This makes the execution 45-50 seconds faster.

All other notes are still created by typing, because it's less flaky imo, and there's no considerable speed gain for smaller notes 

### Changes
The changes are mainly:
- Moved `paste()` and `clearAndEnterText()` extensions to new `UITestHelpers/XCUIElement+Extensions.swift`. 
- Added an optional parameter to all functions that can create notes. The parameter tells if note text should be added by typing or by pasting. Default is by typing.
- `NoteList.clearAndEnterText()` will use typing or pasting based on `viaPaste` parameter.

### Test
1. Running `Screenshots` test on iPhone 8 Plus with no fails (this is a device for which `paste()` is invoked during login)
2. Running `Search` tests, seeing that notes are created via paste.
3. The rest of UI tests does not fail too.

### Review

### Release

### DIFF
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -25,7 +25,6 @@ class NoteEditor {
 
     class func clearAndEnterText(enteredValue: String, viaPaste: Bool = false) {
         let noteContentTextView = app.textViews.element
-        guard noteContentTextView.exists else { return }
 
         if viaPaste {
             UIPasteboard.general.strings = []
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -23,8 +23,14 @@ class NoteEditor {
         app.textViews.element.clearAndEnterText(text: "")
     }
 
-    class func clearAndEnterText(enteredValue: String) {
-        app.textViews.element.clearAndEnterText(text: enteredValue)
+    class func clearAndEnterText(enteredValue: String, viaPaste: Bool = false) {
+        if viaPaste {
+            UIPasteboard.general.strings = []
+            app.textViews.element.paste(text: enteredValue)
+            app.textViews.element.swipeUp(velocity: .fast)
+        } else {
+            app.textViews.element.clearAndEnterText(text: enteredValue)
+        }
     }
 
     class func getEditorText() -> String {
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -24,12 +24,18 @@ class NoteEditor {
     }
 
     class func clearAndEnterText(enteredValue: String, viaPaste: Bool = false) {
+        let noteContentTextView = app.textViews.element
+        guard noteContentTextView.exists else { return }
+
         if viaPaste {
             UIPasteboard.general.strings = []
-            app.textViews.element.paste(text: enteredValue)
-            app.textViews.element.swipeUp(velocity: .fast)
+            noteContentTextView.clearText()
+            noteContentTextView.paste(text: enteredValue)
+            /// Swipe up fast to show tags input, which diappears if pasted text is large
+            /// enough to push it off screen
+            noteContentTextView.swipeUp(velocity: .fast)
         } else {
-            app.textViews.element.clearAndEnterText(text: enteredValue)
+            noteContentTextView.clearAndEnterText(text: enteredValue)
         }
     }
 
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -30,8 +30,8 @@ class NoteEditor {
             UIPasteboard.general.strings = []
             noteContentTextView.clearText()
             noteContentTextView.paste(text: enteredValue)
-            /// Swipe up fast to show tags input, which diappears if pasted text is large
-            /// enough to push it off screen
+            /// Swipe up fast to show tags input, which disappears if pasted text is large
+            /// enough to push tags input off screen
             noteContentTextView.swipeUp(velocity: .fast)
         } else {
             noteContentTextView.clearAndEnterText(text: enteredValue)
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -23,15 +23,19 @@ class NoteEditor {
         app.textViews.element.clearAndEnterText(text: "")
     }
 
-    class func clearAndEnterText(enteredValue: String, viaPaste: Bool = false) {
+    class func clearAndEnterText(enteredValue: String, usingPaste: Bool = false) {
         let noteContentTextView = app.textViews.element
 
-        if viaPaste {
+        if usingPaste {
+            // Clear clipboard before usage to adress flakiness
+            // that appears when it's not done
             UIPasteboard.general.strings = []
+
             noteContentTextView.clearText()
             noteContentTextView.paste(text: enteredValue)
-            /// Swipe up fast to show tags input, which disappears if pasted text is large
-            /// enough to push tags input off screen
+
+            // Swipe up fast to show tags input, which disappears if pasted text is large
+            // enough to push tags input off screen
             noteContentTextView.swipeUp(velocity: .fast)
         } else {
             noteContentTextView.clearAndEnterText(text: enteredValue)
--- a/SimplenoteUITests/SimplenoteUITestsSearch.swift
+++ b/SimplenoteUITests/SimplenoteUITestsSearch.swift
@@ -47,7 +47,7 @@ class SimplenoteUISmokeTestsSearch: XCTestCase {
         Sidebar.tagsDeleteAll()
         NoteList.openAllNotes()
 
-        allNotes.forEach { NoteList.createNoteThenLeaveEditor($0, viaPaste: true) }
+        allNotes.forEach { NoteList.createNoteThenLeaveEditor($0, usingPaste: true) }
 }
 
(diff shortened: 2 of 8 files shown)


SimplenoteUITests/NoteEditor.swift
```diff
@@ -23,8 +23,20 @@ class NoteEditor {
...
     }
 
-    class func clearAndEnterText(enteredValue: String) {
-        app.textViews.element.clearAndEnterText(text: enteredValue)
+    class func clearAndEnterText(enteredValue: String, viaPaste: Bool = false) {
+        let noteContentTextView = app.textViews.element
+        guard noteContentTextView.exists else { return }
+
+        if viaPaste {
+            UIPasteboard.general.strings = []
```
pachlava, line 32: While this theoretically is not needed, there was some flakiness (like using old content for a new note) unless the buffer was cleared.

SimplenoteUITests/SimplenoteUITestsSearch.swift
```diff
@@ -47,7 +47,7 @@ class SimplenoteUISmokeTestsSearch: XCTestCase {
         Sidebar.tagsDeleteAll()
         NoteList.openAllNotes()
 
-        allNotes.forEach { NoteList.createNoteThenLeaveEditor($0) }
+        allNotes.forEach { NoteList.createNoteThenLeaveEditor($0, viaPaste: true) }
```
pachlava: Not sure is `viaPaste` is the best name for this parameter

UITestHelpers/XCUIElement+Extensions.swift
```diff
@@ -0,0 +1,32 @@
...
+    public func clearText() {
+        guard let stringValue = self.value as? String else {
+            XCTFail("Tried to clear and enter text into a non string value")
+            return
+        }
+
+        self.tap()
+        let deleteString = String(repeating: XCUIKeyboardKey.delete.rawValue, count: stringValue.count)
+        self.typeText(deleteString)
+    }
```
pachlava: Moved the part that clears text outside of `clearAndEnterText`

UITestHelpers/XCUIElement+Extensions.swift
```diff
@@ -0,0 +1,32 @@
+import XCTest
+
+extension XCUIElement {
+
+    public func paste(text: String) -> Void {
```
pachlava: Had to change `private func...` to `public func...` to make it usable outside the extension

SimplenoteUITests/NoteEditor.swift
mokagio: Makes sense to me. In the past, I experience some weird behavior with the state of the `UIPastebord.general` too.

I suggest making this "why is this line here" comment part of the code.

SimplenoteUITests/NoteEditor.swift
```diff
@@ -23,8 +23,19 @@ class NoteEditor {
...
-        app.textViews.element.clearAndEnterText(text: enteredValue)
+    class func clearAndEnterText(enteredValue: String, viaPaste: Bool = false) {
+        let noteContentTextView = app.textViews.element
+
+        if viaPaste {
+            UIPasteboard.general.strings = []
+            noteContentTextView.clearText()
+            noteContentTextView.paste(text: enteredValue)
+            /// Swipe up fast to show tags input, which disappears if pasted text is large
+            /// enough to push tags input off screen
```
mokagio: Nitpick.

Using `///` makes Xcode parse the comment as _documentation_ and adds it to Quick help and generated headers. Because this comment is for an internal bit of the implementation, what do you think of using `//`?

### `///` comment
screen shot 2019-02-01 at 10 49 20 am

### `//` comment
screen shot 2019-02-01 at 10 49 36 am

```suggestion
            // Swipe up fast to show tags input, which disappears if pasted text is large
            // enough to push tags input off screen
```

SimplenoteUITests/SimplenoteUITestsSearch.swift
mokagio: It didn't stood out as odd to me, so feel free to keep it.

Other ideas:

```suggestion
        allNotes.forEach { NoteList.createNoteThenLeaveEditor($0, pastingContent: true) }
```

```suggestion
        allNotes.forEach { NoteList.createNoteThenLeaveEditor($0, usingPaste: true) }
```

UITestHelpers/XCUIElement+Extensions.swift
mokagio: Yep 👍 

If you'd like to learn more about `private` vs `internal` (the default if nothing is specified) vs `public`, there is detailed documentation in the Swift website [here](https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html).

UITestHelpers/XCUIElement+Extensions.swift
```diff
@@ -0,0 +1,32 @@
+import XCTest
```
mokagio: A nitpick on the file name. As this library expands, we'll benefit from having dedicated extensions for clusters of functionality with names that make them easy to find.

This file has only logic around managing text, calling it something like `XCUIElement+TextManagement.swift` might be clearer.

Not necessarily something we need to address now, but something to keep in mind as we add more logic in here 👍

UITestHelpers/XCUIElement+Extensions.swift
```diff
@@ -0,0 +1,32 @@
...
+    public func clearAndEnterText(text: String) {
+        self.clearText()
+        self.typeText(text)
+    }
+
+    public func clearText() {
+        guard let stringValue = self.value as? String else {
+            XCTFail("Tried to clear and enter text into a non string value")
+            return
+        }
```
mokagio: I like this pre-check! I think we should use it wherever possible.

I'm not sure why Apple decided to put everything behind the `XCUIElement` proxy instead of having a more refined type structure 🤔 ...

I'm also not sure failing the test from here is the best approach in the context of a library. That's a discussion that's out of scope for this PR, though. I opened https://github.com/Automattic/simplenote-ios/issues/1244 to talk about it.

mokagio (commented):
Love a good test time improvement 🤩 Thanks!

The first time I run the tests, they failed here:

image

I run them again and it worked. I wasn't looking at the Simulator when the failure occurred so I'm not sure what there might have been on the web view... Maybe let's keep an eye out for this and, if it occurs again, let's add a call to take a screenshot in the report so we can tell what was in the web view? My guess would be that it didn't load for some network hiccup.

I run the tests again and they all passed. Both runs were on the usual iPhone 12 14.4 Simulator.

mokagio (approved):

SimplenoteUITests/NoteEditor.swift
pachlava: > Using /// makes Xcode parse the comment as documentation and adds it to Quick help and generated headers.

^ TIL ^. Changed to `//`, thank you!

SimplenoteUITests/NoteEditor.swift
pachlava: Added a comment, thanks!

UITestHelpers/XCUIElement+Extensions.swift
pachlava: I fully support not having assertions anywhere besides asserting methods 👍 Thank you for the issue

UITestHelpers/XCUIElement+Extensions.swift
pachlava: Renamed it to `XCUIElement+TextManagement.swift` as you suggest. Better do it now than forget later.

UITestHelpers/XCUIElement+Extensions.swift
pachlava: > I like this pre-check! I think we should use it wherever possible.

The whole initial method is unfortunately not mine... I just added the credits. I added the method during my trial, when I was super time-limited, and then forgot to do this after

SimplenoteUITests/SimplenoteUITestsSearch.swift
pachlava: Went for `usingPaste`, thanks for the options!

pachlava:
> I run them again and it worked. I wasn't looking at the Simulator when the failure occurred so I'm not sure what there might have been on the web view... Maybe let's keep an eye out for this and, if it occurs again, let's add a call to take a screenshot in the report so we can tell what was in the web view? My guess would be that it didn't load for some network hiccup.

Yes, this is one of the flaky tests. Good idea to have a screenshot!

UITestHelpers/XCUIElement+Extensions.swift
pachlava: Thank you! While the concept of access control looks very straightforward on the surface, for me it gets complicated when getting deeper into details. Mostly, I feel, because of the lack of practice with access control. I keep returning to this page from time to time, and try to understand it more than at the previous time.

mokagio: >  I keep returning to this page from time to time, and try to understand it more than at the previous time.

That's it with most things really. It takes time and practice.

UITestHelpers/XCUIElement+Extensions.swift
mokagio: No worries. I'm a fan of getting things in fast, then refining them. As Kent Beck says, _Make it work, make it right, make it fast_.

PR #1246, opened 2021-04-16
Feature: New Note Bar

### LINKED ISSUE #1044: Editor: New Note Bar

charliescheer:
### Fix
Fixes #1044 
This is a continuation of #1241 After that original PR I started over how I was doing this on a new branch.  This PR responds to the feedback from the previous PR (will comment and respond to the questions on this PR)

This PR serves a dual purpose of Adding the new note bar feature and removing the current add new note button from the top navigation bar.

This new note bar will appear on both the note list and the note editor, making it easy for a user to create a new note where ever they are in Simplenote iOS.  This is added using the navigationController bottom toolbar, same as the search navigation bar, so there is some logic added to switch between the two in the note editor.  This will also share ui with the multiple edit delete mode (the next step in this process) but that is not covered in this PR

Note list:

Note Editor:

### Test
1. Open up the note list and see the new note bar.  Tap on the button and a new note will be created and the editor will load
2. Exit that new note, select one of your notes to see the new note bar there.  Again test tapping on the button to see a new note be created
3. Return to the note list and do a search.  Select a note from your search, you will see the note editor load with the search results navigation.  Tap done to see the toolbar transition to the new note bar
4. Make sure that the new note button no longer appears in the top toolbar in the editor. Check that when the keyboard appears the dismiss keyboard button still appears
5. Go to the trash, check to see that the delete button still appears.  Return to the note list (not trash mode) to see there is no button in the top right.
6. test dark mode to make sure colors/images are correct

### Review

### DIFF
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -97,7 +97,6 @@ CGFloat const SPSelectedAreaPadding = 20;
     [self configureNavigationBarItems];
     [self configureNavigationBarBackground];
     [self configureRootView];
-    [self configureSearchToolbar];
     [self configureLayout];
     [self configureTagListViewController];
     [self configureInterlinksProcessor];
@@ -120,9 +119,9 @@ CGFloat const SPSelectedAreaPadding = 20;
     [self setupNavigationController];
     [self highlightSearchResultsIfNeeded];
     [self startListeningToKeyboardNotifications];
-
     [self refreshNavigationBarButtons];
-
+    [self configureNavigationControllerToolbar];
+    [self setupNavigationController];
     // Async here to make sure all the frames are correct
     dispatch_async(dispatch_get_main_queue(), ^{
         [self restoreScrollPosition];
@@ -132,7 +131,6 @@ CGFloat const SPSelectedAreaPadding = 20;
 - (void)viewDidAppear:(BOOL)animated
 {
     [super viewDidAppear:animated];
-
     self.userActivity = [NSUserActivity openNoteActivityFor:self.note];
     [self ensureEditorIsFirstResponder];
 }
@@ -147,7 +145,13 @@ CGFloat const SPSelectedAreaPadding = 20;
 {
     // Note: Our navigationBar *may* be hidden, as per SPSearchController in the Notes List
     [self.navigationController setNavigationBarHidden:NO animated:YES];
-    [self.navigationController setToolbarHidden:!self.searching animated:YES];
+
+    if (UIDevice.isPad) {
+        [self.navigationController setToolbarHidden:!self.searching animated:YES];
+    } else {
+        [self.navigationController setToolbarHidden:NO animated:YES];
+    }
+
 }
 
 - (void)ensureEditorIsFirstResponder
@@ -304,6 +308,20 @@ CGFloat const SPSelectedAreaPadding = 20;
     [self setToolbarItems:@[self.doneSearchButton, flexibleSpace, detailButton, flexibleSpaceTwo, self.prevSearchButton, self.nextSearchButton] animated:NO];
 }
 
+- (void)configureNavigationControllerToolbar
+{
+    if (UIDevice.isPad) {
+        [self configureSearchToolbar];
+        return;
+    }
+
+    if (self.searching) {
+        [self configureSearchToolbar];
+    } else {
+        [self configureNewNoteBar];
+    }
+}
+
 - (void)ensureNoteIsVisibleInList
 {
     // TODO: This should definitely be handled by the Note List itself. Please!
@@ -407,7 +425,9 @@ CGFloat const SPSelectedAreaPadding = 20;
     NSMutableArray *buttons = [NSMutableArray array];
 
     if (self.shouldHideKeyboardButton) {
-        [buttons addObject:self.createNoteButton];
+        if (UIDevice.isPad) {
+            [buttons addObject:self.createNoteButton];
+        }
     } else {
         [buttons addObject:self.keyboardButton];
     }
@@ -574,7 +594,11 @@ CGFloat const SPSelectedAreaPadding = 20;
     
     self.searching = NO;
 
-    [self.navigationController setToolbarHidden:YES animated:YES];
+    if (UIDevice.isPad) {
+        [self.navigationController setToolbarHidden:YES animated:YES];
+    } else {
+        [self configureNavigationControllerToolbar];
+    }
     self.searchDetailLabel.text = nil;
 }
 
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -96,6 +96,7 @@ CGFloat const SPSelectedAreaPadding = 20;
 
     [self configureNavigationBarItems];
     [self configureNavigationBarBackground];
+    [self configureNavigationControllerToolbar];
     [self configureRootView];
     [self configureLayout];
     [self configureTagListViewController];
@@ -117,10 +118,9 @@ CGFloat const SPSelectedAreaPadding = 20;
     [super viewWillAppear:animated];
 
     [self highlightSearchResultsIfNeeded];
+    [self setupNavigationController];
     [self startListeningToKeyboardNotifications];
     [self refreshNavigationBarButtons];
-    [self configureNavigationControllerToolbar];
-    [self setupNavigationController];
     // Async here to make sure all the frames are correct
     dispatch_async(dispatch_get_main_queue(), ^{
         [self restoreScrollPosition];
--- a/Simplenote/Classes/SPNoteEditorViewController.m
+++ b/Simplenote/Classes/SPNoteEditorViewController.m
@@ -118,7 +118,7 @@ CGFloat const SPSelectedAreaPadding = 20;
     [super viewWillAppear:animated];
 
     [self highlightSearchResultsIfNeeded];
-    [self setupNavigationController];
+    [self configureNavigationController];
     [self startListeningToKeyboardNotifications];
     [self refreshNavigationBarButtons];
     // Async here to make sure all the frames are correct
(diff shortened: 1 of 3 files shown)


charliescheer:
Questions from last PR:
@SylvesterWilmott 

>On iPad, the new note button should always appear in the top right side of the navigation bar. This is less important now but will become important in the future when we implement the split view. This is because for users who use the on-screen keyboard, the bottom bar would almost always be covered. For reference: zpl://screen?pid=5db058eaf5c10222cfd424b6&sid=5fca5da3acf2934fa94c02a6

We chatted a bit about this over slack, and as discussed I have left the iPad new note button un changed, when you check this all out on an iPad view it should look as it did before, the changes should only appear on iPhone.

>The bottom bar should not be visible in the search results. This is because creating a new note is currently not possible from the search results and in the point above, won't be possible to do on iPad.

👍 fixed

@SylvesterWilmott  Question for you.  I noticed that in Zeplin the color for the new note bar in dark mode on iPhone is different than the color for the multi select toolbar.  Is this intentional?  on the iOS mockup the background is almost black (rgba 23 23 23, 0.5) ref: zpl://screen?sid=5fb540acdf8c0d745151c8c8&pid=5db058eaf5c10222cfd424b6
where on the iPad screens the color is a lighter grey color (the same color as the sort bar had been rgba 28 28 30, 0.8) 
ref: zpl://screen?pid=5db058eaf5c10222cfd424b6&sid=5fca5da001fcd454fa5f0b11
Should these be the same?

charliescheer:
@jleandroperez 
>iOS 12
Launch an iOS 12 simulator
Switch themes
As a result, the new Toolbar won't get adjusted

Nice catch.  Fixed it.  Should be fine on iOS 12 now

>Swipe Back
Launch the app
Press over any Note to access the editor
Swipe back
As a result, the toolbar will end up misplaced (reproduced in iOS 14 / device).

That was a bizarre one, fixed it

SylvesterWilmott:
>I noticed that in Zeplin the color for the new note bar in dark mode on iPhone is different than the color for the multi select toolbar.

That's simply a result of me not really knowing what colour we use for the bar haha. They should both match in colour.

SylvesterWilmott:
Looking good! Thanks @charliescheer 

It looks like all of the previous issues are fixed in this PR ie. tags are not covered by the bar and going back from the notes list is smooth.

One thing I noticed was that in the trash, the scrollbar appears in the wrong place. This could be related to the fact we are hiding the bar in the trash? But I think that we shouldn't be hiding the bar in trash anyway.

Also just a little FYI, I am still experiencing the bug with the variable widths in the swipe interactions in the notes list in this PR.

charliescheer:
>One thing I noticed was that in the trash, the scrollbar appears in the wrong place. This could be related to the fact we are hiding the bar in the trash? But I think that we shouldn't be hiding the bar in trash anyway.

Cool.  I will make sure it is not being hidden in the trash

>Also just a little FYI, I am still experiencing the bug with the variable widths in the swipe interactions in the notes list in this PR.

I saw your note about the labels in slack.  that is probably causing it.  I recently added some labels to those buttons which is  probably causing them to appear and also probably causing the change in sizing.  I will try setting those labels to just be accessibility labels which should hide them again.  Thanks!

jleandroperez:
@charliescheer sir!! please LMK whenever this one is ready for another look.

Thank youuuu!!

charliescheer:
Alright, @SylvesterWilmott can you take another look at this?  I have made the new note bar appear in the trashed menu.  I didn't see anything off about the scroll bar in trash, so hopefully it is fixed by adding the new note bar.  If you are still seeing the issue can you grab a screenshot of what you are seeing?

@jleandroperez this is ready to be reviewed again

SylvesterWilmott:
Looks great @charliescheer. The issue mentioned regarding the trash scrollbar seems to be resolved with the latest commit. :shipit:

pachlava:
I tried this:

1. Open `Trash`
2. Create new note
3. Go back ~> selection is `Trash`, new note is not there (because it is in `All Notes`)

This looks both logical and illogical somehow. There was a very similar [issue](https://github.com/Automattic/simplenote-electron/pull/2746) in Electron.

Should this be done for iOS too?

cc @SylvesterWilmott

SylvesterWilmott:
Thanks for taking a look @pachlava. I think switching note list context should not occur on iOS since it would violate the platform's recommended behaviour. This is also true for Android.

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -117,12 +116,11 @@ - (void)viewWillAppear:(BOOL)animated
 {
     [super viewWillAppear:animated];
 
-    [self setupNavigationController];
```
jleandroperez, line 120: Q: why was this one relocated? (If there's some initialization that needs to happen before, mind adding a note in a comment?)

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -147,7 +144,13 @@ - (void)setupNavigationController
 {
     // Note: Our navigationBar *may* be hidden, as per SPSearchController in the Notes List
     [self.navigationController setNavigationBarHidden:NO animated:YES];
-    [self.navigationController setToolbarHidden:!self.searching animated:YES];
+
+    if (UIDevice.isPad) {
+        [self.navigationController setToolbarHidden:!self.searching animated:YES];
+    } else {
+        [self.navigationController setToolbarHidden:NO animated:YES];
+    }
```
jleandroperez: Suggestion!

```
BOOL mustHideToolbar = UIDevice.isPad && !self.searching;
[self.navigationController setToolbarHidden:mustHideToolbar animated:YES];
```

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -117,12 +116,11 @@ - (void)viewWillAppear:(BOOL)animated
...
     [super viewWillAppear:animated];
 
-    [self setupNavigationController];
     [self highlightSearchResultsIfNeeded];
     [self startListeningToKeyboardNotifications];
-
     [self refreshNavigationBarButtons];
-
+    [self configureNavigationControllerToolbar];
+    [self setupNavigationController];
```
jleandroperez: **Nipticky!!**

This API was already there (not really new code!!), buuut for consistency's sake, mind renaming `setupNavigationController` > `configureNavigationController` so that all the things look the same?

**Thank you!!**

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -304,6 +306,20 @@ - (void)configureSearchToolbar
...
+    if (UIDevice.isPad) {
+        [self configureSearchToolbar];
+        return;
+    }
+
+    if (self.searching) {
+        [self configureSearchToolbar];
+    } else {
+        [self configureNewNoteBar];
+    }
```
jleandroperez, line 313: There's a bit of duplication. You caaan burn it as follows:

```
BOOL isPad = UIDevice.isPad;

if (isPad || self.searching) {
   [self configureSearchToolbar];
} else if (!isPad) {
   [self configureNewNoteBar];
}
```

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -574,7 +592,11 @@ - (void)endSearching:(id)sender {
     
     self.searching = NO;
 
-    [self.navigationController setToolbarHidden:YES animated:YES];
+    if (UIDevice.isPad) {
+        [self.navigationController setToolbarHidden:YES animated:YES];
+    } else {
+        [self configureNavigationControllerToolbar];
+    }
```
jleandroperez: Note: You can probably encapsulate this logic in the `configureNavigationControllerToolbar` API 

```
guard !UIDevice.isPad else {
   navigationController.setToolbarHidden(true, animated: true)
   return
}
```

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -229,10 +230,32 @@ - (void)refreshStyle {
...
+- (void)refershNavigationButtons {
+    [self updateNavigationBar];
+    [self refreshNavigationControllerToolbar];
+}
+
+- (void)refreshNavigationControllerToolbar
+{
+    if (UIDevice.isPad) {
+        [self.navigationController setToolbarHidden:YES animated:YES];
+    } else {
```
jleandroperez: Suggestion!!

```
  if (UIDevice.isPad) {
         [self.navigationController setToolbarHidden:YES animated:YES];
         return;
     }
```

That way you can remove the `else` (and unindent a few lines!) 😄

jleandroperez (approved):
Looks good @charliescheer !!

✅ Dark Mode works great on iOS 12 / iOS 14
✅ Verified on iPad 
✅ Verified on iPhone

Few nipticky comments sent your way. Other than that :shipit: sir!!

SylvesterWilmott (approved):

Simplenote/Classes/SPNoteEditorViewController.m
charliescheer: Looking at it, I think that I moved the wrong thing actually.  setupNavigationController was coming after  configureNavigationControllerToolbar which created the buttons for the toolbar.  I moved that to viewDidLoad with the other configuration methods and put setupNavigationController back where it was so that it all fit consistently with the rest of the code flow

Simplenote/Classes/SPNoteEditorViewController.m
charliescheer: Ah cool idea.  There was still some work being done in configureNavigationControllerToolbar that other times needed to be run, but was able to put the display/hide into that work like this:

if (UIDevice.isPad) {
        [self configureSearchToolbar];
        [self.navigationController setToolbarHidden:!self.searching animated:YES];
        return;
    }

which works great.  Thanks!

Simplenote/Classes/SPNoteEditorViewController.m
charliescheer: Thinking about this, and I realized that I had separated setting up the buttons and displaying the buttons into different functions.  
I made this change, but also updated the code for endSearching to reset what the buttons are and hide them using the separate functions instead of inserting hide/don't hide into the setup function

mokagio:
GitHub shows the PR is part of 4.35 but that's not actually the case, see https://github.com/wordpress-mobile/release-toolkit/pull/253.

It will be part of 4.36.

PR #1247, opened 2021-04-16
Use private(set) for var tags property in NoteData

mokagio:
This way, we can use the compiler-generated initializer while also keeping the instance immutable for consumers, which is appropriate for a DTO-like type such as this one.

See https://github.com/Automattic/simplenote-ios/pull/1240#discussion_r612848495

### Test
A green CI is all the validation we need for this compiler-level change.

### Review
Only one developer are required to review these changes, but anyone can perform the review.

### Release

### DIFF
--- a/SimplenoteUITests/NoteData.swift
+++ b/SimplenoteUITests/NoteData.swift
@@ -2,7 +2,7 @@
 struct NoteData {
     let name: String
     let content: String
-    var tags: [String] = []
+    private(set) var tags: [String] = []
 
     var formattedForAutomatedInput: String { "\(name)\n\n\(content)" }
 }


pachlava (approved):
Thank you for the improvement @mokagio ! Checked this out and it works as advertised! 

I'll merge it myself now to keep the ball rolling.

There's one thing I'm not sure if I understand: `private(set)` tells that variable's setter access level is private, while getter still remains default (`internal`). This means, from what I understood, is that it's possible to set only within the struct, and to get within the module. I don't understand why it's still possible to assign/set value to `tags` externally during the creation of an instance? 

Is this because when assigning the `tags` value the first time (during instance creation), it's not a usual 'setting' but 'initial setting/initializing'?

mokagio:
> This means, from what I understood, is that it's possible to set only within the struct, and to get within the module. 

Correct.

> why it's still possible to assign/set value to tags externally during the creation of an instance?
> 
> Is this because when assigning the tags value the first time (during instance creation), it's not a usual 'setting' but 'initial setting/initializing'?

Correct again. I'm not 100% sure of about what's going on under the hood in how Swift manages the memory, but, basically, when we call then `init` method, the properties are instantiated for the first time and is therefore possible to set them even if they're `let` (constants, can set only once) and/or `private` (only visible from within the type implementation).

Also notice that `private(set)` only acts on the property itself. We could always write a method to set the value externally. That would be an _evil_ API, very confusing for a reader.

```swift
struct EvilType {
    private(set) var value = 0

    mutating func set(value newValue: Int) {
        value = newValue
    }
}
    
func t() {
    var e = EvilType(value: 1)
    e.value = 2 // doesn't compile because value is private(set), even though e is a var
    e.set(value: 3)
    print(e.value) // 3
}
```

To be honest, I'm my understanding of these mechanics is a bit superficial. [This talk](https://www.skilled.io/u/playgroundscon/how-i-learned-to-stop-worrying-and-love-mutation) goes a bit in the details, the video is not available (why?!) but the transcript is still there.

PR #1250, opened 2021-04-19
Add UI smoke test as CI gate for release build

mokagio:
_WIP_

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,9 +9,9 @@ orbs:
 parameters:
   beta_build:
     type: boolean
-    default: false 
+    default: false
   release_build:
-    type: boolean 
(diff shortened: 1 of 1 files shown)


mokagio:
This PR has become stale and we never had a proper discussion about having UI tests as CI gates for release.

pachlava:
I only now noticed this PR 😞 

I would not feel safe doing this atm, because when looking [here](https://app.circleci.com/pipelines/github/Automattic/simplenote-ios), I still see sporadic fails here or there in the UI tests. 

I see a point in returning to this once I improve the situation with fails.

mokagio:
> I see a point in returning to this once I improve the situation with fails.

Yeah, totally 👍

PR #1258, opened 2021-04-24
Issue/1248 contexual menu sizing issue

### LINKED ISSUE #1248: Contextual menu items can appear uneven
When you are looking at the note list in Simplenote iOS, if you swipe right to see the contextual menu options, the different menu buttons should be the same size, but in some circumstances they can appear uneven.  This appears to be an issue of the labels appear with the buttons in certain circumstances:

There have been some cases of the sizes appearing differently without the labels, but the labels do seem to be causing the issue.

### Reproduced
1. enable accessibility sizes (XL or larger) on your device which will cause the labels to appear on the buttons
2. Go to the note list and swipe to the left on a note
3. the contextual options will appear and will have different sizes

charliescheer:
### Fix
This PR fixes #1248 

After a recent change to the note list contextual menu, labels were added to each menu item.  Under certain circumstances it was possible for these menu labels to appear, and that was causing the buttons to have uneven sizing:

This PR changes the labels on the buttons from generic labels to accessibility labels.  This way they will not physically appear on screen but will still give correct labels to the users who use voiceover.

### Test
1. Enable voice over on your device and go to the notes list
2. Change the rotor to actions, and select a note.  Listen to the different actions they should be "move to trash, pin to top, share, delete note"
3. Now disable voiceover, return to the note list and swipe to the right to see the contextual menu options. They should be all the same size
4. Enable dynamic type on the device and choose a text size over XL (this would make the labels appear and cause the button sizing issue) make sure the labels to not appear and the buttons are equally sized

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -656,28 +656,33 @@ private extension SPNoteListViewController {
         let pinImageName: UIImageName = note.pinned ? .unpin : .pin
         let pinActionTitle: String = note.pinned ? ActionTitle.unpin : ActionTitle.pin
 
-        return [
-            UIContextualAction(style: .destructive, title: ActionTitle.delete, image: .image(name: .trash), backgroundColor: .simplenoteDestructiveActionColor) { [weak self] (_, _, completion) in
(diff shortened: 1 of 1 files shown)


SylvesterWilmott (approved):
Thanks @charliescheer. Works as expected.

eshurakov (approved):
Looks great, thank you!

:shipit:

mokagio:

It will be part of 4.36.

PR #1260, opened 2021-04-24
Remove all caps on collaboration alert

### LINKED ISSUE #1259: Remove caps lock on alert notification
In the alert message that highlights how collaboration has been moved, the heading text is in all capital letters, which feels a bit shouty.

We should use standard casing for this message instead of all caps.

### Reproduced
1. While editing a note, enter an email address into the tags (the old way to add collaborators)
2. An alert should appear with a title in all caps.

charliescheer:
### Fix
This PR fixes #1259 

This is a quick PR that changes the messaging for the text alert to notify users that collaboration has moved so that it no longer appears in all caps.

After the change:

### Test
1. open up a note
2. in the tags section, enter in an email address other than the email on your account
3. The warning that collaboration has moved should appear.  Make sure the title is not in all caps
4. Double check that there are no lint errors in file NoteEditorTagListViewController.swift

### Review

### DIFF
--- a/Simplenote/Tags/NoteEditorTagListViewController.swift
+++ b/Simplenote/Tags/NoteEditorTagListViewController.swift
@@ -211,8 +211,8 @@ private struct Constants {
 //
 private struct Localization {
     enum CollaborationAlert {
-        static let title = NSLocalizedString("Collaboration has moved", comment: "")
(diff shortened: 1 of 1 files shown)


SylvesterWilmott (approved):
Thanks @charliescheer!

eshurakov (approved):
LGTM! 

:shipit:

mokagio:

It will be part of 4.36.

PR #1262, opened 2021-04-26
Issue/1229 quick dissmiss notices

### LINKED ISSUE #1229: Dismiss notices when app outside of notification is interacted with

charliescheer:
### Fix
This PR Fixes #1229 
is a continuation of #1230 

Currently it is possible to make a series of changes in simplenote iOS that would create more notices, however because you can't queue several of the same notices at once, the subsequent actions receive no notification.  This could be problematic in the case that a user deletes multiple notes really quickly.  In that case the "restore" button on the trashed notice would refer to the first note trashed, not the last note trashed.

In this PR, I have made a change so that if the passthru view that the notice is contained in is interacted with, the notice will be dismissed.  That way you would never be able to create multiple notices at once, as they are dismissed once you interact with the rest of the app in anyway.

### Test
1. From the note list try to delete a note using the contextual actions, a note will appear
2. now delete a another note, you will see the first notice is dismissed before the second one can be trashed.

### Review

### Release
> 
> > Added markdown support

### DIFF
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -64,8 +64,8 @@ class NoticeController {
     // MARK: Dismissing
     //
     private func dismiss(withDuration duration: TimeInterval = UIKitConstants.animationLongDuration, completion: (() -> Void)? = nil) {
-        self.timer = nil
-        self.current = nil
+        timer = nil
+        current = nil
 
         noticePresenter.dismissNotification(withDuration: duration) {
             completion?()
--- a/Simplenote/NoticeController.swift
+++ b/Simplenote/NoticeController.swift
@@ -34,9 +34,8 @@ class NoticeController {
     //
     func present(_ notice: Notice) {
         if isPresenting {
-            dismiss {
+            dismiss(withDuration: UIKitConstants.animationQuickDuration) {
                 self.present(notice)
-                print("canceled")
             }
             return
         }
@@ -48,7 +47,6 @@ class NoticeController {
             let delay = self.current?.action == nil ? Times.shortDelay : Times.longDelay
             self.timer = self.timerFactory.scheduledTimer(with: delay, completion: {
                 self.dismiss {
-                    print("full process")
                 }
             })
         }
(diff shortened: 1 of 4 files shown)


charliescheer:
closing for now.  Had some new ideas I wanted to test before it was reviewed

SylvesterWilmott:
The behaviour looks good to me!

Unfortunately sometimes the notices are still getting stuck. This is usually happening with either the "publish successful" or "publish unsuccessful" ones. It's not really clear how to reproduce but if you publish and unpublish a few times then the issue occurs.

charliescheer:
@SylvesterWilmott Thanks for pointing that out.  I was finally able to identify where the issue was coming from and fix it with @jleandroperez 's help.  If you give it a test now you shouldn't have the issue of notices getting stuck on screen.

@jleandroperez this is ready to be reviewed now

SylvesterWilmott:
>If you give it a test now you shouldn't have the issue of notices getting stuck on screen.

All fixed now, thanks!

SylvesterWilmott (approved):

Simplenote/NoticeController.swift
```diff
@@ -76,16 +63,12 @@ class NoticeController {
 
     // MARK: Dismissing
     //
-    @objc
-    private func dismiss() {
+    private func dismiss(withDuration duration: TimeInterval = UIKitConstants.animationLongDuration, completion: (() -> Void)? = nil) {
         self.timer = nil
+        self.current = nil
```
jleandroperez: No need for `self` (`timer = nil` as well!)

jleandroperez (approved):
Looks good sir!!

:shipit:

mokagio:

It will be part of 4.36.

PR #1264, opened 2021-04-28
Adding support of new note bar

pachlava:
### Why
1. `New Note` button has moved from the Navigation Bar to a dedicated [new note bar](https://github.com/Automattic/simplenote-ios/pull/1246).
2. [This commit](https://github.com/Automattic/simplenote-ios/commit/e7ba032e087ca84813bdb0e8fe2ca054d79367a6) improved accessibility labels for `Trash` contextual menu, which meant having new labels for buttons.
 
### Fix
1. The `New Note` button locator now refers directly to `app`, not to the `navigationBars`. This means that the code will work on older versions too, and as well on iPad, where the `New Note` button was [left](https://github.com/Automattic/simplenote-ios/pull/1246#issuecomment-820861982) on its original place.
2. Renamed several buttons: 
- `noteTrash` ~> `trashNote`
- `trashRestore` ~> `restoreNote`
3. Changed the labels of these buttons
4. Added new `deleteNote` button.

### Test
1. Running `SimplenoteUISmokeTestsTrash` suite on iPhone 12. It involves both `New Note` button and `Trash` contextual menu.
2. Running same suite on iPad 8th Gen, where `New Note` button is still on old place.
3. Optionally, running all other tests on iPhone 12.

### Review
Only one developer and one designer are required to review these changes, but anyone can perform the review. Thank you 🙇 

### Release


mokagio (commented):
I think something has changed in the account, most tests are failing and this is what I see in the screen

Screen Shot 2021-04-29 at 2 39 41 pm


The failures I'm getting are like

> `SimplenoteUISmokeTestsTrash.testCanDeleteNoteForeverIndividually()`: 
Invalid query - string identifier '[
"`/Users/<your user name>/Library/Developer/CoreSimulator/Devices/8E6A226A-5B5E-495A-BFDC-DB871D1468AF/data/Containers/Bundle/Application/A7F50922-FA89-4F19-BB80-C4AEF14AFFF9/SimplenoteUITests-Runner.app/SimplenoteUITests-Runner`", 
"`-NSTreatUnknownArgumentsAsOpen`", "`NO`", "`-ApplePersistenceIgnoreState`",
"`YES`"]' exceeds maximum length of 128 characters.
You can work around this limitation by constructing a query with a custom 
NSPredicate that specifies the property (label, title, value, 
placeholderValue, or identifier) to match against. 
(NSInternalInconsistencyException)
```

**Comment by pachlava** [comment]
@mokagio Thank you 😄! I was working on another task, but the same account was still logged in on Simulator, and the text got into the note. This uncovered a limitation I did not know about, which is not using 128 or more characters when locating an element. Nice!

This happened when the script was trashing all notes, where it firstly saves all notes names, and then tries to find and swipe the corresponding cell for every note name.

I did some testing, and could reproduce this when had a note with a lengthy title. I just [committed](https://github.com/Automattic/simplenote-ios/pull/1264/commits/74f161fddc836bfa1f53b12bb7ff84304632c5c0) a change, and now `Table.getCell` method uses NSPredicate (as failure message suggested), which solved the problem, at least in this place.

File: SimplenoteUITests/Generic.swift [hunk 1]
```diff
@@ -80,15 +80,16 @@ class Table {
...
+            UID.Button.deleteNote : UID.Button.trashNote
         let noteCell = Table.getCell(label: noteName)
 
         noteCell.swipeLeft()
         noteCell.buttons[deleteButtonLabel].tap()
     }
 
     class func getCell(label: String) -> XCUIElement {
-        let cell = app.tables.cells[label]
+        let predicate = NSPredicate(format: "label LIKE '\(label)'")
```
mokagio, line 91: TIL about this limit with long-ish strings too. 👍

mokagio (approved):
Tried if after your fixes. All tests passed on the iPhone 12 Simulator.

PR #1267, opened 2021-05-02
Issue/1043 add multi select

### LINKED ISSUE #1043: List: Multiple Selection Support
### Details:
- [x] New Note Bar UI: Make sure it plays well with the Sort Bar please!
- [x] TableView multiple row selection mechanism
- [x] Display the number of selected notes in the title
- [x] Deletion Confirmation TBD

Filing yet another of the new Spring Cleaning tasks!!
cc @SylvesterWilmott @eshurakov

charliescheer:
### Fix
This PR fixes #1043 

This PR is intended to add multiple selection support to Simplenote iOS.  Currently there is not way to select and trash multiple notes at a time in SNiOS.  In this PR the ability to enter an edit mode, select multiple notes at a time and then trash them is now possible!

### Test
1. enable and disable edit mode.  Make sure that the mode animates in and out.  While active ensure that you can select multiple notes, that you can't transition to seeing a note and that entering search mode dismisses edit mode
2. Make sure when you go in and out of edit mode the button in the new note bar at the bottom transitions from the new notice icon to the trash icon
3. While in edit mode, select multiple notes to make sure that the view title changes from the standard note list name to "X Selected" Exit edit mode to ensure that the note list returns to its regular name.
4. Test selecting multiple notes and then using the trash button to trash the notes
5. Test changing the note list tag while in edit mode to ensure that edit mode exits
6. While in edit mode you should see a Select All/Deselect all button in the top left where the main menu usually is.  Enable/disable edit mode to make sure that appears and disappears
7. Tap on the Select All button and confirm all notes are selected.  Tap Deselect and make sure they are deselected
8. Delete a selection of notes to see a new in app notification.  Make sure the correct number of notes is marked as being deleted in the notification.  Test the undo button to make sure that all the bulk deleted notes are restored

### Review

### Release
> 
> > Added markdown support

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -572,8 +572,16 @@ extension SPNoteListViewController: UITableViewDelegate {
     }
 
     public func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+        guard let cell = cell as? SPNoteTableViewCell else {
+            return
+        }
         var insets = SPNoteTableViewCell.separatorInsets
         insets.left -= cell.layoutMargins.left
+
+        if isEditing {
+            insets.left += cell.checkboxContainingView.frame.width
+        }
+
         cell.separatorInset = insets
     }
 }
@@ -666,11 +674,11 @@ extension SPNoteListViewController {
         super.setEditing(editing, animated: animated)
 
         // Reloading data ensures that no cells are selected and the current data is up to date
-        tableView.reloadData()
+        let rows = tableView.indexPathsForVisibleRows
+        tableView.reloadRows(at: rows!, with: .automatic)
 
         // Dispatches to asyncAfter because reloadData will override the animations if not complete
         DispatchQueue.main.asyncAfter(deadline: .now() + 0.1, execute: {
-            self.tableView.deselectSelectedRow(animated: true)
             self.tableView.setEditing(editing, animated: animated)
             self.configureNavigationToolbarButton()
         })
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -606,10 +606,18 @@ extension SPNoteListViewController: UITableViewDelegate {
     }
 
     public func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+        guard let cell = cell as? SPNoteTableViewCell else {
+            return
+        }
+        
         var insets = SPNoteTableViewCell.separatorInsets
         insets.left -= cell.layoutMargins.left
 
         cell.separatorInset = insets
+
+        if indexPath.row == notesListController.numberOfObjects - 1 {
+            cell.shouldDisplayBottomSeparator = false
+        }
     }
 }
 
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -258,12 +258,8 @@ extension SPNoteListViewController {
     }
 
     func refreshSelectAllLabels() {
-        var deselect = false
         let numberOfSelectedRows = tableView.indexPathsForSelectedRows?.count ?? 0
-
-        if notesListController.numberOfObjects == numberOfSelectedRows {
-            deselect = true
-        }
+        let deselect = notesListController.numberOfObjects == numberOfSelectedRows
 
         selectAllButton.title = Localization.selectAllLabel(deselect: deselect)
         selectAllButton.isAccessibilityElement = true
@@ -388,9 +384,8 @@ extension SPNoteListViewController {
 
     @objc
     func selectAllWasTapped() {
-
         if notesListController.numberOfObjects == tableView.indexPathsForSelectedRows?.count {
-            tableView.deselectAllRows(inSection: 0, animated: false)
+            tableView.deselectAllRows(inSection: .zero, animated: false)
             refreshNavigationBarLabels()
 
             return
@@ -709,18 +704,18 @@ extension SPNoteListViewController {
     open override func setEditing(_ editing: Bool, animated: Bool) {
         super.setEditing(editing, animated: animated)
         refreshEditButtonTitle()
-        
-        tableView.deselectAllRows(inSection: 0, animated: false)
+
+        tableView.deselectAllRows(inSection: .zero, animated: false)
         updateNavigationBar()
 
         tableView.allowsMultipleSelection = editing
-        toggleMultiSelectEditingTo(editing)
+        toggleVisibleCells(to: editing)
 
         configureNavigationToolbarButton()
         setListViewTitle()
     }
 
-    func toggleMultiSelectEditingTo(_ editing: Bool) {
+    func toggleVisibleCells(to editing: Bool) {
         guard let noteCells = tableView.visibleCells as? [SPNoteTableViewCell] else {
             return
         }
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -709,6 +709,7 @@ extension SPNoteListViewController {
         super.setEditing(editing, animated: animated)
         tableView.setEditing(editing, animated: animated)
         refreshEditButtonTitle()
+        refreshSelectAllLabels()
 
         tableView.deselectAllRows(inSection: .zero, animated: false)
         updateNavigationBar()
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -257,6 +257,20 @@ extension SPNoteListViewController {
         updatePlaceholderPosition()
     }
 
+    func refreshSelectAllLabels() {
+        var deselect = false
+        let numberOfSelectedRows = tableView.indexPathsForSelectedRows?.count ?? 0
+
+        if notesListController.numberOfObjects == numberOfSelectedRows {
+            deselect = true
+        }
+
+        selectAllButton.title = Localization.selectAllLabel(deselect: deselect)
+        selectAllButton.isAccessibilityElement = true
+        selectAllButton.accessibilityLabel = Localization.selectAllLabel(deselect: deselect)
+        selectAllButton.accessibilityHint = Localization.selectAllAccessibilityHint
+    }
+
     private var placeholderDisplayMode: SPPlaceholderView.DisplayMode {
         if isIndexingNotes || SPAppDelegate.shared().bSigningUserOut {
             return .generic
@@ -378,6 +392,20 @@ extension SPNoteListViewController {
         let flexibleSpace = UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
         let actionButton = isEditing ? trashButton : addButton
         setToolbarItems([flexibleSpace, actionButton], animated: true)
+    }
+
+    @objc
+    func selectAllWasTapped() {
+
+        if notesListController.numberOfObjects == tableView.indexPathsForSelectedRows?.count {
+            tableView.deselectSelectedRows(animated: true)
+            refreshSelectAllLabels()
+
+            return
+        }
+
+        tableView.selectAll(nil)
+        refreshSelectAllLabels()
     }
 }
 
@@ -532,6 +560,7 @@ extension SPNoteListViewController: UITableViewDelegate {
     public func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
         if isEditing {
             setListViewTitle()
+            refreshSelectAllLabels()
             return
         }
 
@@ -552,6 +581,7 @@ extension SPNoteListViewController: UITableViewDelegate {
     public func tableView(_ tableView: UITableView, didDeselectRowAt indexPath: IndexPath) {
         if isEditing {
             setListViewTitle()
+            refreshSelectAllLabels()
         }
     }
 
@@ -682,6 +712,7 @@ extension SPNoteListViewController {
 
     open override func setEditing(_ editing: Bool, animated: Bool) {
         super.setEditing(editing, animated: animated)
+        updateNavigationBar()
 
         // Reloading data ensures that no cells are selected and the current data is up to date
         let rows = tableView.indexPathsForVisibleRows
@@ -1097,4 +1128,12 @@ private enum Localization {
         let string = NSLocalizedString("%i Selected", comment: "Count of currently selected notes")
         return String(format: string, count)
     }
+
+    static func selectAllLabel(deselect: Bool) -> String {
+        let selectLabel = NSLocalizedString("Select All", comment: "Select all Button Label")
+        let deselectLabel = NSLocalizedString("Deselect All", comment: "Deselect all Button Label")
+        return deselect ? deselectLabel : selectLabel
+    }
+
+    static let selectAllAccessibilityHint = NSLocalizedString("Tap button to select or deselect all notes", comment: "Accessibility hint for the select/deselect all button")
 }
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -35,6 +35,8 @@ extension SPNoteListViewController {
         tableView.register(SPNoteTableViewCell.loadNib(), forCellReuseIdentifier: SPNoteTableViewCell.reuseIdentifier)
         tableView.register(SPTagTableViewCell.loadNib(), forCellReuseIdentifier: SPTagTableViewCell.reuseIdentifier)
         tableView.register(SPSectionHeaderView.self, forHeaderFooterViewReuseIdentifier: SPSectionHeaderView.reuseIdentifier)
+
+        tableView.allowsMultipleSelectionDuringEditing = true
     }
 
     /// Sets up the Results Controller
@@ -495,16 +497,18 @@ extension SPNoteListViewController: UITableViewDelegate {
     }
 
     public func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
-        selectedNote = nil
+        if !isEditing {
+            selectedNote = nil
 
-        switch notesListController.object(at: indexPath) {
-        case let note as Note:
-            SPRatingsHelper.sharedInstance()?.incrementSignificantEvent()
-            open(note, animated: true)
-        case let tag as Tag:
-            refreshSearchText(appendFilterFor: tag)
-        default:
-            break
+            switch notesListController.object(at: indexPath) {
+            case let note as Note:
+                SPRatingsHelper.sharedInstance()?.incrementSignificantEvent()
+                open(note, animated: true)
+            case let tag as Tag:
+                refreshSearchText(appendFilterFor: tag)
+            default:
+                break
+            }
         }
     }
 
@@ -569,6 +573,8 @@ private extension SPNoteListViewController {
         cell.keywordsTintColor = .simplenoteTintColor
 
         cell.prefixText = prefixText(for: note)
+
+        cell.multiSelectTintColor = .simplenoteNotePinStatusImageColor
 
         cell.refreshAttributedStrings()
 
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -405,6 +405,11 @@ extension SPNoteListViewController {
         setListViewTitle()
         refreshSelectAllLabels()
     }
+
+    @objc
+    func refreshEditButtonTitle() {
+        editButtonItem.title = isEditing ? Localization.cancelTitle : Localization.editTitle
+    }
 }
 
 
@@ -611,14 +616,7 @@ extension SPNoteListViewController: UITableViewDelegate {
         guard let cell = cell as? SPNoteTableViewCell else {
             return
         }
-        var insets = SPNoteTableViewCell.separatorInsets
-        insets.left -= cell.layoutMargins.left
-
-        if isEditing {
-            insets.left += cell.checkboxContainingView.frame.width
-        }
-
-        cell.separatorInset = insets
+        cell.refreshEdgeInsets(editing: isEditing)
     }
 }
 
@@ -710,7 +708,8 @@ extension SPNoteListViewController {
 
     open override func setEditing(_ editing: Bool, animated: Bool) {
         super.setEditing(editing, animated: animated)
-
+        refreshEditButtonTitle()
+        
         tableView.deselectAllRows(inSection: 0, animated: false)
         updateNavigationBar()
 
@@ -1143,4 +1142,8 @@ private enum Localization {
     }
 
     static let selectAllAccessibilityHint = NSLocalizedString("Tap button to select or deselect all notes", comment: "Accessibility hint for the select/deselect all button")
+
+    static let editTitle = NSLocalizedString("Edit", comment: "Edit button title")
+
+    static let cancelTitle = NSLocalizedString("Cancel", comment: "Cancel button title")
 }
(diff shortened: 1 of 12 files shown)


Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -344,6 +346,39 @@ extension SPNoteListViewController {
...
+            return
+        }
+
+        var selectedNotes: [Note] = []
+
+        for indexPath in selectedIndicies {
+            if let note = notesListController.object(at: indexPath) as? Note {
+                selectedNotes.append(note)
+            }
+        }
```
jleandroperez: Use the force @charliescheer . Use the force! :yoda:

(Huh we could use a Yoda emoji!!)

```
    func trashSelectedNotes() {
        guard let notes = tableView.indexPathsForSelectedRows?.compactMap({ notesListController.object(at: $0) as? Note }) else {
            return
        }

        delete(notes: notes)
        setEditing(false, animated: true)
    }
```

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -622,6 +679,29 @@ extension SPNoteListViewController {
...
+
+    open override func setEditing(_ editing: Bool, animated: Bool) {
+        super.setEditing(editing, animated: animated)
+
+        // Reloading data ensures that no cells are selected and the current data is up to date
+        let rows = tableView.indexPathsForVisibleRows
+        tableView.reloadRows(at: rows!, with: .automatic)
+
+        // Dispatches to asyncAfter because reloadData will override the animations if not complete
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1, execute: {
```
jleandroperez: We'll prooooobably need to check if there's a way to do this without a Dispatch Async.

Generally speaking, we only use this kind of scenario when working around an SDK bug

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -622,6 +679,29 @@ extension SPNoteListViewController {
             notesListController.object(at: indexPath) as? Note
         }
     }
+
+    open override func setEditing(_ editing: Bool, animated: Bool) {
+        super.setEditing(editing, animated: animated)
+
+        // Reloading data ensures that no cells are selected and the current data is up to date
+        let rows = tableView.indexPathsForVisibleRows
+        tableView.reloadRows(at: rows!, with: .automatic)
```
jleandroperez: Can we drop the IUO?

jleandroperez (commented):
Sending few nipticky comments!

Please LMK whenever it's ready for review. Thank you sir!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
charliescheer: nice!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
charliescheer: I had poked around for some other ways to get around this, but I didn't find any while also reloading the rows.  This was in there because if we batch deleted some rows and then returned to edit mode, the cells that were deleted are the only ones that would enter edit mode.  

The fix was to reload the cells before changing edit mode, and that overrides the animations.  But I will take a look and see what I can find

Simplenote/Classes/SPNoteListViewController+Extensions.swift
charliescheer: yup

charliescheer:
@jleandroperez  this is ready to be reviewed now.  I have added a few steps in the testing after the updates I made today

eshurakov:
Stopping by to say that it's also good to test that keyboard navigation still works as expected when not in edit mode and doesn't break anything when in edit mode.

Would be great to support keyboard navigation in edit mode, but it may be tricky to show both selected notes and keyboard selection at the same time.

charliescheer:
Nice catch @eshurakov I did a bit of testing on the keyboard shortcuts and realized that one of the changes I made to the cells to get the multi select styling to work removed the highlighting for cells when not in edit mode.  This made using the keyboard shortcuts to navigate notes pretty much impossible.  I fixed that now.

As for using the keyboard shortcuts in edit mode, currently that works but sort of weirdly.  Hitting enter does nothing, and as you use the arrow keys it moves the selected note to the next one 


I'm not sure what using the keyboard shortcuts in edit mode would be like.  @SylvesterWilmott do you have any thoughts on how that interaction might work?

SylvesterWilmott:
Looking good so far and seems to be working well.

**Behaviour**

- I noticed that there is a delay between pressing "Edit" and revealing the multiple select UI, is it possible to eliminate this delay? Can we use the 'system' transition for example the edit mode in the mail app?
- There is a small bug when swiping a note list row. If you swipe then tap outside of the row then the selection checkbox is shown momentarily.

**UI**
- Can we change the checkbox icons to 24pt
- When in edit mode, I think the rows don't move over far enough. This could be due to the above though so we can re-asses once the icon sizes are updated.
- When a note is trashed, it seems there is a crossfade transition. It would be great if we can use the typical transition used when deleting notes from the swipe menu in this spot as well.
- On iPad, the "Edit" button can be shown in the top right of the note list UI - to the left of the new note icon.
- Can we try applying the "pressed" background colour to the selected note rows?

>I'm not sure what using the keyboard shortcuts in edit mode would be like.

I tested this interaction in some of the Apple apps, in some instances it behaves really strangely like in the mail app. In the Apple notes it behaves similarly to what is shown in the gif you shared. We wont be allowing edit mode to be accessible by shortcut (yet) so I don't feel much pressure to build a keyboard experience for it at this stage. What is shown in the gif is what I would expect.

charliescheer:
@SylvesterWilmott this is ready to be looked at again.  Check out some specific responses below:

>I noticed that there is a delay between pressing "Edit" and revealing the multiple select UI, is it possible to eliminate this delay? Can we use the 'system' transition for example the edit mode in the mail app?

Yes we can fix that.  I had a delay in there for animations reasons, but I think I have fixed that by fixing the next thing

>There is a small bug when swiping a note list row. If you swipe then tap outside of the row then the selection checkbox is shown momentarily.

I found this bug yesterday and fixed it but didn't push the changes at the end of the day.  If you check it again now that has been fixed

>Can we change the checkbox icons to 24pt
When in edit mode, I think the rows don't move over far enough. This could be due to the above though so we can re-asses once the icon sizes are updated.

Sure thing.  I made them bigger:


>Can we try applying the "pressed" background colour to the selected note rows?

Do you mean like this?  That is easy enough to do if we want to make that change.  In this most recent version I have left it so that the select highlighting appears, if you want to go back to not highlighting we can restore it np


>When a note is trashed, it seems there is a crossfade transition. It would be great if we can use the typical transition used when deleting notes from the swipe menu in this spot as well.
On iPad, the "Edit" button can be shown in the top right of the note list UI - to the left of the new note icon.

I think both of these issues were fixed with some changes that I made yesterday but didn't push.  I will dig around and see if I can locate them, but check it out again and let me know what you think!

SylvesterWilmott:
Thanks @charliescheer 

All of the issues are fixed! I think by enabling the selected background, it caused the separators to remain in the same place between normal > edit mode. Can we try to let the separators always align with the text?

After using this a few times, I think we should change the "Done" label to "Cancel". Since we auto-dismiss the UI if you delete something, the only real scenarios for using the "Done" button are when you want to manually dismiss the checkboxes so in this case I think "Cancel" makes more sense.

charliescheer:
>All of the issues are fixed! I think by enabling the selected background, it caused the separators to remain in the same place between normal > edit mode. Can we try to let the separators always align with the text?

That was me.  I had fixed a different issue in a way that broke my fix for the separators.  It is fixed now

> I think we should change the "Done" label to "Cancel". Since we auto-dismiss the UI if you delete something, the only real scenarios for using the "Done" button are when you want to manually dismiss the checkboxes so in this case I think "Cancel" makes more sense.

Sure thing.  Made that change as well.

charliescheer:
@SylvesterWilmott there should be a new installable build for the above changes if you want to check it out :)

SylvesterWilmott (approved):
Looks great, works as expected. Green light from me. :shipit:

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -255,6 +257,20 @@ extension SPNoteListViewController {
         updatePlaceholderPosition()
     }
 
+    func refreshSelectAllLabels() {
+        var deselect = false
+        let numberOfSelectedRows = tableView.indexPathsForSelectedRows?.count ?? 0
+
+        if notesListController.numberOfObjects == numberOfSelectedRows {
+            deselect = true
+        }
```
jleandroperez:
```
let deselect = notesListController.numberOfObjects == numberOfSelectedRows
```

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -344,6 +360,56 @@ extension SPNoteListViewController {
...
+            return
+        }
+        let flexibleSpace = UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil)
+        let actionButton = isEditing ? trashButton : addButton
+        setToolbarItems([flexibleSpace, actionButton], animated: true)
+    }
+
+    @objc
+    func selectAllWasTapped() {
+
```
jleandroperez: Nipticky: Extra space

Simplenote/Classes/SPNoteTableViewCell.swift
```diff
@@ -23,6 +23,12 @@ class SPNoteTableViewCell: UITableViewCell {
     ///
     @IBOutlet private var accessoryRightImageView: UIImageView!
 
+    /// Multi Select Checkbox ImageView
+    ///
+    @IBOutlet private weak var multiSelectCheckbox: UIImageView!
+
+    @IBOutlet weak var checkboxContainingView: UIView!
```
jleandroperez: Nipticky: private!

Simplenote/Classes/SPNoteTableViewCell.swift
```diff
@@ -200,6 +232,24 @@ class SPNoteTableViewCell: UITableViewCell {
 
         bodyLabel.attributedText = bodyString
     }
+
+    func setMultiSelectEditing(_ editing: Bool) {
+
```
jleandroperez: Nipticky: Extra space

Simplenote/Classes/UITableView+Simplenote.swift
```diff
@@ -124,6 +124,37 @@ extension UITableView {
...
+        }
+    }
+
+    func deselectAllRows(inSection section: Int, animated: Bool) {
+        forEachIndexPathIn(section: section) { indexPath in
+            deselectRow(at: indexPath, animated: animated)
+        }
+    }
+
+    func forEachIndexPathIn(section: Int, doAction action: (IndexPath)-> Void) {
```
jleandroperez:
```suggestion
    func forEachIndexPath(in section: Int, doAction action: (IndexPath)-> Void) {
```

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -344,6 +360,56 @@ extension SPNoteListViewController {
...
+        if notesListController.numberOfObjects == tableView.indexPathsForSelectedRows?.count {
+            tableView.deselectAllRows(inSection: 0, animated: false)
```
jleandroperez: Please bear in mind you can use `.zero` everywhere there's a `0`.

Yeah it's definitely the same, buuut feels a whole lot less like magic numbers 🔥

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -622,6 +705,40 @@ extension SPNoteListViewController {
             notesListController.object(at: indexPath) as? Note
         }
     }
+
+    open override func setEditing(_ editing: Bool, animated: Bool) {
+        super.setEditing(editing, animated: animated)
+        refreshEditButtonTitle()
+
```
jleandroperez: I'm getting a `Trailing Space Violation` at this spot. No local changes. Mind confirming?

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -622,6 +705,40 @@ extension SPNoteListViewController {
...
+    }
+
+    private func setListViewTitle() {
+        if isEditing {
+            let count = tableView.indexPathsForSelectedRows?.count ?? 0
+            title = count > 0 ? Localization.selectedTitle(with: count) : notesListController.filter.title
+            return
+        }
+
+        title = notesListController.filter.title
```
jleandroperez:
```
        title = {
            guard isEditing else {
                return notesListController.filter.title
            }

            let count = tableView.indexPathsForSelectedRows?.count ?? .zero
            return count > 0 ? Localization.selectedTitle(with: count) : notesListController.filter.title
        }()
```

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -622,6 +705,40 @@ extension SPNoteListViewController {
...
     }
+
+    open override func setEditing(_ editing: Bool, animated: Bool) {
+        super.setEditing(editing, animated: animated)
+        refreshEditButtonTitle()
+        
+        tableView.deselectAllRows(inSection: 0, animated: false)
+        updateNavigationBar()
+
+        tableView.allowsMultipleSelection = editing
```
jleandroperez: I **think** we can drop this? (during initialization it's not explicitly set).

When we're not in edition mode, pressing over any row causes the Editor to get pushed, and we immediately deselect the selected row.

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -622,6 +705,40 @@ extension SPNoteListViewController {
...
+        toggleMultiSelectEditingTo(editing)
+
+        configureNavigationToolbarButton()
+        setListViewTitle()
+    }
+
+    func toggleMultiSelectEditingTo(_ editing: Bool) {
+        guard let noteCells = tableView.visibleCells as? [SPNoteTableViewCell] else {
```
jleandroperez: Beware this could be a problem, under certain conditions. `visibleCells` returns only the reusable cells that are onscreen, which **not necessarily** will always be all of the cells in the table. ([Ref. here](https://stackoverflow.com/questions/39458544/uitableview-visiblecells-is-not-returning-all-cells)).

I've tested this PR in a large account (+500 notes), and appears to be working fine. But to be safe, we should consider:

1. Setting the `multiEdit` toggle in `configureCell`
2. Switch to `reloadRowsAtIndexPaths` (this could be an API in the UITableView extension)

WDYT?

Simplenote/Classes/SPNoteTableViewCell.swift
```diff
@@ -200,6 +232,24 @@ class SPNoteTableViewCell: UITableViewCell {
...
+        UIView.animate(
+            withDuration: UIKitConstants.animationQuickDuration,
+            animations: {
+                self.refreshEdgeInsets(editing: editing)
+                self.checkboxContainingView.isHidden = !editing
+                self.checkboxContainingView.superview?.layoutIfNeeded()
```
jleandroperez:
```suggestion
                self.contentView.layoutIfNeeded()
```

jleandroperez (commented):
@charliescheer I LOVE this PR!!!

Sending you few nipticky notes. Plus: I managed to get the TableView into a broken state, see the video below. I pressed, repeatedly, `Edit` / `Cancel`.

https://user-images.githubusercontent.com/1195260/117506320-c7913880-af5b-11eb-8e60-eeece3dca7ea.MP4

Simplenote/Classes/SPNoteTableViewCell.swift
jleandroperez: Note: `contentVIew != .superview` though!!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
charliescheer: I had it setup that way before, but the problem with reloading all of the cells is that is changes the animation.  As opposed to the checkboxes animating in from the side, the entire cell is replaced which could be fine, but isn't what I would expect.  

I wonder if a compromise situation could be made where in configure we set the multi edit mode for the cell, but still animate the currently visible cells.  That should cover them all as new ones will dequeue in the correct mode while still giving us the slide out animation.

WDYT?

charliescheer:
@jleandroperez thanks for taking a look.  This is ready to be looked at again!

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -344,6 +354,55 @@ extension SPNoteListViewController {
...
+    func selectAllWasTapped() {
+        if notesListController.numberOfObjects == tableView.indexPathsForSelectedRows?.count {
+            tableView.deselectAllRows(inSection: .zero, animated: false)
+            refreshNavigationBarLabels()
+
+            return
+        }
+
+        tableView.selectAllRows(inSection: 0, animated: false)
+        refreshNavigationBarLabels()
```
jleandroperez, line 392: Apologies I've missed this! there's a bit of code duplication:

```
if notesListController.numberOfObjects == tableView.indexPathsForSelectedRows?.count {
   tableView.deselectAllRows(inSection: .zero, animated: false)
} else {
   tableView.selectAllRows(inSection: .zero, animated: false)
}

refreshNavigationBarLabels()
```

charliescheer:
@SylvesterWilmott this should be ready to review again now that we have moved to the system standard selection mechanism, if you would like to check it out.

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -535,9 +606,18 @@ extension SPNoteListViewController: UITableViewDelegate {
...
+        }
+
         var insets = SPNoteTableViewCell.separatorInsets
         insets.left -= cell.layoutMargins.left
+
         cell.separatorInset = insets
+
+        if indexPath.row == notesListController.numberOfObjects - 1 {
+            cell.shouldDisplayBottomSeparator = false
+        }
```
jleandroperez: Because of cell reuse, a given Cell might have the `shouldDisplayBottomSeparator` set to `false`, and that value is never set back to true.

Always always consider doing something like:

```
cell.shouldDisplayBottomSeparator = indexPath.row < notesListController.numberOfObjects - 1 
```

**Furthermore** the AppStore build doesn't hide the cell separator for the last row. I think you can drop this snippet?

Simplenote/SeparatorsView.swift
```diff
@@ -0,0 +1,131 @@
...
+        super.init(coder: aDecoder)!
+        setupView()
+    }
+
+    open override func draw(_ rect: CGRect) {
+        super.draw(rect)
+
+        let scale = UIScreen.main.scale
+        let ctx = UIGraphicsGetCurrentContext()
+        ctx!.clear(rect)
```
jleandroperez: I know this was part of the original code (and I have NO IDEA why there's an IUO there).

Can we please `guard let else`, and drop the force unwraps?

Simplenote/Classes/SPNoteTableViewCell.swift
```diff
@@ -226,6 +246,21 @@ private extension SPNoteTableViewCell {
         bodyLabel.numberOfLines = Style.maximumNumberOfBodyLines
         bodyLabel.lineBreakMode = .byWordWrapping
     }
+
+    private func setupCellSeparatorView() {
+        separatorsView = SeparatorsView()
```
jleandroperez: You can initialize this one directly at the property level, and you can get away without the `var` and the IUO

```
private let separatorsView = SeparatorsView()
```

Simplenote/Classes/SPNoteTableViewCell.swift
```diff
@@ -226,6 +246,21 @@ private extension SPNoteTableViewCell {
...
+        separatorsView.translatesAutoresizingMaskIntoConstraints = false
+        contentView.addSubview(separatorsView)
+        NSLayoutConstraint.activate([
+            separatorsView.bottomAnchor.constraint(equalTo: contentView.bottomAnchor),
+            separatorsView.trailingAnchor.constraint(equalTo: contentView.trailingAnchor),
+            separatorsView.topAnchor.constraint(equalTo: contentView.topAnchor),
+            separatorsView.leadingAnchor.constraint(equalTo: textAreaStackView.leadingAnchor)
```
jleandroperez, line 257: Beautiful!!

jleandroperez (commented):
Thank you for all the work Charlie!! Few details I've noticed:

### Issue 1: Incorrect Deselect Label
1. Press **Edit** (top right)
2. Press **Select All** (top left)
3. Press **Cancel** (top right)
4. Press **Edit** again

As a result, no rows will be selected, but the top left action will read `Deselect All`

### Issue 2: Peek + Pop
1. Press **Edit** (top right)
2. While in Select Mode, long press over any cell

As a result, you'll get the Note Preview onscreen. This allows you to access the Editor (or trash the Note, among others!)

Simplenote/Classes/SPNoteTableViewCell.swift
charliescheer: Thanks 😄

Simplenote/Classes/SPNoteListViewController+Extensions.swift
charliescheer: 👍

Simplenote/Classes/SPNoteListViewController+Extensions.swift
charliescheer: Asked @SylvesterWilmott if we wanted to have the last separator and he said that we shouldn't currently have it and that I could drop it if I wanted.  My sense was that with the new note bar, the two separators join weirdly and it would be better without it.

Made this change

charliescheer:
@jleandroperez thanks for taking a look.  I made the changes that you suggested, and I found/fixed both bugs.  This is ready to be looked at a again

SylvesterWilmott:
Looks great! Thanks @charliescheer. I had noticed a couple of issues in the build I installed a couple of days ago but those issues seem to have gone in the latest commits.

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -622,6 +697,38 @@ extension SPNoteListViewController {
...
         }
     }
+
+    open override func setEditing(_ editing: Bool, animated: Bool) {
+        // tableView could be in edit mode already if a single cell is displaying a slide out menu
+        // Check if tableView and NoteListVC have the same isEditing value
+        // If not set them to the same before continuing
+        if tableView.isEditing != isEditing {
+            tableView.setEditing(isEditing, animated: true)
+        }
```
jleandroperez: Q: `tableView.setEditing()` is done, again, in L710.

Do we need to run **both?**

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -622,6 +697,38 @@ extension SPNoteListViewController {
...
+
+        tableView.deselectAllRows(inSection: .zero, animated: false)
+        updateNavigationBar()
+
+        configureNavigationToolbarButton()
+        setListViewTitle()
+    }
+
+
+    private func setListViewTitle() {
```
jleandroperez:
```suggestion
    private func refreshListViewTitle() {
```

Since `setListViewTitle` implies the method is a **setter** (and you would expect a parameter!)

jleandroperez (commented):
Looking super sharp!!!

I'm afraid I found a glitch on iPad devices:

1. Launch the app (iPad Pro 9.7 / iOS 14.5)
2. Press over any note in your list
3. Press back: Notes List should be onscreen
4. Press the top left button: Tag List should be onscreen
5. Reveal the Notes list again

As a result, the top right **Add new Note** button will be gone. 
**Confirmed** this does not happen in `develop`.

---

Plus: Multi-Selection is iPhone only right?

charliescheer:
>I'm afraid I found a glitch on iPad devices:

Nice catch.  Solved.  Ready to be checked again @jleandroperez

jleandroperez (approved):
Beautiful work @charliescheer !!!

:shipit:

PR #1268, opened 2021-05-03
Fixed an issue where tags would not be deleted on trashed notes

### LINKED ISSUE #1234: Tag deletion from tags list should affect the notes in `Trash` too
Same issue for macOS [here](https://github.com/Automattic/simplenote-macos/issues/934)

### Expected
If we delete a tag from the tags list, the tag should also be removed from notes in Trash (happens in Electron and Android)

### Observed
https://user-images.githubusercontent.com/73365754/114040488-54c56c80-988c-11eb-939b-187b977e4a7f.MP4

### Reproduced
1. Open a note with no tags
2. Add a new unique tag to the note
3. Move note to trash
4. Delete the tag mentioned in step 2
5. Restore the note ~> note has a tag, but tag does not appear in tags list

-|-|-|-
Apple|iPhone XR|14.4.1|4.34

charliescheer:
### Fix
This PR fixes #1234 

Currently if you delete a note that has a tag on it, and then later delete that tag, the trashed note will continue being associated with the deleted tag.  If later you restore that note to the note list it will still have reference to the tag, but the tag will not appear in the tag list.

With this PR, when removing a tag, the tag will be removed from all notes including notes that are in the trash.

### Test
1. Go to a note and give it a tag.
2. Trash that same note
3. Go to the tag list and delete the tag
4. Restore the deleted note
5. If you edit that note the tag that was removed will not appear in the notes tag list.

### Review

### DIFF
--- a/Simplenote/Classes/SPObjectManager+Simplenote.swift
+++ b/Simplenote/Classes/SPObjectManager+Simplenote.swift
@@ -25,7 +25,7 @@ extension SPObjectManager {
     }
 
     @objc
-    func notesWithTag(_ tag: Tag?) -> [Note] {
+    func notesWithTag(_ tag: Tag?, includeDeleted: Bool) -> [Note] {
         guard let tagName = tag?.name else {
             return []
         }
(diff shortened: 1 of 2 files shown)


jleandroperez (approved):
I was in the neighborhood, aaaaand KINDA verified this PR before realizing I wasn't supposed to review it 😂 

:shipit: on my end!! (Hope you don't mind Evgeny!!)

mokagio:

It will be part of 4.36.

PR #1271, opened 2021-05-04
Parameterising the test account for UI Tests

pachlava:
This is a preliminary PR before a PR that will attempt to solve the account race condition.
Dividing it in two in order to keep the noise down for the next PR.

### Changes:

1. `run_ui_tests` lane now accepts a `test_account` parameter, and passes it to `SimplenoteUITests` scheme
2. `SimplenoteUITests` scheme contains an environment variable for test account
3. `EmailLogin.logIn()` method can now be used with no parameters. In such case, it will check if the test account was passed via environment variable. Otherwise, it will use the 'default' credentials that were used previously.

### Test
1. Run `bundle exec fastlane run_ui_tests` (alternatively, run the `SimplenoteUISmokeTestsTrash` tests from Xcode) and make sure they are executed using the 'default' credentials
2. Run `bundle exec fastlane run_ui_tests test_account:"<test_account_email>"` and make sure same tests are executed using the email passed as a parameter.

### Review

### Release

### DIFF
--- a/SimplenoteUITests/EmailLogin.swift
+++ b/SimplenoteUITests/EmailLogin.swift
@@ -14,7 +14,18 @@ class EmailLogin {
         backButton.tap()
     }
 
-    class func logIn(email: String, password: String) {
+    class func logIn() {
+        var testAccount = ProcessInfo.processInfo.environment["UI_TEST_ACCOUNT"]!
+
+        // Use 'default' account if test account was not passed via environment variable
+        if testAccount.isEmpty {
+            testAccount = testDataEmail
+        }
+
+        EmailLogin.logIn(testAccount, testDataPassword)
+    }
+
+    class func logIn(_ email: String, _ password: String) {
         enterEmail(enteredValue: email)
         enterPassword(enteredValue: password)
         app.buttons[UID.Button.logIn].tap()
--- a/SimplenoteUITests/EmailLogin.swift
+++ b/SimplenoteUITests/EmailLogin.swift
@@ -15,11 +15,15 @@ class EmailLogin {
     }
 
     class func logIn() {
-        var testAccount = ProcessInfo.processInfo.environment["UI_TEST_ACCOUNT"]!
+        let testAccountKey = "UI_TEST_ACCOUNT"
+        let testAccount: String
 
-        // Use 'default' account if test account was not passed via environment variable
-        if testAccount.isEmpty {
-            testAccount = testDataEmail
+        switch ProcessInfo.processInfo.environment[testAccountKey] {
+        case .none:
+            fatalError("Expected \(testAccountKey) environment variable to be defined in the scheme")
+        case .some(let value):
+            // Use 'default' account if test account was not passed via environment variable
+            testAccount = value.isEmpty ? testDataEmail : value
         }
 
         EmailLogin.logIn(testAccount, testDataPassword)
--- a/SimplenoteUITests/EmailLogin.swift
+++ b/SimplenoteUITests/EmailLogin.swift
@@ -26,10 +26,10 @@ class EmailLogin {
             testAccount = value.isEmpty ? testDataEmail : value
         }
 
-        EmailLogin.logIn(testAccount, testDataPassword)
+        EmailLogin.logIn(email: testAccount, password: testDataPassword)
     }
 
-    class func logIn(_ email: String, _ password: String) {
+    class func logIn(email: String, password: String) {
         enterEmail(enteredValue: email)
         enterPassword(enteredValue: password)
         app.buttons[UID.Button.logIn].tap()
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -713,7 +713,7 @@ end
         prelaunch_simulator: true,
         reset_simulator: true,
         only_testing: "SimplenoteUITests/SimplenoteUISmokeTestsTrash",
-        xcargs: "UI_TEST_ACCOUNT='#{options[:test_account] || ""}'"
+        xcargs: { UI_TEST_ACCOUNT: #{options[:test_account] || '' }
       )
     rescue => ex
        UI.error(ex)
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -713,7 +713,7 @@ end
         prelaunch_simulator: true,
         reset_simulator: true,
         only_testing: "SimplenoteUITests/SimplenoteUISmokeTestsTrash",
-        xcargs: { UI_TEST_ACCOUNT: #{options[:test_account] || '' }
+        xcargs: { UI_TEST_ACCOUNT: options[:test_account] || '' }
       )
     rescue => ex
        UI.error(ex)
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -705,19 +705,15 @@ end
   #####################################################################################
   desc "Run UI tests"
   lane :run_ui_tests do | options |
-    begin
-      scan(
-        workspace: workspace_name,
-        scheme: UI_TESTS_SCHEME,
-        device: options[:device] || "iPhone 12",
-        prelaunch_simulator: true,
-        reset_simulator: true,
-        only_testing: "SimplenoteUITests/SimplenoteUISmokeTestsTrash",
-        xcargs: { UI_TEST_ACCOUNT: options[:test_account] || '' }
-      )
-    rescue => ex
-       UI.error(ex)
-    end
+    scan(
+      workspace: workspace_name,
+      scheme: UI_TESTS_SCHEME,
+      device: options[:device] || "iPhone 12",
+      prelaunch_simulator: true,
+      reset_simulator: true,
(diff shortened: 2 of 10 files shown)


SimplenoteUITests/EmailLogin.swift
```diff
@@ -14,7 +14,18 @@ class EmailLogin {
         backButton.tap()
     }
 
-    class func logIn(email: String, password: String) {
+    class func logIn() {
+        var testAccount = ProcessInfo.processInfo.environment["UI_TEST_ACCOUNT"]!
```
pachlava: It seems like using force unwrap is not a problem here, because the environment variable exists in the scheme for sure 🤔

fastlane/Fastfile
```diff
@@ -698,21 +698,26 @@ end
...
-  )
+    begin
+      scan(
+        workspace: workspace_name,
+        scheme: UI_TESTS_SCHEME,
+        device: options[:device] || "iPhone 12",
+        prelaunch_simulator: true,
+        reset_simulator: true,
+        only_testing: "SimplenoteUITests/SimplenoteUISmokeTestsTrash",
+        xcargs: "UI_TEST_ACCOUNT='#{options[:test_account] || ""}'"
```
pachlava: Currently, passing the account email as it is. In future, it can be changed to passing a test account ordinal number only instead.

fastlane/Fastfile
```diff
@@ -698,21 +698,26 @@ end
...
-    only_testing: "SimplenoteUITests/SimplenoteUISmokeTestsLogin"
-  )
+    begin
+      scan(
+        workspace: workspace_name,
+        scheme: UI_TESTS_SCHEME,
+        device: options[:device] || "iPhone 12",
+        prelaunch_simulator: true,
+        reset_simulator: true,
+        only_testing: "SimplenoteUITests/SimplenoteUISmokeTestsTrash",
```
pachlava: Changed the tests selection to `Trash` in order to make it easier to see which account is used during tests.

SimplenoteUITests/EmailLogin.swift
```diff
@@ -14,7 +14,18 @@ class EmailLogin {
...
+
+        // Use 'default' account if test account was not passed via environment variable
+        if testAccount.isEmpty {
+            testAccount = testDataEmail
+        }
+
+        EmailLogin.logIn(testAccount, testDataPassword)
```
pachlava: To minimize the changes not directly related to the main logic that will come next, the tests currently assume that all test credentials share the same password. Because of this, we only have to pass the email at the moment.

fastlane/Fastfile
mokagio: I recently [learned](https://github.com/Automattic/simplenote-macos/pull/870#discussion_r578376864) that we can pass an hash to `xcargs`.

```suggestion
        xcargs: { UI_TEST_ACCOUNT: #{options[:test_account] || '' }
```

_I wonder what would happen if we also removed the `|| ''`_ 🤔

fastlane/Fastfile
```diff
@@ -698,21 +698,26 @@ end
...
+        xcargs: "UI_TEST_ACCOUNT='#{options[:test_account] || ""}'"
+      )
+    rescue => ex
+       UI.error(ex)
+    end
```
mokagio: Is there a particular purpose for wrapping the call to `scan` in a `begin - rescue` block? I thought `scan` already captured and reported errors.

SimplenoteUITests/EmailLogin.swift
```diff
@@ -14,7 +14,18 @@ class EmailLogin {
...
+    }
+
+    class func logIn(_ email: String, _ password: String) {
```
mokagio: Nitpick:

```suggestion
    class func logIn(email: String, password: String) {
```

The [Swift API Guidelines](https://swift.org/documentation/api-design-guidelines/#weak-type-information) recommend to precede each weakly typed parameter, such as `Any`, `Int`, or `String, with a noun describing its role:

screen shot 2019-02-06 at 1 43 58 pm

It's true that Xcode's code completion would suggest that the first argument should be email and the second password, but I find having them explicit helps with readability.

SimplenoteUITests/EmailLogin.swift
mokagio: For some reason, opening the project in Xcode removed the scheme change and the test crashed here.

Even without the Xcode quirk I experienced, I find it's best not to make assumptions on input parameters in the code. This one in particular lives in a scheme in the form of a string. There is no compile-time check that would tell a developer that removed it that it was expected here.

I would suggest explicitly handling the nil scenario. Something like

```swift
let testAccountKey = "UI_TEST_ACCOUNT"
let testAccount: String
switch ProcessInfo.processInfo.environment[testAccountKey] {
case .none:
    fatalError("Expected \(testAccountKey) environment variable to be defined in the scheme")
case .some(let value):
    // Use 'default' account if test account was not passed via environment variable
    testAccount = value.isEmpty ? testDataEmail : value
}
```

This way, it something goes wrong, like it happened to me, the error message in the console will be more informative.
Screen Shot 2021-05-05 at 1 06 35 pm

fastlane/Fastfile
mokagio: Update: while testing the `fastlane` invocation, I passed a wrong email on purpose. This code did what it was told to do, that is, captured and reported the error, but did not fail the lane.

image

From my point of view, this is not a desirable behavior because, if tests fail, we want to know about it. Am I missing some context?

Thanks 😄

mokagio (commented):
- Tested it from Xcode: it fell back to the default value and the tests passed
- Tested it with Fastlane invocation and a working email: it successfully used the email and the tests passed
- Tested it with Fastlane invocation and a non-working email: it successfully used the email and the tests failed because they could not login with that email and password combo, as expected.

Waiting for an answer to my question regarding the the `begin - rescue`. Other than that, this is good to go.

fastlane/Fastfile
pachlava: Thank you @mokagio ! Yes, there is really a reason behind this: in my next PR there will be another lane, that picks the free account, and calls `run_ui_tests` lane and passes the user email to it. Regardless of tests execution status, I need to make sure that the test account is always freed at the end. Without `begin - rescue` block, if the there's a fail, the execution will not cruise back to the `pick_test_account_and_run_ui_tests` lane in order to mark the account as free.

This might be not the only way to achieve this, but the one that I found working. If this goes against the best practices, I can remove this, and we can discuss possible options in the next PR?

SimplenoteUITests/EmailLogin.swift
pachlava: Thank you! I agree, another reason of doing it the way you suggest is that tests might be executed using the other scheme. I used you code 🙇

pachlava: Thank you! I agree, another reason of doing it the way you suggest is that tests might be executed using the other scheme. I used you code 🙇

fastlane/Fastfile
pachlava: > I recently learned that we can pass an hash to xcargs.

Thank you 🙂 Generally speaking, I consider myself extremely lucky to be able to work with people like you and Olivier, and learn. It amplifies my impostor syndrome a bit, though 😅 

> I wonder what would happen if we also removed the `|| ''`

In my case, this happened:
```
[14:18:44]: You passed invalid parameters to 'scan'.
[14:18:44]: Check out the error below and available options by running `fastlane action scan`
[14:18:44]: undefined method `shellescape' for nil:NilClass
```

SimplenoteUITests/EmailLogin.swift
pachlava: This is nice to know, thanks! I previously thought that using nameless parameters is something dictated by gut feeling rather than having a more well-define rule, like not using it for weakly typed parameters.

pachlava:
@mokagio thanks for feedback! All you suggestions for changes were applied, and the reply to `begin - rescue` question is [here](https://github.com/Automattic/simplenote-ios/pull/1271#discussion_r626391386). Please let me know what do you think 🙂

fastlane/Fastfile
mokagio: Thank you for adding extra context. 

I see what you mean: we need a way to run a "tear down" kind of process after the lane execution, regardless of whether it succeeds.

I think we could use the [`after_each` API](http://docs.fastlane.tools/advanced/lanes/#before_each-and-after_each-blocks) to achieve this without loosing the important failure information.

```ruby
after_each do |lane, options|
  free_test_account if lane == :run_ui_tests
end
```

mokagio (approved):
Hey @pachlava! Sorry I did review this yesterday but my submit didn't go through 😳 

:shipit:

pachlava:
```ruby
after_each do |lane, options|
  free_test_account if lane == :run_ui_tests
end
```

Oh perfect! I've read about `after_each`, but did not realize there's a way to let `after_each` know which lane was previously used. I've removed the `begin - rescue`. Thanks a lot Gio!

PR #1272, opened 2021-05-10
Lowercase signup emails before sending to server

### LINKED ISSUE #1252: Signup: Emails should be lowercased before sending to the backend
### Expected

If a user enters an email using some capital letters, it should be converted to lowercase on form submission.

### Observed

The app sends a mixed-case email to the backend for registration.

Further context in pbMoDN-2oI-p2

charliescheer:
### Fix
fixes #1252 

This PR changes the case on email addresses before they are sent to the server during signup.  Previously there was an issue where mixed case email addresses where causing accounts to be created immediately borked.  Setting the emails to lowercase on signup is part of fixing this problem.

There isn't much to see in this PR as the email case is being set just before being sent to the server.  As such to verify the changes I have created unit tests to confirm what is being sent.  While I was there I also wrote tests for the rest of SignupRemote

### Test
1. run the unit tests for SNiOS
2. Check the tests on SignupRemoteTests.swift, they should all pass and confirm the correct casing of the email address.

### Review

### DIFF
--- a/Simplenote/Classes/SignupRemote.swift
+++ b/Simplenote/Classes/SignupRemote.swift
@@ -8,8 +8,8 @@ class SignupRemote {
             switch (lhs, rhs) {
             case (.success, .success):
                 return true
-            case (.failure(let code1, let error1), .failure(let code2, let error2)):
-                return code1 == code2 && error1?.localizedDescription == error2?.localizedDescription
+            case (.failure(let code1, _), .failure(let code2, _)):
+                return code1 == code2
             default:
                 return false
             }
--- a/Simplenote/Classes/SignupRemote.swift
+++ b/Simplenote/Classes/SignupRemote.swift
@@ -3,7 +3,19 @@ import Foundation
 // MARK: - SignupRemote
 //
 class SignupRemote {
-    enum Result {
+    enum Result: Equatable {
+        static func == (lhs: SignupRemote.Result, rhs: SignupRemote.Result) -> Bool {
+            switch (lhs, rhs) {
+            case (.success, .success):
+                return true
+            case (.failure(let code1, let error1), .failure(let code2, let error2)):
+                return code1 == code2 && error1?.localizedDescription == error2?.localizedDescription
+            default:
+                return false
+            }
+
+        }
+
         case success
         case failure(_ statusCode: Int, _ error: Error?)
     }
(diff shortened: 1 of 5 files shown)


Simplenote/Classes/SignupRemote.swift
```diff
@@ -3,7 +3,19 @@ import Foundation
 // MARK: - SignupRemote
 //
 class SignupRemote {
-    enum Result {
+    enum Result: Equatable {
```
jleandroperez, line 6: Swift introduced its own [Result Type](https://developer.apple.com/documentation/swift/result) (we should eventually drop this one). 

But yeah... Error itself isn't equatable, so that prompts for another problem.

Simplenote/Classes/SignupRemote.swift
```diff
@@ -3,7 +3,19 @@ import Foundation
...
+        static func == (lhs: SignupRemote.Result, rhs: SignupRemote.Result) -> Bool {
+            switch (lhs, rhs) {
+            case (.success, .success):
+                return true
+            case (.failure(let code1, let error1), .failure(let code2, let error2)):
+                return code1 == code2 && error1?.localizedDescription == error2?.localizedDescription
```
jleandroperez: Not sure if comparing by `error.localizedDescription` is worth it TBH! (the code should just suffice)

SimplenoteTests/SignupRemoteTests.swift
```diff
@@ -0,0 +1,75 @@
...
+                           nil)
+
+        let expectation = self.expectation(description: "Verify is called")
+
+        signupRemote.signup(with: email) { (result) in
+            XCTAssertEqual(result, expectedResult)
+            expectation.fulfill()
+        }
+
+        waitForExpectations(timeout: 1, handler: nil)
```
jleandroperez:
```suggestion
        waitForExpectations(timeout: Constants.expectationTimeout, handler: nil)
```

SimplenoteTests/Network/MockURLSession.swift
```diff
@@ -4,10 +4,21 @@ import Foundation
...
     }
+
+    func decodedHtmlBodyInLastRequest<T: Decodable>(outputType: T.Type) throws -> T? {
+        guard let lastRequest = lastRequest,
+              let data = lastRequest.httpBody else {
+            return nil
+        }
+
+        return try JSONDecoder().decode(outputType, from: data)
+    }
```
jleandroperez: Suggestion: rather than having MockURLSession do this, you caaan move this into an URLRequest extension:

```
extension URLRequest {
   func decodeHtmlBody<T: Decodable>() throws -> T? {
         guard let data = httpBody else {
             return nil
         }

         return httpBody?.map {
             try JSONDecoder().decode(outputType, from: $0)
         }
     }
}
```

jleandroperez (approved):
You killed it!! Nice work!!

:shipit:

PR #1273, opened 2021-05-11
Merge finalized 4.35 into develop

mokagio:
Alongside the usual release finalization businesses:

- Update the copyright year in the `Deliverfile` to 2021. 🤔 _Thinking about it... Maybe I should have used code to compute it from the system time, so it's always up to date?_
- Opted out from `deliver` IAP checks. The other apps already do this, too.
- Updated the release toolkit


fastlane/Deliverfile
```diff
@@ -17,8 +17,18 @@ privacy_url({
...
+
+# We are using the API key to upload the metadata to ASC via Deliver, but, as
+# the tool reports:
+#
+# > [!] Precheck cannot check In-app purchases with the App Store Connect API
+# > Key (yet). Exclude In-app purchases from precheck, disable the precheck step
+# > in your build step, or use Apple ID login
+#
+# Simplenote iOS doesn't use IAPs, for the moment, so we can safely skip them.
+precheck_include_in_app_purchases false
```
AliSoftware, line 34: 👍

config/Version.Public.xcconfig
```diff
@@ -1,4 +1,4 @@
 VERSION_SHORT=4.35
 
 // Public long version example: VERSION_LONG=4.8.1.1
-VERSION_LONG=4.35.0.0
+VERSION_LONG=4.35.0.2
```
loremattei, line 4: What happened with `4.35.0.1`?

loremattei (approved):
:shipit:

fastlane/Deliverfile
```diff
@@ -17,8 +17,18 @@ privacy_url({
   'default' => 'https://simplenote.com/privacy/',
 })
 
-copyright('2020 Automattic')
+copyright('2021 Automattic')
```
AliSoftware, line 20: FWIW we also do it manually in other apps, so it's ok to keep it hardcoded, after all we only need to updated it once a year, so I am not sure it's worth it trying to auto-compute it as you pondered about in the PR description.

I'm surprised though that `fastlane deliver` didn't warn you about it sooner about it being outdated 🤔

AliSoftware (approved):

config/Version.Public.xcconfig
mokagio: Good point. I should have explained it in here.

After I finalized 4.35 and our automation generated `4.35.0.1` in 3b9f7f6f3e60b2307ab41e11b61fdd84a9f6c76e, I realized there was an issue with the localizations (#1266). Once then missing localizations came in, I had to generated a new build for ASC. I did not merge the `.1` build into `develop` because the branch at that point was in an "incomplete" state.

fastlane/Deliverfile
mokagio: Yeah... 🤔 It's not unlikely it did warn before and I didn't notice it in the noise of the output 😳 .

PR #1275, opened 2021-05-11
New lane and functions managing test accounts for UI tests

pachlava:
The PR adds a new lane and several supporting functions in `Fastfile` that manage the test accounts' availability, in order to prevent the account race condition when UI tests are executed on CI.

The information about test account state (free/taken) is stored at the online key-value [service](https://keyvalue.xyz/). I found no better means to have this information shared between different runs on CI.

The state is represented by a string of `TFFFF` format, where `T` is 'taken' and `F` is 'free'. In the given example, the account with index 0 is currently in use, so the new parallel run will occupy the next account. This solution provides easy scalability: the string can be lengthened to match the number of test accounts created for the purpose.

If this approach in overall is acceptable, the next step will be making the address of key-value and accounts credentials less exposed.

### What's new
Changes in `Fastfile`:
- New `pick_test_account_and_run_ui_tests` lane.
- Added `error` block, that is executed after an error occurs in a lane. This block is needed to make sure an occupied test account is always freed even if the tests fail (this causes an error in lane). `after_each` block is not executed in the case of error.
- A number of functions that take care of getting/setting test account state.

Changes in `config.yml`:
- `pick_test_account_and_run_ui_tests` lane runs instead of `run_ui_tests`

### Test
1. Running `bundle exec fastlane pick_test_account_and_run_ui_tests` and seeing in fastlane log that it uses a free account, and deoccupies it at the end of execution.
2. Doing the same, but messing up the execution of UI tests by navigating to unexpected area of the app (e.g. Settings), which results in several tests fail. Test account should be freed in this case as well via `error` block.
3. Optionally committing something here, just to trigger the `UI Tests` job while 1 or 2 from above runs, and seeing that parallel execution causes no issues.
4. Checking the latest `UI Tests` execution in this PR and seeing that it was selecting the test account using the key-value.

**Hint**: to make testing of account states easier, it's possible to get the current value by running `curl https://api.keyvalue.xyz/82649691/simplenoteuitest-ios`

### Review

### Release

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -893,9 +893,9 @@ end
 # expected to look like a series of Fs and Ts.
 # Example:
 #   get_test_accounts_state
-#     => "TFF"
+#     => "TFFFF"
 def get_test_accounts_state
-  uri = URI.parse(UI_TESTS_ACCOUNTS_KEY)
+  uri = URI.parse(UI_TEST_ACCOUNTS_KEY)
   response = Net::HTTP.get_response(uri)
   accounts_state = response.body.gsub("\n","")
   UI.message("State of test accounts: #{accounts_state}")
@@ -916,7 +916,7 @@ end
 # Replaces a character at `index` with `availability` value.
 # Example:
 #   change_test_account_availability("F")
-#     => FTT 
+#     => FFFFF
 def change_test_account_availability(availability)
   accounts_state = get_test_accounts_state
   accounts_state[$used_test_account_index] = availability
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -895,7 +895,7 @@ end
 #   get_test_accounts_state
 #     => "TFFFF"
 def get_test_accounts_state
-  uri = URI.parse(UI_TEST_ACCOUNTS_KEY)
+  uri = URI.parse(UI_TESTS_ACCOUNTS_KEY)
   response = Net::HTTP.get_response(uri)
   accounts_state = response.body.gsub("\n","")
   UI.message("State of test accounts: #{accounts_state}")
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -10,6 +10,8 @@ end
 USER_ENV_FILE_PATH = File.join(Dir.home, '.simplenoteios-env.default')
 PROJECT_ENV_FILE_PATH = File.expand_path(File.join(Dir.pwd, '../.configure-files/project.env'))
 UI_TESTS_SCHEME = "SimplenoteUITests"
+UI_TESTS_ACCOUNTS_KEY = "https://api.keyvalue.xyz/82649691/simplenoteuitest-ios"
+$used_test_account_index = nil
 
 before_all do
   # Check that the env files exist
@@ -24,6 +26,14 @@ before_all do
   # It is skipped if this isn't running on CI
   # See https://circleci.com/docs/2.0/ios-codesigning/
   setup_circle_ci
+end
+
+after_each do |lane, options|
+  free_occupied_test_account if lane == :run_ui_tests
+end
+
+error do |lane, exception, options|
+  free_occupied_test_account if lane == :run_ui_tests
 end
 
 platform :ios do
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -850,3 +875,62 @@ end
...
+  response = Net::HTTP.get_response(uri)
+  accounts_state = response.body.gsub("\n","")
+  UI.message("State of test accounts: #{accounts_state}")
+  accounts_state
+end
+
+# Returns the index of first "F" (Free) occurence in the test
+# accounts keyvalue, or crashes fastlane if it was not found.
+# Example:
+#   get_free_test_account_index
```
mokagio: Nitpick. This documentation is out of sync with the name of the method.

I actually like this name better.

fastlane/Fastfile
```diff
@@ -10,6 +10,8 @@ end
 USER_ENV_FILE_PATH = File.join(Dir.home, '.simplenoteios-env.default')
 PROJECT_ENV_FILE_PATH = File.expand_path(File.join(Dir.pwd, '../.configure-files/project.env'))
 UI_TESTS_SCHEME = "SimplenoteUITests"
+UI_TESTS_ACCOUNTS_KEY = "https://api.keyvalue.xyz/82649691/simplenoteuitest-ios"
```
mokagio: Nitpick. What do you think of using `URL` in the name? The value is a URL more than a key.

```suggestion
UI_TESTS_ACCOUNTS_URL = "https://api.keyvalue.xyz/82649691/simplenoteuitest-ios"
```

fastlane/Fastfile
```diff
@@ -715,6 +721,25 @@ end
...
+  # Usage:
+  # bundle exec fastlane pick_test_account_and_run_ui_tests [device:<Name of the iOS Simulator>]
+  #
+  # Example:
+  # bundle exec fastlane pick_test_account_and_run_ui_tests device:"iPhone 12"
+  #####################################################################################
+  desc "Occupy a free test account and run UI tests with it"
+  lane :pick_test_account_and_run_ui_tests do | options |
+    occupy_test_account
+    run_ui_tests(options.merge({ test_account: "simplenoteuitest-ios-#{$used_test_account_index+1}@a8c.com" }))
```
mokagio: Nitpick.

It might be a tiny bit simpler to avoid the index shifting logic. What do you think?

```suggestion
    run_ui_tests(options.merge({ test_account: "simplenoteuitest-ios-#{$used_test_account_index}@a8c.com" }))
```

Or is there a reason why we shouldn't have a `simplenoteuitest-ios-0@a8c.com` test account?

fastlane/Fastfile
```diff
@@ -850,3 +875,62 @@ end
...
+
+# Returns the value of test accounts key, which is 
+# expected to look like a series of Fs and Ts.
+# Example:
+#   get_test_accounts_state
+#     => "TFFFF"
+def get_test_accounts_state
+  uri = URI.parse(UI_TESTS_ACCOUNTS_KEY)
+  response = Net::HTTP.get_response(uri)
+  accounts_state = response.body.gsub("\n","")
```
mokagio: Is this to remote the trailing `\n`?

If so, there's also [`chomp`](https://apidock.com/ruby/String/chomp) that could be used and is more "idiomatic".

```suggestion
  accounts_state = response.body.chomp
```

mokagio (approved):
I'm a bit worried we're adding extra brittleness with the management of the accounts state in a remote service. But it's hard to tell until we actually try this out.

The implementation looks good. I only left a few nitpicks.

Let's see how this performs for us once we put some load on it 😄

fastlane/Fastfile
pachlava: Thank you, Gio! It works here!

fastlane/Fastfile
pachlava: Yes, of course! I did not think this through when asked JB to create 1-3 accounts. The zero-account should be ready today

fastlane/Fastfile
pachlava: Done! Makes sense, thank you

fastlane/Fastfile
pachlava: Honesty, I also like the initial name better, but it does not reflect what the method does any more. 

In the first iteration the index was returned by / fed to methods, but now the methods read/write the global variable instead. I fixed the name in documentation, thanks for the notice!

fastlane/Fastfile
pachlava: Account created already - so no index shift anymore! 🙇

pachlava:
> I'm a bit worried we're adding extra brittleness with the management of the accounts state in a remote service. But it's hard to tell until we actually try this out.

@mokagio I'm also worried by this, and also by the fact that this solution looks ugly to me, for some reason, even if it proves to be working.

Thank you for all your notes and nitpicks: all was committed. With the next PR I'd like to create a separate scheme in Xcode with the selection of several tests, and finally make this work not only on `UI-tests` branches.

fastlane/Fastfile
mokagio: Cool. Thanks for that.

mokagio:
> With the next PR I'd like to create a separate scheme in Xcode with the selection of several tests, and finally make this work not only on UI-tests branches.

Might want to look into [test plans](https://useyourloaf.com/blog/xcode-test-plans/), too:

> I like to have a quick set of tests I run by default when I hit command-U. I also want a full set of unit, UI and performance tests running against several configurations. Test plans make it much easier to set that up.

Somehow, I never had a chance to use test plans, so I'm not 100% sure they're the best solution for this 😅

Issue #1270, opened 2021-05-03
Some in app notifications are not sizing correctly


charliescheer:
Some of the in app notifications have started using their max sizing instead of using the intrinsic size of the contained text:

### Expected
When a notice is created, the size of the notice should be the size of the text or less than 350 pt in width.

<img src="https://cdn-std.droplr.net/files/acc_593859/ZF9HK4" />

### Observed
Certain notices are appearing with additional space in them and do not shrink down to match the size of the text.

<img src="https://cdn-std.droplr.net/files/acc_593859/hrevVM" />

### Reproduced
This only seems to happen with notices that have a button in them, so to test this trigger any notice that has an action:
Trashed note
Publish Successful
Unpublish Successful

You will see there is extra space between the button and the text instead of having the correct spacing of 16 pt

charliescheer:
This issue should have been closed by #1278 
Closing

PR #1278, opened 2021-05-14
Change how notices auto size

charliescheer:
###Fix
Fixes issue #1270 

Currently there is an issue with the in app notifications where they do not size correctly when there is an action button visible.
 

This problem came up while trying to fix an issue with how the notices render when on the largest of accessibility sizes.  Per a conversation about how best to approach this (internal ref p2XJRt-2Za ) the idea was to first fix how the sizing works for the normal size, and then to change the layout of the stack view for the largest of accessibility sizes.  This PR makes both changes.

(in order from left to right: Normal size, Accessibility Medium, Accessibility Large, Accessibility extra extra extra large)

### Test
1. Trigger a notice that has a button (trashed note, publish successful, unpublish successful) make sure that the layout sizes correctly
2. Using Publish Successful (the largest notice currently) on a few different accessibility sizes.
3. Make sure to check sizes below Accessibility Medium (the last size that will keep the label and button horizontal)
4. Check accessibility sizes that are Accessibility Large or bigger (all sizes over Accessibility large will change the axis of the stack view to vertical)

### Review

### DIFF
--- a/Simplenote/NoticeView.swift
+++ b/Simplenote/NoticeView.swift
@@ -50,6 +50,7 @@ class NoticeView: UIView {
         translatesAutoresizingMaskIntoConstraints = false
         setupViewStyles()
         setupGestureRecognizers()
+        configureAccessibilityIfNeeded()
 
         layoutIfNeeded()
     }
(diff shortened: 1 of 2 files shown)


SylvesterWilmott:
Thanks @charliescheer it looks great and works as described.

eshurakov (approved):
Thank you, looks good!

:shipit:

PR #1281, opened 2021-05-17
Fix xcode compiler warnings

charliescheer:
### Fix
With a recent update to Xcode, there were a few changes that were causing the complier to throw up some warnings.  This PR make the changes to remove those warnings.

Two things needed to be changed:
1.For delegates the class keyword was depreciated, suggestion was to replace those with AnyClass
2. in SPEntryListViewController we were setting the selectionStyle for the cell using an enum that required an implicit conversion.  Changing that from "UITableViewCellAccessoryNone" to "UITableViewCellSelectionStyleNone" silenced the warning.

### Test
1. build the project and confirm there are no warnings

Nothing should have changed about how the app operates, but do a quick smoke test to confirm

### Review

### DIFF
--- a/Simplenote/Classes/KeyboardObservable.swift
+++ b/Simplenote/Classes/KeyboardObservable.swift
@@ -5,7 +5,7 @@ import UIKit
 /// Adopters of this protocol will recieve interactive keyboard-based notifications
 /// by implmenting the provided functions within.
 ///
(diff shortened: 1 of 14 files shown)


jleandroperez (approved):
Farewell, compiler warnings!! we won't miss you!! 😂 

**Thank you** for doing this sir!!

:shipit:

PR #1284, opened 2021-05-19
Move secrets out of the repository – Part 1

mokagio:
### Fix

Moves the secrets to `~/.configure/simplenote-ios/secrets/`. ~~All except the screenshots credentials one, which for some reason is [failing](https://app.circleci.com/pipelines/github/Automattic/simplenote-ios/3722/workflows/40efa934-39e4-44e1-85c0-24a204854fde/jobs/3919) to execute its build phase (see inline comments for `SPCrendentials.swift`).~~

### Test
**First**. See the [CI run](https://app.circleci.com/pipelines/github/Automattic/simplenote-ios/3731/workflows/32dc9aa4-5625-494b-b767-36c53a0e0156/jobs/3929) for the companion [test branch](https://github.com/Automattic/simplenote-ios/pull/1285).

**Second**. Checkout this branch and run `bundle exec fastlane run configure_apply && ls ~/.configure/simplenote-ios/secrets` it should succeed and list:

```
➜ ls ~/.configure/simplenote-ios/secrets/
SPCredentials.swift                     google_cloud_keys.json
ScreenshotsCredentials.swift            project.env
app_store_connect_fastlane_api_key.json
```

**Third**. Play around with the build phase script for `SPCredentials.swift` and `ScreeshotsCredentials.swift` to verify the errors:

_When called with missing arguments:_

image

_When the source path can't be found:_

image

_When the target path is not ignored_ (note that the whole `Simplenote/Credentials` folder is already ignored):

image

### Review

### Release


Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 GIT
   remote: https://github.com/wordpress-mobile/release-toolkit
-  revision: eb0f24ec6c73b3454d711cc4b3276c50b9655861
-  tag: 0.18.1
+  revision: 8534f11250be45e2b848b9ccbeea6b474a9b63d3
+  branch: allow-paths-outside-project-folder
```
mokagio: Waiting for https://github.com/wordpress-mobile/release-toolkit/pull/252 to be merged.

Scripts/Build-Phases/copy-secret.sh
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+if [ ! -f ${SOURCE_PATH} ]; then
+    echo "error: Unable to copy credentials. Could not find ${SOURCE_PATH}."
+    exit 1
+else
+    echo "Copying credentials..."
+    mkdir -p ${TARGET_FOLDER}
+    cp ${SOURCE_PATH} ${TARGET_PATH}
+fi
```
mokagio, line 32: If we like this approach to copy over secret `.swift` files, we could move this logic into the release toolkit.

Something I don't like, though, is that in its current form it is not very helpful for external contributors. 🤔

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -22,7 +22,7 @@
 			isa = PBXAggregateTarget;
 			buildConfigurationList = B5C7BF98230C59F5000DEC91 /* Build configuration list for PBXAggregateTarget "SimplenoteSecrets" */;
 			buildPhases = (
-				B5C7BF9E230C5A79000DEC91 /* ShellScript */,
+				B5C7BF9E230C5A79000DEC91 /* Run Script */,
```
mokagio: Can you believe that the Xcode UI shows "Run Script" in the name of the build phase by default, but, until you actually go in and touch the string, doesn't set the value in here?

Maybe that was a quirk of the version in which the aggregate target for the secrets was added, but still...

fastlane/Matchfile
```diff
@@ -1,6 +1,10 @@
...
-storage_mode("google_cloud")
-google_cloud_bucket_name("a8c-fastlane-match")
-google_cloud_keys_file(".configure-files/google_cloud_keys.json")
+storage_mode('google_cloud')
+google_cloud_bucket_name('a8c-fastlane-match')
+# Interestingly, Using the path with '~/.configure/...' results in Fastlane
+# failing to find the file.
+SECRETS_ROOT = File.join(Dir.home, '.configure/simplenote-ios/secrets')
+google_cloud_keys_file(File.join(SECRETS_ROOT, 'google_cloud_keys.json'))
+api_key_path(File.join(SECRETS_ROOT, 'app_store_connect_fastlane_api_key.json'))
```
mokagio, line 10: While I'm not a fan of spreading configurations across multiple `-file` in Fastlane, it was handy to have this here considering we have three different calls to `match` in this project.

fastlane/Matchfile
```diff
@@ -1,6 +1,10 @@
...
 
 # Store certs/profiles encrypted in Google Cloud
-storage_mode("google_cloud")
-google_cloud_bucket_name("a8c-fastlane-match")
-google_cloud_keys_file(".configure-files/google_cloud_keys.json")
+storage_mode('google_cloud')
+google_cloud_bucket_name('a8c-fastlane-match')
+# Interestingly, Using the path with '~/.configure/...' results in Fastlane
+# failing to find the file.
+SECRETS_ROOT = File.join(Dir.home, '.configure/simplenote-ios/secrets')
```
mokagio, line 8: I experienced that with the Google Cloud Key value. Sounds like a bug in Fastlane that we could help address. I haven't had the chance to dig into it, yet.

fastlane/Pluginfile
```diff
@@ -2,7 +2,9 @@
 #
 # Ensure this file is checked in to source control!
 
-gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.18.1'
+gem 'fastlane-plugin-wpmreleasetoolkit',
+  git: 'https://github.com/wordpress-mobile/release-toolkit',
+  branch: 'allow-paths-outside-project-folder'
```
mokagio: I was tired of seeing a whole line diff that I had store horizontally scroll whenever going from a tag to a branch and vice versa... 🤓

.circleci/config.yml
```diff
@@ -18,6 +18,11 @@ xcode_version: &xcode_version
   xcode-version: "12.0"
 
 commands:
+  copy_secrets:
+    steps:
+      - run:
+          name: Copy Secrets
+          command: bundle exec fastlane run configure_apply
```
mokagio, line 25: DRYed this repeated step

.circleci/config.yml
```diff
@@ -42,9 +47,7 @@ jobs:
       - ios/install-dependencies:
           bundle-install: true
           pod-install: true
-      - run:
-          name: Copy demo SPCredentials
-          command: mkdir -p Simplenote/Credentials && cp Simplenote/SPCredentials-demo.swift Simplenote/Credentials/SPCredentials.swift
+      - copy_secrets
```
mokagio, line 50: It was neat to copy the demo credentials over, but now that we fully rely on the aggregate target build phase, this is not sufficient anymore.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -22,7 +22,7 @@
 			isa = PBXAggregateTarget;
 			buildConfigurationList = B5C7BF98230C59F5000DEC91 /* Build configuration list for PBXAggregateTarget "SimplenoteSecrets" */;
 			buildPhases = (
-				B5C7BF9E230C5A79000DEC91 /* ShellScript */,
+				B5C7BF9E230C5A79000DEC91 /* Copy Secrets */,
```
mokagio, line 25: I had forgotten about this, but we had this neat aggregate target that runs before the Simplenote target build and copies the `SPCredentials.swift` file into the path that the codebase expects.

What I did in this PR was to move the logic into a dedicated script (see above) to make it easier to work on and point the source file to `~/.configure/simplenote-ios/secrets/`

image

.configure
```diff
@@ -5,31 +5,31 @@
   "files_to_copy": [
     {
       "file": "iOS/simplenote/SPCredentials.swift",
-      "destination": "Simplenote/Credentials/SPCredentials.swift",
+      "destination": "~/.configure/simplenote-ios/secrets/SPCredentials.swift",
```
mokagio, line 8: As a future improvement, I think we could make `configure` default to `~/.configure/<"project_name" key value>/secrets` if no `destination` is specified.

One reason _not_ to do it is that, while it would keep the config file DRY and avoid the chance of typos, it would make it less clear to newcomers where the files go. ⚖️

Scripts/Build-Phases/copy-secret.sh
mokagio: An idea I have for a next iteration on this is to provide another parameter, maybe optional, to the script to fallback to if the source can't be found. That could be the value of the example file and the script could echo a warning:

> Could not find secret at `$SOURCE_PATH`. If you are a maintainer of this project, please ensure you [... instructions for `configure` ...]. In case you are an external contributor instead, I copied `$FALLBACK_PATH` to `$TARGET_PATH` to allow the build to proceed.

Another option could be to not copy the example file ourselves, to ensure everyone is aware of this, and instead print the suggestion as part of the error.

> Unable to copy credentials. Could not find `${SOURCE_PATH}`. If you are a maintainer of this project, [... instructions ...]. If you are an external contributor, copy `$FALLBACK_PATH` to `$TARGET_PATH` instead to allow the build to continue.

mokagio:
@pachlava I think these changes should address the issues you described in https://github.com/Automattic/simplenote-ios/pull/1242#issuecomment-819123843.

.configure
jkmassel: I like the idea of doing this, but making the logging more verbose so it's clear what the tool is doing under the hood?

Scripts/Build-Phases/copy-secret.sh
jkmassel: Something else for consideration – adding an `example` file to the `.configure` file entry, so that it copies that one if the real one can't be decrypted?

I like the idea of sharing this with the release toolkit, but that puts a dependency on Ruby in every project (which isn't as nice).

jkmassel (changes requested):
Probably a bootstrapping issue:

```
ls: /Users/jkmassel/.configure/simplenote-ios/secrets: No such file or directory
```

I'm guessing there isn't a "create directory recursively" option specified somewhere?

Scripts/Build-Phases/copy-secret.sh
mokagio: Regarding the extra Ruby dependency: dang I didn't consider that... 🤔 It shouldn't be hard to make it part of the upcoming Rust [`configure`](https://github.com/Automattic/configure/) itself, though.

mokagio: And now that I think about it, this is meaningful only for Xcode projects, which already depend on Ruby anyways, so we might be able to get away with it for some time. 🤷‍♂️

.configure
mokagio: Great point!

mokagio:
> Probably a bootstrapping issue:
> 
> ```
> ls: /Users/jkmassel/.configure/simplenote-ios/secrets: No such file or directory
> ```
> 
> I'm guessing there isn't a "create directory recursively" option specified somewhere?

🤔 🤔 🤔 

That's odd that you are seeing that. [It works on CI](https://github.com/Automattic/simplenote-ios/pull/1285) after all.

What I find it even weirder it that your comment is on a diff for outdated code:

Screen Shot 2021-05-27 at 10 36 20 am

The code from your comment (right) was wrong but I did fix it eventually (left). GitHub should show and serve the up to date version. 

Moreover, my fix is earlier than your review. `:confused_parrot:`.

jkmassel (approved):
Retried and it worked great

PR #1286, opened 2021-05-19
Add tracks to notices

charliescheer:
### Fix
This PR adds tracks events for the recently added in app notifications

### Review

### Release
> 
> > Added markdown support

### DIFF
--- a/Simplenote/Classes/OptionsViewController.swift
+++ b/Simplenote/Classes/OptionsViewController.swift
@@ -356,7 +356,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedInternalLink()
         UIPasteboard.general.copyInternalLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
-        SPTracker.trackNoticeLinkCopied()
+        SPTracker.trackPresentedNotice(kind: .linkCopied)
         dismiss(animated: true, completion: nil)
     }
 
@@ -387,7 +387,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedPublicLink()
         UIPasteboard.general.copyPublicLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
-        SPTracker.trackNoticeLinkCopied()
+        SPTracker.trackPresentedNotice(kind: .linkCopied)
         dismiss(animated: true, completion: nil)
     }
 
--- a/Simplenote/Classes/OptionsViewController.swift
+++ b/Simplenote/Classes/OptionsViewController.swift
@@ -356,7 +356,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedInternalLink()
         UIPasteboard.general.copyInternalLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
-        SPTracker.trackPresentedNotice(kind: .linkCopied, additionalProperties: ["link_type": "internal_link"])
+        SPTracker.trackPresentedNotice(kind: .internalLinkCopied)
         dismiss(animated: true, completion: nil)
     }
 
@@ -387,7 +387,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedPublicLink()
         UIPasteboard.general.copyPublicLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
-        SPTracker.trackPresentedNotice(kind: .linkCopied, additionalProperties: ["link_type": "public_link"])
+        SPTracker.trackPresentedNotice(kind: .publicLinkCopied)
         dismiss(animated: true, completion: nil)
     }
 
--- a/Simplenote/Classes/OptionsViewController.swift
+++ b/Simplenote/Classes/OptionsViewController.swift
@@ -356,7 +356,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedInternalLink()
         UIPasteboard.general.copyInternalLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
-        SPTracker.trackPresentedNotice(kind: .linkCopied)
+        SPTracker.trackPresentedNotice(kind: .linkCopied, additionalProperties: ["link_type": "internal_link"])
         dismiss(animated: true, completion: nil)
     }
 
@@ -387,7 +387,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedPublicLink()
         UIPasteboard.general.copyPublicLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
-        SPTracker.trackPresentedNotice(kind: .linkCopied)
+        SPTracker.trackPresentedNotice(kind: .linkCopied, additionalProperties: ["link_type": "public_link"])
         dismiss(animated: true, completion: nil)
     }
 
--- a/Simplenote/Classes/OptionsViewController.swift
+++ b/Simplenote/Classes/OptionsViewController.swift
@@ -356,7 +356,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedInternalLink()
         UIPasteboard.general.copyInternalLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
-        SPTracker.trackPresentedNotice(kind: .internalLinkCopied)
+        SPTracker.trackPresentedNotice(ofType: .internalLinkCopied)
         dismiss(animated: true, completion: nil)
     }
 
@@ -387,7 +387,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedPublicLink()
         UIPasteboard.general.copyPublicLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
-        SPTracker.trackPresentedNotice(kind: .publicLinkCopied)
+        SPTracker.trackPresentedNotice(ofType: .publicLinkCopied)
         dismiss(animated: true, completion: nil)
     }
 
--- a/Simplenote/Classes/OptionsViewController.swift
+++ b/Simplenote/Classes/OptionsViewController.swift
@@ -356,6 +356,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedInternalLink()
         UIPasteboard.general.copyInternalLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
+        SPTracker.trackNoticeLinkCopied()
         dismiss(animated: true, completion: nil)
     }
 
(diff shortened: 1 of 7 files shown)


Simplenote/Classes/OptionsViewController.swift
```diff
@@ -386,6 +387,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedPublicLink()
         UIPasteboard.general.copyPublicLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
+        SPTracker.trackNoticeLinkCopied()
```
jleandroperez: `copyInterlinkWasPressed` also calls `SPTracker.trackNoticeLinkCopied`

Note that:

- Copy Interlink: Copies the Simplenote internal reference to a note
- Copy Link (this precise API) copies the **Public Link**. That is the published URL

Perhaps this one should be `trackPublicLinkCopied` and `trackNoticeLinkCopied` should be `trackInterlinkCopied` ?

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -521,7 +521,9 @@ extension SPNoteEditorViewController {
         CSSearchableIndex.default().deleteSearchableNote(note)
         NoticeController.shared.present(NoticeFactory.noteTrashed(onUndo: {
             SPObjectManager.shared().restoreNote(note)
+            SPTracker.trackNoticeActionMultipleTrashUndo()
```
jleandroperez: There is no multiple undo in this one

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -777,8 +777,10 @@ private extension SPNoteListViewController {
                 self?.delete(note: note)
                 NoticeController.shared.present(NoticeFactory.noteTrashed(onUndo: {
                     SPObjectManager.shared().restoreNote(note)
+                    SPTracker.trackNoticeNoteTrashed()
```
jleandroperez: Should this be `Restored` ?

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -819,6 +822,7 @@ private extension SPNoteListViewController {
         let copy = UIAction(title: ActionTitle.copyLink, image: .image(name: .link)) { [weak self] _ in
             self?.copyInternalLink(to: note)
             NoticeController.shared.present(NoticeFactory.linkCopied())
+            SPTracker.trackNoticeLinkCopied()
```
jleandroperez: Same note as above, this should probably be `trackInterlinkCopied`

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -792,6 +794,7 @@ private extension SPNoteListViewController {
         let copyAction = UIContextualAction(style: .normal, title: nil, image: .image(name: .link), backgroundColor: .simplenoteTertiaryActionColor) { [weak self] (_, _, completion) in
                 self?.copyInternalLink(to: note)
                 NoticeController.shared.present(NoticeFactory.linkCopied())
+                SPTracker.trackNoticeLinkCopied()
```
jleandroperez: This should probably be `trackInterlinkCopied`

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -911,7 +916,9 @@ private extension SPNoteListViewController {
             for note in notes {
                 SPObjectManager.shared().restoreNote(note)
             }
+            SPTracker.trackNoticeActionMultipleTrashUndo()
         }))
+        SPTracker.trackNoticeNotesTrashed()
```
jleandroperez: We could align the name of both. Something like:

```
let multiple = notes.count > 1
SPTracker.trackNoticeTrash(multiple: multiple)
```

Same for the Undo? maybe Undo could even be a parameter (undoing). Not sure about the last part!

Simplenote/Controllers/PublishNoticePresenter.swift
```diff
@@ -5,19 +5,25 @@ class PublishNoticePresenter {
         switch note.publishState {
         case .publishing:
             NoticeController.shared.present(NoticeFactory.publishing())
+            SPTracker.trackNoticePublishing()
         case .published:
             let notice = NoticeFactory.published(note, onCopy: {
                 UIPasteboard.general.copyPublicLink(to: note)
                 NoticeController.shared.present(NoticeFactory.linkCopied())
+                SPTracker.trackNoticeActionPublished()
```
jleandroperez: `trackPublickLinkCopied` perhaps?

Simplenote/SPTracker+Extensions.swift
```diff
@@ -27,6 +27,58 @@ extension SPTracker {
...
+    static func trackNotice(_ value: String) {
+        trackAutomatticEvent(withName: "notice_presented", properties: ["notice": value])
+    }
+
+    static func trackNoticeAction(_ value: String, action: String) {
+        trackAutomatticEvent(withName: "notice_action_tapped", properties: ["notice": value, "notice_action": action])
+    }
+
+    static func trackNoticeLinkCopied() {
+        trackNotice("Link Copied")
```
jleandroperez: For consistency reasons (with the naming) perhaps:

- All of the `notice names` should be lowecase
- Preferrably a single word, if multiple, maybe using `_` as a word spacer

Simplenote/SPTracker+Extensions.swift
```diff
@@ -27,6 +27,58 @@ extension SPTracker {
...
+        trackNoticeAction("Notes Trashed", action: "Undo")
+    }
+
+    static func trackNoticeActionUnpublished() {
+        trackNoticeAction("Unpublished", action: "Undo")
+    }
+
+    static func trackNoticeActionPublished() {
+        trackNoticeAction("Published", action: "Copy Link")
+    }
```
jleandroperez, line 58: Since all of the code that invoke these API(s) is pure Swift, how about something like this?

```
enum NoticeKind: String {
   case trashed
   case published
   case unpublished
   //..
}

enum NoticeActionKind {
     case undo
     case interlinkCopied
     case publicLinkCopied
}

func trackPresentedNotice(kind: NoticeKind) {
   // ..
}

func trackPerformedAction(noticeAction: NoticeActionKind, noticeKind: NoticeKind ) {
  // ..
}
```


Naming can probably be improved, but you'd be able to get rid of similar looking methods. Maybe?

jleandroperez (commented):
Sending few notes your way sir!

Simplenote/Classes/OptionsViewController.swift
charliescheer: Hmmm, I think I am not clear on what are we trying to track.  

If we are tracking that the notice has been displayed, both notices for public link and inter link are the same.  So I would think that we wouldn't want to differentiate the two.  And since public link copied and interlink copied also create track events, it feels a bit like duplication to me.

But I can also see how seeing the extra details of which action caused the creation of the copy link notice to appear could be useful.  🤔 

@SylvesterWilmott do you have any thoughts on the different types of "Link Copied" notices.  Should they all be treated the same?  Or should the different types of Link Copied be noted with what link was copied

Simplenote/Classes/SPNoteListViewController+Extensions.swift
charliescheer: I kinda feel about this one the same way that I do with the link copied notices.  We could combine these track events into a single method, but the only time this notice would be presented is if we were doing a multi edit delete, and the notice is specific to that action.  

If we are tracking the presentation of the notice, I feel like multi delete and undo should have their own track methods.

Simplenote/SPTracker+Extensions.swift
charliescheer: I like it!

mokagio:
I'm going to bump this to the next release because we'll be code freezing 4.37 next week and it seems like this needs a bit more touches.

It doesn't seem the case, but if it's very important that this makes it into this release, let me know and we'll organize a new beta once ready.

jleandroperez:
Thank you Gio!

Simplenote/Classes/OptionsViewController.swift
SylvesterWilmott: Sorry for the delay on this. Tracking only the type of notice is okay but we should differentiate between public link and internal link. We can use one event for both since the notices are identical and a "type" property to distinguish between the two.

charliescheer:
This got a little lost in the shuffle but it is ready to be reviewed again @jleandroperez

Simplenote/Classes/OptionsViewController.swift
```diff
@@ -356,6 +356,7 @@ private extension OptionsViewController {
         SPTracker.trackEditorCopiedInternalLink()
         UIPasteboard.general.copyInternalLink(to: note)
         NoticeController.shared.present(NoticeFactory.linkCopied())
+        SPTracker.trackPresentedNotice(kind: .linkCopied, additionalProperties: ["link_type": "internal_link"])
```
jleandroperez: `link_type` is used in 4 spots. It'd be neat not to repeat those strings.

Suggestion: flatten it out, less code required!

```
 enum NoticeType: String {
         case internalLinkCopied = "internal_link_copied"
         case publicLinkCopied = "public_link_copied"
```

Simplenote/SPTracker+Extensions.swift
```diff
@@ -29,6 +29,41 @@ extension SPTracker {
     }
 }
 
+// MARK: - In App Notifications
+//
+extension SPTracker {
+
+    static func trackPresentedNotice(kind: NoticeType, additionalProperties: [String: String]? = nil) {
```
jleandroperez:
```suggestion
    static func trackPresentedNotice(ofType type: NoticeType) {
```

Simplenote/SPTracker+Extensions.swift
```diff
@@ -29,6 +29,41 @@ extension SPTracker {
...
+
+        if let additionalProperties = additionalProperties {
+            for key in additionalProperties.keys {
+                properties[key] = additionalProperties[key]
+            }
+        }
+        trackAutomatticEvent(withName: "notice_presented", properties: properties)
+    }
+
+    static func trackPreformedNoticeAction(kind: NoticeType, action: NoticeActionType) {
```
jleandroperez:
```suggestion
    static func trackPreformedNoticeAction(ofType actionType: NoticeActionType, noticeType: NoticeType) {
```

jleandroperez: Or something like that! I'm leaning towards `Type` because both enums read `NoticeXXXType`

jleandroperez (commented):
Few nipticky comments! only important note is: let's drop the `additionalProperties` payload, since it's just meant to differentiate what kind of link was copied.

If you flatten out, everything gets simpler 😄

jleandroperez (approved):
:shipit: 

(Please make sure to set the milestone!). Thank you sir!!

PR #1288, opened 2021-05-22
Changed in app notification dark mode background color

charliescheer:
### Fix
Under certain conditions, the in app notifications have the same background color as other elements, which makes them sort of disappear into their background

This PR changes the color for the notifications in dark mode so that they do not blend into the background.

### Test
This problem is easiest to notice when in dark mode and in large accessibility sizes, but this fix should work for all sizes and visual modes.

1. go to the options menu while in dark mode
2. change the publish state of a notice
3. make sure that it can be seen against the background colors of the cells in the table

### Review


jleandroperez (approved):
:shipit:

PR #1292, opened 2021-05-27
Run `Trash` UI tests on every commit

pachlava:
This is an unexpected PR prior to the one introducing a subset of tests to be executed on every commit.

### Fix
- Since the previously used third party key-value service went down, I switched to another one, which looks more reliable
- The sensitive information, such as the URL of the key, and the test account login was moved to CircleCI environment variables `UI_TESTS_ACCOUNTS_URL` and `UI_TESTS_ACCOUNT_EMAIL`
- `pick_test_account_and_run_ui_tests` lane now runs only on CI
- Previously existing selection of tests (`Trash`) runs now on every commit. The tests selection is temporary until next PR with a new scheme.

As for the upcoming subset of tests: it looks like test plans will not really fit the purpose here, because they are more about running tests under different configurations, rather than having different selections of tests. It looks like having additional scheme(s) is still the easiest way to accomplish this.

### Test
1. Running `bundle exec fastlane pick_test_account_and_run_ui_tests` is no longer possible locally and results in `This lane runs only on CI` message 
2. Rerunning `UI Tests` job here is successful, Fastlane informs about accounts state before and after running the tests.

### Review

### Release

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -10,7 +10,6 @@ end
 USER_ENV_FILE_PATH = File.join(Dir.home, '.simplenoteios-env.default')
 PROJECT_ENV_FILE_PATH = File.expand_path(File.join(Dir.pwd, '../.configure-files/project.env'))
 UI_TESTS_SCHEME = "SimplenoteUITests"
-UI_TESTS_ACCOUNTS_URL = "https://api.keyvalue.xyz/82649691/simplenoteuitest-ios"
 $used_test_account_index = nil
 
 before_all do
@@ -736,9 +735,13 @@ end
   #####################################################################################
   desc "Occupy a free test account and run UI tests with it"
   lane :pick_test_account_and_run_ui_tests do | options |
-    occupy_test_account
-    run_ui_tests(options.merge({ test_account: "simplenoteuitest-ios-#{$used_test_account_index}@a8c.com" }))
-    deoccupy_test_account
+    if is_ci
+      occupy_test_account
+      run_ui_tests(options.merge({ test_account: get_required_env("UI_TESTS_ACCOUNT_EMAIL").sub("X", $used_test_account_index.to_s)}))
+      deoccupy_test_account
+    else
+      UI.message("This lane runs only on CI")
+    end
   end
 end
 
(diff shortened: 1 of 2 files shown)


mokagio:
Hey @pachlava 👋 Sorry this one fell to the bottom of my review list and didn't get to it till now. 😳 

Is it still relevant after the changes in #1301?

pachlava:
Hey @mokagio ! The changes here are not as relevant by now, but I think it still makes sense to have it reviewed and merged because #1301 is based on this one. At the moment, the only meaningful change here is the removal of branch filter for `UI Tests` job.

fastlane/Fastfile
```diff
@@ -736,9 +735,13 @@ end
...
   lane :pick_test_account_and_run_ui_tests do | options |
-    occupy_test_account
-    run_ui_tests(options.merge({ test_account: "simplenoteuitest-ios-#{$used_test_account_index}@a8c.com" }))
-    deoccupy_test_account
+    if is_ci
+      occupy_test_account
+      run_ui_tests(options.merge({ test_account: get_required_env("UI_TESTS_ACCOUNT_EMAIL").sub("X", $used_test_account_index.to_s)}))
+      deoccupy_test_account
+    else
+      UI.message("This lane runs only on CI")
```
mokagio: How do you feel about changing this to a user error?

```suggestion
      UI.user_error!("This lane should be run only from CI")
```

I think a failure would make it clear to a developer calling this lane locally that they're doing something inappropriate. Otherwise, the `fastlane` call will succeed and they might not realize the side effect they expected didn't occur.

fastlane/Fastfile
```diff
@@ -10,7 +10,6 @@ end
 USER_ENV_FILE_PATH = File.join(Dir.home, '.simplenoteios-env.default')
 PROJECT_ENV_FILE_PATH = File.expand_path(File.join(Dir.pwd, '../.configure-files/project.env'))
 UI_TESTS_SCHEME = "SimplenoteUITests"
-UI_TESTS_ACCOUNTS_URL = "https://api.keyvalue.xyz/82649691/simplenoteuitest-ios"
```
mokagio, line 13: > The sensitive information, such as the URL of the key, and the test account login was moved to CircleCI environment variables UI_TESTS_ACCOUNTS_URL and UI_TESTS_ACCOUNT_EMAIL

👍

fastlane/Fastfile
mokagio: As an alternative that doesn't fail the lane, this could become:

```suggestion
      UI.important("This lane runs only on CI")
```

This way, the output will be colored and the caller will be more likely to notice it:

image

mokagio (approved):
Nice. :shipit:

fastlane/Fastfile
pachlava: Thank you @mokagio ! Makes sense, I went for the `UI.user_error!`

Issue #1289, opened 2021-05-23
Project: Widgets


charliescheer:
This is a project post for adding iOS 14 style widgets to Simplenote.  

### Tasks

- [x] New Note Widget
- [x] Single Note Widget
- [x] Note List Widget

### Description

There are three different widgets we want to add to Simplenote.  The first is a new note button that can live on the home screen of an iPhone. This widget is only available in the smallest widget family, so the entire widget will open up into a new note.

<img src="https://cdn-std.droplr.net/files/acc_593859/eCLMhS" />

The next widget is a single note widget that displays the content of a user selected note on the home screen.  This widget is available in Small and Medium widget families.  Tapping on this widget will take the user into the selected note

<img src="https://cdn-std.droplr.net/files/acc_593859/xQ1Stc" />

<img src="https://cdn-std.droplr.net/files/acc_593859/nh6IOk" />

Lastly we have a widget that shows a list of notes in a tag.  This multiple tap targets, and will provide a shortcut into the notes that are listed.  This widget is available in medium and large widget sizes.

<img src="https://cdn-std.droplr.net/files/acc_593859/IlydIZ" />

<img src="https://cdn-std.droplr.net/files/acc_593859/COcZj8" />

hardevine:
👍  waiting for it

PR #1293, opened 2021-05-27
New note widget

charliescheer:
### Fix
This PR is the first part of the iOS widget project #1289 

This is PR add the basic infrastructure to add widgets into Simplenote (adding widget target, widget group etc)

It also includes the New Note Widget.  This widget is a home screen button to quickly jump straight into the new note screen.

The widget it's self is pretty simple, it has no timeline, and displays no details.  It simply triggers opening the app.

### Test
1. Build the app on your device/simulator
2. Go to the home screen and press and hold to make the homepage edit appear, tap the plus button
3. Find Simplenote and add the new note widget
4. tap on it, you should be taken to new note page

### Review

Also, @mokagio  would you please take a look at this PR to see if there are any changes to the SimplenoteWidgetExtension configuration that we should make.

### Release
> `RELEASE-NOTES.txt` was updated in 216f7d with:
> 
> > Adds a home screen widget to start a new note with a single tap


pachlava:
I've noticed that new note created via widget has a difference compared to 'traditional' creation:
- If new note is created traditionally, and left empty (user goes back to note list) ~> it goes to `Trash` momentarily and is not shown in note list. Actually it should even [skip the `Trash`](https://github.com/Automattic/Simplenote-United/issues/116), but it's planned for future.
- If a note is created via widget, and left empty, it will stay in the note list:

https://user-images.githubusercontent.com/73365754/119973451-2c204000-bfbc-11eb-95a2-f3973b3305b1.mov

pachlava:
Did some testing on iPhone and iPad sims, in usual mode as well as in logged out state and with passcode enabled. Found no issues besides the one described above.

Side note: maybe it makes sense to decrease the `IPHONEOS_DEPLOYMENT_TARGET` in `SimplenoteWidgetsExtension` from 14.5 to 14.4, to avoid potential issues with CircleCI VMs, stat still run Xcode 12.4

mokagio:
I'm going to bump this to Future because we'll be code freezing 4.37 next week and it seems like this needs a bit more touches.

It doesn't seem the case, but if it's very important that this makes it into 4.37, let me know and we'll organize a new beta once ready.

charliescheer:
>I've noticed that new note created via widget has a difference compared to 'traditional' creation:

Thanks for catching that @pachlava!  I am using the new note URL to access creating this note via 
`-(BOOL)application:(UIApplication *)app openURL:(NSURL *)url`

I did a bit of testing and confirmed that the issue is with how this method is setup currently.  When you create a new note, the managed object is given a temporary object ID, that is replaced when the object is saved.  in the openURL function we are saving the note to simperium (which also saves to core data) before we load the editor.  

Looking at the history of the method that gets called with simplenote://new that has not been updated since before 2016 when SN went open source.

@jleandroperez is there any reason we can't remove the save call in the SPAppDelegate.m method 
`- (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options`  

I tested creating a new note without the call to " [_simperium save];"  (line 602) and it works just fine.  Thoughts?

charliescheer:
On further thought, I went ahead and removed the saving in the new note method and pushed that change, so it can be tested again and no longer persists an empty note.  Since the note is held with a temporary ID, which is what we would expect it to do with any new note, there is no need to save it on creation unless the user makes changes to the note.

@jleandroperez @eshurakov  do you agree?  Or is there something I am missing, and we should save the new note that is loaded by url before displaying?

charliescheer:
>Side note: maybe it makes sense to decrease the IPHONEOS_DEPLOYMENT_TARGET in SimplenoteWidgetsExtension from 14.5 to 14.4, to avoid potential issues with CircleCI VMs, stat still run Xcode 12.4

That was my mistake, I think apple updated my Xcode without asking.  I have downgraded the target to 14.4

Simplenote/SPAppDelegate.m
```diff
@@ -599,7 +599,6 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
                 }
             }
         }
-        [_simperium save];
```
eshurakov, line 602: I think it is fine to remove `save` from here, but check that if we pass content via query string (not sure if anyone uses this) the note is not deleted after we go back to the list.

While we're here, it may be nice to refactor this code to use [URLComponents](https://developer.apple.com/documentation/foundation/urlcomponents) instead of manually parsing the query string.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -4499,6 +4640,201 @@
...
+				IPHONEOS_DEPLOYMENT_TARGET = 14.4;
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/Frameworks",
+					"@executable_path/../../Frameworks",
+				);
+				MARKETING_VERSION = 4.36;
+				MTL_ENABLE_DEBUG_INFO = NO;
+				MTL_FAST_MATH = YES;
+				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.Internal.SimplenoteWidgets;
```
jleandroperez:
```suggestion
				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.Alpha.SimplenoteWidgets;
```

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -4499,6 +4640,201 @@
...
+				IPHONEOS_DEPLOYMENT_TARGET = 14.4;
+				LD_RUNPATH_SEARCH_PATHS = (
+					"$(inherited)",
+					"@executable_path/Frameworks",
+					"@executable_path/../../Frameworks",
+				);
+				MARKETING_VERSION = 4.36;
+				MTL_ENABLE_DEBUG_INFO = NO;
+				MTL_FAST_MATH = YES;
+				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.Development.SimplenoteWidgets;
```
jleandroperez:
```suggestion
				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.SimplenoteWidgets;
```

Simplenote/Classes/UIColor+Helpers.swift
```diff
@@ -43,12 +43,12 @@ extension UIColor {
     convenience init(hexString: String, alpha: CGFloat = 1.0) {
         let hexString = hexString.trimmingCharacters(in: .whitespacesAndNewlines)
         let scanner = Scanner(string: hexString)
-        if hexString.hasPrefix("#") {
-            scanner.scanLocation = 1
-        }
 
-        var color: UInt32 = 0
-        scanner.scanHexInt32(&color)
+        let characters = CharacterSet.init(charactersIn: "#")
```
jleandroperez:
```suggestion
        let characters = CharacterSet(charactersIn: "#")
```

SimplenoteWidgets/Widgets/NewNoteWidget.swift
```diff
@@ -0,0 +1,21 @@
...
+        .description(Constants.description)
+        .supportedFamilies([.systemSmall])
+
+    }
+}
+
+private struct Constants {
+    static let configurationKind = "NewNoteWidget"
+    static let displayName = "New Note"
+    static let description = "Create a new note instantly with one tap."
```
jleandroperez: 1. I'm assuming we need to use NSLocalizedString? 
2. Or the above is not needed, because of SwiftUI?
2. We also need to make sure that the localization scripts pick these up

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -4499,6 +4640,201 @@
...
+				CLANG_ENABLE_OBJC_WEAK = YES;
+				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
+				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
+				CODE_SIGN_IDENTITY = "iPhone Distribution: Automattic, Inc.";
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution: Automattic, Inc.";
+				CODE_SIGN_STYLE = Automatic;
+				COPY_PHASE_STRIP = NO;
+				CURRENT_PROJECT_VERSION = 4.36.0.0;
+				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
+				DEVELOPMENT_TEAM = PZYM8XX95Q;
```
jleandroperez: Please make sure the `Development Team` is set to **Automattic Enterprise** for the **Internal / Alpha** builds

SimplenoteWidgets/Providers/NewNoteWidgetProvider.swift
```diff
@@ -0,0 +1,23 @@
+import WidgetKit
+
+struct NewNoteWidgetEntry: TimelineEntry {
+    var date: Date
```
jleandroperez: Can this be `let date: Date` ?

jleandroperez (commented):
First pass looking good, sending you few comments sir!

Simplenote/SPAppDelegate.m
charliescheer: >I think it is fine to remove save from here, but check that if we pass content via query string (not sure if anyone uses this) the note is not deleted after we go back to the list.

Tested and the note persists if there is content in the note.

>While we're here, it may be nice to refactor this code to use URLComponents instead of manually parsing the query string.

refactored the creation to use URL components, much cleaner now.

SimplenoteWidgets/Providers/NewNoteWidgetProvider.swift
charliescheer: yes it can :)

SimplenoteWidgets/Widgets/NewNoteWidget.swift
charliescheer: I will need to dig into this a bit.  SwiftUI does handle the localizing a little differently, but since this isn't going into a Text view, it is a description, we may be able to do what we normally do and use NSLocalizedString.  I have done that now, but one thing to note is that our localizable files are not part of the widget extension target yet.  Would imagine that it will need to be.

charliescheer:
Note that I have updated the merge target to feature/ios-widgets

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -4659,7 +4659,7 @@
 				CLANG_ENABLE_OBJC_WEAK = YES;
 				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
-				CODE_SIGN_ENTITLEMENTS = SimplenoteWidgetsExtensionDebug.entitlements;
+				CODE_SIGN_ENTITLEMENTS = Simplenote/SimplenoteWidgetsExtensionDebug.entitlements;
```
mokagio, line 4662: I got this error [in CI](https://app.circleci.com/pipelines/github/Automattic/simplenote-ios/3838/workflows/199eebfe-d403-4ca1-a651-8496d8998ecf/jobs/4113):

Screen Shot 2021-06-03 at 12 41 41 pm

Matched by this one locally:

Screen Shot 2021-06-03 at 12 37 58 pm

mokagio:
@charliescheer @jkmassel 

I was curious to see the reason for the error [in CI](https://circleci.com/gh/Automattic/simplenote-ios/4113?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) for 9120223 was so I run the lane that exercise `match` locally and because locally `readonly` is `true` (which is something we want to move away from, BTW) I went through and generated the missing provisioning profiles.

With `match` sorted, another minor issue was revealed: the path to the entitlements for the new extension was missing the `Simplenote/` component. Once I fixed that in 6dab269, the installable build worked 😄 

I was able to install it and to use the new note widget

<table>
<tr>
	<td>

IMG_4999

</td>
	<td>

IMG_5001


</td>
</table>

jkmassel:
Thanks for sorting that @mokagio!! I think I had to mess with the app groups, and I probably forgot to regenerate the internal/alpha profiles due to keychain issues I had today 😅

hardevine:
👍

SylvesterWilmott (approved):

PR #1297, opened 2021-05-30
Feature/background sync

charliescheer:
### Fix
This is a draft PR for adding background sync back into SN iOS.  At this point all this does is implement the BackgroundTasks API into the app and setup Simperium to do a background sync every 30 minutes or more (sync will be triggered by apple when they feel like it, but not before 30 minutes)

### Test
The background sync will print to the logs, so there are two options for testing.  First, run the simulator to a device (must be at least iOS 13) and leave it alone for a while.  As mentioned above the earliest begin date on the refresh is 30 minutes, if you want to make that faster go to 

`    @objc
    func scheduleAppRefresh() {
        let request = BGAppRefreshTaskRequest(identifier: BackgroundRefreshConstants.identifier)
        request.earliestBeginDate = Date(timeIntervalSinceNow: BackgroundRefreshConstants.earliestBeginDate)
        do {
            try BGTaskScheduler.shared.submit(request)
        } catch {
            print("Couldn't schedule app refersh: \(error)")
        }
    }`

and change request.earliestBeginDate to Date(timeIntervalSinceNow: 60) which will make it happen in a minute or more.  After giving it some time you should see some events appear in the logs for showing how the background sync went

OR,
1. run the app while connected to Xcode
2. pause execution in the debug tools
3. in lldb run this command 

`e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"com.codality.NotationalFlow.backgroundRefresh"]`

4. unpause running.  When you do this a background refresh will be simulated.  

### Review

### Release
> 
> > Added markdown support

### DIFF
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -425,11 +425,10 @@ extension SPAppDelegate {
         }
     }
 
-
     @objc
     func scheduleAppRefresh() {
         let request = BGAppRefreshTaskRequest(identifier: BackgroundRefreshConstants.identifier)
-        request.earliestBeginDate = Date(timeIntervalSinceNow: BackgroundRefreshConstants.earliestBeginDate) //15 minutes
+        request.earliestBeginDate = Date(timeIntervalSinceNow: BackgroundRefreshConstants.earliestBeginDate)
         do {
             try BGTaskScheduler.shared.submit(request)
         } catch {
@@ -440,5 +439,5 @@ extension SPAppDelegate {
 
 private struct BackgroundRefreshConstants {
     static let identifier = "com.codality.NotationalFlow.backgroundRefresh"
-    static let earliestBeginDate = 900.0
+    static let earliestBeginDate = 1800.0 //30 minutes
 }
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -1,4 +1,5 @@
 import Foundation
+import BackgroundTasks
 
 
 // MARK: - Initialization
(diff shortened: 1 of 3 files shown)


Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -394,3 +395,49 @@ extension SPAppDelegate {
...
+        simperium.backgroundFetch { result in
+            if result == .failed {
+                NSLog("<< Background Fetch: New Data Received")
+                task.setTaskCompleted(success: false)
+                return
+            } else if result == .noData {
+                NSLog("<< Background Fetch: No Data Received")
+            }
+            task.setTaskCompleted(success: true)
+        }
```
eshurakov, line 425: Thoughts:
* We should handle expiration of the task. `task.expirationHandler = { ... }`. Not sure if we can/should tell simperium to stop it's networking.
* When the app is waken up and you have a valid token, simperium library automatically connects to simperium and  starts syncing. Here we trigger another sync, which I think may result in 2 sync processes being run at the same time and possibly data corruption? Maybe we can "ride-along" somehow on the normal sync process and add some sort of notification when we think everything is synced 🤔 

/cc @jleandroperez

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -394,3 +395,49 @@ extension SPAppDelegate {
...
+        do {
+            try BGTaskScheduler.shared.submit(request)
+        } catch {
+            print("Couldn't schedule app refersh: \(error)")
+        }
+    }
+}
+
+private struct BackgroundRefreshConstants {
+    static let identifier = "com.codality.NotationalFlow.backgroundRefresh"
```
jleandroperez, line 441: We should avoid hardcoding the bundleIdentifier, since it may actually differ, depending on which target we're using:

```
let bundleIdentifier = Bundle.main.bundleIdentifier ?? "com.codality.NotationalFlow"
let bgTaskIdentifier = bundleIdentifier + ".refresh"
```

Note that the `bundleIdentifier` getter should never fail, but adding a fallback just in case

jleandroperez: Screen Shot 2021-05-31 at 2 37 12 PM

Simplenote/Simplenote-Info.plist
```diff
@@ -78,8 +78,16 @@
...
 	</array>
+	<key>UIBackgroundModes</key>
+	<array>
+		<string>fetch</string>
+	</array>
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
+	<key>BGTaskSchedulerPermittedIdentifiers</key>
+	<array>
+		<string>com.codality.NotationalFlow.backgroundRefresh</string>
```
jleandroperez, line 89:
```suggestion
		<string>$(PRODUCT_BUNDLE_IDENTIFIER).refresh</string>
```

jleandroperez: **Why:** Because we've got multiple Bundle Identifiers, and I'm not sure how iOS behaves if the Task Identifier does not contain our actual fully qualitifed Bundle ID

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -394,3 +395,49 @@ extension SPAppDelegate {
...
+            guard let task = task as? BGAppRefreshTask else {
+                return
+            }
+            self.handleAppRefresh(task: task)
+        }
+    }
+
+    private func handleAppRefresh(task: BGAppRefreshTask) {
+        scheduleAppRefresh()
+        print("Ran handle app refersh")
```
jleandroperez, line 415:
```suggestion
        NSLog("Ran handle app refresh")
```

Simplenote/SPAppDelegate.m
```diff
@@ -189,6 +194,9 @@ - (void)applicationDidEnterBackground:(UIApplication *)application
 
     [self showPasscodeLockIfNecessary];
     [self cleanupScrollPositionCache];
+
+    //Schedule background refresh
+    [self scheduleAppRefresh];
```
jleandroperez, line 199: Note that `BGAppRefreshTaskRequest` is iOS +13 as well, we should guard this invocation

Simplenote/SPAppDelegate.m
```diff
@@ -189,6 +194,9 @@ - (void)applicationDidEnterBackground:(UIApplication *)application
 
     [self showPasscodeLockIfNecessary];
     [self cleanupScrollPositionCache];
+
+    //Schedule background refresh
```
jleandroperez, line 198: Nipticky: Missing space between `//` and `Text`

jleandroperez (commented):
We've been discussing this one with @eshurakov, and I think we should:

1. Remap this PR against `develop` (since it's standalone!)
2. Add a guard so that this only works when `BuildConfiguration.debug`
3. "Ride Along":

```
 private func handleAppRefresh(task: BGAppRefreshTask) {
     scheduleAppRefresh()
     
     DispatchQueue.main.asyncAfter(deadline: .now() + timeOut) {
         task.setTaskCompleted(success: true)
     }
}
```

This is in the lines of "just have the app launched, and wait for Simperium to do the rest".

As far as I understand, when BG fetch happens, iOS launches the app in Backgrounf mode. The lib should open the websocket interface, and sync **should happen** without further action required.

What do you folks think?

PR #1299, opened 2021-05-30
Alert users that biometric lock requires setting up a pin code

### LINKED ISSUE #1256: Security: Improve Flow
### Details:

Whenever the **FaceID** row is pressed and there is no Passcode set, we should display an Alert as follows:

cc @SylvesterWilmott

charliescheer:
### Fix
This PR fixes #1256 

To be able to use the biometric locks on iOS (face ID or touch ID) a user needs to also setup a pin lock code.  Currently when you try to set it up without having a pin code setup, the pin code setup appears without any notification.  

In this PR I have added an alert that pops up informing the user that they need a pin code to use the biometrics and giving them a chance to concur or to dismiss the setup.

Note that in the original issue the request was just for this alert to setup face ID, but touch id has the same need of setting up a pin code, so I extended the change to deal with touch ID as well. cc: @SylvesterWilmott 

Face ID:

Touch ID:

### Test
1. on a device with face id go to simplenote > sidebar > Settings
2. once there disable the pin lock if it is currently enabled
3. tap on the biometric switch. An alert asking you to setup the pin code will appear
4. test canceling the setup (the switch should switch to off) and test setting up a pin code
5. Repeat steps with a touch id device if possible  (make sure the touch id/face id titles show up correctly based on the device used)

### Review

### DIFF
--- a/Simplenote/Classes/SPSettingsViewController.m
+++ b/Simplenote/Classes/SPSettingsViewController.m
@@ -668,18 +668,19 @@ typedef NS_ENUM(NSInteger, SPOptionsDebugRow) {
 
 - (UIAlertController*)pinLockRequiredAlert
 {
-    NSString* alertTitle = NSLocalizedString(([NSString stringWithFormat:@"To enable %@, you must have a passcode setup first.", self.biometryTitle]),
-                                             @"Prompt to setup passcode before biomtery lock");
-    UIAlertController* alert = [UIAlertController alertControllerWithTitle:self.biometryTitle
+    NSString *alertTitleTemplate = NSLocalizedString(@"To enable %1$@, you must have a passcode setup first.",
+                                                     @"Prompt to setup passcode before biomtery lock. Parameters: %1$@ - biometry type (Face ID or Touch ID");
+    NSString *alertTitle = [NSString stringWithFormat:alertTitleTemplate, self.biometryTitle];
+    UIAlertController *alert = [UIAlertController alertControllerWithTitle:self.biometryTitle
                                                                    message:alertTitle
                                                             preferredStyle:UIAlertControllerStyleAlert];
 
-    NSString* setupTitle = NSLocalizedString(@"Set up Passcode", @"Setup a passcode action button");
+    NSString *setupTitle = NSLocalizedString(@"Set up Passcode", @"Setup a passcode action button");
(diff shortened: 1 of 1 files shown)


Simplenote/Classes/SPSettingsViewController.m
```diff
@@ -660,11 +660,33 @@ - (void)touchIdSwitchDidChangeValue:(UISwitch *)sender
 {
     SPPinLockManager.shared.shouldUseBiometry = sender.on;
 
-    if (![self isPinLockEnabled]) {
-        [self showPinLockViewController];
+    if (![self isPinLockEnabled] && [self.biometrySwitch isOn]) {
+        UIAlertController* alert = [self pinLockRequiredAlert];
```
eshurakov, line 664: Nitpick: `*` should be glued to the variable name: `UIAlertController *alert`

Simplenote/Classes/SPSettingsViewController.m
```diff
@@ -660,11 +660,33 @@ - (void)touchIdSwitchDidChangeValue:(UISwitch *)sender
...
+        [self presentViewController:alert animated:YES completion:nil];
     }
 }
 
+- (UIAlertController*)pinLockRequiredAlert
+{
+    NSString* alertTitle = NSLocalizedString(([NSString stringWithFormat:@"To enable %@, you must have a passcode setup first.", self.biometryTitle]),
+                                             @"Prompt to setup passcode before biomtery lock");
```
eshurakov: Note: this localization string won't be correctly exported because localization "key" is known only in runtime.
Basically it is now like this:
```
NSString *localizationKey = can-be-anything;
NSString *alertTitle = NSLocalizedString(localizationKey, @"...");
```

When we try to export localization keys to get them translated, the tool won't know what is the key. It is a variable, so it can be anything.

Localization keys have to be explicit. In this particular case we can rewrite it like this:

```
NSString *alertTitleTemplate = NSLocalizedString(@"To enable %1$@, you must have a passcode setup first.", @"Prompt to setup passcode before biomtery lock. Paramteres: %1$@ - biometry type (Face ID or Touch ID)");
NSString *alertTitle = [NSString stringWithFormat: alertTitleTemplate, self.biometryTitle];
```

eshurakov (approved):
Works as advertised 👍 

Pre-approving now. After fixing the issue with localized string should be good to go!

PR #1301, opened 2021-05-31
New scheme with a subset of UI tests + test accounts sanitiser

pachlava:
This PR ended up being more bloated that I expected initially, because of the issue I've found and had to solve after the PR was created. I'm sorry if PR is too noisy.

**Initial:** This PR contains a new scheme with a subset of UI tests. The current selection takes less than 5 min to run on CI, while touches most parts of the app.

Besides a new scheme, there's also a new optional CircleCI job, that will run _all_ UI tests.

When I tested the execution on CI, I initially ran into flakiness for preview and search tests. It looks like the cause is in performance of VMs: it takes more time to render the markdown and to display the search results. For now, I 'fixed' this by increasing a delay after 'search' / 'swipe to preview' is executed.

Also, I experimented with parallel execution of tests (not using tests account in parallel, but using several Simulators in parallel). Theoretically, the set of login tests can run in parallel to any other set of tests, but actually there were fails even locally. Taking into account the lower performance of VM, I decided not to play with multiple simulators for now.

**Addition:** I've missed a very obvious case of accounts status getting stuck as occupied in case if the job was cancelled (e.g. by a new commit to PR, while the UI tests were still running). After the discussion in p1622548675025100-slack-CT80U6X0W, thanks to the idea from @AliSoftware, I've made a number of additional changes, listed as 4-6 below.

### Changes
1. The `UI Tests` job in `.circleci/config.yml` now accepts Xcode scheme name as a parameter. This allows to run the job both on a subset of tests (required), and full set of tests (optional).
2. New file for `SimplenoteUITests_Subset` scheme.
3. Increased sleep from one/two seconds to five for 'swipe to preview', 'search', and `link tap`.
4. Account data is now stored in the form of json formatted string. Example: `{"0":"free","1":"4098","2":"free","3":"free"}`. If the account is taken, it's marked with current job number.
5. New method `sanitize_test_accounts` is executed at the beginning of `pick_test_account_and_run_ui_tests` lane, and checks if the job numbers mentioned in the json are still running. If they are not running, the account will be marked as free.
6. In order to do 4 and 5, I've added a few new methods and changed the old ones to improve logging.

### Test
Now it's finally possible to test simultaneous account selection/usage easily:
1. Running `UI Tests Subset` (7 tests) causes no fails
2. Running `UI Tests Full` (33 tests) in parallel to ^ causes no fails either (sometimes it might, for the search tests, I feel it's still about the rendering performance)
3. While tests are running, the workflow rerun will cancel the job. The 'locked' accounts will be freed up during the next `UI Tests...` job and used.

### Review

### Release


SimplenoteUITests/NoteEditor.swift
```diff
@@ -57,7 +57,7 @@ class NoteEditor {
 
     class func swipeToPreview() {
         app.textViews.firstMatch.swipeLeft()
-        sleep(1)
+        sleep(5)
```
mokagio, line 60: At some point, it would be great to replace these `sleep()` calls with something that polls the UI looking for a criteria that can validate whether the element has loaded. Doing so would make the tests more robust.

Having said that, I acknowledge it's a challenging thing to do and sometime may require a lot of extra work on the production code to expose hooks we can use to validate the UI from here.

fastlane/Fastfile
```diff
@@ -703,20 +702,19 @@ end
   # selected for now, there will be no credentials race condition.
   # -----------------------------------------------------------------------------------
   # Usage:
-  # bundle exec fastlane run_ui_tests [device:<Name of the iOS Simulator>] [test_account:<Test account email>]
+  # bundle exec fastlane run_ui_tests [sheme:<Xcode scheme with UI tests>] [device:<Name of the iOS Simulator>] [test_account:<Test account email>]
```
mokagio:
```suggestion
  # bundle exec fastlane run_ui_tests [scheme:<Xcode scheme with UI tests>] [device:<Name of the iOS Simulator>] [test_account:<Test account email>]
```

fastlane/Fastfile
```diff
@@ -703,20 +702,19 @@ end
...
   #
   # Example:
-  # bundle exec fastlane run_ui_tests device:"iPhone 12" test_account:"test.account@test.com"
+  # bundle exec fastlane run_ui_tests sheme:"SimplenoteUITests_Subset" device:"iPhone 12" test_account:"test.account@test.com"
```
mokagio:
```suggestion
  # bundle exec fastlane run_ui_tests scheme:"SimplenoteUITests_Subset" device:"iPhone 12" test_account:"test.account@test.com"
```

fastlane/Fastfile
```diff
@@ -936,5 +964,5 @@ def change_test_account_availability(availability)
     http.request(request)
   end
 
-  get_test_accounts_state
-end
+  UI.message("Accounts state after write: #{get_test_accounts_hash}")
+end
```
mokagio: Informational nitpick. 

With this symbol, GitHub is telling us that there is no new line at the end of the file.

image

The convention of having a new line at the end of the files comes from an old C decision that has been passed down through Unix history:

> A source file that is not empty shall end in a new-line character, which shall not be immediately preceded by a backslash character.
> 
> Since this is a “shall” clause, we must emit a diagnostic message for a violation of this rule.

So, it turns out that, according to POSIX, every text file (including Ruby and JavaScript source files) should end with a `\n`, or “newline” (not “a new line”) character. This acts as the `eol`, or the “end of line” character. It is a line “terminator”. ([Source](https://thoughtbot.com/blog/no-newline-at-end-of-file))

**No need to address this now**, the next time someone with an editor configured to fix this will touch the file, it'll go away. 👍

mokagio (commented):
Thanks for this work @pachlava!

I just run the "subset" tests locally and they failed. I need to investigate why, maybe it has something to do with my setup or the state of my repo.

In the meantime, I wanted to log these comments.

.circleci/config.yml
```diff
@@ -145,7 +145,9 @@ jobs:
       device:
         description: The device (e.g. "iPhone 12") to use when running tests.
         type: string
-        default: "iPhone 12"
+      scheme:
+        description: Xcode scheme with tests.
```
mokagio: What do you think of this, to keep it consistent with the other parameter?

```suggestion
        description: Xcode scheme to use to run the tests.
```

mokagio (commented):
> The issue was on my end 😄 My simulator didn't have the software keyboard enabled so the automation could type into the login text fields...

Screen Shot 2021-06-03 at 2 47 29 pm

Once I toggled the keyboard, the "subset" scheme passed. The full suite failed on the web view test:

I just run the "subset" tests locally and they failed. I need to investigate why, maybe it has something to do with my setup or the state of my repo.

Screen Shot 2021-06-03 at 2 43 18 pm

That looked to me like some kind of timeout, or bug with how the simulator handles long presses on links (which I experienced in the past). I re-run the tests by itself and it passed 👍 

I'm going to hold approving this till I reviewed the base one.

.circleci/config.yml
pachlava: Fully agree, thanks!

pachlava:
>  I just run the "subset" tests locally and they failed. I need to investigate why, maybe it has something to do with my setup or the state of my repo.
> That looked to me like some kind of timeout, or bug with how the simulator handles long presses on links (which I experienced in the past). I re-run the tests by itself and it passed 👍

This is exactly the problem that started occurring as soon as full set started running on CI, and I tried to 'solve' this with longer sleeps.

SimplenoteUITests/NoteEditor.swift
pachlava: I agree. This is a thing I'd like to address in one of the upcoming PRs. `sleep(1)` looks harmless and easy to justify, until the moment when it turns to `sleep(11)`. As always, here I did it to quickly unblock the 'main' task for the PR - which was the accounts handling.

pachlava:
@mokagio huge thanks for the review, as always 🌟! I have a feeling that I have polluted the `Fastfile` by adding ~100 lines of code that is only related to account selection. Do you see a point in moving this to a separate file?

AliSoftware:
I won't have time to review these changes today tbh, but just wanted to drop a thought after reading the PR description:

Is there a reason why you created a dedicated scheme for the test subset, as opposed to using an `xctestplan` to provide different configurations and subsets of tests being selected to run for the same scheme?

I have to admit that I haven't played with test plans much, but even if using a separate scheme for this should also work, I think test plans would fit the use case better here, as they are kinda made for this, right? 🤔 

---

[EDIT] Just saw this paragraph [in the other PR's description](https://github.com/Automattic/simplenote-ios/pull/1292)… which I guess answers my question above after all 😅 👍 

> As for the upcoming subset of tests: it looks like test plans will not really fit the purpose here, because they are more about running tests under different configurations, rather than having different selections of tests. It looks like having additional scheme(s) is still the easiest way to accomplish this.

fastlane/Fastfile
```diff
@@ -879,54 +878,83 @@ def simulator_version
...
+  UI.message("Freeing used test account #{$used_test_account_index}")
+  change_test_account_availability("free")
   $used_test_account_index = nil
 end
 
-# Returns the value of test accounts key, which is 
-# expected to look like a series of Fs and Ts.
+# Test accounts info is stored in JSON-formatted string, where keys represent
+# the account index, and values are either "free" or CIRCLE_BUILD_NUM, in case if account
+# was occupied by a buld. Returns the hash.
```
AliSoftware:
```suggestion
# was occupied by a build. Returns the hash.
```

fastlane/Fastfile
```diff
@@ -879,54 +878,83 @@ def simulator_version
...
+# Crashed fastlane if free account was not found.
 def find_free_test_account
-  $used_test_account_index = get_test_accounts_state.index("F")
+  accounts_hash = get_test_accounts_hash
+  UI.message("Looking for a free test account...")
+  UI.message("Accounts state: #{accounts_hash}")
+
+  accounts_hash.each do |key, value|
+    break $used_test_account_index = key.to_i if value == "free"
+  end
```
AliSoftware: iinm you could use [`Hash#key()`](https://ruby-doc.org/core-2.7.0/Hash.html#method-i-key) for a simpler code here
```suggestion
  $used_test_account_index = accounts_hash.key('free')
```

tested in the `pry` REPL
```ruby
$ pry
[1] pry(main)> json = { "0":"12843", "1":"12345", "2":"free", "3":"14032", "4":"free" }
=> {:"0"=>"12843", :"1"=>"12345", :"2"=>"free", :"3"=>"14032", :"4"=>"free"}
[2] pry(main)> json.key('free')
=> :"2"
[3] pry(main)> json2 = { "0": "1234", "1":"5676" }
=> {:"0"=>"1234", :"1"=>"5676"}
[4] pry(main)> json2.key('free')
=> nil
```

fastlane/Fastfile
```diff
@@ -879,54 +878,83 @@ def simulator_version
...
+  accounts_hash = get_test_accounts_hash
+  UI.message("Sanitizing stale test accounts...")
+  UI.message("Accounts before sanitizing: #{accounts_hash}")
+
+  accounts_hash.each do |key, value|
+    if value != "free" && is_job_running(value) == false
+      accounts_hash[key] = "free"
+      submit_accounts_hash(accounts_hash)
+    end
+  end
```
AliSoftware: You seem to post (submit) the JSON on each iteration in which you find a case to sanitize, is that on purpose, to limit the race condition given `is_job_running` requires a network call and can thus take time? (As opposed to only submitting the updated hash at the end)

```suggestion
  accounts_hash.each do |key, value|
    next if value == "free" || is_job_running(value)
    accounts_hash[key] = "free"
    submit_accounts_hash(accounts_hash)
  end
```

pachlava: This is a beautiful improvement of readability, thanks @AliSoftware (and thanks for jumping in, in general)!

> You seem to post (submit) the JSON on each iteration in which you find a case to sanitize, is that on purpose, to limit the race condition given is_job_running requires a network call and can thus take time? (As opposed to only submitting the updated hash at the end)

I was thinking of limiting the race condition initially, yes. Now, after trying submitting the hash only once at the end, I realize that another reason for doing this the initial way is that we are sure we submit changes only when needed. If no changes were made to the hash, but we move to submit to the end, we will need to additionally make sure that we only submit if the final hash is different from the initial, i.e. `submit_accounts_hash(final_accounts_hash) if final_accounts_hash != accounts_hash`. And this is actually what I did 🙂 

I'm really not sure what is better - having less code/variables or having less requests? 🤔

AliSoftware: That's a good question. I think updating on every loop rather than at the end will still be safer tho.

Especially imagine job 1 starts and do the cleanup only after checking the status of _all_ jobs occupying an account, which could take a couple of seconds in total. While it does this, job 2 starts and start doing the same.

After the few seconds that was required by job 1 to do all the requests to check the status of all jobs occupying an account, it finally uploads the updated JSON, picks an account for itself, and upload again.

In parallel job 2 will still be processing the JSON from before job 1 updated anything and will not see that job 1 has since taken one of the accounts for itself…

pachlava:
> I have to admit that I haven't played with test plans much, but even if using a separate scheme for this should also work, I think test plans would fit the use case better here, as they are kinda made for this, right? 🤔
> 
> [EDIT] Just saw this paragraph in the other PR's description… which I guess answers my question above after all 😅 👍

@AliSoftware  I also have to admit that I might have missed something, and there is a way to change the tests selection in a test plan, still. I'll recheck it again now.

fastlane/Fastfile
pachlava: Thanks a lot @AliSoftware ! It's amazing to see these improvements, for such a Ruby noob like me. Also, I think I like Ruby quite a lot, at least on the surface that I have seen.

pachlava:
@AliSoftware after some more reading/trying, it looks like the only (if I'm not mistaken) way to have different subsets of tests via test plans is having different test plans. A single test plan can have multiple configurations, when it comes to properties like localization, screenshots generation, etc. But the tests selection is not among these properties.

i.e. I'll need to create two test plans, add them to a scheme, and then it will be possible to define the test plan to run inside a scheme. So, again, it brings a similar question - what is better: having one scheme and two test plans, or two schemes and zero test plans? 🙂 💭

AliSoftware:
Yeah given that, at that point it's simpler to keep the 2 schemes and no test plan then!

fastlane/Fastfile
```diff
@@ -879,54 +878,80 @@ def simulator_version
...
-  $used_test_account_index = get_test_accounts_state.index("F")
+  accounts_hash = get_test_accounts_hash
+  UI.message("Looking for a free test account...")
+  UI.message("Accounts state: #{accounts_hash}")
+
+  $used_test_account_index = accounts_hash.key('free')
+
   UI.user_error!("Could not find free UI Test account. Quitting.") if $used_test_account_index.nil? 
   UI.message("Free account index: #{$used_test_account_index}")
+  change_test_account_availability("#{get_required_env("CIRCLE_BUILD_NUM")}")
```
mokagio, line 921: Using `CIRCLE_BUILD_NUM` is pretty neat 👍

fastlane/Fastfile
```diff
@@ -879,54 +878,80 @@ def simulator_version
...
+  end  
+end
+
+def is_job_running(job_number)
+  uri = URI.parse("https://circleci.com/api/v2/project/gh/Automattic/simplenote-ios/job/#{job_number}")
+  response = Net::HTTP.get_response(uri)
+  return if response.code != "200"
+
+  response_hash = JSON.parse(response.body)
+  puts "Status of Job Number #{job_number} is #{response_hash["status"]}"
```
mokagio: Maybe this `puts` was a debugging feature, but I think it might be useful to keep it in here. It could become:

```suggestion
  UI.message("Status of CircleCI Job Number #{job_number} is #{response_hash['status']}")
```

Or maybe `verbose` if we want to print it only when a developer is looking for extra information.

```suggestion
  UI.verbose("Status of CircleCI Job Number #{job_number} is #{response_hash['status']}")
```

---

Given the logs are already "noisy" and that if a CI job fails it's helpful to have as much information about why in the log for that job itself, because re-running the job might result into a different behavior, I'd say let's use `message`.

mokagio (approved):
👍 

Looking at the CI logs, we can see the accounts management working as desired  👏 

Here's the subset UI tests lane starting up:

```
[13:52:27]: Sanitizing stale test accounts...
[13:52:27]: Accounts before sanitizing: {"0"=>"4143", "1"=>"free", "2"=>"free", "3"=>"free"}
[13:52:27]: Status of Job Number 4143 is canceled
[13:52:27]: Writing new accounts state: {"0"=>"free", "1"=>"free", "2"=>"free", "3"=>"free"}
[13:52:27]: Accounts state after write: {"0"=>"free", "1"=>"free", "2"=>"free", "3"=>"free"}
[13:52:27]: Looking for a free test account...
[13:52:27]: Accounts state: {"0"=>"free", "1"=>"free", "2"=>"free", "3"=>"free"}
[13:52:27]: Free account index: 0
[13:52:28]: Writing new accounts state: {"0"=>"4146", "1"=>"free", "2"=>"free", "3"=>"free"}
[13:52:28]: Accounts state after write: {"0"=>"4146", "1"=>"free", "2"=>"free", "3"=>"free"}
```

While that run, you unblocked the full UI tests step. When they started, they found one account busy, as configured by the subset lane just before.

```
[13:55:09]: Sanitizing stale test accounts...
[13:55:09]: Accounts before sanitizing: {"0"=>"4146", "1"=>"free", "2"=>"free", "3"=>"free"}
[13:55:09]: Status of Job Number 4146 is running
[13:55:10]: Looking for a free test account...
[13:55:10]: Accounts state: {"0"=>"4146", "1"=>"free", "2"=>"free", "3"=>"free"}
[13:55:10]: Free account index: 1
[13:55:10]: Writing new accounts state: {"0"=>"4146", "1"=>"4147", "2"=>"free", "3"=>"free"}
[13:55:10]: Accounts state after write: {"0"=>"4146", "1"=>"4147", "2"=>"free", "3"=>"free"}
```

When the subset UI tests finish and cleanup their state, they find the account `1` busy:

```
[14:02:45]: Cruising back to lane 'ios pick_test_account_and_run_ui_tests' 🚘
[14:02:45]: Freeing used test account 0
[14:02:46]: Writing new accounts state: {"0"=>"free", "1"=>"4147", "2"=>"free", "3"=>"free"}
[14:02:46]: Accounts state after write: {"0"=>"free", "1"=>"4147", "2"=>"free", "3"=>"free"}
```

Finally, the full UI tests lane cleans up once its tests passed

```
[14:21:11]: Cruising back to lane 'ios pick_test_account_and_run_ui_tests' 🚘
[14:21:11]: Freeing used test account 1
[14:21:11]: Writing new accounts state: {"0"=>"free", "1"=>"free", "2"=>"free", "3"=>"free"}
[14:21:11]: Accounts state after write: {"0"=>"free", "1"=>"free", "2"=>"free", "3"=>"free"}
```

Beautiful!

fastlane/Fastfile
pachlava: Yes, the `puts` slipped in here from the local file where I was testing this. I intended to keep the message, but did not change it to `UI.message`. Thank you!

fastlane/Fastfile
pachlava: Thank you Gio! 🎆

PR #1303, opened 2021-06-01
Feature/bg fetch

charliescheer:
### Fix
This is a continuation of the original PR #1297  I moved the changes to a new branch, so created a new PR to isolate those features from the widgets changes.

Note that in this new PR I have changed what the fetch is doing.  Now the fetch simply wakes the app and lets it run for a time, shutting down before we hit the apple limit.  I am still testing if it is actually doing the bg fetch on it's own.  If triggered from lldb everything seems to work fine, but iOS hasn't wanted to do background updates for me today, so still trying to see how it responds.  Will comment as I know more about what happens if I can get it to actually log a fetch

This is a draft PR for adding background sync back into SN iOS.  At this point all this does is implement the BackgroundTasks API into the app and setup Simperium to do a background sync every 30 minutes or more (sync will be triggered by apple when they feel like it, but not before 30 minutes)

### Test
The background sync will print to the logs, so there are two options for testing.  First, run the simulator to a device (must be at least iOS 13) and leave it alone for a while.  As mentioned above the earliest begin date on the refresh is 30 minutes, if you want to make that faster go to 

`    @objc
    func scheduleAppRefresh() {
        let request = BGAppRefreshTaskRequest(identifier: BackgroundRefreshConstants.identifier)
        request.earliestBeginDate = Date(timeIntervalSinceNow: BackgroundRefreshConstants.earliestBeginDate)
        do {
            try BGTaskScheduler.shared.submit(request)
        } catch {
            print("Couldn't schedule app refersh: \(error)")
        }
    }`

and change request.earliestBeginDate to Date(timeIntervalSinceNow: 60) which will make it happen in a minute or more.  After giving it some time you should see some events appear in the logs for showing how the background sync went

OR,
1. run the app while connected to Xcode
2. pause execution in the debug tools
3. in lldb run this command 

`e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"com.codality.NotationalFlow.backgroundRefresh"]`

4. unpause running.  When you do this a background refresh will be simulated.  

### Review

### Release
> 
> > Added markdown support


charliescheer:
Okay, so it turns out when I leave iOS alone and let it do it's thing the BG Fetch does happen when iOS is ready for it.  I was able to see it do a few cycles of refreshing in the background and from what I am seeing in the console logs it seems to be working as expected.  The app wakes up, reconnects to the web socket, does some updating, and sits quietly till we shut it down. 

The way I tested this was:
1.  connecting my device to my computer and opening the console app.  
2. Selecting the device from the left panel and then searching for "background refresh" (there should be nothing there now
3. Run the app from Xcode to the device.  Once it loads, return the to the home screen on the device.  You should see "Background Refresh Scheduled" in the console now
4. Walk away for a while.  In the PR the delay time is set to 30 minutes, but you can drop that down to 1 minute and it should cause the sync more frequently.  I was seeing it once every ten minutes or so when I had it set to trigger each minute.
5. When you return, you should see more background refresh messages.  Select one of the "background refresh initiated" messages and change the search to "Simplenote" to see how it responds.

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -405,3 +406,61 @@ extension SPAppDelegate {
...
+        }
+
+        NSLog("Background refresh intiated")
+        scheduleAppRefresh()
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + BackgroundRefreshConstants.timeOut) {
+            NSLog("Background refresh finishing")
+            task.setTaskCompleted(success: true)
+        }
+    }
```
eshurakov: We should also set an expiration handler, because according to [Apple Docs](https://developer.apple.com/documentation/backgroundtasks/bgtask/3142236-settaskcompleted):
> If you don’t set an expiration handler, the system will mark your task as complete and unsuccessful instead of sending a warning.

If we set an expiration handler I think there is no reason to have our own timeout.

---
Other ideas:
* We can extract all BG related code to a separate class to clean up app delegate
* To avoid always waiting for BG task to expire we can try to be smart: re-set a timer for x seconds on every change that we receive (`- (void)bucket:(SPBucket *)bucket didChangeObjectForKey:(NSString *)key....`). So when all changes are received and processed, we wait a bit and finish BG task.

Simplenote/SPAppDelegate.m
```diff
@@ -420,6 +430,10 @@ - (void)setSelectedTag:(NSString *)selectedTag {
 
 - (void)bucket:(SPBucket *)bucket didChangeObjectForKey:(NSString *)key forChangeType:(SPBucketChangeType)change memberNames:(NSArray *)memberNames
 {
+    if (@available(iOS 13.0, *)) {
+        [self.refreshManager refreshTimer];
```
eshurakov: I think it's better to name this method by saying what is happening instead of what should be done. 

For example if we say `[self.refreshManager onSimperiumObjectChange]`, `refreshManager` can decide what it wants to do in this case (and maybe it doesn't even want to `refreshTimer`)

Simplenote/SPAppDelegate.m
```diff
@@ -190,6 +195,11 @@ - (void)applicationDidEnterBackground:(UIApplication *)application
 
     [self showPasscodeLockIfNecessary];
     [self cleanupScrollPositionCache];
+
+    // Schedule background refresh
+    if (@available(iOS 13.0, *)) {
+        [self.refreshManager scheduleAppRefresh];
```
eshurakov, line 201: We probably should also cancel tasks if the app goes back to foreground.

Simplenote/BackgroundRefreshManager.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    private var handler: (()->Void)?
+
+    var finished: Bool = false {
+        didSet {
+            if finished == true {
+                handler?()
+            }
+        }
+    }
```
eshurakov: I don't think we need to track `finished`

Simplenote/BackgroundRefreshManager.swift
```diff
@@ -0,0 +1,101 @@
...
+    private func finishRefresh() {
+        guard let handler = handler else {
+            return
+        }
+        NSLog("Finish Refresh Called")
+
+        handler()
+
+        finished = false
+        self.handler = nil
```
eshurakov, line 31: it's good to reset a timer here as well

Simplenote/SPAppDelegate.m
charliescheer: Yeah, I was thinking that, but didn't see anything in the docs that required it.  What I did notice digging into the logs yesterday is that the background service only allows for one scheduled event at a time. If you schedule a new one it removes the old one first.  But it makes sense to me to clear out any pending ones when the app enters the foreground.  Will add that today

Simplenote/BackgroundRefreshManager.swift
charliescheer: A totally.  The finished variable was something I had put in because I was having issues making an instance of the backgroundFetchManager in the app delegate.  Was trying to use changing it to trigger ending the refresh, but using the handler instead can serve that same purpose.  I have removed the finished var.

charliescheer:
Thanks for the suggestions @eshurakov  I have updated the PR per these suggestions. 

Did some more testing today and it is working for me again!  Seems that I made iOS mad yesterday by not shutting down the BG task

PR #1315, opened 2021-06-07
Replace UI tests helper local framework with remote Swift package

mokagio:
As part of the 2nd HACK week for 2021, I'm working on UI testing shared infrastructure.

As step towards sharing more code across projects was extracting general-purpose helpers present in the Simplenote iOS project into a dedicated framework.

The second was to create a [Swift package](https://github.com/Automattic/SwiftUITestsHelpers) from that code.

With this diff, we replace the local framework with its Swift package version.

### Test
Verify that the UI tests pass in CI.

Optionally, checkout this branch and try to build for testing both "SimplenoteUITests" and "SimplenoteScreenshots" targets.

### Review

### Release

---

Leaving this as a draft till later on in the HACK week because we're pointing to `trunk` for the package and I'm not sure about the package and repository name.


pachlava (approved):
Thank you for doing this @mokagio!

I've executed the Screenshots and UI Tests locally, with no fails. The only issue was that fastlane forced me to update Snapshot Helper in order to run. I saw no related warnings when executed screenshots from Xcode. But, this is not related to the change.

One thing I've noticed is that Xcode is showing this warning about `Info.plist`:

Screenshot_2021-06-08 00 31 21_SktsIf

I did not run the tests on CI, since this branch does not include the latest changes in CI/Fastfile for this, but I see no reasons why this should not work on CI.

mokagio:
Thanks for reporting those @pachlava. I'll take a look 👀

mokagio:
FYI, I addressed the `snapshot` update issue [here](https://github.com/Automattic/simplenote-ios/pull/1316) and the `Info.plist` warning in 81924734.

mokagio:
I'm going to bank on @pachlava's previous approval because the changes I made since then are not functionality related:

- Renamed the package from SwiftUITestsHelpers to XCUITestHelpers
- Switch to SemVer with `~> 0.1` rule instead of fetching from `trunk` to manage the dependency

PR #1316, opened 2021-06-08
Update Fastlane and its Simulator snapshot helper

mokagio:
Just a routine update. @pachlava noticed that `snapshot` was out of date and out of sync with Fastlane, which blocks the execution of the screenshots automation.

### Test
`bf take_screenshots devices:'iPhone Xs Max'` should succeed on your machine. I constrained the build to "iPhone Xs Max" just to save time locally.

Screen Shot 2021-06-08 at 2 41 04 pm

### Review

### Release


pachlava (approved):
Ran successfully on first try 🥳, with iPhone XS Max, firstly using light mode, then dark. After both iterations, the available screenshots were opened in the browser: very handy!

SimplenoteScreenshots/SnapshotHelper.swift
```diff
@@ -176,7 +180,11 @@ open class Snapshot: NSObject {
                 simulator = regex.stringByReplacingMatches(in: simulator, range: range, withTemplate: "")
 
                 let path = screenshotsDir.appendingPathComponent("\(simulator)-\(name).png")
-                try image.pngData()?.write(to: path, options: .atomic)
+                #if swift(<5.0)
+                    UIImagePNGRepresentation(image)?.write(to: path, options: .atomic)
+                #else
+                    try image.pngData()?.write(to: path, options: .atomic)
+                #endif
```
pachlava, line 187: Just curious what is the difference between those two? The parameters and return values description for [UIImagePNGRepresentation](https://developer.apple.com/documentation/uikit/1624096-uiimagepngrepresentation) and [pngData()](https://developer.apple.com/documentation/uikit/uiimage/1624096-pngdata) are the same. Is it about one being a method and the other being a function, hence using `pndData()` after it was introduced really made the code easier to read?

SimplenoteScreenshots/SnapshotHelper.swift
```diff
@@ -219,7 +227,7 @@ open class Snapshot: NSObject {
         #if os(OSX)
             let homeDir = URL(fileURLWithPath: NSHomeDirectory())
             return homeDir.appendingPathComponent(cachePath)
-        #elseif arch(i386) || arch(x86_64)
+        #elseif arch(i386) || arch(x86_64) || arch(arm64)
```
pachlava, line 230: This line made me go into another 🕳️ , and learn some new things! Nice!

pachlava: While I was reading about this, I saw such examples. Looks like this might do the similar trick?

```suggestion
        #elseif targetEnvironment(simulator)
```

I tested this locally and it worked. I might miss some context, though 🤔

SimplenoteScreenshots/SnapshotHelper.swift
mokagio: This code is automatically generated by Fastlane, so I'm not sure about the design decisions the authors took.

>  hence using `pndData()` after it was introduced really made the code easier to read?

I'd say so. It's also a more "Swifty" way of writing the code. It calls a method on an object instead of using a global free function (which the Swift API guide lines recommend not to use) and it also `throws` giving better error management capabilities.

My first guess was that the `try image.pngData...` call is something only available in Swift 5.0+, but that doesn't seem the case looking at the docs 🤔 

The fact that there is that `#if swift(<5.0)` check, though, definitely makes me think the authors had some reason to prefer `try image.pngData` over `UIImagePNGRepresentation` in code running on a runtime equal to or later than Swift 5.0.

SimplenoteScreenshots/SnapshotHelper.swift
mokagio: This code is automatically generated by Fastlane, so I'm not sure about the design decisions the authors took. Also, we should not touch it because the screenshots automation expects it work in a certain way.

Is your to use `targetEnvironment(simulator)` because it's platform agnostic? I think it's a great idea. You should open a pull request for that in Fastlane itself. The source for this file is [here](https://github.com/fastlane/fastlane/blob/master/snapshot/lib/assets/SnapshotHelper.swift).

👍

SimplenoteScreenshots/SnapshotHelper.swift
pachlava: Thanks for the extended answer! 😊

SimplenoteScreenshots/SnapshotHelper.swift
pachlava: > You should open a pull request for that in Fastlane itself.

This sounds exciting! Yes, it is to make it platform-agnostic

PR #1317, opened 2021-06-09
Update logic to sanitize keywords via Fastlane

mokagio:
App Store Connect requires a keyword string that is 100 characters or less with keywords separated by "English comma, Chinese comma, or any combination of the two."

This change makes sure that sanitized keywords use the English comma.

Note that I actually updated the source of truth for the keywords in [GlotPress](https://translate.wordpress.com/projects/simplenote/ios/release-notes/ar/default/), so this script should not be necessary anymore, but I'm pushing this just in case. Also, this is the code I used to generate the new keywords, so it's handy to have it here for reference, at least till we're sure they're good on GlotPress.

### Test

1. Checkout this branch
1. `git revert 222ef611` to bring the keywords back to a non compliant state
2. `bundle exec fastlane sanitize_appstore_keywords`

The result should match this:

```
برنامج simplenote, والملحوظة, والتدوين, والمُدوِّن, والملحوظات, والمزامنة, والمشاركة, وبسيط, والضبط
```

Notice that I didn't update the keywords metadata file. The reason is that, ideally, we'll download up to date keywords from GlotPress when it's time to finalize the release.

You can verify this by via `cd fastlane && ./download_metadata.swift`. The `metadata/ar-SA/keywords.txt` will update because of the changes I made in GlotPress.

### Review

### Release


fastlane/Fastfile
```diff
@@ -138,16 +138,20 @@ end
...
       else
-        separator = ','
+        # Nothing do do here, but we might have more languages
+        # with different commas in the future
       end
 
-      until keywords.length <= app_store_connect_keywords_length_limit do
-        keywords = keywords.split(separator)[0...-1].join(separator)
-        puts keywords
+      until keywords.length <= app_store_connect_keywords_length_limit
```
mokagio, line 150: The only thing changed in this line is that there isn't the `do` anymore. It's not necessary because the `until` block has only one line. Thanks Rubocop.

The reason it now has only one line is because I removed that debugging `puts` that I didn't notice when submitting the first version of this work.

fastlane/Fastfile
```diff
@@ -138,16 +138,20 @@ end
...
         separator = '،'
+        keywords = keywords.split(separator).join(english_comma)
       else
-        separator = ','
+        # Nothing do do here, but we might have more languages
+        # with different commas in the future
       end
 
-      until keywords.length <= app_store_connect_keywords_length_limit do
-        keywords = keywords.split(separator)[0...-1].join(separator)
```
mokagio, line 149: In the original version, I didn't know about / realize that ASC wants English commas. The code was trimming the keywords length all right, but because it was putting them back together with the Arabic comma, to ASC they would have looked like a single giant keyword 🤦

mokagio: (Why aren't they using an array? _shakes fist at the sky_)

oguzkocer (approved):
@mokagio I am a little sad to admit that it took me a while to understand what's going on with this code. The line comments, although helpful in providing context, was making it hard to read the changes, but even after turning off the comments, it took me a second to connect the dots.

This ^ might be just my morning slowness, but I wonder while fixing this code, did we maybe stick to the original a bit too much? As in, would it be better to simply replace the illegal separators with the `english_comma`, so there is no more `separator` keyword we are dealing with? In short, the following feels a bit much for just a simple character replacement:

```
when 'ar-SA'
   # Replace Arabic comma with English one as per App Store
   # Connect requirement
   separator = '،'
   keywords = keywords.split(separator).join(english_comma)
else
   # Nothing do do here, but we might have more languages
   # with different commas in the future
end
```

---

As for testing instructions, the 1-2-3 step one went OK. To my eye, the string you pasted and the string generated are the same. `git diff` doesn't agree, but I assume it's an invalid config on my part, so I think it's fine.

> You can verify this by via cd fastlane && ./download_metadata.swift. The metadata/ar-SA/keywords.txt will update because of the changes I made in GlotPress.

I first didn't understand why we did `cd` and not run the command in root, but looks like it just spits out the data to its folder :joy: This did change the `metadata/ar-SA/keywords.txt` and deleted a bunch of unrelated stuff. So, I think this is fine as well.

mokagio:
> This ^ might be just my morning slowness...

Nah, you're right, that code was a bit too convoluted. Using the switch statement would make sense if there were more cases, but with only one it adds cruft.

> In short, the following feels a bit much for just a simple character replacement

Totally. I should have taken a further step backward 😄 This is why it's great to do code reviews, hey!

I'll requested a new review from you to check my simplified code and enabled auto-merge to close the loop on this.

oguzkocer:
@mokagio It looks like auto-merge might have merged your changes prematurely. I see that there has been a force push as well, so.. not sure how to proceed with this one.

oguzkocer:
@mokagio fwiw, I got a ping in this commit d589edc309cc0976beb6ee5cacb8e43b1daa6fb4 and assuming you've tested it, that commit seems fine to me 🤷

And as I was typing that, I just realized this all got merged to `trunk`, because Github updated it when `release/4.37` was removed 😱 🤦

oguzkocer:
Sorry for the 3rd comment in 7 minutes, I just compared trunk and develop and the diff is what this PR is supposed to be. @mokagio I suggest we open a PR from `trunk` to `develop` and merge it in with a PR review. `trunk` is ahead of where it's supposed to be at this time of code freeze, but `trunk` is not a very useful branch, so not a big loss 🙈

mokagio:
Can't believe I opened this against `trunk` and didn't realize it 😢 

I don't want `trunk` to be ahead and I worry that merging the next release into it might be annoying.

What do you think of:

1. Opening a PR to revert this merge into `trunk` (I thought of rewriting the history, should be safe because who's checking out `trunk` anyways, but don't know... a revert seems safer)
2. Open a PR against `develop` (`develop`, Gio! `d e v e l o p`) for this fix?

Issue #1311, opened 2021-06-04
Adding/removing notes during multiselect brings weird reselection/UI


pachlava:
### Expected
The notes that are added to the list during edit mode should not affect the existing selection.

### Observed
When note list is in edit mode, and there is a selection, new notes that are added via other device will alter the selection:
(Notice how the new note also does not have the checkbox (it's not always like that).)

https://user-images.githubusercontent.com/73365754/120797279-7cf7e180-c544-11eb-823f-ff1f7e14f956.MP4

### Reproduced
1. iOS: open notes list, start edit mode, select a few notes
2. other device: add new note ~> the selection will be altered on iOS

Make|Model|iOS Version|App Version
-|-|-|-
iPhone|XR|14.4.2|4.37.0.0

Issue #1308, opened 2021-06-03
More enhancements to notes multi-selection


pachlava:
### What

After more comparison with similar behaviour in Messages, it feels like doing this will improve the experience:

1. Delete button should be disabled when no note is selected
2. 'Edit' button should be disabled when there are no notes in notes list
3. Search bar should be disabled when in Edit mode (tapping the search bar accidentally will end the edit mode)

Debatable, but:

4. Additional highlighting of selected note cell looks redundant

### Why
Improved usability

cc @SylvesterWilmott

charliescheer:
>Delete button should be disabled when no note is selected

Great suggestion! Fixed this one.

>'Edit' button should be disabled when there are no notes in notes list

Moving the edit mode into the context menu should fix this one as the only way to access the edit mode would require long pressing on an existing note

>Search bar should be disabled when in Edit mode (tapping the search bar accidentally will end the edit mode)

I had set this up so that the search mode would cancel edit mode.  For me I found it more convenient to exit the edit mode if the search was tapped on.  But that being said I can also see how disabling the search is a good option.  I checked the Apple notes app and when in edit for that app the search is disabled.  @SylvesterWilmott  WDYT should search be disabled when in edit mode?

SylvesterWilmott:
>WDYT should search be disabled when in edit mode?

I don't feel too strongly about it. I would say let's follow what Apple does and disable the search so we have behaviour that's consistent with the system. Disabling search also seems to be the case in Pages, Mail, Numbers, Keynote etc.

pachlava:
This issue can be closed since everything mentioned here was fixed in #1328

Issue #1307, opened 2021-06-03
Multiselect: note should be selected on releasing the finger, not right after the touch


pachlava:
### What
It feels like it would be more natural and consistent to have the note selected in edit mode on ‘fingerup’ (if that’s a thing), not on ‘fingerdown’. 

### Why
Two reasons:

1. This is done so in the such apps as Messages, Photos...
2. Currently, we have a weird case of instant selection followed by deselection when the finger touches the note and then moves:

https://user-images.githubusercontent.com/73365754/120716180-56975f00-c4ce-11eb-985f-461dc7f7e60e.mov

SylvesterWilmott:
Have to agree! I must say, I didn't even notice but now that it's called out, I can't unsee.

charliescheer:
I've been look into this, and from what I have been able to find there is not a way to choose selection on "press up" for tableview cell interaction (@eshurakov  @jleandroperez correct me if I am wrong)  Selection/Deselection is handled by a tap gesture which only happens on an instant.  If you tap it fires the method, continuing to press is a different gesture recognizer.  So what I think is happening in the tap and drag example is that the table view thinks you are tap selecting the cell, but when you start dragging, it changes it's mind/deselects because you didn't do a tap/begins scrolling through the tableview content instead.

What do you think instead about this idea.  What if we allowed for a tap and drag selection.  So any cell that is long pressed-and-dragged on will get temporarily selected, as the press continues and the cell changes the selected cell will change.  Finally when the tap is released the last cell that was pressed on will remain selected.  I think that would be possible.

SylvesterWilmott:
I would say this issue is low priority. I also don't think that we shouldn't make any deliberate changes to the default table cell selection interaction and I think this issue is more an observation that there is a difference between which event (finger up/finger down) is used to initiate the selection. In normal usage it's not much of an issue. The problem mentioned above regarding the blinking cells can actually be reproduced in mail.app as well by tapping and then scrolling.

charliescheer:
Sounds good.  I will drop this one off my list for now.  My sense is that we should close this one since it isn't a matter of choosing when the selection takes place, and would involve overwriting the system selection mechanism to make any changes.  What do you think @SylvesterWilmott ?

charliescheer:
fixed in https://github.com/Automattic/simplenote-ios/pull/1560

Issue #1306, opened 2021-06-03
Notes multiselection is discarded after switching the apps


pachlava:
Labelled this as a bug, because it can potentially cause the loss of time for the user

### Expected
Notes selection is not lost after app is minimized/sent to background. E.g. see native Messages.

### Observed
When notes are selected in edit mode, and the app is minimized/someone calls/user switches the app, the selection is lost:

https://user-images.githubusercontent.com/73365754/120713976-4c279600-c4cb-11eb-93bc-a9495e7b5480.MP4

### Reproduced
1. Have some notes
2. Open All Notes
3. Start `Edit mode`
4. Select some notes
5. Switch to the other app and back ~> selection is lost 

Make|Model|iOS Version|App Version
-|-|-|-
iPhone | XR | 14.4.2 |4.37.0.0 Nü

PR #1328, opened 2021-06-15
Edit mode/Action Bar updates

### LINKED ISSUE #1326: New Note/Edit mode changes
Per ref-p2XJRt-36O we are going to move the multi select edit mode into the long press contextual menu and return the new note button to the top right of the list and editor views.

To make this change we need to:

- [x] Move the multi select edit mode into the contextual menu
- [x] put the new note button in the top navigation menu
- [x] remove the toolbar from the list view and edit view
- [x] make the cancel button replace the new note bar while in edit mode
- [x] remove the logic that makes the new note bar/edit mode changes only appear on iPhone

charliescheer:
### Fix
Fixes #1326 

This PR removes the Action bar from Simplenote iOS.  This has affects the new note button and edit mode.  

Edit mode has been moved into the long press context menu

when in edit mode the new note button is replaced with the Cancel button.  When edit mode is exited the new note button returns.

This also puts the new note button back in the top navigation bar

### Test
1. Load the app and confirm that in the note list and the editor you don't see the bottom bar and that the new note button is in the top right navigation toolbar
2. Long press on a note from the note list  to see the select option
3. tap on the select option.  Notice the new note button changes to cancel.  
4. Try canceling and select/deleting to make sure the new note button returns after edit mode is exited

### Review

### Release

> `RELEASE-NOTES.txt` was updated in 4e5ad8 with:
> 
> > -   Moved edit mode into a long press action and removed the new note bar #1328


eshurakov (approved):
Looks good to me!

Questions:
* should we preselect the long-tapped note after user chooses "select"?
* should we disable "Delete" button if nothing is selected?

cc @SylvesterWilmott

pachlava:
> should we preselect the long-tapped note after user chooses "select"?

This was my assumption here as well, I would expect this as a user

> should we disable "Delete" button if nothing is selected?

this is tracked in #1308

Generally, all issues that were created for the edit mode: #1306, #1307, #1308, #1311 are still applicable here

pachlava:
Tried this scenario (assuming that tapping `Select Notes` should auto-select the note):

1. Have so many notes that scrolling the list is needed
2. Scroll to the very bottom
3. Select the last note ~> this note is now hidden behind the bottom bar

https://user-images.githubusercontent.com/73365754/122200700-f4facb80-cea3-11eb-8a2d-eece4fcdd42e.mov

I would expect that just selected note should always be visible on screen.

SylvesterWilmott:
Looks good!

In reference to what @eshurakov mentioned about pre-selection, the label in the contextual menu should be "Select" instead of "Select notes" and therefore indicate that the current note will be selected. So, you should always enter edit mode with one note selected.

charliescheer:
Thanks for the feedback y'all.  I have fixed the two issues raised:

>should we preselect the long-tapped note after user chooses "select"?
>should we disable "Delete" button if nothing is selected?

I also fixed a few of the things mentioned in #1308 

I looked into #1307 #1306 and #1311 I will need a bit more time to dig into those, I think it is better to exclude those from this PR, but it is on my radar to circle back to those soon

charliescheer:
@SylvesterWilmott when you get a chance can you take another look at the selecting behavior, it will now select the note that was long pressed when entering edit mode

pachlava:
Did a bit of testing with recent additions - looks great! One question we can discuss here maybe: should edit mode be stopped once note is deleted via edit mode? This is what happens now.

cc @SylvesterWilmott

SylvesterWilmott:
Look great!

>should edit mode be stopped once note is deleted via edit mode?

That's the expected behaviour yes, based on what generally happens in system selections of this type.

One last thing, unless there is a strong objection to it. Can we change the label "Select Notes" to "Select"? I think this better implies that the current note will be selected after pressing.

charliescheer:
Those changes have been made, search is now disabled when editing and the menu item is now "Select" @SylvesterWilmott can you take another look when you have a chance

@eshurakov can you also take another look at how I am disabling the search bar.  It is a fairly straight forward change, but I wasn't sure if we wanted to disable the search bar directly or use the search bar controller (I used the search bar controller to manage disabling the search)

TY!!

SylvesterWilmott:
Looks great, thanks @charliescheer .

A couple things:

Since the search bar has some degree of transparency already and therefore blurred elements show through it, can we instead only apply transparency to the search placeholder label while search is disabled? So to clarify. When the search bar is disabled, placeholder text goes to 0.5 opacity.

When searching, you are able to enter edit mode which then results in not being able to exit. I think at this time, removing or disabling the "Select" option from the contextual menu while in search is fine. There are plans to move the tag search out of the main search results view which will have an effect on how this works and will need to be defined in the design too.

Simplenote/Classes/SearchDisplayController.swift
```diff
@@ -79,6 +79,11 @@ class SearchDisplayController: NSObject {
     func hideNavigationBarIfNecessary() {
         updateNavigationBar(hidden: active)
     }
+
+    func setEnabled(_ enabled: Bool) {
+        searchBar.isUserInteractionEnabled = enabled
+        searchBar.alpha = enabled ? UIKitConstants.alpha1_0 : UIKitConstants.alpha0_5
+    }
```
eshurakov: I think this is good 👍 

We may also want to disable search activation via shortcut when in edit mode, right now it is possible.

Simplenote/Classes/SPNoteListViewController+Extensions.swift
```diff
@@ -830,14 +840,20 @@ private extension SPNoteListViewController {
             self?.togglePinnedState(note: note)
         }
 
+        let select = UIAction(title: ActionTitle.select, image: .image(name: .success)) { [weak self] _ in
+            self?.setEditing(true, animated: true)
+            self?.tableView.selectRow(at: indexPath, animated: false, scrollPosition: .bottom)
```
eshurakov: I checked with @SylvesterWilmott and we don't want to always scroll selected row to the bottom.

I guess that you wanted to fix this issue: https://github.com/Automattic/simplenote-ios/pull/1328#issuecomment-862232153 and I would expect `scrollPosition: .none` to work, but for some reason it doesn't :-(

I tried a few things and the following combo worked, it only scrolls the row if it is covered by the bottom toolbar, otherwise scroll position is kept intact

```
self?.tableView.selectRow(at: indexPath, animated: false, scrollPosition: .none)
self?.tableView.scrollToNearestSelectedRow(at: .none, animated: false)
```

charliescheer: Great.  TY!

Simplenote/Classes/SearchDisplayController.swift
charliescheer: Good suggestion.  Disabled

charliescheer:
@SylvesterWilmott @eshurakov Thanks for the suggestions!  I have updated the PR to reflect those changes.  When you have a chance this is ready to be reviewed again

eshurakov (approved):
Looks good to me 👍

SylvesterWilmott:
Looks great @charliescheer thanks!

SylvesterWilmott (approved):

pachlava:
Some additions from my side:

1. When we select notes, the navbar title usually changes to 'X Selected', where X is the number of selected notes. But this is not working for the very first selection. In the video you can see that with the 1st selected note, the title is still 'All Notes':

https://user-images.githubusercontent.com/73365754/122923445-6687bd80-d36d-11eb-8d4e-a14430c65533.mov

2. Taking iPad+keyboard setup possibility into account, I think it's worth disabling the new note shortcut, to avoid weird situations. Generally, shortcuts are very unreliable when used for Simulator, because of this I could reproduce new note creation from the edit mode only once. As for search shortcut, I did not reproduce it, either because it's disabled, or because it was not properly handled by Simulator.

charliescheer:
@pachlava thanks for those suggestions!  I updated the PR to fix those issues

mokagio:
The failures in both UI tests CI jobs are due to know issues that we're trying to iron out. It's actually possible that if you do another merge of `develop` into this branch (or even better _IMHO_ rebase this onto `develop` to keep the history linear) the tests will pass thanks to @pachlava's changes from #1335.

I say let's just merge this in, though, so it's in before the 4.39 code freeze.

PR #1333, opened 2021-06-21
Upgrade CircleCI to use Xcode 12.5

mokagio:
### Fix
Updates the Fastlane and CircleCI config file to use Xcode 12.5 and iOS 14.4. Also adds and `.xcversion` file, which we'll use in the future as the single source of truth for the Xcode toolchain version.

### Test
If CI is green, that's all we need because this change affects CI only.

### Review

### Release


mokagio:
🤔 the UI tests are failing in CI, but they're passing for me locally:

Screen Shot 2021-06-22 at 1 02 13 pm

Maybe @pachlava you have some ideas about this?

SimplenoteUITests/UIDs.swift
```diff
@@ -38,7 +38,7 @@ enum UID {
         static let trashNote = "Move to Trash"
         static let restoreNote = "Restore Note"
         static let deleteNote = "Delete Note"
-        static let trashEmptyTrash = "Empty"
+        static let trashEmptyTrash = "Empty trash"
```
mokagio, line 41: I'm guessing this is a change that made it into `develop` without us noticing it. Doesn't seem something that the Xcode 12.5 toolchain would introduce 😄

.circleci/config.yml
```diff
@@ -9,13 +9,13 @@ orbs:
 parameters:
   beta_build:
     type: boolean
-    default: false 
+    default: false
```
mokagio, line 12: Sorry for the noise here. You can use the "Hide whitespace changes" option to get rid of it in the diff.

Screen Shot 2021-06-22 at 3 13 25 pm

pachlava:
Both fails are happening for the UI Tests that involve markdown rendering. I suspect this is about performance. Will take a closer look now.

SimplenoteUITests/UIDs.swift
pachlava, line 41:
```suggestion
        static let trashEmptyTrash = "Empty"
```

Can the newer version of Xcode generate the button label differently? 

The code from this branch, running on my laptop (Xcode 12.4), with empty trash button label being set to `Empty trash` fails, because the actual label is still `Empty`. I had to change it back to `Empty` in code to have it working. At the same time, the same code runs on CI (with Xcode 12.5, from what I see) with no fails in this place.

Since I'm now at coworking, I'd postpone downloading 11 Gb of Xcode 12.5 to check my theory about label difference 😅 

Can I suggest keeping the old 'Empty' label, but extending the locator for this button to work with predicate instead of exact match, so that it works for both `Empty` and `Empty trash`?

in `Trash.swft`:
```swift
    class func empty() {
        Trash.open()
        let predicate = NSPredicate(format: "label LIKE '\(UID.Button.trashEmptyTrash)'")
        let emptyTrashButton = app.buttons.element(matching: predicate).firstMatch
        guard emptyTrashButton.isEnabled else { return }

        emptyTrashButton.tap()
        app.alerts.scrollViews.otherElements.buttons[UID.Button.yes].tap()
    }
```

pachlava: On the second thought, feel free to keep it the way it is now, since it works on CI anyway, and I doubt anyone besides me will run these test locally using Xcode 12.4 in the nearest few days.

 Today I'll be creating the PR that solves the instability you mention, and will include the change there.

In other aspects, good to go from my side :shipit: 
Thank you!

pachlava (approved):

SimplenoteUITests/UIDs.swift
pachlava: Just updated to 12.5, and now the label is `Empty trash` 🧙‍♂️

mokagio: Wow! So it _was_ because of the new Xcode 😱 Go figure...

@pachlava, for future reference, something we could have done during a transition period is adopt the approach from [the Xcode 12.5 WooCommerce PR](https://github.com/woocommerce/woocommerce-ios/pull/4459/files#diff-7c41cb2b963b257aa8753b3c474f818eee259dc7ea8d8eb7653b7031423e2affR69): `if #available(iOS 14.4, *) { ... } else { ... }`

Something like:

```swift
static private(set) lazy var trashEmptyTrash: String = {
  if #available(iOS 14.4, *) {
    "Empty trash"
  } else }
    "Empty
  }
}()
```

mokagio:
Thanks as always @pachlava 🙇 

I agree we should merge this in this instance as opposed to invest in supporting Xcode 12.4 and 12.5 both, because it's about time we all move to 12.5 😄 

Let me know if you need help with the UI tests.

SimplenoteUITests/UIDs.swift
pachlava: Did this as a part of #1335, thanks! I agree that this is better than using `*Empty*`. But, it took me some time to understand that asterisk means 'specified iOS and higher'. Spent about an hour trying to understand why the execution goes inside `if #available(iOS 14.4, *)` block while running 14.5 😅 Final code I used is:

```swift
static private(set) var trashEmptyTrash: String = {
    if #available(iOS 14.5, *) {
        return "Empty trash"
    } else {
        return "Empty"
    }
}()
```

PR #1334, opened 2021-06-22
Update AppStoreStrings to mention English comma keywords requirement

mokagio:
App Store Connect requires keywords to be separated by "English comma, Chinese comma, or a mix of both". For consistency and simplicity, I specify only English comma here.

### Test
Nothing to test, this change will be picked up as the new comment for translators in GlotPress.

### Review

### Release

### DIFF
--- a/Simplenote/Resources/AppStoreStrings.pot
+++ b/Simplenote/Resources/AppStoreStrings.pot
@@ -61,7 +61,7 @@ msgid ""
 msgstr ""
 
 #. translators: Keywords used in the App Store search engine to find the app.
-#. Delimit with a comma between each keyword. Limit to 100 characters including spaces and commas.
+#. Delimit with an English comma between each keyword. Limit to 100 characters including spaces and commas.
 msgctxt "app_store_keywords"
 msgid "simplenote,note,taking,taker,notes,sync,share,simple,markdown,tag,to-do,list,lock,cloud,writing"
 msgstr ""


Simplenote/Resources/AppStoreStrings.pot
```diff
@@ -61,7 +61,7 @@ msgid ""
 msgstr ""
 
 #. translators: Keywords used in the App Store search engine to find the app.
-#. Delimit with a comma between each keyword. Limit to 100 characters including spaces and commas.
+#. Delimit with an English comma between each keyword. Limit to 100 characters including spaces and commas.
```
mokagio, line 64: image

AliSoftware (approved):
👍  💯  for the mention of this in the translator comment.

I fail to remember, did we also implement a check in all the Fastfiles or in the action to at least warn us about any non-English comma accidentally used at beta/release time? i remember we talked about this but can't remember if we implemented it, and if so if we did it in all repos.

mokagio:
Thank @AliSoftware 🙇 

> did we also implement a check in all the Fastfiles or in the action to at least warn us about any non-English comma accidentally used at beta/release time?

No, but it's a neat idea

> i remember we talked about this but can't remember if we implemented it, and if so if we did it in all repos.

We talked about something similar in the context of adding a post-processing step to do the replacement here on the repo. Discussion breadcrumbs can be found in #1206.

My recollection of the discussion, and the place I'm currently at, is that:

- Post-processing is okay at a temporary fix to unblock a time sensitive release; but
- It's better to fix the issue at the source, i.e.: GlotPress

Your suggestion to add a check step during beta builds is great and in line with helping us fixing the issue at the GlotPress level 👍 

As an aside, I did fix the current version of the Arabic keywords in GlotPress myself. Given the slow pace at which keywords get updated, we might not need the check for a while.

PR #1335, opened 2021-06-22
Reduce flakiness of `testBulletedLists()` and `testLongTappingOnLinkOpensLinkInNewWindow()`

pachlava:
### Fix
This hopefully fixes random fails of mentioned two tests during execution on CI. The first test belongs to both `Full` and `Subset` schemes, and the second was only in the `Full` scheme. While I never saw them fail locally, this is quite frequent for CI. Both tests involve either Markdown rendering or actual Web-page loading. My assumption was that the cause was in lower performance of CI VMs. To address this, here are the changes:
1. For several methods that count the occurrences of texts with a certain value or label, a preliminary `waitForExistence` call is added. I.e. when we look how many "Hello" texts are on screen, we now firstly wait for at least one to appear (5 sec timeout). The downside is that we now have one extra search executed before the actual counting takes place. This results in 20 seconds longer execution for a test with 18 such calls.
2. The `Subset` scheme was extended with one more test: `testLongTappingOnLinkOpensLinkInNewWindow()` to make the possible false fail catching easier/faster.
3. `Empty Trash` button is now located in a softer way, i.e. 'Empty*' instead of exact match. This should address the difference in button label generation between Xcodes 12.4 and 12.5
4. `All Notes` button is tapped only after `waitForExistence` call (I noticed one more test failing once on CI because `All Notes` button was not rendered when Sidebar was opened).

### Test
Rerun both `Full` and `Subset` jobs here ~> there should be no fails.
Also, you can see from the [history](https://app.circleci.com/pipelines/github/Automattic/simplenote-ios?branch=reduce-uitests-flakiness-mk1), that there were no fails for this branch on CI.

Current approach for this PR is 'fix what we know is failing'. More preventive approach requires more time investments. Here I wanted to fix only the tests currently causing issues on CI.

### Review

### Release


Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteUITests_Subset.xcscheme
```diff
@@ -52,9 +52,6 @@
                <Test
                   Identifier = "SimplenoteUISmokeTestsNoteEditor/testCreateUncheckedItem()">
                </Test>
-               <Test
-                  Identifier = "SimplenoteUISmokeTestsNoteEditor/testLongTappingOnLinkOpensLinkInNewWindow()">
-               </Test>
```
mokagio, line 57: Seeing this deleted when the PR description says it was added confused me, but expanding the diff shows that the node containing this is `<SkippedTests>`. It's been removed from the skipped list, which means it was added to the list of tests to execute 😅 👍

SimplenoteUITests/NoteEditor.swift
```diff
@@ -117,6 +117,14 @@ class NoteEditor {
     }
 
     class func getTextViewsWithExactValueCount(value: String) -> Int {
+        // We wait for at least one element with exact value to appear before counting
+        // all occurences. The downside is having one extra call before the actual count.
+        let predicate = NSPredicate(format: "value == '" + value + "'")
+        let _ = app
```
mokagio: Nitpick. Since we're ignoring the result, we can also omit the `let` definition.

```suggestion
        _ = app
```

SimplenoteUITests/NoteEditor.swift
```diff
@@ -117,6 +117,14 @@ class NoteEditor {
     }
 
     class func getTextViewsWithExactValueCount(value: String) -> Int {
+        // We wait for at least one element with exact value to appear before counting
+        // all occurences. The downside is having one extra call before the actual count.
+        let predicate = NSPredicate(format: "value == '" + value + "'")
```
mokagio: Nitpick. While the name `predicate` is correct, it doesn't say much about what the predicate is about. Resulting in a "mysterious" usage below `.element(matching: predicate)`.

I'd suggest a more verbose name for this, like `equalValuePredicate`, which would have a slightly clearer usage: `.element(matching: equalValuePredicate)`.

mokagio: If we wanted to be fancy, we could even do something like this

```swift
extension NSPredicate {

  static func valueEquals(_ value: String) -> NSPredicate {
    NSPredicate(format: "value == '\(value)'")
  }
}
```

Then use it like `.element(matching: .valueEquals(value))`.

🤓 

This actually sounds like something we could put in [XCUITestsHelper](https://github.com/Automattic/XCUITestHelpers/) 🤔

SimplenoteUITests/NoteList.swift
```diff
@@ -39,7 +39,9 @@ class NoteList {
         guard !NoteList.isAllNotesListOpen() else { return }
         print(">>> Opening \"All Notes\"")
         Sidebar.open()
-        app.tables.staticTexts[UID.Button.allNotes].tap()
+        let allNotesButton = app.tables.staticTexts[UID.Button.allNotes]
+        guard allNotesButton.waitForExistence(timeout: averageLoadTimeout) else { return }
```
mokagio, line 43: Should we make the test fail if the button isn't found?

SimplenoteUITests/Trash.swift
```diff
@@ -19,7 +19,8 @@ class Trash {
 
     class func empty() {
         Trash.open()
-        let emptyTrashButton = app.buttons[UID.Button.trashEmptyTrash]
+        let predicate = NSPredicate(format: "label LIKE '\(UID.Button.trashEmptyTrash)'")
+        let emptyTrashButton = app.buttons.element(matching: predicate).firstMatch
```
mokagio: Let's please revert this change.

The rationale is discussed [here](https://github.com/Automattic/simplenote-ios/pull/1333#discussion_r656020546), and the summary is that leaving it as it is fine because both CI and developers are on Xcode 12.5. No one that needs to run the UI tests will do so on the older version of Xcode which would require this looser check for compatibility.

mokagio (changes requested):
Requesting change just because of the code to get the trash button and because it's best to have stricter queries, just in case a new button would appear in the UI with a label matching that query.

Once that's addressed this is good to merge. Great work!

> Current approach for this PR is 'fix what we know is failing'. More preventive approach requires more time investments. Here I wanted to fix only the tests currently causing issues on CI.

Totally makes sense!

> a preliminary `waitForExistence` call is added

In my limited experience with UI tests, this seems to be the way to go. It makes tests slower, but also more robust. 

For the most common actions, like tapping a button, we should probably build a dedicated method in [XCUITestHelpers](https://github.com/Automattic/XCUITestHelpers/) so that we can do the wait and tap in a one liner.

I launched a new UI tests run just for fun and... [they failed](https://app.circleci.com/pipelines/github/Automattic/simplenote-ios/3943/workflows/8c0c1dcd-ae71-4926-b5c8-1eb7c268f4cf/jobs/4345/tests)! 😱 I must be cursed 😅 But ☝️ , they didn't fail in the method you worked on in this PR, so I wouldn't count that as a regression 👍

Simplenote.xcodeproj/xcshareddata/xcschemes/SimplenoteUITests_Subset.xcscheme
pachlava: Yep, not the most intuitive thing 😅

SimplenoteUITests/NoteEditor.swift
pachlava: Thank you! Done everywhere.

SimplenoteUITests/Trash.swift
pachlava: Done, thank you! More details in [this comment](https://github.com/Automattic/simplenote-ios/pull/1333#discussion_r656963600).

SimplenoteUITests/NoteList.swift
pachlava: That's a good question... Failing it makes sense, because the execution will go wrong for sure if it's not found, so we will save time before reaching the next inevitable error/fail. On the other hand, the current approach in the project is having assertions only inside asserting methods or tests, and not having them in methods that 'do stuff'. What is your opinion about this?

pachlava:
> I launched a new UI tests run just for fun and... they failed! 😱 I must be cursed 😅

I saw exactly this happen once before. From the selection of failed search tests, it looks like `setUp()` failed to create the notes for search(?). This is worth keeping an eye on.

SimplenoteUITests/NoteEditor.swift
pachlava: I've updated the name as you suggest - thanks for noticing! 

> This actually sounds like something we could put in XCUITestsHelper

👍  I'm thinking of making this a topic for my next PR, once this is done.

pachlava:
> I launched a new UI tests run just for fun and... they failed! 😱 I must be cursed 😅

Tried to address this here by increasing the `minLoadTimeout` from one sec to five. It was used in a lot of places. I think one second is a meaningless timeout, anyway. Also, I'm not sure if it makes any sense now to have three different options for timeout:

```swift
let maxLoadTimeout = 20.0,
    averageLoadTimeout = 10.0,
    minLoadTimeout = 5.0
```
and use them with no strategy. I'm thinking of using the max option everywhere, unless it's a test that explicitly tests the performance.

SimplenoteUITests/NoteEditor.swift
mokagio: Sound good to me 👍

SimplenoteUITests/NoteList.swift
mokagio: > On the other hand, the current approach in the project is having assertions only inside asserting methods or tests, and not having them in methods that 'do stuff'.

Good point. Let's stick with that for the moment and have a dedicated experimentation round for (possibly) changing the approach in the future.

I think the sooner we fail the better. That's particularly true for UI tests, which are notoriously time consuming.

mokagio: This is a short WWDC session on the topic of handling failures in tests: https://developer.apple.com/videos/play/wwdc2020/10091/

There may be something useful for us in there. Just because Apple does something it doesn't mean it's right, still, it may be good food for thought. (_I could be wrong, though, I remember liking it, but I don't really remember the exact recommendations. I should re-watch it..._)

mokagio (approved):
Thank you for addressing the trash handling comment 🙇 

> I'm thinking of using the max option everywhere, unless it's a test that explicitly tests the performance.

Yeah that's reasonable. Let's try it out and see how it plays out for us.

PR #1337, opened 2021-06-24
Fixed an issue where cell separators appeared on ipad

charliescheer:
### Fix
This is a quick PR to fix an issue that was added with the inclusion of the edit mode.  On iPad the note list does not usually have cell separators.  To simplify the animations on iPhone with edit mode, I added a separator view, but didn't realize that the separators didn't appear on iPad.  

This change adds a check to check device for iPad and doesn't activate the separators if it is.

### Test
1. on an iPad load the note list.  You shouldn't see separators between the notes
2. On iPhone check the note list again, there should be separators in between the notes
3. on iPhone, check and make sure there is not a separator after the last cell.

### Review

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -614,7 +614,7 @@ extension SPNoteListViewController: UITableViewDelegate {
 
         cell.separatorInset = insets
 
-        cell.shouldDisplayBottomSeparator = indexPath.row < notesListController.numberOfObjects - 1
+        cell.shouldDisplayBottomSeparator = indexPath.row < notesListController.numberOfObjects - 1 && !UIDevice.isPad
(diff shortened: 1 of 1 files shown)


eshurakov (approved):
LGTM 👍 

:shipit:

mokagio:
@charliescheer the Optional Full UI Test is yellow because the full UI tests run hasn't been started. Those checks need to be unlocked manually, but they are _optional_, so it's up to the author or the reviewers to start them if they see fit.

The fact that it's yellow is not an issue. In fact, as you can see, the check doesn't have the "Required" label and the merge button is enabled.

In other projects, we actually have automation that toggles the state to green to avoid this kind of confusing situations but we haven't gotten to set it up for this one yet. I don't actually think we should, at this point, because we'll soon be moving away from CircleCI in favor of Buildkite.

Feel free to merge 😉

PR #1341, opened 2021-06-30
DRAFT: connect widgets to db

charliescheer:
### Fix
This is a proof of concept connecting the core data DB to the intents extension

### Test
Not much to test at this point.  The PR only connects the shared DB to the intents extension.  So you can see that the notes are selectable but they are not yet wired into the widget it's self.

To see the intent extension in action....
1. load and run this branch
2. log into simplenote on device/simulator if not already logged in
3. minimize app and long press on home screen to bring up edit mode
4. select +, find simplenote, add one of the note preview widgets
5. once added tap on the note preview widget to see the configuration menu
6. finally select the configuration menu for note and you should see a list of the notes pulled from your core data DB

Screen Shot 2021-06-30 at 6 10 11 PM

### Review

### DIFF
--- a/Simplenote/CoreDataManager.swift
+++ b/Simplenote/CoreDataManager.swift
@@ -1,6 +1,8 @@
 import Foundation
 import CoreData
-import AutomatticTracks
+// Commented out for now,
+// TODO: figure out if I want to include tracks here
+//import AutomatticTracks
 
 @objcMembers
 class CoreDataManager: NSObject {
@@ -47,7 +49,8 @@ class CoreDataManager: NSObject {
                                        options: options)
         } catch {
             NSLog("Error loading PersistentStore at URL: \(storageSettings.storageURL)")
-            CrashLogging.crash()
+//            CrashLogging.crash()
+            fatalError()
         }
 
         return psc
--- a/Simplenote/CoreDataManager.swift
+++ b/Simplenote/CoreDataManager.swift
@@ -34,7 +34,7 @@ class CoreDataManager: NSObject {
 
         let options = [NSMigratePersistentStoresAutomaticallyOption: true,
                        NSInferMappingModelAutomaticallyOption: true,
-                       NSSQLitePragmasOption: [Constants.journalMode: Constants.journalSetting]] as [AnyHashable: Any]
+                       NSSQLitePragmasOption: storageSettings.journalModeDisabled] as [AnyHashable: Any]
 
         // Testing logs
         //
@@ -52,9 +52,4 @@ class CoreDataManager: NSObject {
 
         return psc
     }()
-}
-
-private struct Constants {
-    static let journalMode = "journal_mode"
-    static let journalSetting = "DELETE"
 }
--- a/Simplenote/StorageSettings.swift
+++ b/Simplenote/StorageSettings.swift
@@ -29,10 +29,14 @@ class StorageSettings: NSObject {
         }
         return sharedStorageURL
     }
+
+    let journalModeDisabled = [Constants.journalMode: Constants.journalSetting]
 }
 
 private struct Constants {
     static let sqlFile = "Simplenote.sqlite"
     static let resourceName = "Simplenote"
     static let resourceType = "momd"
+    static let journalMode = "journal_mode"
+    static let journalSetting = "DELETE"
 }
--- a/SimplenoteIntents/IntentHandler.swift
+++ b/SimplenoteIntents/IntentHandler.swift
@@ -1,6 +1,27 @@
 import Intents
+import CoreData
 
 class IntentHandler: INExtension {
+    private var dataManager: CoreDataManager = {
+        let storageSettings = StorageSettings()
+        let coreDataManager = CoreDataManager(storageSettings: storageSettings)
+        coreDataManager.managedObjectContext.persistentStoreCoordinator = coreDataManager.persistentStoreCoordinator
+
+        return coreDataManager
+    }()
+
+    private func fetchedResultsController<T: SPManagedObject>(ofType type: T.Type) -> NSFetchedResultsController<T> {
+        let fetchRequest = NSFetchRequest<T>()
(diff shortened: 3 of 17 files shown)


Simplenote/CoreDataManager.swift
```diff
@@ -1,6 +1,8 @@
 import Foundation
 import CoreData
-import AutomatticTracks
+// Commented out for now,
+// TODO: figure out if I want to include tracks here
```
jleandroperez, line 4: Discussed over Slack with Jeremy, nope, we should not include Tracks!

SimplenoteIntents/IntentHandler.swift
```diff
@@ -1,23 +1,44 @@
...
+        let storageSettings = StorageSettings()
+        let coreDataManager = CoreDataManager(storageSettings: storageSettings)
+        coreDataManager.managedObjectContext.persistentStoreCoordinator = coreDataManager.persistentStoreCoordinator
+
+        return coreDataManager
+    }()
+
+    private func fetchedResultsController<T: SPManagedObject>(ofType type: T.Type) -> NSFetchedResultsController<T> {
+        let fetchRequest = NSFetchRequest<T>()
+        let sortDescriptor = NSSortDescriptor(key: "modificationDate", ascending: false)
```
jleandroperez, line 15:
```suggestion
        let sortDescriptor = NSSortDescriptor(keyPath: \Note.modificationDate, ascending: false)
```

jleandroperez: It's always best to stick to KeyPaths. Compiler checked!

Simplenote/StorageSettings.swift
```diff
@@ -29,10 +29,14 @@ class StorageSettings: NSObject {
         }
         return sharedStorageURL
     }
+
+    let journalModeDisabled = [Constants.journalMode: Constants.journalSetting]
```
jleandroperez, line 33: (Thinking) perhaps this class should only expose the URL(s), and this detail should be private to both SharedStorageMigrator and CoreDataStorage

SimplenoteIntents/IntentHandler.swift
```diff
@@ -1,23 +1,44 @@
...
+        fetchRequest.sortDescriptors = [sortDescriptor]
+
+        let entityDescription = NSEntityDescription.entity(forEntityName: String(describing: type.self), in: dataManager.managedObjectContext)
```
jleandroperez, line 18: Let's link `NSObject+Helpers` and use `.classNameWithoutNamespaces`

SimplenoteIntents/IntentHandler.swift
```diff
@@ -1,23 +1,44 @@
...
+        try? frc.performFetch()
 
         var notes: [SPNote] = []
-
-        for placeholder in placeholderNoteData {
-            let spNote = SPNote(identifier: placeholder, display: placeholder)
-            notes.append(spNote)
+        if let fetchedObjects = frc.fetchedObjects {
+            for object in fetchedObjects {
+                let spNote = SPNote(identifier: object.remoteId, display: object.content ?? "New Note")
```
jleandroperez, line 39: Is `SPNote` autogenerated? If so, could we use something like `WidgetNote` or so?

Let's drop the `SP` prefix from everywhere. Historically, classes were prefixed with two characters, to indicate the module / framework / app in which they belong.

Swift happened, and now we have namespaces. For instance, if you add a new class called `Notification`, it will collide with Foundation.Notification.

Such collision is solved by simply using the namespace as a prefix (that is: `Foundation.Notification`).

jleandroperez: Plus: Missing `NSLocalizedString` for `New Note`

SimplenoteWidgets/Models/Note+CoreDataClass.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//
+
+import Foundation
+import CoreData
+
+@objc(Note)
+public class Note: SPManagedObject {
+
+}
```
jleandroperez, line 15: These were autogenerated classes (by Xcode). Let's please merge both the `Note+CoreDataClass` and `Note+CoreDataProperties` into a single file.

**Plus** please also:

- Drop the dummy comments
- Add a note mentioning this class is there for us to avoid linking Simperium / SPManagedObject, in the widget target

jleandroperez (commented):
Sending you few comments Charlie!!

Plus: 
- For this SimplenoteIntents target, we'll probably need to poke Platform9 again, the Provisioning Profiles don't have the `match` fastlane keyword
- Please make sure, also, to track the String Localization required changes (if any). That is: we need to make sure the new Strings are extracted for localization!

SimplenoteIntents/IntentHandler.swift
```diff
@@ -1,23 +1,44 @@
...
+import CoreData
 
 class IntentHandler: INExtension {
+    private var dataManager: CoreDataManager = {
+        let storageSettings = StorageSettings()
+        let coreDataManager = CoreDataManager(storageSettings: storageSettings)
+        coreDataManager.managedObjectContext.persistentStoreCoordinator = coreDataManager.persistentStoreCoordinator
+
+        return coreDataManager
+    }()
```
eshurakov, line 11: A note: we need to be careful here. If migration in the main app failed, this code will probably create a new empty database in the shared container? And this will make the app to "loose" it's data once it sees that there is a database in the shared container.

jleandroperez: +1'ing to what Evgeny mentioned. We need to handle the "Non cconfigured State". That is:

1. Safeguard against the DB not existing (otherwise spawning the CoreDataManager would create the DB!)
2. Probably display a placeholder in the UI, saying "please log into your Simplenote account first"

charliescheer:
This PR was a proof of concept at the time, and is no longer valid.  Gonna close this

PR #1336, opened 2021-06-23
Add SimplenoteIntents extension to app

charliescheer:
### Fix
This PR lays the ground work for adding intent configuration widgets into Simplenote.

As compared to the new note widget, which displays no data, to be able to setup widgets that can be configured by the user we need to create another extension for the app that defines an Intent configuration.  This PR sets that up.

### Test
There isn't much to test with this PR, but you can confirm that the option to configure a widget is there...
There is a placeholder for the note preview widget that was added to be able to confirm the intents were working, so to test this pr, make sure that you have configuration options.  To do that.
1. run the app on device or in the sim to install the widgets
2. On the home screen long press till the home screen editor appears, tap on the plus
3. find simplenote and select the note preview widget
4. Once it the widget is added to your home screen, while home screen edit is enabled, tap on note preview widget.
5. you will see an option to select the note.  Tap on that and you will be given four choices Note 1-4
6. select one and see that the text appears in the widget.

PR #1344, opened 2021-07-05
Update Fastlane code signing automation to include Intents extension

mokagio:
Builds on top of #1336 to ensure the app can build with the new intents extension.

### Test
This is time consuming to test manually because of the multiple targets and because we want to ensure code signing work.

Here's what I did:

- Try to archive the app with the "Simplenote"
- Modify the "Simplenote" scheme to use "Distribution Alpha" and try to archive
- Repeat with "Distribution Internal"
- Notice how I launched an Installable Build and it [succeeded](https://app.circleci.com/pipelines/github/Automattic/simplenote-ios/3988/workflows/b35025e1-82ba-47d6-a494-7d7b0c8f9ab1/jobs/4450), this exercises the Alpha build configuration in CI
- Comment the ASC and App Center upload in `Fastfile` (see aac0063) and run `bundle exec fastlane build_and_upload_beta_release skip_confirm:true create_gh_release:false`
- Run `bundle exec fastlane build_and_upload_beta_release skip_confirm:true create_gh_release:false`

### Review

### Release

@charilescheer: could you please try this out and let me know if it work for you? Thanks!

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -4764,7 +4764,7 @@
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
-				CODE_SIGN_STYLE = Automatic;
+				CODE_SIGN_STYLE = Manual;
 				DEBUG_INFORMATION_FORMAT = dwarf;
 				DEVELOPMENT_TEAM = PZYM8XX95Q;
 				GCC_C_LANGUAGE_STANDARD = gnu11;
@@ -4803,7 +4803,7 @@
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
-				CODE_SIGN_STYLE = Automatic;
+				CODE_SIGN_STYLE = Manual;
 				COPY_PHASE_STRIP = NO;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				DEVELOPMENT_TEAM = PZYM8XX95Q;
@@ -4820,6 +4820,7 @@
 				MTL_FAST_MATH = YES;
 				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.Intents;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				PROVISIONING_PROFILE_SPECIFIER = "match AppStore com.codality.NotationalFlow.Intents";
 				SKIP_INSTALL = YES;
 				SWIFT_VERSION = 5.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
@@ -4837,7 +4838,7 @@
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_IDENTITY = "iPhone Distribution: Automattic, Inc.";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution: Automattic, Inc.";
-				CODE_SIGN_STYLE = Automatic;
+				CODE_SIGN_STYLE = Manual;
 				COPY_PHASE_STRIP = NO;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				DEVELOPMENT_TEAM = 99KV9Z6BKV;
@@ -4854,6 +4855,7 @@
 				MTL_FAST_MATH = YES;
 				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.Internal.Intents;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				PROVISIONING_PROFILE_SPECIFIER = "match InHouse com.codality.NotationalFlow.Internal.Intents";
 				SKIP_INSTALL = YES;
 				SWIFT_VERSION = 5.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
@@ -4871,7 +4873,7 @@
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_IDENTITY = "iPhone Distribution: Automattic, Inc.";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Distribution: Automattic, Inc.";
-				CODE_SIGN_STYLE = Automatic;
+				CODE_SIGN_STYLE = Manual;
 				COPY_PHASE_STRIP = NO;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				DEVELOPMENT_TEAM = 99KV9Z6BKV;
@@ -4888,6 +4890,7 @@
 				MTL_FAST_MATH = YES;
 				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.Alpha.Intents;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				PROVISIONING_PROFILE_SPECIFIER = "match InHouse com.codality.NotationalFlow.Alpha.Intents";
 				SKIP_INSTALL = YES;
 				SWIFT_VERSION = 5.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
@@ -4905,7 +4908,7 @@
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_IDENTITY = "Apple Distribution: Automattic, Inc. (PZYM8XX95Q)";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "Apple Distribution: Automattic, Inc. (PZYM8XX95Q)";
-				CODE_SIGN_STYLE = Automatic;
+				CODE_SIGN_STYLE = Manual;
 				COPY_PHASE_STRIP = NO;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				DEVELOPMENT_TEAM = PZYM8XX95Q;
@@ -4922,6 +4925,7 @@
 				MTL_FAST_MATH = YES;
 				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.Intents;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				PROVISIONING_PROFILE_SPECIFIER = "match AppStore com.codality.NotationalFlow.Intents";
 				SKIP_INSTALL = YES;
 				SWIFT_VERSION = 5.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -4784,6 +4784,7 @@
 				MTL_FAST_MATH = YES;
 				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.Development.Intents;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				PROVISIONING_PROFILE_SPECIFIER = "Simplenote Development Intents";
 				SKIP_INSTALL = YES;
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -350,7 +350,8 @@ end
         method: "enterprise",
         provisioningProfiles: {
           "com.codality.NotationalFlow.Internal" => "match InHouse com.codality.NotationalFlow.Internal",
-          "com.codality.NotationalFlow.Internal.Share" => "match InHouse com.codality.NotationalFlow.Internal.Share"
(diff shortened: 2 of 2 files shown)


fastlane/Fastfile
```diff
@@ -350,26 +350,27 @@ end
         method: "enterprise",
         provisioningProfiles: {
           "com.codality.NotationalFlow.Internal" => "match InHouse com.codality.NotationalFlow.Internal",
-          "com.codality.NotationalFlow.Internal.Share" => "match InHouse com.codality.NotationalFlow.Internal.Share"
+          "com.codality.NotationalFlow.Internal.Share" => "match InHouse com.codality.NotationalFlow.Internal.Share",
+          "#{APP_STORE_BUNDLE_IDENTIFIER}.Internal.Intents" => "match InHouse #{APP_STORE_BUNDLE_IDENTIFIER}.Internal.Intents"
```
mokagio: The code here could have been updated a bit more, like I did for the calls to `match`, but I decided not to go down that way and leave the diff smaller.

fastlane/Fastfile
```diff
@@ -829,15 +827,20 @@ end
...
   end
 
+  def simplenote_app_identifiers(root_bundle_id:)
+    [
+      root_bundle_id,
+      "#{root_bundle_id}.Share",
+      "#{root_bundle_id}.Widgets",
+      "#{root_bundle_id}.Intents"
+    ]
+  end
```
mokagio: I decided to DRY the code defining the bundle identifiers a bit because of their constant structure. It made it simpler to add the "Intents" one and be sure I didn't forget it from one build configuration entry.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -4922,6 +4925,7 @@
 				MTL_FAST_MATH = YES;
 				PRODUCT_BUNDLE_IDENTIFIER = com.codality.NotationalFlow.Intents;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				PROVISIONING_PROFILE_SPECIFIER = "match AppStore com.codality.NotationalFlow.Intents";
```
mokagio, line 4929: The best explanation I have for the changes in this diff is: Xcode needed them to make the process work.

charliescheer (commented):
Hey @mokagio I checked out the branch and I am not able to build it.  I am seeing a couple of things, the first is that "com.codality.NotationalFlow.Development.Intents" has no provisioning profile


I am also seeing  no matching profile warnings after downloading the manual certificates (but I may be missing something to update the certs) :

mokagio:
@charliescheer sorry about the hiccup.

I addressed the development provisioning profile situation in 4a285b2f.

I'm pretty sure the reason the Distribution Internal and Distribution Alpha certificates are missing on your end is because you weren't part of the Enterprise account, which is where those are defined. I sent you an invite. After you accept the invitation, you should have a new Automattic entry in your Xcode > Preferences > Account and you may need to "Download Manual Profiles" again. Unfortunately, Apple doesn't show which one is the Enterprise one 😞 

Screen Shot 2021-07-06 at 4 33 04 pm


The final one is the Release build configuration. That's showing an error on my end, too, but I don't think we need to address it now. You might notice that it's showing errors in the other targets, too. That's not a problem because we are not actually using the Release build configuration for App Store releases, but the Distribution App Store one.

As a side note, I'm not sure why we have Release _and_ Distribution App Store (maybe @jleandroperez knows?) If there's no reason, or if the reason we had in the past is no longer valid (which often happens as codebases and their team grow and the tech stack evolves) we should set aside some time to consolidate, ideally fold Distribution App Store into Release, which is a more standard setup for App Store distribution, IMHO.

charliescheer (approved):
That did it.  Compiling now without an issue.  Thanks @mokagio !!!

mokagio:
@charliescheer I'll leave it up to you whether to merge this in your branch now or wait till it's been code reviewed / approved. Holding on the merge would keep the diff in the PR clearer, but it might prevent reviewers from running the app, so there's no "good" answer 😄

fastlane/Fastfile
AliSoftware: I actually just suggested a very similar approach to how you did it here in a comment on the other PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16805/files#r666021212 👍 

I like this. Maybe for this PR here you can just apply the same `map` trick to DRY the repetition of the prefix like how we applied it in the other PR?

fastlane/Fastfile
```diff
@@ -400,7 +401,8 @@ end
         export_team_id: ENV["EXT_EXPORT_TEAM_ID"],
         provisioningProfiles: {
           "com.codality.NotationalFlow" => "match AppStore com.codality.NotationalFlow",
-          "com.codality.NotationalFlow.Share" => "match AppStore com.codality.NotationalFlow.Share"
+          "com.codality.NotationalFlow.Share" => "match AppStore com.codality.NotationalFlow.Share",
+          "#{APP_STORE_BUNDLE_IDENTIFIER}.Intents" => "match InHouse #{APP_STORE_BUNDLE_IDENTIFIER}.Intents"
```
AliSoftware: ⚠️  I think this should be `match AppStore …` instead of `match InHouse`

fastlane/Fastfile
AliSoftware: Given that `APP_STORE_BUNDLE_IDENTIFIER` is equal to `com.codality.NotationalFlow`, wouldn't it make sense to use `#{APP_STORE_BUNDLE_IDENTIFIER}.Internal` in line 353 too, instead of repeating `com.codality.NotationalFlow.Internal` as a literal?

AliSoftware: Also, I'd love to see more DRY here tho, like having a `simplenote_provisioning_profiles` method to provide those.

My attempt at this:
```ruby
# in_house should be `nil` for AppStore, and `"Internal"` or `"Alpha"` for InHouse builds
def simplenote_provisioning_profiles(in_house_suffix = nil)
  root_bundle_id = [APP_STORE_BUNDLE_IDENTIFIER, in_house_suffix].compact.join('.')
  match_type = in_house_suffix.nil? ? "AppStore" : "InHouse"
  bundle_ids = [
    root_bundle_id,
    "#{root_bundle_id}.Share",
    "#{root_bundle_id}.Intents"
  ]
  bundle_ids.map { |key| [key, "match #{match_type} #{key}"] }.to_h
end
```

* Usage here: `provisioningProfiles: simplenote_provisioning_profiles('Internal')`
* Usage on line 402: `provisioningProfiles: simplenote_proviosining_profiles`

---

Alternatively, if you prefer, you could also make the method take the whole `root_bundle_id` as a parameter of the method, maybe making `APP_STORE_BUNDLE_IDENTIFIER` the default value for that parameter, and compare the param against that `APP_STORE_BUNDLE_IDENTIFIER` to determine the `match_type`. Actually, that alternative solution would allow you to reuse your `simplenote_app_identifiers` method to determine the keys – and would allow you to [avoid forgetting about the Widgets case for example](https://github.com/Automattic/simplenote-ios/pull/1344#discussion_r666070924) – so maybe even better?

```ruby
def simplenote_provisioning_profiles(root_bundle_id: APP_STORE_BUNDLE_IDENTIFIER)
    bundle_ids = simplenote_app_identifiers(root_bundle_id: root_bundle_id)
    match_type = root_bundle_id == APP_STORE_BUNDLE_IDENTIFIER ? "AppStore" : "InHouse"
    bundle_ids.map { |key| [key, "match #{match_type} #{key}"] }.to_h
end
```

fastlane/Fastfile
```diff
@@ -798,16 +801,15 @@ end
 ########################################################################
 # Fastlane match code signing
 ########################################################################
+
+  APP_STORE_BUNDLE_IDENTIFIER = 'com.codality.NotationalFlow'.freeze
```
AliSoftware: Seems strange to have this constant declared here in the middle of the file while it's used _before_.

I can totally see the appeal of making it close to the "Fastlane match code signing" section, but since it's also used in lanes that are declared earlier in the file, I think this should be moved at the top of the Fastfile instead.

AliSoftware (changes requested):
Left a couple of nitpicks and suggestions to improve the code, though those are not blockers. 

But marking this as "Request changes" still, because I think there is an incorrect InHouse profile used for AppStore case, that does need to be fixed.

fastlane/Fastfile
AliSoftware: ⚠️  Also, is this normal that we have `#{root_bundle_id}.Widgets"` listed in `simplenote_app_identifiers` but that we dont have them listed as participating in provisioningProfiles mapping here? Aren't those missing?

mokagio: > ⚠️  I think this should be `match AppStore …` instead of `match InHouse`

ZOMG 😱 Thanks @AliSoftware great catch!

mokagio: > ⚠️  Also, is this normal that we have `#{root_bundle_id}.Widgets"` listed in `simplenote_app_identifiers` but that we dont have them listed as participating in provisioningProfiles mapping here? Aren't those missing?

I asked myself the same question when touching this code, but then I figured that "it works in production (right?)" so maybe we don't need them?

It might be interesting to add them and see what happens to the code signing and ASC submission, but that might be a change for a different PR.

fastlane/Fastfile
mokagio: > wouldn't it make sense to use... Also, I'd love to see more DRY here tho...

Yeah, totally! I didn't go down that path because I felt like I had already made too many changes in this PR, which was about adding support for Intents. I probably should have done the refactoring step in isolation first, then add the support on top of it.

I've already spent time to write this commend and DRY the implementation, though, so I might as well keep going and implement your suggestion 😄

fastlane/Fastfile
mokagio: Good point 👍

fastlane/Fastfile
```diff
@@ -829,15 +824,36 @@ end
...
+      %w[Share Widgets Intents]
+      .map { |suffix| "#{root_bundle_id}.#{suffix}" }
+
+    [root_bundle_id, *extension_bundle_ids]
+  end
+
+  # Compiles the dictionary mapping of the bundle identifiers and provisioning
+  # profiles to be used in the `export_options > provisioningProfiles`
+  # parameter for the `build_app`/`gym` action.
+  def simplenote_provisioning_profiles(root_bundle_id: APP_STORE_BUNDLE_IDENTIFIER, match_type: 'AppStore')
```
mokagio, line 846: Leaving the caller the ability to specify the match type (`AppStore`, `InHouse`, etc) as a String leaves the door open for typos and use of unsupported types (maybe during a migration to a new major version of the tool?)

I decided to stick with this approach for now to avoid adding extra complexity to the code here by defining something like a symbol for the type and then switching on its value.

Definitely open to the idea of doing it, but maybe in a followup step or as a more abstract method in the release toolkit.

mokagio:
@AliSoftware, I think I addressed all your comments. Could you have another look? Thanks sir!

charliescheer (approved):

fastlane/Fastfile
```diff
@@ -829,15 +824,36 @@ end
...
 
+  # Compiles the array of bundle identifiers for the different targets that
+  # make up the Simplenote app, to be used as the `app_identifier` parameter
+  # for the `match` action.
+  def simplenote_app_identifiers(root_bundle_id: APP_STORE_BUNDLE_IDENTIFIER)
+    extension_bundle_ids =
+      %w[Share Widgets Intents]
+      .map { |suffix| "#{root_bundle_id}.#{suffix}" }
+
+    [root_bundle_id, *extension_bundle_ids]
```
AliSoftware, line 840: Super-nit: I'm not sure what rubocop would say, but iirc the ruby conventions would tell to not let the `=` hanging at the end of the line, but instead write it this way:

```ruby
    extension_bundle_ids = %w[Share Widgets Intents]
      .map { |suffix| "#{root_bundle_id}.#{suffix}" }
```
Or alternatively this way:
```ruby
    extension_bundle_ids = %w[Share Widgets Intents].map do |suffix|
      "#{root_bundle_id}.#{suffix}"
    end
```

Not a big deal, only style, just thought I'd mention fyi.

Alternative idea, if we want to be very nitpicky (personally I like it even better, but it's a matter of taste, so take it or leave it):
```ruby
  def simplenote_app_identifiers(root_bundle_id: APP_STORE_BUNDLE_IDENTIFIER)
    [nil, 'Share', 'Widgets', 'Intents'].map do |suffix|
      [root_bundle_id, suffix].compact.join('.')
    end
  end
```

fastlane/Fastfile
AliSoftware: I think this is fine to keep it that way especially because this is an internal function and not a lane, so that value will only be provided by callers elsewhere in the `Fastfile`.

Alternatively you could make this parameter be a boolean `in_house: false` and use `match_type = in_house ? 'InHouse' : 'AppStore'` in the implementation. Or even not have that parameter at all, and ise `match_type = root_bundle_id == APP_STORE_BUNDLE_IDENTIFIER ? 'AppStore' : 'InHouse'` (though maybe it's taking it too far?)  🤷

Simplenote.xcodeproj/project.pbxproj
AliSoftware: One thing that could be fancy would be to make this setting be:
```suggestion
				PROVISIONING_PROFILE_SPECIFIER = "match AppStore ${PRODUCT_BUNDLE_IDENTIFIER}";
```
(And same for other instances of this for other targets and configs ofc)

---

Side note: In other projects in past company I've even gone as far as making a user-defined build setting at project level named `BASE_BUNDLE_ID = com.codality.NotationalFlow` and then at target level make each target's `PRODUCT_BUNDLE_IDENTIFIER` use `${BASE_BUNDLE_ID}` to construct its own, e.g. `${BASE_BUNDLE_ID}.Intents`… which would match (pun intended) how we do it in the `Fastfile` too.

That's probably taking it too far, especially just for this PR, and not really needed (as the pbxproj works well just the way it is, so…); just thought you'd be curious about this possibility and idea 🙂

AliSoftware (approved):
Left a couple of either ideas but that are more FYIs, nothing really _requiring_ any change.

:shipit:

charliescheer:
Thanks @mokagio and @AliSoftware for working on this.

I am gonna go ahead and merge this back into the feature branch and continue on with the widget project.  Thanks again for your help!!!

PR #1347, opened 2021-07-06
Previously viewed tag loaded on app launch

### LINKED ISSUE #1218: Simplenote Always Opens to a Tag List Instead of All Notes
### Expected
Switching to All Notes before closing the app should force the app to open to All Notes again when reopening.

### Observed
If you force quit the app, when relaunching the app, it will always launch to a tag list.

### Reproduced
1. Open a note from All Notes.
2. Add a tag to the note.
3. Go back to All Notes
4. Tap on Menu then tap on a tag from the list.
5. Force quit the app.
6. Relaunch Simplenote.
7. Switch to the All Notes view.
8. Force quit the app.

It will now always open to the tag view, even if the tag is removed from the note and the tag is deleted from the tag list.

-|-|-|-
Apple|iPhone 11 Pro|14.4.1|4.33.0.0

charliescheer:
### Fix
This PR fixes #1218 

Currently when Simplenote terminates the previously viewed tag is restored so the app launches into a specific tag, even if that tag has been removed.  This includes going into the trash.  What is expected is the app would launch into All Notes instead.

This PR removes the logic that was saving the last selected tag (there was something odd going on with user defaults and clearing this value) so the app will launch into All Notes

### Test
1. Open the app and select a tag
2. Use the app switcher to close out of Simplenote completely
3. Launch Simplenote
4. You should be taken to All notes when the app loads.

### DIFF
--- a/Simplenote/SPAppDelegate.m
+++ b/Simplenote/SPAppDelegate.m
@@ -133,7 +133,7 @@
     [self configurePublishController];
     [self setupDefaultWindow];
     [self configureStateRestoration];
-    NSLog(@"selected tag: %@", self.selectedTag);
+    
     return YES;
 }
 
--- a/Simplenote/SPAppDelegate.m
+++ b/Simplenote/SPAppDelegate.m
@@ -79,12 +79,6 @@
     self.window.backgroundColor = [UIColor simplenoteWindowBackgroundColor];
     self.window.tintColor = [UIColor simplenoteTintColor];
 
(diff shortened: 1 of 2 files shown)


jleandroperez (approved):
@charliescheer can we also please delete the `kSelectedTagKey` constant (SPConstants), and the snippet that clears that out upon logout?

Thank you sir!!

:shipit:

PR #1348, opened 2021-07-07
Search Highlighting out of place after rotation

### LINKED ISSUE #1255: Search match highlighter is displaced when device is rotated
### Expected
Search match highlighter should be positioned above the keyword in all cases

### Observed
When there is a matching keyword selected, and the device is rotated, the highlighting word becomes strongly unaligned with the word it tries to highlight.

Video:

https://user-images.githubusercontent.com/73365754/115753182-96cdd280-a3a3-11eb-8643-af47d060887a.MP4

Picture:
Screenshot_2021-04-22 19 50 41_smF8cN

### Reproduced
1. Have a note with specific keyword
2. Allow automatic device rotation
3. Start searching for the keyword, open the note
4. Rotate the device ~> the highlighting keyword is not above the keyword it tries to highlight.

-|-|-|-
iPhone|XR|14.4.2|4.35.0.0
iPad|8Gen Sim|14.4|4.35.0.0

charliescheer:
### Fix
This PR fixes #1255 

When you load a note while search is enabled all of the instances of the search term are highlighted in that note and the current instance is given a special highlight to make it stand out.  Currently if you rotate the device while looking at the note editor in search mode the special highlighting becomes out of place.

In this PR, when the device is rotated, the current highlight is removed and then replaced with a new one in the correct location.

### Test
1. Search for some text that is in one of your notes, open the note and notice that the first instance of the search term is highlighted
2. rotate the device
3. notice that the highlight is still correctly highlighted after rotation

### Review


Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -848,6 +848,12 @@ extension SPNoteEditorViewController {
         ]
     }
 
+    @objc
+    func refershSearchHighlight() {
```
eshurakov: `refersh` => `refresh` :-)

Simplenote/Classes/SPNoteEditorViewController.m
```diff
@@ -260,6 +260,17 @@ - (void)refreshTagEditorOffsetWithCoordinator:(id<UIViewControllerTransitionCoor
...
 
+- (void)refreshSearchHighlightIfNeededWithCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
+{
+    if (!self.searching) {
+        return;
+    }
+
+    [coordinator animateAlongsideTransition:nil completion:^(id<UIViewControllerTransitionCoordinatorContext>  _Nonnull context) {
+        [self refershSearchHighlight];
+    }];
```
eshurakov: Right now you can see the highlight jumping from one position to another because we refresh it in `completion` block. What we can do is refresh it in `animation` block, like this:
```
[coordinator animateAlongsideTransitionInView:nil animation:^(id<UIViewControllerTransitionCoordinatorContext>  _Nonnull context) {
            [self refershSearchHighlight];
    } completion:nil];
```

eshurakov (approved):
Looks good, works good!

Sent a few nitpicky comments, otherwise :shipit:

Simplenote/Classes/SPNoteEditorViewController.m
charliescheer: nice worked like a charm!

PR #1349, opened 2021-07-08
Migrate CoreData to shared app group directory

charliescheer:
This PR is a continuation of #1329 I have cherry picked the commits from the feature/widgets-shared-db branch to a new branch that is separate form the widgets and only focuses on the migration of the database.

### Fix
This PR makes changes to the CoreData database for Simplenote iOS.  It approaches two changes, first it moves the CoreData stack into its own class, and second it migrates the data from the in app storage to a shared directory that can be accessed by other SNiOS extensions.

### Test
To test this PR you need to have a simulator or device setup with a database that was configured in the previous location first.  So..
1. Load the develop branch on your device/sim and log into simplenote
2. Stop operation of the app on that device
3. Change branches to feature/shared-db-migration
4. Compile this branch to your device.
5. When the app loads the migration should happen automatically.

In the app if the migration succeeds or fails it shouldn't be visible to the user as if it fails then the app should default to using the previous database location.  To tell if this has worked, check the console logs, there should be prints from the app that the database was migrated successfully.
If you run the app again after a successful migration you should see a message "Core Data Migration not required" and the app should load normally

### Review

### DIFF
--- a/Simplenote/Classes/FileManager+Simplenote.swift
+++ b/Simplenote/Classes/FileManager+Simplenote.swift
@@ -13,4 +13,23 @@ extension FileManager {
 
          return url
      }
+
+    /// URL for Simplenote's shared app group directory
+    ///
+    var groupDirectory: URL {
+        containerURL(forSecurityApplicationGroupIdentifier: Constants.sharedDirectoryDomain + Constants.groupIdentifier)!
+    }
+
+    /// URL for Simplenote's shared app group documents directory
+    ///
+    var groupDocumentsDirectory: URL {
+        groupDirectory.appendingPathComponent(Constants.documentDirectory)
+    }
+}
+
+private struct Constants {
+    static let defaultBundleIdentifier = "com.codality.NationalFlow"
+    static let groupIdentifier = Bundle.main.bundleIdentifier ?? Constants.defaultBundleIdentifier
+    static let sharedDirectoryDomain = "group."
+    static let documentDirectory = "Documents"
 }
--- a/Simplenote/Classes/FileManager+Simplenote.swift
+++ b/Simplenote/Classes/FileManager+Simplenote.swift
@@ -19,17 +19,10 @@ extension FileManager {
     var sharedContainerURL: URL {
         containerURL(forSecurityApplicationGroupIdentifier: Constants.sharedDirectoryDomain + Constants.groupIdentifier)!
     }
-
-    /// URL for Simplenote's shared app group documents directory
-    ///
-    var sharedDocumentsDirectory: URL {
-        sharedContainerURL.appendingPathComponent(Constants.documentDirectory)
-    }
 }
 
 private struct Constants {
     static let defaultBundleIdentifier = "com.codality.NationalFlow"
     static let groupIdentifier = Bundle.main.bundleIdentifier ?? Constants.defaultBundleIdentifier
     static let sharedDirectoryDomain = "group."
-    static let documentDirectory = "Documents"
 }
--- a/Simplenote/Classes/FileManager+Simplenote.swift
+++ b/Simplenote/Classes/FileManager+Simplenote.swift
@@ -23,6 +23,19 @@ extension FileManager {
 
 private struct Constants {
     static let defaultBundleIdentifier = "com.codality.NotationalFlow"
-    static let groupIdentifier = Bundle.main.bundleIdentifier ?? Constants.defaultBundleIdentifier
+    static let groupIdentifier = rootBundleIdentifier() ?? Constants.defaultBundleIdentifier
     static let sharedDirectoryDomain = "group."
+
+    static func rootBundleIdentifier() -> String? {
+        if Bundle.main.bundleURL.pathExtension != "appex" {
+            return Bundle.main.bundleIdentifier
+        }
+
+        let url = Bundle.main.bundleURL.deletingLastPathComponent().deletingLastPathComponent()
+
+        guard let bundle = Bundle(url: url) else {
+            return nil
+        }
+        return bundle.object(forInfoDictionaryKey: "CFBundleIdentifier") as? String
+    }
 }
--- a/Simplenote/Classes/FileManager+Simplenote.swift
+++ b/Simplenote/Classes/FileManager+Simplenote.swift
@@ -23,19 +23,6 @@ extension FileManager {
 
 private struct Constants {
     static let defaultBundleIdentifier = "com.codality.NotationalFlow"
-    static let groupIdentifier = rootBundleIdentifier() ?? Constants.defaultBundleIdentifier
+    static let groupIdentifier = Bundle.main.rootBundleIdentifier ?? Constants.defaultBundleIdentifier
     static let sharedDirectoryDomain = "group."
-
-    static func rootBundleIdentifier() -> String? {
-        if Bundle.main.bundleURL.pathExtension != "appex" {
-            return Bundle.main.bundleIdentifier
-        }
-
-        let url = Bundle.main.bundleURL.deletingLastPathComponent().deletingLastPathComponent()
-
-        guard let bundle = Bundle(url: url) else {
-            return nil
-        }
-        return bundle.object(forInfoDictionaryKey: "CFBundleIdentifier") as? String
-    }
 }
--- a/Simplenote/Classes/FileManager+Simplenote.swift
+++ b/Simplenote/Classes/FileManager+Simplenote.swift
@@ -22,7 +22,7 @@ extension FileManager {
 }
 
 private struct Constants {
-    static let defaultBundleIdentifier = "com.codality.NationalFlow"
+    static let defaultBundleIdentifier = "com.codality.NotationalFlow"
     static let groupIdentifier = Bundle.main.bundleIdentifier ?? Constants.defaultBundleIdentifier
     static let sharedDirectoryDomain = "group."
 }
--- a/Simplenote/Classes/FileManager+Simplenote.swift
+++ b/Simplenote/Classes/FileManager+Simplenote.swift
@@ -16,14 +16,14 @@ extension FileManager {
 
     /// URL for Simplenote's shared app group directory
     ///
-    var groupDirectory: URL {
+    var sharedContainerURL: URL {
         containerURL(forSecurityApplicationGroupIdentifier: Constants.sharedDirectoryDomain + Constants.groupIdentifier)!
     }
 
     /// URL for Simplenote's shared app group documents directory
     ///
-    var groupDocumentsDirectory: URL {
-        groupDirectory.appendingPathComponent(Constants.documentDirectory)
+    var sharedDocumentsDirectory: URL {
+        sharedContainerURL.appendingPathComponent(Constants.documentDirectory)
     }
 }
 
--- a/Simplenote/CoreDataManager.swift
+++ b/Simplenote/CoreDataManager.swift
@@ -73,7 +73,8 @@ class CoreDataManager: NSObject {
     lazy var persistentStoreCoordinator: NSPersistentStoreCoordinator = {
         let psc = NSPersistentStoreCoordinator(managedObjectModel: managedObjectModel)
 
-        var storeURL: URL = CoreDataManager.appGroupDbExists ? CoreDataManager.groupStorageURL : CoreDataManager.appStorageURL
+        let mustUseOldDb = CoreDataManager.oldDbExists && !CoreDataManager.appGroupDbExists
+        var storeURL: URL = mustUseOldDb ? CoreDataManager.appStorageURL : CoreDataManager.appStorageURL
         let options = [NSMigratePersistentStoresAutomaticallyOption: true, NSInferMappingModelAutomaticallyOption: true]
 
         // Testing logs
--- a/Simplenote/CoreDataManager.swift
+++ b/Simplenote/CoreDataManager.swift
@@ -4,15 +4,10 @@ import CoreData
 @objcMembers
 class CoreDataManager: NSObject {
 
-    /// Storage Settings
-    ///
-    private let storageSettings: StorageSettings
-
     init(_ storageURL: URL, storageSettings: StorageSettings = StorageSettings()) throws {
-        self.storageSettings = storageSettings
         super.init()
 
-        try setupCoreDataStack(at: storageURL)
+        try setupCoreDataStack(at: storageURL, modelURL: storageSettings.modelURL)
     }
 
     // MARK: Core Data
@@ -20,8 +15,8 @@ class CoreDataManager: NSObject {
     private(set) var managedObjectContext: NSManagedObjectContext!
     private(set) var persistentStoreCoordinator: NSPersistentStoreCoordinator!
 
-    private func setupCoreDataStack(at url: URL) throws {
-        guard let mom = NSManagedObjectModel(contentsOf: storageSettings.modelURL) else {
+    private func setupCoreDataStack(at url: URL, modelURL: URL) throws {
+        guard let mom = NSManagedObjectModel(contentsOf: modelURL) else {
             fatalError("Cannot load model")
         }
         managedObjectModel = mom
--- a/Simplenote/CoreDataManager.swift
+++ b/Simplenote/CoreDataManager.swift
@@ -31,8 +31,7 @@ class CoreDataManager: NSObject {
 
         let psc = NSPersistentStoreCoordinator(managedObjectModel: managedObjectModel)
         let options = [NSMigratePersistentStoresAutomaticallyOption: true,
-                       NSInferMappingModelAutomaticallyOption: true,
-                       NSSQLitePragmasOption: storageSettings.journalModeDisabled] as [AnyHashable: Any]
+                       NSInferMappingModelAutomaticallyOption: true]
 
         // Testing logs
         //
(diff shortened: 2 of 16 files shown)


Simplenote/StorageSettings.swift
```diff
@@ -0,0 +1,37 @@
...
+        fileManager.documentsURL.appendingPathComponent(Constants.sqlFile)
+    }
+
+    /// URL for core data storage in shared app group documents directory
+    ///
+    var sharedStorageURL: URL {
+        fileManager.sharedContainerURL.appendingPathComponent(Constants.sqlFile)
+    }
+
+    let journalModeDisabled = [Constants.journalMode: Constants.journalSetting]
```
jleandroperez: Can we drop this from `StorageSettings` and keep it private to both CoreDataManager and the migrator?

(It's okay to duplicate those two IMO!)

Simplenote/StorageSettings.swift
```diff
@@ -0,0 +1,37 @@
...
+}
+
+private struct Constants {
+    static let sqlFile = "Simplenote.sqlite"
+    static let resourceName = "Simplenote"
+    static let resourceType = "momd"
```
jleandroperez: These two should prob. be `modelName` and `modelExtension` (or so?)

Simplenote/Classes/FileManager+Simplenote.swift
```diff
@@ -13,4 +13,29 @@ extension FileManager {
...
+        containerURL(forSecurityApplicationGroupIdentifier: Constants.sharedDirectoryDomain + Constants.groupIdentifier)!
+    }
+}
+
+private struct Constants {
+    static let defaultBundleIdentifier = "com.codality.NotationalFlow"
+    static let groupIdentifier = rootBundleIdentifier() ?? Constants.defaultBundleIdentifier
+    static let sharedDirectoryDomain = "group."
+
+    static func rootBundleIdentifier() -> String? {
```
jleandroperez: ### Few notes!!
- This mechanism is fragile / error prone, since it depends on the `appex` path extension to be present (if that ever, ever changes, it'll stop working properly)
- Most of the code relies on `Bundle`, so we should prob. relocate over to the `Bundle+Simplenote` extension we've got

Pasting an approach a bit more robust below. In order to get this working properly, we also need to add the `APP_EXTENSION` constant in the Extension targets (screenshot below!):

Screen Shot 2021-07-08 at 11 56 07 AM



```
extension Bundle {

    /// Returns the BundleIdentifier
    /// - Important:
    ///     - When invoked from an App Extension, this API will attempt to determine the "Parent App" Bundle identifier
    ///     - You must also make sure App Extension targets contain the `APP_EXTENSION` constant, in the "active compilation conditions" project settings
    ///
    var rootBundleIdentifier: String? {
        guard isAppExtensionBundle else {
            return bundleIdentifier
        }

        let url = bundleURL.deletingLastPathComponent().deletingLastPathComponent()
        let rootIdentifier = Bundle(url: url)?.object(forInfoDictionaryKey: kCFBundleIdentifierKey as String)
        return rootIdentifier as? String
    }
}


// MARK: - Private Helpers
//
private extension Bundle {

    var isAppExtensionBundle: Bool {
    #if APP_EXTENSION
        return true
    #else
        return false
    #endif
    }
}
```

YES this approach requires a bit of more work, and we'd need to add this `APP_EXTENSION` constant in every extension we've got.  But once it's working, there is no reason for it to ever break (Alledgedly)

Simplenote/CoreDataManager.swift
```diff
@@ -0,0 +1,47 @@
...
+        }
+        managedObjectModel = mom
+
+        managedObjectContext = NSManagedObjectContext(concurrencyType: .mainQueueConcurrencyType)
+        managedObjectContext.undoManager = nil
+
+        let psc = NSPersistentStoreCoordinator(managedObjectModel: managedObjectModel)
+        let options = [NSMigratePersistentStoresAutomaticallyOption: true,
+                       NSInferMappingModelAutomaticallyOption: true,
+                       NSSQLitePragmasOption: storageSettings.journalModeDisabled] as [AnyHashable: Any]
```
jleandroperez: (Sorry, duplicating the comment!) let's keep the Journaling detail private (good to duplicate here and in the migrator).

Simplenote/CoreDataManager.swift
```diff
@@ -0,0 +1,47 @@
...
+
+@objcMembers
+class CoreDataManager: NSObject {
+
+    /// Storage Settings
+    ///
+    private let storageSettings: StorageSettings
+
+    init(_ storageURL: URL, storageSettings: StorageSettings = StorageSettings()) throws {
+        self.storageSettings = storageSettings
```
jleandroperez: Since only the `modelURL` will be required, we could probably just pass along the Model URL (and drop the StorageSettings ivar?).

Furthermore, you can probably move `setupCoreDataStack` API in the initializer. Or not, irrelevant!! :joy:

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -405,3 +405,38 @@ extension SPAppDelegate {
...
+        switch migrationResult {
+        case .notNeeded, .success:
+            do {
+                coreDataManager = try CoreDataManager(settings.sharedStorageURL)
+            } catch {
+                coreDataManager = try CoreDataManager(settings.legacyStorageURL)
+            }
+        case .failed:
+            coreDataManager = try CoreDataManager(settings.legacyStorageURL)
+        }
```
jleandroperez, line 436: Pasting @eshurakov's [original comment here](https://github.com/Automattic/simplenote-ios/pull/1329#discussion_r665145788).

The man has bright ideas, indeed!

---

Instead of trying to init a CoreDataManger here and backing off to the legacy storage, we can do that as last step of migration:

- Try to init CoreDataManger with the file that we copied
- if it fails, delete new file, return .failed
- If it succeeds, delete old file (or move it to a backup folder), return .success

And then this code can be simplified to:

```
switch migrationResult {
     case .notNeeded, .success:
          coreDataManager = try CoreDataManager(settings.sharedStorageURL)
     case .failed:
          coreDataManager = try CoreDataManager(settings.legacyStorageURL)
}
```

Simplenote/CoreDataManager.swift
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    // MARK: Core Data
+    private(set) var managedObjectModel: NSManagedObjectModel!
+    private(set) var managedObjectContext: NSManagedObjectContext!
+    private(set) var persistentStoreCoordinator: NSPersistentStoreCoordinator!
+
+    private func setupCoreDataStack(at url: URL) throws {
+        guard let mom = NSManagedObjectModel(contentsOf: storageSettings.modelURL) else {
+            throw NSError(domain: "CoreDataManager", code: 100, userInfo: nil)
```
jleandroperez: This should be a `fatalError("Cannot load Model")`. 100% Irrecoverable

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,99 @@
...
+    init(storageSettings: StorageSettings = StorageSettings(), fileManager: FileManager = FileManager.default) {
+        self.storageSettings = storageSettings
+        self.fileManager = fileManager
+    }
+
+    var legacyStorageExists: Bool {
+        fileManager.fileExists(atPath: storageSettings.legacyStorageURL.path)
+    }
+
+    var sharedStorageExists: Bool {
```
jleandroperez: These two can be private

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,99 @@
...
+        return legacyStorageExists && !sharedStorageExists
+    }
+
+    private func migrateCoreDataToAppGroup() -> MigrationResult {
+        // Testing prints
+        // TODO: Remove prints later
+        print("oldDb exists \(FileManager.default.fileExists(atPath: storageSettings.legacyStorageURL.path))")
+        print(storageSettings.legacyStorageURL.path)
+        print("newDb exists \(FileManager.default.fileExists(atPath: storageSettings.sharedStorageURL.path))")
+        print(storageSettings.sharedStorageURL.path)
```
jleandroperez: Can we clean this up?

You can leave two or three actual `NSLog` statements, something in the lines of `Beginning database migration from [source] to [target]` ?

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,99 @@
...
+        NSLog("Beginning database migration")
+
+        do {
+            try disableJournaling()
+            try migrateCoreDataFiles()
+            NSLog("Database migration successful!!")
+            return .success
+        } catch {
+            NSLog("Could not migrate database to app group")
+            NSLog(error.localizedDescription)
```
jleandroperez:
```
NSLog("Could not migrate database to app group " + error.localizedDescription)
```

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,99 @@
...
+                try psc.remove(store)
+            }
+        } catch {
+            NSLog("Could not remove temporary persistent Store")
+            NSLog(error.localizedDescription)
+        }
+    }
+
+    private func migrateCoreDataFiles() throws {
+        try FileManager.default.copyItem(at: storageSettings.legacyStorageURL, to: storageSettings.sharedStorageURL)
```
jleandroperez:
```suggestion
        try fileManager.copyItem(at: storageSettings.legacyStorageURL, to: storageSettings.sharedStorageURL)
```

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,99 @@
...
+
+        // Remove the persistent store before exiting
+        // If removing fails, the migration can still continue so not throwing the errors
+        do {
+            for store in psc.persistentStores {
+                try psc.remove(store)
+            }
+        } catch {
+            NSLog("Could not remove temporary persistent Store")
+            NSLog(error.localizedDescription)
```
jleandroperez:
```
NSLog("Could not remove temporary persistent Store " + error.localizedDescription)
```

jleandroperez (commented):
Impecable work @charliescheer !!

✅ Verified that `disableJournaling` **effectively** deletes the `-WAL` file (in the sim)
✅ Verified that the `sqlite` file is effectively copied

Only pending questions I've got in mind are:

1. Move or copy?
2. Should we always preserve Journaling disabled? (`CoreDataManager` could just work as it always worked, and journaling could be disabled merely for relocation)
3. Under what conditions should we drop the legacy sqlite file? (risky business). Maybe we should just call move, and period

Thanks for all the effort you've invested into making this PR rock solid!!

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,99 @@
...
+            return .success
+        } catch {
+            NSLog("Could not migrate database to app group")
+            NSLog(error.localizedDescription)
+            CrashLogging.logError(error)
+            return .failed
+        }
+    }
+
+    private func disableJournaling() throws {
```
jleandroperez, line 65: Addendum: can we please add a link to [this Apple TechNote](https://developer.apple.com/library/archive/qa/qa1809/_index.html) for further reference?

(Thank you!!)

Simplenote/SPAppDelegate+Extensions.swift
charliescheer: Love it! and done!

Simplenote/Classes/FileManager+Simplenote.swift
charliescheer: Like this much much better!  I have added this extension to bundle.  Since we have rebased this off develop the widget/intent extensions don't exist here yet, but I will add the APP_EXTENSION constant to them in the feature branch

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    /// This method disables journaling on the core data database
+    /// Per this doc: https://developer.apple.com/library/archive/qa/qa1809/_index.html
+    /// Core data databases have journaling enabled by default, without disabling the journaling first it is possible some notes may get lost in migration
+    ///
+    private func disableJournaling() throws {
+        guard let mom = NSManagedObjectModel(contentsOf: storageSettings.modelURL) else {
+            throw NSError(domain: "SharedStorageMigrator", code: 100, userInfo: nil)
```
jleandroperez: This one would be a fatal error, if ever happens. You can also use force unwrap for this one BTW

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    private func migrateCoreDataFiles() throws {
+        try fileManager.copyItem(at: storageSettings.legacyStorageURL, to: storageSettings.sharedStorageURL)
+    }
+
+    private func attemptCreationOfCoreDataStack() throws {
+        NSLog("Confirming migrated database can be loaded at: \(storageSettings.sharedStorageURL)")
+        let _ = try CoreDataManager(storageSettings.sharedStorageURL)
```
jleandroperez: WDYT if weee

1. Rename `func disableJournaling` > `private func loadPersistentStorage(at storagePath: URL, journaling: Bool) throws`. And parametrize journaling + the Storage URL
2. Invoke `loadPersistentStorage(at: journaling)` from the two spots where `attemptCreationOfCoreDataStack`and `disableJournaling` are called now?

Only reason why doing this might be a good idea is that the current `disableJournaling` code **explicitly** removes the Persistent Store (and CoreDataManager does not!)

jleandroperez (approved):
✅ Verified fresh install on iOS 12 / 13 / 14
✅ Verified migration flows in iOS 12 / 13 / 14

Few nipticky notes, but other than that, :shipit: on my end.
Thank you for the attention to detail on this one!!

Please let's wait for @eshurakov's triple check on this one, just in case. 

IMHO I'd feel much comfortable knowing that the original storage is still there, if anything goes wrong. Worst case scenario, migration will fail each time, until the device has available space.

Simplenote/SharedStorageMigrator.swift
charliescheer: Should it though?  In this case if the MOM isn't loaded the migration fails and the app can continue with launching normally.  I mean.... if it fails here it will probably also fail when the core data manager tries the same thing to launch the app, but my thought here was to allow the migrator to fail and the app continue, so I wouldn't think fatalError() would be the right choice for this MOM

Simplenote/SharedStorageMigrator.swift
charliescheer: Is it wasteful create `private func loadPersistentStorage(at storagePath: URL, journaling: Bool) throws `  as it's own method and then call it from  `func disableJournaling` and `attemptCreationOfCoreDataStack` (maybe change that to something like `verifyDataBaseLoadable`) and then in those methods print a few log messages

My thought here is that the migration becomes less readable to me like this:
    do {
            try loadPersistentStorage(at: URL, journaling: true)()
            try migrateCoreDataFiles()
            try loadPersistentStorage(at: URL, journaling: false)()
            NSLog("Database migration successful!!")
            return .success
}
compared to:
     do {
            try disableJournaling()
            try migrateCoreDataFiles()
            try attemptCreationOfCoreDataStack()
            NSLog("Database migration successful!!")
            return .success
}

charliescheer: I made a quick update that shows what I was thinking with this.  WDYT?

Simplenote/SharedStorageMigrator.swift
eshurakov: If MOM is missing, something is really wrong with the app and continuing won't do any good. I also think we should just crash here.

eshurakov (approved):
Looks good to me 👍 

> IMHO I'd feel much comfortable knowing that the original storage is still there, if anything goes wrong. Worst case scenario, migration will fail each time, until the device has available space.

I feel uneasy if we keep the old DB at the same location as it was. Feels like it's still in use. Considering that we checked that new DB can be initialised, I say we can either delete old DB or move it to some other location in the documents folder (consider it being a backup 😸)

Simplenote/SharedStorageMigrator.swift
charliescheer: makes sense.  Have changed that.

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,125 @@
...
+enum MigrationResult {
+    case success
+    case notNeeded
+    case failed
+}
+
+private struct Constants {
+    static let journalMode = "journal_mode"
+    static let journalSetting = "DELETE"
+    static let journalModeDisabled = [Constants.journalMode: Constants.journalSetting]
```
jleandroperez: You can prob. just use `"journal_mode": "DELETE"` (since the sub-constants aren't used anywhere else)

charliescheer:
I added a method to the end of the migration process that moves the legacy DB file from /appDocuments/Simplenote.sqlite to /appDocuments/Simplenote.sqlite.old/  So the file is maintained but no longer accessible by the app without us reversing the migration in some way.
@eshurakov  @jleandroperez  WDYT?

eshurakov:
> moves the legacy DB file from /appDocuments/Simplenote.sqlite to /appDocuments/Simplenote.sqlite.old

I like it!

jleandroperez (approved):
Congratz!! :shipit:

PR #1354, opened 2021-07-12
In app account deletion

### LINKED ISSUE #1353: In app account deletion
### What
add the ability to delete accounts from within Simplenote.  Details  of this project can be found at: p2XJRt-3bK-p2

UI:

Tasks:

- [x] create new row/alert UI for settings page
- [x] create API method to trigger account deletion request
- [x] tracks events to track delete request

### Why
Upcoming update to Apple's use guidelines require apps on the app store that have accounts to allow them to be deleted from within the app

charliescheer:
### Fix
This PR Fixes #1353 

Per upcoming changes to the app store use guidelines we need to add In app account deletion.  This PR adds this feature to SNiOS

Note that currently the end point we are using to trigger deletion is in progress so a temporary url was added for testing purposes.

Note that currently the PR is missing some details needed to make it run.  Ping me for the details needed to make it compile
Discussion here p1627443887039600-slack-C01KRRK0MLJ
### Test
1. Log into simplenote to an account you don't mind deleting
2. Go to Settings -> Delete Account
3. Tap on the delete account button, confirm deletion
4. After a few seconds you should see a confirmation button.  Click on Okay
5. Check the email associated with the account you should receive a verification email to confirm deleting.

Because of the changes to Remote and Result, also test signing up to Simplenote from the app to ensure the SignupRemote still operates correctly

### Release
> `RELEASE-NOTES.txt` was updated in 37db49 with:
> 
> > -   Added In app account deletion #1354


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -948,6 +951,9 @@
 		BA55124D2600210B00D8F882 /* TimerFactory.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TimerFactory.swift; sourceTree = "<group>"; };
 		BA55B05925F067DF0042582B /* NoticePresenter.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NoticePresenter.swift; sourceTree = "<group>"; };
 		BA55B06225F068650042582B /* NoticeController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = NoticeController.swift; sourceTree = "<group>"; };
+		BA5768D9269A7E90008B510E /* DeleteAccountRemote.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DeleteAccountRemote.swift; sourceTree = "<group>"; };
+		BA5768E2269A803F008B510E /* Remote.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Remote.swift; sourceTree = "<group>"; };
+		BA5768EB269BE4D0008B510E /* AccountDeletionHandler.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AccountDeletionHandler.swift; sourceTree = "<group>"; };
```
jleandroperez: Nipticky: Can we group all of the `*Remote` classes in the same Xcode folder?

Simplenote/Remote.swift
```diff
@@ -0,0 +1,47 @@
...
+    private let urlSession: URLSession
+
+    init(urlSession: URLSession = URLSession.shared) {
+        self.urlSession = urlSession
+    }
+
+    /// Send  task for remote
+    /// Sublcassing Notes: To be able to send a task it is required to first setup the URL request for the task to use
+    ///
+    func task(with request: URLRequest, completion: @escaping (_ result: Result) -> Void) {
```
jleandroperez: This API signature would imply that a `task` is returned.

WDYT about something like... `func performDataTask(with request: URLRequest, completion:..)` ?

Simplenote/DeleteAccountRemote.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+class DeleteAccountRemote: Remote {
```
jleandroperez: WDYT about `AccountRemote` (and merge with `AccountVerificationRemote`), or `AccountDeletionRemote` (for consistency?)

Simplenote.xcodeproj/project.pbxproj
jleandroperez: P.s.: There's also `AccountVerificationRemote` 🔥

Simplenote/AccountDeletionHandler.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+class AccountDeletionController {
```
jleandroperez: Mixed feelings about this controller!! (because it really does not add value). Maybe we should keep it, for consistency reasons (no other *Remote* is used directly in any ViewController).

If we do keep it:

1. The filename doesn't match the class name
2. We should move it to the `Controllers` logic folder?
3. The `SPTracker` invocation should go at the ViewController level.
3.1. Why: because `requestAccountDeletion` doesn't really have "knowledge" of whether this happened because of a Button Tap, or else.

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,55 @@ private extension BiometricAuthentication.Biometry {
...
+        present(alert, animated: true, completion: nil)
+    }
+
+    private func accountDeletionController(for user: SPUser) -> AccountDeletionController {
+        let deletionController = AccountDeletionController()
+        deletionController.successHandler = {
+            self.presentSuccessAlert(for: user)
+        }
+
+        return deletionController
```
jleandroperez: For consistency, we could probably move the `AccountDeletionController` initialization directly into `deleteAccount`, and move the confirmation Alert to `func persentDeletionConfirmationAlert(onConfirm: ())`

That would leave you with two `present...` API(s), and "the core business logic" handled by `deleteAccount`

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,55 @@ private extension BiometricAuthentication.Biometry {
         }
     }
 }
+
+// MARK: - Account Deletion
+extension SPSettingsViewController {
+
+    @objc
+    func deleteAccount() {
```
jleandroperez: Suggestion:

A. `func deleteAccountWasPressed()`
B. `func presentAccountDeletionConfirmation`

Why: because `deleteAccount` would imply that the account is immediately deleted, but there's a confirmation step beforehand

jleandroperez (commented):
Sending few really nipticky notes, looking good Charlie!!

Simplenote/DeleteAccountRemote.swift
charliescheer: I like the idea of combining the account remotes :)

Simplenote/AccountDeletionHandler.swift
charliescheer: >Mixed feelings about this controller!! (because it really does not add value). Maybe we should keep it, for consistency reasons (no other Remote is used directly in any ViewController).

This is the thought process I went though.  Ultimately I decided to keep the controller so that the VC didn't know anything about the remote, but it also feels like a waste... I've made the requested changes.  

_Maybe_ we could look at combining the AccountDeletionController into AccountVerificationController for something like `AccountStateController`  I'm not really in love with the idea, but since I combined the account remotes it would at least be consistent.  WDYT?

Simplenote/Classes/SPSettingsViewController+Extensions.swift
charliescheer: I dig it

Simplenote/Remote.swift
```diff
@@ -0,0 +1,56 @@
...
+        case failure(_ statusCode: Int, _ error: Error?)
+
+        static func random() -> Result {
+            let random = arc4random_uniform(1)
+            if random == 0 {
+                return .failure(0, nil)
+            } else {
+                return .success
+            }
+        }
```
eshurakov: I suggest to move this to a separate extension in unit tests

Simplenote/SPAppDelegate.m
```diff
@@ -195,6 +195,7 @@ - (void)applicationDidEnterBackground:(UIApplication *)application
 - (void)applicationWillEnterForeground:(UIApplication *)application
 {
     [self dismissPasscodeLockIfPossible];
+    [self.simperium authenticateIfNecessary];
```
eshurakov: Wonder if there is a better way 🤔 
I think this method works because it forces web socket reconnection.

jleandroperez: YES we should avoid doing this. Ongoing conversation via P2, outcome still TBD!

charliescheer: Per the continuing conversation I have added a timestamp stored in app delegate to mark a deletion request.  Only if a deletion request was made in the last 24 hours will the [self.simperium authenticateIfNecessary]; get called

SylvesterWilmott:
Thanks @charliescheer 

The design all looks great to me. One small request. Can we update the red dialog button - "Request Account Deletion" to use our red tint?

I tried pressing the delete account button multiple times and after a few times, the second dialog was not shown. After trying a few more times I was able to see the second dialog again. Not sure if this is a consequence of having already had the email sent but thought I was mention it just in case.

I also tried turning off my data to see the error but after pressing "Request Account Deletion" the app crashes. This is on iOS 15, iPhone X.

charliescheer:
@SylvesterWilmott thanks for taking a look! 

> I tried pressing the delete account button multiple times and after a few times, the second dialog was not shown. After trying a few more times I was able to see the second dialog again. Not sure if this is a consequence of having already had the email sent but thought I was mention it just in case.

I am able to replicate the same thing.  I suspect that what is going on here is that the request is getting marked as spam and then getting rejected.  @sandymcfadden is that something that might happen with the current setup of the endpoint?  If so, is there any sort of failure response that we would get back if the request failed?

> I also tried turning off my data to see the error but after pressing "Request Account Deletion" the app crashes. This is on iOS 15, iPhone X.

Nice catch.  I fixed that in the process of adding the failure messages.  

If you want to check it out again it's ready to go 
cc: @jleandroperez

SylvesterWilmott:
Looking good @charliescheer 

Did we add a slide-in transition to the loader or maybe I missed that the first time round. I would suggest simply displaying the loader without any transition in or out since sometimes there's barely enough time to see the loader before it moves out of view again.

charliescheer:
>Did we add a slide-in transition to the loader or maybe I missed that the first time round. I would suggest simply displaying the loader without any transition in or out since sometimes there's barely enough time to see the loader before it moves out of view again.

I changed the way the wait alert mechanism was being created from using a built in UI alert to creating a custom view which was being presented modally.  The animations came along with the modal presentation.  I removed the animation so that will just pop in and pop out now @SylvesterWilmott

SylvesterWilmott (approved):
The design looks good. Thanks @charliescheer.

sandymcfadden:
Hey @charliescheer and @SylvesterWilmott 

>I am able to replicate the same thing. I suspect that what is going on here is that the request is getting marked as spam and then getting rejected. @sandymcfadden is that something that might happen with the current setup of the endpoint? If so, is there any sort of failure response that we would get back if the request failed?

I did notice that I was doing something wrong with the return when requesting a to delete when a valid request already exists. I was returning an error but with a 202 status code. I've updated it so it shouldn't show an error but return a 202 status code. I'm not sure if that helps with this issue, but it's the only thing I could see that might cause an issue.

charliescheer:
@eshurakov @jleandroperez can I get another look at this when you have a chance.  I think it is good to go, just waiting on the end point, but want to make sure it is ready as soon as the backend changes are set

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,77 @@ private extension BiometricAuthentication.Biometry {
...
+        deletionController.failureHandler = { status in
+            if status == 0 {
+                pendingAlert.stopAnimating()
+                pendingAlert.dismiss(animated: false, completion: nil)
+                NoticeController.shared.present(NoticeFactory.networkError())
+                return
+            }
+
+            let alert = UIAlertController(title: AccountDeletion.errorTitle, message: AccountDeletion.errorMessage, preferredStyle: .alert)
+            alert.addDefaultActionWithTitle(AccountDeletion.ok)
```
eshurakov: Looks like either dead code or sad alert that never got presented

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,77 @@ private extension BiometricAuthentication.Biometry {
...
+            self.presentSuccessAlert(for: user)
+        }
+
+        deletionController.failureHandler = { status in
+            if status == 0 {
+                pendingAlert.stopAnimating()
+                pendingAlert.dismiss(animated: false, completion: nil)
+                NoticeController.shared.present(NoticeFactory.networkError())
+                return
+            }
```
eshurakov: If status is not 0, pending alert will keep spinning.

Simplenote/Controllers/AccountDeletionController.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+class AccountDeletionController {
+    var successHandler: (() -> Void)?
+    var failureHandler: ((Int) -> Void)?
```
eshurakov: Curious if there is a benefit of using two separate handlers and not send a completion directly to `requestAccountDeletion`. I feel that would be a more common pattern.

It's also hard to tell what `Int` is in `((Int) -> Void)?`.

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -405,3 +405,25 @@ extension SPAppDelegate {
         EditorFactory.shared.scrollPositionCache.cleanup(keeping: allIdentifiers)
     }
 }
+
+// MARK: - Account Deletion
+//
+extension SPAppDelegate {
+    @objc
+    func authenticateLoginTokenIfNeeded() {
+        guard let accountDeletionRequestDate = accountDeletionRequestDate else {
```
eshurakov: Probably can write
`guard let accountDeletionRequestDate = accountDeletionRequestDate as Date else {`

Simplenote/Controllers/AccountDeletionController.swift
```diff
@@ -0,0 +1,21 @@
...
+
+    func requestAccountDeletion(_ user: SPUser) {
+        SPAppDelegate.shared().accountDeletionRequestDate = Date()
```
eshurakov: I think we should set it only if the request was actually successful.

Simplenote/ActivityIndicatorAlertViewController.swift
```diff
@@ -0,0 +1,68 @@
+import UIKit
+
+class ActivityIndicatorAlertViewController: UIViewController {
```
eshurakov: I got confused by `AlertView` in the name :-)

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,77 @@ private extension BiometricAuthentication.Biometry {
...
+
+            let alert = UIAlertController(title: AccountDeletion.errorTitle, message: AccountDeletion.errorMessage, preferredStyle: .alert)
+            alert.addDefaultActionWithTitle(AccountDeletion.ok)
+        }
+
+        presentAccountDeletionConfirmation { (_) in
```
eshurakov: I think it's a good idea to keep `onConfirm` here for clarity, otherwise it's not clear what the callback is.
And if we don't use `UIAlertAction`, let's remove it.

Simplenote/AccountRemote.swift
```diff
@@ -0,0 +1,64 @@
...
+// MARK: - AccountVerificationRemote
+//
+class AccountRemote: Remote {
+    // MARK: Performing tasks
+
+    /// Send verification request for specified email address
+    ///
+    func verify(email: String, completion: @escaping (_ result: Result) -> Void) {
+        guard let request = verificationURLRequest(with: email) else {
+            return
```
jleandroperez, line 13: Missing callback on failure

Simplenote/AccountRemote.swift
```diff
@@ -0,0 +1,64 @@
+import Foundation
+
+// MARK: - AccountVerificationRemote
+//
+class AccountRemote: Remote {
+    // MARK: Performing tasks
+
+    /// Send verification request for specified email address
+    ///
+    func verify(email: String, completion: @escaping (_ result: Result) -> Void) {
```
jleandroperez: Since we're changing things around, let's switch to Swift's [Result Type](https://developer.apple.com/documentation/swift/result) here

Simplenote/ActivityIndicatorAlertViewController.swift
jleandroperez: Yup! How about `SpinnerViewController` or something like that?

Simplenote/ActivityIndicatorAlertViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+    private func setupViewAppearance() {
+        modalPresentationStyle = .overFullScreen
+
+        if #available(iOS 13.0, *) {
+            activityIndicator.style = .large
+        } else {
+            activityIndicator.style = .gray
+        }
+
+        view.backgroundColor = UIColor(studioColor: .gray50, alpha: 0.5)
```
jleandroperez:
```suggestion
        view.backgroundColor = UIColor(studioColor: .gray50, alpha: UIKitConstants.alpha0_5)
```

Simplenote/ActivityIndicatorAlertViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+        alertView.backgroundColor = UIColor(studioColor: .white, alpha: 1)
```
jleandroperez:
```suggestion
        alertView.backgroundColor = UIColor(studioColor: .white, alpha: UIKitConstants.alpha1_0)
```

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,77 @@ private extension BiometricAuthentication.Biometry {
...
+            deletionController.requestAccountDeletion(user)
+        }
+    }
+
+    private func presentAccountDeletionConfirmation(onConfirm: @escaping ((UIAlertAction) -> Void)) {
+        let alert = UIAlertController(title: AccountDeletion.deleteAccount,
+                                      message: AccountDeletion.confirmAlertMessage,
+                                      preferredStyle: .alert)
+        let deleteAccountButton = UIAlertAction(title: AccountDeletion.deleteAccountButton, style: .destructive, handler: onConfirm)
+        deleteAccountButton.setValue(UIColor.simplenoteRedButtonColor, forKey: "titleTextColor")
```
jleandroperez: Let's avoid using private API(s)

Simplenote/Controllers/AccountDeletionController.swift
jleandroperez: ^ I was gonna suggest the same thing. Swift's [Result Type](https://developer.apple.com/documentation/swift/result), and a single callback, passed directly to `requestAccountDeletion`

Simplenote/Controllers/AccountDeletionController.swift
jleandroperez: Plus perhaps the caller should do this. Feels a bit odd to do it here?

Simplenote/Date+Simplenote.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+extension Date {
+    func increased(by days: Int) -> Date? {
```
jleandroperez:
```suggestion
    func increased(byDays days: Int) -> Date? {
```

jleandroperez: Otherwise when you invoke `date.increased(by: 123)` it's not clear what's going on

Simplenote/Remote.swift
```diff
@@ -0,0 +1,47 @@
...
+                return code1 == code2
+            default:
+                return false
+            }
+
+        }
+
+        case success
+        case failure(_ statusCode: Int, _ error: Error?)
+    }
```
jleandroperez: Can we use `https://developer.apple.com/documentation/swift/result` instead?

Since it's got Generics support, should be `Result<Int, Error>` I believe?

SimplenoteTests/Verification/AccountVerificationRemoteTests.swift
```diff
@@ -5,25 +5,26 @@ import XCTest
...
         for _ in 0..<5 {
-            test(withStatusCode: Int.random(in: 400..<600), expectedResult: false)
+            let statusCode = Int.random(in: 400..<600)
+            test(withStatusCode: statusCode, expectedResult: Remote.Result.failure(statusCode, nil))
         }
     }
 
     func testFailureWhenNoResponse() {
-        test(withStatusCode: nil, expectedResult: false)
+        test(withStatusCode: nil, expectedResult: Remote.Result.failure(0, nil))
```
jleandroperez: By switching to Swift Result Type, this would end up looking like:

```
test(withStatusCode: nil, expectedResult: .failure(unknown))
```

With the detail that we should, probably, not allow for null errors.

Simplenote/Classes/SPSettingsViewController+Extensions.swift
charliescheer: That should get presented if the response code from the server is anything other than 2xx or 0
I am not sure that there are any cases where we would get back a not 2xx return value.

Simplenote/ActivityIndicatorAlertViewController.swift
charliescheer: Works for me :)

Simplenote/Classes/SPSettingsViewController+Extensions.swift
eshurakov: There is no code to present it, current code only creates an alert but doesn't present it.

charliescheer: Oh shoot, how did I miss that?  Thank you!!

charliescheer:
Thanks for the feedback @eshurakov  and @jleandroperez 

I have made the changes that you were asking about.  Check it out know and let me know what you think about the new changes!

charliescheer:
Just dropping a note here that with the creation of the Remote super class and the change to using the Swift Result, I have added to the testing steps to confirm that signup still works.

This PR should be ready to be reviewed again.  Thanks!

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,85 @@ private extension BiometricAuthentication.Biometry {
...
+// MARK: - Account Deletion
+extension SPSettingsViewController {
+
+    @objc
+    func deleteAccountWasPressed() {
+        guard let user = SPAppDelegate.shared().simperium.user else {
+            return
+        }
+        SPTracker.trackDeleteAccountButttonTapped()
+        spinnerViewController = SpinnerViewController()
```
jleandroperez: An alternative that would allow you to drop the `spinnerViewController` ivar could be:

```
extension UIViewController {

    func addChildAndPresent(_ childViewController: UIViewController, animated: Bool) {
        addChild(childViewController)
        present(childViewController, animated: animated, completion: nil)
    }

    func dismissAndRemoveFromParent(animated: Bool) {
        childViewController.dismiss(animated: animated, completion: nil)
        childViewController.removeFromParent
    }
}
```

These API(s) would allow you to do something like this:

```
let spinnerViewController = SpinnerViewController()
addChildAndPresent(spinnerViewController, animated: false)

/// ...

presentAccountDeletionConfirmation { (_) in
   self.present(spinnerViewController, animated: false, completion: nil)

   deletionController.requestAccountDeletion(user) { [weak self] (result) in
      spinnerViewController.dismissAndRemoveFromParent(animated: false)
      self?.handleDeletionResult(user: user, result)
   }
 }
```

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,85 @@ private extension BiometricAuthentication.Biometry {
...
+        spinnerViewController.startAnimating()
+
+        let deletionController = AccountDeletionController()
+        SPAppDelegate.shared().accountDeletionController = deletionController
```
jleandroperez: This instance should probably be already initialized. Check out `setupVerificationController`

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,85 @@ private extension BiometricAuthentication.Biometry {
...
+extension SPSettingsViewController {
+
+    @objc
+    func deleteAccountWasPressed() {
+        guard let user = SPAppDelegate.shared().simperium.user else {
+            return
+        }
+        SPTracker.trackDeleteAccountButttonTapped()
+        spinnerViewController = SpinnerViewController()
+        spinnerViewController.startAnimating()
```
jleandroperez: This probably belongs to SpinnerViewController. (`viewWillAppear` maybe?)

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -88,3 +88,85 @@ private extension BiometricAuthentication.Biometry {
...
+    }
+
+    private func handleDeletionResult(user: SPUser, _ result: Result<Data?, RemoteError>) {
+        spinnerViewController.stopAnimating()
+        spinnerViewController.dismiss(animated: false, completion: nil)
+        switch result {
+        case .success:
+            presentSuccessAlert(for: user)
+        case .failure(let error):
+            if error.statusCode == 0 {
```
jleandroperez: This one is coupled. Ideally the Error type, itself, lets you determine this (the knowledge that `statusCode == 0` mean Network Error is a bit obscure).

No need to change anything ATM, though, since this is in line with pre-existant code

Simplenote/Date+Simplenote.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+extension Date {
+    func increased(byDays days: Int) -> Date? {
+        var components = DateComponents()
+        components.day = days
+        let calendar = Calendar.current
```
jleandroperez: No need to assign `let calendar` since it'll only be used one line below

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -405,3 +405,21 @@ extension SPAppDelegate {
         EditorFactory.shared.scrollPositionCache.cleanup(keeping: allIdentifiers)
     }
 }
+
+// MARK: - Account Deletion
+//
+extension SPAppDelegate {
+    @objc
+    func authenticateLoginTokenIfNeeded() {
```
jleandroperez: Nipticky: the name doesn't match the purpose of the API, since no login token is authenticated.

Maybe either a more verbose name, or just a comment detailing why this is here?

Simplenote/SpinnerViewController.swift
```diff
@@ -0,0 +1,68 @@
+import UIKit
+
+class SpinnerViewController: UIViewController {
+    private var alertView: UIView!
+    private var activityIndicator: UIActivityIndicatorView!
```
jleandroperez: Nipticky: since you're not using nibs, you can drop the Implicitly Unwrapped Optionals:

```
private lazy var alertView = UIView()
private lazy var activityIndicator = UIActivityIndicatorView()
```

Simplenote/SpinnerViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+        alertView.translatesAutoresizingMaskIntoConstraints = false
+        activityIndicator.translatesAutoresizingMaskIntoConstraints = false
+
+        view.addSubview(alertView)
+
+        NSLayoutConstraint.activate([
+            alertView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+            alertView.centerYAnchor.constraint(equalTo: view.centerYAnchor),
+            alertView.widthAnchor.constraint(equalToConstant: 150),
+            alertView.heightAnchor.constraint(equalToConstant: 115)
```
jleandroperez: Magic numbers 🔥

SimplenoteTests/RemoteResult+TestHelpers.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+@testable import Simplenote
+
+extension Remote {
+    static func randomResult() -> Result<Data?, RemoteError> {
+        let random = arc4random_uniform(1)
+        if random == 0 {
+            return .failure(RemoteError(statusCode: 0))
+        } else {
```
jleandroperez: Nipticky: No need for `else`, since you're returning in L8

SimplenoteTests/RemoteResult+TestHelpers.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+@testable import Simplenote
+
+extension Remote {
+    static func randomResult() -> Result<Data?, RemoteError> {
+        let random = arc4random_uniform(1)
```
jleandroperez: Out of curiosity!!! I've noticed the original code uses `Bool.random()`. What'd be the reason for this change? (purely curious!!)

jleandroperez (commented):
Hey there @charliescheer !!

Verified over here, the flow is super **SUPER** smooth!!. Sending several nipticky notes your way, no blockers.

Please also check these API(s) from UIVIewController+Simplenote, since they may help you display the spinner with a smooth fadeIn animation!

```
func attach(to parent: UIViewController, attachmentView: AttachmentView? = nil, animated: Bool = false) { }
func detach(animated: Bool = false) { }
```

Simplenote/Classes/SPSettingsViewController+Extensions.swift
charliescheer: I had left this un initialized cause we would only need it if someone had requested an account deletion.  Otherwise the app delegate doesn't need to have an active deletion controller.  

I can set it up, but since initializing didn't require any configuration I felt that setting it up here when the user went to delete was an okay time to create this object, and other wise leave it as nil

WDYT?

Simplenote/Classes/SPSettingsViewController+Extensions.swift
charliescheer: I made some changes to the remote error that removes the need for the == 0 and instead defines the error as a enum.
I think this makes it a bit better, and by using
`
case requestError(Int, Error?) 
`

we can still pass the error details and status code to the existing remotes.

Simplenote/SPAppDelegate+Extensions.swift
charliescheer: Changed the name to
`
authenticateSimperiumIfNeeded
`

SimplenoteTests/RemoteResult+TestHelpers.swift
charliescheer: Previously the result of the closures were capturing a Bool and an Error, and this function.

Since I was changing it to no longer returning bool I went to do a random in a different way.  But since you asked, I realized that it was still a binary choice and I could use Bool.Random to choose success or failure, and then return the correct value. Have updated the function

Simplenote/AccountRemote.swift
```diff
@@ -0,0 +1,65 @@
...
+// MARK: - AccountVerificationRemote
+//
+class AccountRemote: Remote {
+    // MARK: Performing tasks
+
+    /// Send verification request for specified email address
+    ///
+    func verify(email: String, completion: @escaping (_ result: Result<Data?, RemoteError>) -> Void) {
+        guard let request = verificationURLRequest(with: email) else {
+            completion(.failure(RemoteError.network))
```
jleandroperez: BTW this is slightly off, since this specific failure would occur if the URL cannot be constructed. Not really a network error.

Simplenote/AccountRemote.swift
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        performDataTask(with: request, completion: completion)
+    }
+
+    /// Send account deletion request for user
+    ///
+    func requestDelete(_ user: SPUser, completion: @escaping (_ result: Result<Data?, RemoteError>) -> Void) {
+        guard let request = deleteRequest(with: user) else {
+            completion(.failure(RemoteError.network))
```
jleandroperez: Same note as above

Simplenote/Classes/SPSettingsViewController.h
```diff
@@ -1,6 +1,8 @@
 #import <UIKit/UIKit.h>
 #import "SPTableViewController.h"
 
+@class SpinnerViewController;
```
jleandroperez: You can drop this one, since it's not required here

Simplenote/RemoteError.swift
```diff
@@ -0,0 +1,21 @@
...
+    case network
+    case requestError(Int, Error?)
+}
+
+extension RemoteError: Equatable {
+    static func == (lhs: RemoteError, rhs: RemoteError) -> Bool {
+        lhs.localizedDescription == rhs.localizedDescription
+    }
+
+    init(statusCode: Int, dataTaskError: Error? = nil) {
```
jleandroperez: It'd be cleaner to drop this constructor, and explicitly use `.network` or `.requestError`.

Otherwise, knowledge of `0 == .network` remains coupled

Simplenote/SpinnerViewController.swift
```diff
@@ -0,0 +1,74 @@
...
+
+        view.addSubview(alertView)
+
+        NSLayoutConstraint.activate([
+            alertView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+            alertView.centerYAnchor.constraint(equalTo: view.centerYAnchor),
+            alertView.widthAnchor.constraint(equalToConstant: Constants.width),
+            alertView.heightAnchor.constraint(equalToConstant: Constants.height)
+        ])
+        alertView.layer.cornerRadius = 15
```
jleandroperez: Magic number 🔥

Simplenote/Classes/SPAuthHandler.swift
```diff
@@ -72,13 +72,20 @@ class SPAuthHandler {
...
-            case .failure(let statusCode, let error):
-                let error = SPAuthError(signupErrorCode: statusCode, response: nil, error: error)
-                onCompletion(error)
+            case .failure(let error):
+                onCompletion(self.handleRemoteError(error))
             }
         }
     }
 
+    private func handleRemoteError(_ error: RemoteError) -> SPAuthError {
```
jleandroperez: Note that this API doesn't really handles remote errors, so the name is a bit misleading. 

-`errorFromRemoteError()`
- `authenticationError(for remoteError: RemoteError)`

Or so?

Simplenote/Classes/SPAuthHandler.swift
```diff
@@ -72,13 +72,20 @@ class SPAuthHandler {
...
+        switch error {
+        case .network:
+            return SPAuthError(signupErrorCode: 0, response: nil, error: nil)
```
jleandroperez: Rather than coupling `SPAuthError` we could maybe add a new case in there?

Simplenote/Classes/SignupRemote.swift
```diff
@@ -2,53 +2,14 @@ import Foundation
...
-    }
-
-    /// Send signup request for specified email address
-    ///
-    func signup(with email: String, completion: @escaping (_ result: Result) -> Void) {
+class SignupRemote: Remote {
+    func signup(with email: String, completion: @escaping (_ result: Result<Data?, RemoteError>) -> Void) {
         guard let request = request(with: email) else {
-            completion(.failure(0, nil))
+            completion(.failure(RemoteError.network))
```
jleandroperez: Note that this isn't really a network error (but a URLRequest that couldn't be built)

jleandroperez (commented):
Few nipticky ones, only real important one is that the Dark Mode's spinner BG is light.

Thank you Charlieee

charliescheer:
Thanks for the feedback @jleandroperez .  Dark mode problem fixed and updated for the rest of the comments.

Simplenote/Classes/SimplenoteConstants.swift
```diff
@@ -36,4 +36,5 @@ class SimplenoteConstants: NSObject {
     static let settingsURL          = currentEngineBaseURL.appendingPathComponent("/settings")
     static let signupURL            = currentEngineBaseURL.appendingPathComponent("/account/request-signup")
     static let verificationURL      = currentEngineBaseURL.appendingPathComponent("/account/verify-email/")
+    static let accountDeletion      = currentEngineBaseURL.appendingPathComponent("/account/request-delete/")
```
jleandroperez:
```suggestion
    static let accountDeletionURL   = currentEngineBaseURL.appendingPathComponent("/account/request-delete/")
```

jleandroperez (approved):
Nice work @charliescheer !!

:shipit: sir!!

jleandroperez:
P.s.: make sure to set the correct release version in the PR itself 🔥

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -399,6 +399,24 @@ extension SPAppDelegate {
...
+            return
+        }
+
+        if deletionController.deletionTokenHasExpired {
+            self.accountDeletionController = nil
+            return
+        }
+
+        simperium.authenticateIfNecessary()
+    }
```
eshurakov, line 417: Unfortunately this won't work as expected and can end up crashing the app.
When accessing objc code via swift it will go via property getter, which will create a new `accountDeletionController`.

Imagine the following flow:
* Launch the app
* Put into background
* Put into foreground, `authenticateSimperiumIfNeeded` is invoked and `accountDeletionController` is created. `accountDeletionRequestDate` is set to the current date.
* Because the token is not expired, we call `simperium.authenticateIfNecessary()` every time the app is opened (and we didn't initiate account deletion)
* If we request account deletion within a day, the expiration date won't be correct as it was already set when `accountDeletionController` was created
* If we request account deletion after a day has passed (considering the app wasn't killed), the app will crash as `accountDeletionController` will be nil.

RELEASE-NOTES.txt
```diff
@@ -9,6 +9,7 @@
 
 4.40
 -----
+-   Added In app account deletion #1354
```
oguzkocer, line 12: @charliescheer Isn't this supposed to be in `4.42` instead?

PR #1357, opened 2021-07-13
Fixing failing Search tests and `addBulletedLists` test (again)

pachlava:
### Fix
This PR hopefully fixes two issues:
1. `testBulletedLists` test still continues to fail on CI, and now it looks like the reason was in trying to enter the text into note editor before the text view actually appeared. Wait for text view to appear has been added.
**Edit:** besides this, moving the call to enable Markdown in the note to an earlier stage (as done for all other tests) seems to finally solve the issue. 🧙 

2. `setUp()` part of the Search tests started failing in the recent version, because of the text bubble (can be seen on picture) covering NavBar button after pasting text from clipboard:

Screenshot 2021-07-13 at 23 39 37

The fail occurred at the point of tapping the `Dismiss Keyboard` button - it was not tappable. In this case, I decided to wait for `Dimiss Keyboard` button to become hittable (which means the bubble is gone), instead of hiding the bubble with a swipe.

### Test
Seeing `Subset` and `Full` suits passing on CI should be enough, I believe. [Here](https://app.circleci.com/pipelines/github/Automattic/simplenote-ios?branch=UI-tests-fix-failing-search-tests) you can see that there are:
- 5 successful `Subset` runs since [this commit](https://github.com/Automattic/simplenote-ios/commit/ddb92552b6df87f34409186c72f5fe0378a57537). `testBulletedLists` is a part of this suite.
- One `Full` run, which contains Search tests (and all the other tests).

### Review
I've assigned two reviewers: @thehenrybyrd and @mokagio, but one is enough. _Thank you!_ 🙇 

### Release

### DIFF
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -34,6 +34,16 @@ class NoteEditor {
             noteContentTextView.clearText()
             noteContentTextView.paste(text: enteredValue)
 
+            // Once the text is pasted, there might be an info bubble shown at the note top,
+            // saying "Simplenote pasted from Simplenote-UITestsRunner". Since it covers all
+            // navbar buttons, we have to wait for it to disappear:
+            let isHittablePredicate = NSPredicate { _, _ in
+                app.buttons[UID.Button.dismissKeyboard].isHittable == true
+            }
+
+            let expectation = XCTNSPredicateExpectation(predicate: isHittablePredicate, object: .none)
+            XCTWaiter().wait(for: [expectation], timeout: 0.5)
+
             // Swipe up fast to show tags input, which disappears if pasted text is large
             // enough to push tags input off screen
             noteContentTextView.swipeUp(velocity: .fast)
@@ -47,6 +57,7 @@ class NoteEditor {
     }
 
     class func setFocus() {
+        _ = app.textViews.firstMatch.waitForExistence(timeout: averageLoadTimeout)
         app.textViews.firstMatch.tap()
     }
 
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -36,7 +36,7 @@ class NoteEditor {
 
             // Once the text is pasted, there might be an info bubble shown at the note top,
             // saying "Simplenote pasted from Simplenote-UITestsRunner". Since it covers all
-            // navbar buttons, we have to wait for it to disappear:
+            // NavBar buttons, we have to wait for it to disappear:
             let isHittablePredicate = NSPredicate { _, _ in
                 app.buttons[UID.Button.dismissKeyboard].isHittable == true
             }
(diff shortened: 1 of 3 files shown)


SimplenoteUITests/NoteEditor.swift
```diff
@@ -34,6 +34,17 @@ class NoteEditor {
...
+            // Once the text is pasted, there might be an info bubble shown at the note top,
+            // saying "Simplenote pasted from Simplenote-UITestsRunner". Since it covers all
+            // NavBar buttons, we have to wait for it to disappear:
+            let isHittablePredicate = NSPredicate { _, _ in
+                app.buttons[UID.Button.dismissKeyboard].isHittable == true
+            }
+
+            let expectation = XCTNSPredicateExpectation(predicate: isHittablePredicate, object: .none)
+            XCTWaiter().wait(for: [expectation], timeout: averageLoadTimeout)
+            // end of wait
```
mokagio: Nitpick: Unlike the comment above about the bubble blocking the navigation bar button, which is super useful, this comment doesn't add much information:

```suggestion
```

SimplenoteUITests/NoteEditor.swift
```diff
@@ -47,6 +58,15 @@ class NoteEditor {
...
 
     class func setFocus() {
+        // Waiting for the TextView to become hittable before using it
+        let isHittablePredicate = NSPredicate { _, _ in
+            app.textViews.firstMatch.isHittable == true
+        }
+
+        let expectation = XCTNSPredicateExpectation(predicate: isHittablePredicate, object: .none)
+        XCTWaiter().wait(for: [expectation], timeout: averageLoadTimeout)
+        // end of wait
```
mokagio: Same as above.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
```diff
@@ -276,7 +277,6 @@ class SimplenoteUISmokeTestsNoteEditor: XCTestCase {
         NoteEditorAssert.textViewWithExactLabelsShownOnce(labels: ["* Asterisk1", "* Asterisk2", "* Asterisk3"])
 
         trackStep()
-        NoteEditor.markdownEnable()
```
mokagio, line 279: > Edit: besides this, moving the call to enable Markdown in the note to an earlier stage (as done for all other tests) seems to finally solve the issue. 🧙

Maybe this has to do with us trying to swipe immediately after, but the screen doing a little bounce animation when Markdown gets enabled? _Animations are the crux of UI automation_.


https://user-images.githubusercontent.com/1218433/126108050-138d8791-e267-4667-a5fa-c4ae2fbaa7b3.mov

mokagio (approved):
:shipit:

SimplenoteUITests/NoteEditor.swift
pachlava: True. Thank you Gio!

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
pachlava: > Animations are the crux of UI automation.

Yes, I think this was the case here. Thank you for the review @mokagio, as always! 🙂

mokagio: Likewise. Thank you for caring and for keeping this in good shape!

PR #1359, opened 2021-07-14
H5/H6 remove force capitalized in markdown preview

### LINKED ISSUE #1304: Markdown H5 is automatically converted into upper case
Reported on 4010099-zen 

When using markdown, H5 defaults to uppercase no matter the case that is used:

### Expected
That H5 works as other headings, using upper or lowercase when required.

### Reproduced
- 1 Open a note or edit an existing one
- 2 Write `##### my h5 heading` in the editor
- 3 Switch to markdown preview
- 4 See `MY H5 HEADING`

Screenshot 2021-06-02 at 16 42 30

Apple | iPhone | iOS 14.2 | Simplenote 4.35

charliescheer:
### Fix
With this PR I have made two changes, first fixes #1304  where all H5 and H6 text in markdown preview were set to uppercase.  That will not happen now.

The other change in this PR changes the organization of the CSS and how it is loaded.  Previously all the CSS was in two separate files, default and dark.  Depending on if the device was in dark mode or not it would load one file or another.  But that also meant that we needed to maintain two files to update the CSS correctly.

What I have done is removed the color CSS from the default file and made two files for the colors.  So now the main CSS is loaded and then depending on the dark mode setting the correct color css is loaded and appended to the main css.  This change isn't visible in the app,  just an convenience for maintaining the markdown css

### Test
1. create a note that has mixed case H5 and H6 text in it
2. activate markdown on that note
3. go to the preview and check to make sure the heading text case is unchanged
4. Check dark mode and light mode

### Review


jleandroperez (approved):
Thank you for doing this sir!!

:shipit:

PR #1361, opened 2021-07-15
Track XCUITestHelpers using version 0.1.0

mokagio:
This should have been part of 39e10c34, but I somehow missed it.

### Test
If CI builds, we're good.

### Review

### Release

### DIFF
--- a/Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -32,9 +32,9 @@
         "package": "XCUITestHelpers",
         "repositoryURL": "https://github.com/Automattic/XCUITestHelpers",
         "state": {
-          "branch": "trunk",
+          "branch": null,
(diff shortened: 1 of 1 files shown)


Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -32,9 +32,9 @@
         "package": "XCUITestHelpers",
         "repositoryURL": "https://github.com/Automattic/XCUITestHelpers",
         "state": {
-          "branch": "trunk",
+          "branch": null,
           "revision": "47d688b0b2a6356361a16836c4d23f12d6643a73",
-          "version": null
+          "version": "0.1.0"
```
mokagio, line 37: This JSON format is horrendous. `branch` is now `null` because there is a `version` set. You may notice that before it was the opposite. The `revision` is the same because 0.1.0 was tagged on top of the tip of the package `trunk` to which this project was pointing at the time.

charliescheer (approved):
LGTM
Xcode didn't automatically update the package file.  Thanks for taking a look at this!

PR #1365, opened 2021-07-25
Connect shared db to widgets

charliescheer:
### Fix
This PR is the next stage in adding widgets to Simplenote iOS.  In this PR I have added the things that we need to connect the coreData database (now in the shared directory) to the widgets and intents.

By adding this connection we will be able to query the database for the notes and display them in the widgets, or use them to configure the widgets.

Most of this PR is laying the groundwork for the next steps, so there isn't a lot to see, but I have added some placeholder connections that allow seeing that it is working by trying to configure the note widget.  Note that I have NOT set it up to actually display a note in the widget, only in the configuration.

### Test
1. log into a simplenote account on a device or sim
2. Go to the home screen and long press to edit the home screen layout
3. tap the app button -> Simplenote -> Note Preview widget
4. tap on the newly added widget to get it's configuration settings.
5. you will see a Note option.  Tap on that. 

Notice that this displays the titles of the notes saved in the shared database
Again note that once selected the widget can not currently display the selected note.

### Review


SimplenoteWidgets/Models/Note+CoreDataProperties.swift
```diff
@@ -0,0 +1,25 @@
...
+    @NSManaged public var noteSynced: NSNumber?
+    @NSManaged public var owner: String?
+    @NSManaged public var pinned: NSNumber?
+    @NSManaged public var publishURL: String?
+    @NSManaged public var remoteId: String?
+    @NSManaged public var shareURL: String?
+    @NSManaged public var systemTags: String?
+    @NSManaged public var tags: String?
+
+}
```
jleandroperez: Since `Note+CoreDataClass.swift` + `Note+CoreDataProperties.swift` are meant to be a workaround, and aren't supposed to contain code in the class level, mind collapsing that to `Note+Widget.swift` maybe?

Background:
When using code gen, Core Data splits your NSManagedObject classes in two files. One is the concrete class definition, the other is an extension with the properties.

Goal is: if you need to throw in custom code into the Class, you just do so in `+CoreDataClass`. Whenever the model is updated, and you need to regen the files, your code is safe there (and shouldn't be overwritten).

Since we won't happen, let's collapse them both. **Plus:** please also add a note mentioning why this is required

SimplenoteWidgets/Models/SPManagedObject+CoreDataClass.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+import CoreData
+
+@objc(SPManagedObject)
+public class SPManagedObject: NSManagedObject {
+
+}
```
jleandroperez: **Quesiton:**

Since Intents only requires Note, can we check what happens if the other (unused) "workaround NSMO Classes" are removed?

- `SPManagedObject`
- `Tag`
- `Settings`
- `Preferences`

SimplenoteWidgets/Tools/BuildConfiguration.swift
```diff
@@ -0,0 +1,35 @@
...
+    static func `is`(_ configuration: BuildConfiguration) -> Bool {
+        return configuration == current
+    }
+}
+
+extension BuildConfiguration: CustomStringConvertible {
+    var description: String {
+        return self.rawValue
+    }
+}
```
jleandroperez: This file is duplicated (Original is in `Simplenote/BuildConfiguration`). Can we drop this copy and update the Target Membership of the former file?

jleandroperez (commented):
Sending you few notes sir!

SimplenoteWidgets/Models/SPManagedObject+CoreDataClass.swift
charliescheer: Intents will also need Tag, but can see if the others can be removed

SimplenoteWidgets/Tools/BuildConfiguration.swift
charliescheer: I had done that originally, but Simplenote/BuildConfiguration imports Tracks, which we had previously decided we did not want to include in the extensions.  I can have it import Tracks and not use it if that would be preferred.  I was trying to avoid bundling the extra package if it wasn't needed

SimplenoteWidgets/Models/SPManagedObject+CoreDataClass.swift
charliescheer: Seems to make Tag and Note compile we also need SPManagedObject, but the others can be removed

SimplenoteWidgets/Tools/BuildConfiguration.swift
jleandroperez: Tracks appears not to be used anywhere in `BuildConfiguration`.

No idea exactly why there's an `import AutomatticTracks` in there, but you can just remove that import, and update the target membership all over 😄

SimplenoteIntents/IntentHandler.swift
```diff
@@ -8,15 +9,28 @@ class IntentHandler: INExtension {
 
 extension IntentHandler: NoteWidgetIntentHandling {
     func provideNoteOptionsCollection(for intent: NoteWidgetIntent, with completion: @escaping (INObjectCollection<WidgetNote>?, Error?) -> Void) {
-        // This is placeholder code to confirm the dynamic intent selection is working
-        // TODO: add logic to fetch the available notes for account
+        // NOTE: This core data implementation is just to prove that the shared database is accessible and is not intended to be used in production.
```
jleandroperez: (As discussed on Slack!) let's drop these changes from this PR, since it's PoC code.

Thank you sir!

SimplenoteWidgets/Models/Note+Widget.swift
```diff
@@ -0,0 +1,33 @@
...
+    @NSManaged public var modificationDate: Date?
+    @NSManaged public var noteSynced: NSNumber?
+    @NSManaged public var owner: String?
+    @NSManaged public var pinned: NSNumber?
+    @NSManaged public var publishURL: String?
+    @NSManaged public var remoteId: String?
+    @NSManaged public var shareURL: String?
+    @NSManaged public var systemTags: String?
+    @NSManaged public var tags: String?
+
```
jleandroperez: Nipticky: Extra newline

SimplenoteWidgets/Models/SPManagedObject+Widget.swift
```diff
@@ -0,0 +1,22 @@
...
+
+extension SPManagedObject {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<SPManagedObject> {
+        return NSFetchRequest<SPManagedObject>(entityName: "SPManagedObject")
+    }
+
+    @NSManaged public var ghostData: String?
+    @NSManaged public var simperiumKey: String?
+
```
jleandroperez: Nipticky: Extra newline

SimplenoteWidgets/Models/Tag+Widget.swift
```diff
@@ -0,0 +1,24 @@
...
+extension Tag {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<Tag> {
+        return NSFetchRequest<Tag>(entityName: "Tag")
+    }
+
+    @NSManaged public var index: NSNumber?
+    @NSManaged public var name: String?
+    @NSManaged public var share: String?
+
```
jleandroperez: Nipticky: Extra newline

SimplenoteWidgets/Models/Tag+Widget.swift
```diff
@@ -0,0 +1,24 @@
...
+import Foundation
+import CoreData
+
+@objc(Tag)
+public class Tag: SPManagedObject {
+
+}
+
+
```
jleandroperez: Nipticky: Extra newline

jleandroperez (approved):
Feel free to :shipit: once the `CoreDataManager` snippet is gone sir!

Sidenote. I'm thinking, if the Model+Widget entities (let's call them "Readonly NSManagedObjects" maybe) are to be consumed by Intents + Widget, maybe they should be located inside a top level folder called `Unsyncable Models` or `Readonly Entities` or else.

Or maybe not. Just thinking out loud, I'm not 100% sold by the idea!

Nice work Charlie!!

PR #1375, opened 2021-08-02
Note Widget

charliescheer:
### Fix
This PR is the next step in the widgets project.  Here I have added the ability to connect to the shared database with intents and widgets, parse those details, and then display them on the note widget.

### Test
1. Open simplenote while logged out
2. try to setup the note widget and confirm that you get an error
3. log into simplenote
4. return to the widget setup and tap on the Note option, you should see a list of all of your not deleted notes
5. choose a note with a title and some content in it
6. tap outside of the setup to return to the homescreen
7. after a second your note content should appear

### Review


SimplenoteWidgets/Tools/WidgetDataController.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// Initialization
+    ///
+    init(coreDataManager: CoreDataManager) throws {
+        guard let isLoggedIn = UserDefaults(suiteName: SimplenoteConstants.sharedGroupDomain)?.bool(forKey: .accountIsLoggedIn),
+              isLoggedIn else {
+            throw StorageError.appConfigurationError
+        }
+
+        self.coreDataManager = coreDataManager
+        coreDataManager.managedObjectContext.persistentStoreCoordinator = coreDataManager.persistentStoreCoordinator
```
jleandroperez: This shouldn't be done in here. This specific initialization probably deserves a new CoreDataManager initializer?

SimplenoteWidgets/Providers/NoteWidgetProvider.swift
```diff
@@ -0,0 +1,52 @@
...
+    func getTimeline(for configuration: NoteWidgetIntent, in context: Context, completion: @escaping (Timeline<NoteWidgetEntry>) -> Void) {
+        // Confirm valid configuration
+        guard let widgetNote = configuration.note,
+              let simperiumKey = widgetNote.identifier else {
+            NSLog("Couldn't find configuration or identifier")
+            return
+        }
+
+        // Prepare data controller
+        var dataController: WidgetDataController
```
jleandroperez:
```suggestion
        let dataController: WidgetDataController
```

Simplenote/Options.swift
```diff
@@ -68,6 +68,7 @@ extension Options {
         }
         set {
             defaults.set(newValue.rawValue, forKey: .listSortMode)
+            UserDefaults(suiteName: SimplenoteConstants.sharedGroupDomain)?.set(newValue.rawValue, forKey: .listSortMode)
```
jleandroperez: so if the Defaults aren't initialized with `suiteName`, the value isn't accessible?

Problem with this approach is: unless you set the `listSortMode` again, then, it won't be accessible by the extension.

charliescheer: Good point.  I added a method to sync the setting in `applicationDidEnterBackground` as well.  That should solve the issue of the setting not syncing to the widgets/intents until it is updated

charliescheer:
@jleandroperez I made the changes that we discussed over slack.  This is ready for another look

SylvesterWilmott:
Looking good @charliescheer 

When first adding the widget, are we able to show the last modified note already in the widget? I think this experience would be better than seeing only the placeholder when first adding the widget.

I also noticed we have some space below the title. Can we remove this space? Often users ad an empty line under the note title which will result in a double empty line in the widget.

charliescheer:
@SylvesterWilmott yes, we can remove the padding below the title, that is easily done.

Choosing a default note... I will have to look into that.  The setting up of a widget and the widget presenting details are completely separate processes.... I don't know if there is a way to give it a default.  But I will look and see if possible

charliescheer:
@SylvesterWilmott Turns out yes it is pretty easy to setup a default value for widgets.  It is currently setup to display the note from the top of the user's list (will use the user's sort settings to figure that out)

Also removed the padding.  This is ready to be checked again

SylvesterWilmott:
Thanks @charliescheer 

I think there is an issue in displaying notes with checklists that wasn't present prior to the latest commits. Notes with checklists seem to have a large space to the left of the text and checklist icons appear in plain text. Do you think we can include checklist icons here as well as eliminating the space on the left?

Screenshot 2021-08-06 at 10 47 31

Also, when you visit the UI where you choose from our list of widgets, it would be great to have the content from the first note in the users list populating those previews - similar to Apple Notes.

Screenshot 2021-08-06 at 10 53 41

charliescheer:
@SylvesterWilmott Thanks for that feedback!

>Do you think we can include checklist icons here as well as eliminating the space on the left?

Odd that it wasn't doing this before but is now.  I have added some rules to the UI which makes the content fill the space, so that large space will not appear anymore.

As for adding the checklists, that is a bit more complicated.  Currently we generate the checklists by inserting them into AttributedStrings, but widgets can't read attributed strings by default.  We should be able to create a new SwiftUI view that will be able to process the checkboxes as they are, but that will take a bit of time to generate.  I wonder if we should make this a GH issue and handle it after the widgets are released.  WDYT?

> when you visit the UI where you choose from our list of widgets, it would be great to have the content from the first note in the users list populating those previews - similar to Apple Notes.

Totally.  That has been done, you will now see the preview of the note which is at the top of the users note list:

SylvesterWilmott:
Looks great @charliescheer 

>We should be able to create a new SwiftUI view that will be able to process the checkboxes as they are, but that will take a bit of time to generate. I wonder if we should make this a GH issue and handle it after the widgets are released.

Understood. Let's create an issue and tackle that after we release widgets. I am happy with the spot we are in with the note widget so it's a :shipit: from me.

SylvesterWilmott (approved):

charliescheer:
@SylvesterWilmott  awesome!
Here is that issue https://github.com/Automattic/simplenote-ios/issues/1388
I think I know how we can do this, it is just might take some time to get it working just right

Simplenote/Classes/SimplenoteConstants.swift
```diff
@@ -26,6 +26,8 @@ class SimplenoteConstants: NSObject {
     ///
     static let simplenotePublishedBaseURL = "http://simp.ly/publish/"
 
+    static let sharedGroupDomain = "group.\(Bundle.main.rootBundleIdentifier ?? "com.codality.NotationalFlow")"
```
jleandroperez, line 35: Note that this duplicates `defaultBundleIdentifier` in `FileManager+Simplenote` (Constants struct).

Since `SimplenoteConstants` is the proper spot to share these things, mind updating FileManager to use this? 

**PLUS:** Nipticky, for consistency with the rest, mind adding a comment?

Thank yooouuuu!!!

Simplenote/Color+Simplenote.swift
```diff
@@ -0,0 +1,17 @@
+import SwiftUI
+
+@available(iOS 13.0, *)
+extension Color {
+    /// Convenience initializers to get and use Simplenote color studio colors
+    ///
+
```
jleandroperez: Nipticky: Extra newline

Simplenote/Controllers/WidgetController.swift
```diff
@@ -0,0 +1,37 @@
...
+
+struct WidgetController {
+    @available(iOS 14.0, *)
+    static func resetWidgetTimelines() {
+        WidgetCenter.shared.getCurrentConfigurations { result in
+            guard case .success(let widgets) = result else {
+                return
+            }
+
+            if widgets.contains(where: { (widget) -> Bool in
```
jleandroperez, line 11: Nipticky: No need to specify the closure return type (it's implicit in the API signature). Plus no need to wrap `(widget)` in parentheses

Simplenote/Controllers/WidgetController.swift
```diff
@@ -0,0 +1,37 @@
...
+                if widget.configuration as? NoteWidgetIntent != nil {
+                    return true
+                }
+
+                // Add check for ListWidgetIntent when created later
+
+                return false
```
jleandroperez: Nipticky: 

Since `widget.configuration as? NoteWidgetIntent != nil` already yields a boolean, you can collapse the entire closure as follows:

```
if widgets.contains(where: { widget in
    widget.configuration as? NoteWidgetIntent != nil
})
```

Simplenote/Controllers/WidgetController.swift
```diff
@@ -0,0 +1,37 @@
...
+            }) {
+                NSLog("Reloading all widgets")
```
jleandroperez: Let's 🔥 this one!

Simplenote/Controllers/WidgetController.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    static func syncWidgetDefaults(isLoggedIn loggedIn: Bool) {
+        guard let widgetDefaults = UserDefaults(suiteName: SimplenoteConstants.sharedGroupDomain) else {
+            return
+        }
+
+        // Set listsortmode default
+        widgetDefaults.set(UserDefaults.standard.integer(forKey: .listSortMode), forKey: .listSortMode)
+
+        // User is logged in
```
jleandroperez: Nipticky: No need to add comments for these (since it's slightly redundant!)

Simplenote/Controllers/WidgetController.swift
```diff
@@ -0,0 +1,37 @@
...
+
+                return false
+            }) {
+                NSLog("Reloading all widgets")
+                WidgetCenter.shared.reloadAllTimelines()
+            }
+        }
+    }
+
+    static func syncWidgetDefaults(isLoggedIn loggedIn: Bool) {
```
jleandroperez:
```suggestion
    static func syncWidgetDefaults(authenticated: Bool) {
```

Really, really, like, really nipticky!!

Simplenote/Date+Simplenote.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+extension Date {
+    func increased(byHours hours: Int) -> Date? {
+        var components = DateComponents()
+        components.hour = hours
+        return Calendar.current.date(byAdding: components, to: self)
+
```
jleandroperez, line 9: Nipticky: Extra newline!

Simplenote/CoreDataManager.swift
```diff
@@ -9,27 +15,44 @@ class CoreDataManager: NSObject {
...
-        guard let mom = NSManagedObjectModel(contentsOf: storageSettings.modelURL) else {
-            fatalError("Cannot load model")
-        }
-        managedObjectModel = mom
+        setupStackIfNeeded(mainContext: managedObjectContext, psc: persistentStoreCoordinator, for: usageType)
+    }
 
-        managedObjectContext = NSManagedObjectContext(concurrencyType: .mainQueueConcurrencyType)
-        managedObjectContext.undoManager = nil
+    private func objectContext() -> NSManagedObjectContext {
```
jleandroperez: We should prob. rename `objectContext()`, since its purpose is not clear: it's building and returning a new instance. `objectContext` would be proper for a property name.

How abooouttttt:

- `buildMainContext`
- `newMainContext`
- `createMainContext`

?

Simplenote/CoreDataManager.swift
```diff
@@ -9,27 +15,44 @@ class CoreDataManager: NSObject {
...
+        let moc = NSManagedObjectContext(concurrencyType: .mainQueueConcurrencyType)
+        moc.undoManager = nil
+        return moc
+    }
 
+    private func storeCoordinator(with model: NSManagedObjectModel, storageURL: URL) throws -> NSPersistentStoreCoordinator {
```
jleandroperez: Same note as `objectContext()` above. How about:

- buildStoreCoordinator
- newStoreCoordinator
- createStoreCoordinator?

TY sir!!

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -436,3 +437,18 @@ extension SPAppDelegate {
...
+@available(iOS 14.0, *)
+extension SPAppDelegate {
+    @objc
+    func resetWidgetTimelines() {
+        WidgetController.resetWidgetTimelines()
+    }
+
+    @objc
+    func syncWidgetDefaults() {
+        WidgetController.syncWidgetDefaults(isLoggedIn: simperium.user != nil)
```
jleandroperez: Actually there could be a `user`, and it could have no token.

```
let authenticated = simperium.user.authenticated
```

Simplenote/SPAppDelegate.m
```diff
@@ -319,6 +322,8 @@ - (void)logoutAndReset:(id)sender
                 [self.simperium authenticateIfNecessary];
                 self.bSigningUserOut = NO;
 			}];
+
+            [[[NSUserDefaults alloc] initWithSuiteName:[SimplenoteConstants sharedGroupDomain]] setValue:false forKey:@"accountIsLoggedIn"];
```
jleandroperez: This should be `syncWidgetDefaults` instead right?

jleandroperez: Actually:`syncWidgetDefaults` should be invoked in three spots?

- applicationDidEnterBackground
- simperiumDidLogin **(it's in SPAppDelegate+Extensions!)**
- simperiumDidLogout **(same as above)**

Simplenote/StorageError.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+
+enum StorageError: Error {
+    case appConfigurationError
+    case attachingPersistentStoreFailure
+    case fetchError
+}
+
+extension StorageError {
```
jleandroperez: Nitpciky: Most of these errors aren't really Storage Errors:

- `appConfigurationError` is thrown when the app is logged off
- `fetchError` is thrown by WidgetDataController if fetch fails

I think we could prob. get rid of this enum. Where is it actually used? (still reading!!)

SimplenoteIntents/IntentHandler.swift
```diff
@@ -2,18 +2,53 @@ import Intents
...
-        var notes: [WidgetNote] = []
 
-        // TODO: add intent core data implementation.
+        // Fetch notes
+        var notes: [Note] = []
+        do {
+            notes = try dataController.notes()
+        } catch {
+            completion(nil, error)
+        }
```
jleandroperez, line 29: Question: where is the `completion` closure used? is the Error actually passed over to the user?

- If the error is presented somehow, we should prob. have our own (rather than relaying the Core Data one).
- If not, we could just drop it, and pass an empty array?

```
guard let notes = try? dataController.notes() else {
     completion([])
     return
}
```

SimplenoteIntents/IntentHandler.swift
```diff
@@ -2,18 +2,53 @@ import Intents
...
     func provideNoteOptionsCollection(for intent: NoteWidgetIntent, with completion: @escaping (INObjectCollection<WidgetNote>?, Error?) -> Void) {
-        var notes: [WidgetNote] = []
 
-        // TODO: add intent core data implementation.
+        // Fetch notes
+        var notes: [Note] = []
+        do {
+            notes = try dataController.notes()
+        } catch {
+            completion(nil, error)
```
jleandroperez: Note: regardless of whether we keep this or not, this completion runs, and then execution proceeds. This would need a `return` statement

SimplenoteWidgets/Providers/NoteWidgetProvider.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        // Fetch note
+        guard let note = dataController.note(forSimperiumKey: simperiumKey) else {
+            return
+        }
+
+        // Prepare timeline entry for every hour for the next 6 hours
+        // Create a new set of entries at the end of the 6 entries
+        var entries: [NoteWidgetEntry] = []
+        for int in 0..<6 {
```
jleandroperez: Let's move the range to a constant below

SimplenoteWidgets/Tools/DemoContent.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+struct DemoContent {
+    static let singleNoteTitle = "Twelve Tone Serialism"
+    static let singleNoteContent = "The twelve-tone technique is often closely related with the compositional style, serialism. Fundamentally, twelve-tone serialism is a compositional technique where all 12 notes of the chromatic scale are played with equal frequency throughout the piece without any emphasis on any one note. For this reason, twelve-tone serialism avoids being in any key. Arnold Schoenberg was arguably the most influential composers who embraced the twelve-tone technique. Schoenberg described the system as a “Method of composing with twelve tones which are related only with one another.”"
+    static let singleNoteContentAlt = "- [] Check item one \n- [] check item two\n- [] check item three"
```
jleandroperez, line 6: Localized?

jleandroperez (commented):
Few incoming nipticky notes! Thank you sir!!

SimplenoteIntents/IntentHandler.swift
charliescheer: What appears to the user is a generic error at widget configuration.  I don't think the user is presented any of our details, but it tells the configurator there is an error and then logs the error output.  Returning an empty array would allow the configuration to load but without their notes.

🤔 let me break the widgets so I can show this

charliescheer: ah, it does display:

charliescheer: Will add some sort of error case to be handled for this

SimplenoteWidgets/Tools/DemoContent.swift
charliescheer: I'm not sure if this needs to be localized.  _**Most**_ of what this does is power the SwiftUI previews. 

I was thinking that I would remove all the UI previews, but we would still need this demo content because it servers a few other purposes.  It gives the widget placeholder some sort of shape if the placeholder appears, and if for some reason there is a failure fetching the user's most recent note then this text will appear.  My feeling is that we don't really need this to be localized, WDYT?

Simplenote/StorageError.swift
charliescheer: Currently this gets called when the core data manager fails to load the persistent store, if the widget data controller fails to fetch, and if the app is not logged in.

So here is my proposal for these error comments.  First if I remove the persistentStore error case from this all of these errors are specific to the widgets.  So I could change the name of the error to WidgetError and still use the configuration and fetch errors in the places they currently are.
Then in the core data manager I can throw `throw NSError(domain: Constants.errorDomain, code: .zero, userInfo: nil)` instead

I have updated the codes for that.  WDYT?

SimplenoteWidgets/Tools/DemoContent.swift
jleandroperez: TBH if there's a chance for this to be displayed, it should be localized! (Rule of thumb is: any string that may go to screen, NSLocalized!)

Localizations are extracted by the `Scripts/localize.py`. You'll prob. neeed to patch [this spot, exactly](https://github.com/Automattic/simplenote-ios/blob/develop/Scripts/localize.py#L124) as follows:

```
    genstrings_cmd = 'genstrings -q -o "%s" `find ./Simplenote ./SimplenoteShare ./SimplenoteWidgets -name "*.m" -o -name "*.swift"`'
```

(Otherwise SimplenoteWidgets won't be picked up). Same for SimplenoteShare!!

Simplenote/StorageError.swift
jleandroperez: Swift Errors are absolutely great. Specifically, the one thing that caught my eye is  the name. It's called `StorageError`, but it may be thrown by WidgetDataController.

StorageError would maybe imply it came from the Storage* (StorageManager perhaps?)

Simplenote/SPAppDelegate.m
jleandroperez: I think this line should be removed?

SimplenoteWidgets/Models/Note+Widget.swift
```diff
@@ -30,3 +30,52 @@ extension Note {
...
     @NSManaged public var tags: String?
 }
+
+extension Note {
+    private var lines: [String]? {
+        guard let content = content else {
+            return nil
+        }
+
+        return content.components(separatedBy: .newlines)
```
jleandroperez: You may collapse everything into:

```
   private var lines: [String]? {
         content?.components(separatedBy: .newlines)
   }
```

SimplenoteWidgets/Models/Note+Widget.swift
```diff
@@ -30,3 +30,52 @@ extension Note {
...
+    }
+
+    private var firstLine: String {
+        guard let lines = lines else {
+            return Constants.defaultTitle
+        }
+        return lines.first ?? String()
```
jleandroperez: You can simplify as follows:

```
lines?.first ?? Constants.defaultTitle
```

SimplenoteWidgets/Providers/NoteWidgetProvider.swift
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    func placeholder(in context: Context) -> NoteWidgetEntry {
+        return NoteWidgetEntry.placeholder
+    }
+
+    func getSnapshot(for configuration: NoteWidgetIntent, in context: Context, completion: @escaping (NoteWidgetEntry) -> Void) {
+        guard let dataController = dataController,
+              let note = try? dataController.firstNote() else {
```
jleandroperez: You may collapse the two first lines as follows (Unless I'm missing something!!)

```
guard let note = try? dataController?.firstNote() else {
```

SimplenoteWidgets/Providers/NoteWidgetProvider.swift
```diff
@@ -0,0 +1,92 @@
...
+
+        // Prepare timeline entry for every hour for the next 6 hours
+        // Create a new set of entries at the end of the 6 entries
+        var entries: [NoteWidgetEntry] = []
+        for int in Constants.entryRange {
+            if let date = Date().increased(byHours: int) {
+                entries.append(NoteWidgetEntry(date: date, note: note))
+            }
+        }
+        let timeline = Timeline(entries: entries, policy: .atEnd)
```
jleandroperez, line 72: Swifty Alternative!!

```
        let entries: [NoteWidgetEntry] = Constants.entryRange.compactMap { index in
            guard let date = Date().increased(byHours: index) else {
                return nil
            }

            return NoteWidgetEntry(date: date, note: note)
        }

        let timeline = Timeline(entries: entries, policy: .atEnd)
        completion(timeline)
```

SimplenoteWidgets/Tools/WidgetDataController.swift
```diff
@@ -0,0 +1,100 @@
...
+            self.coreDataManager = coreDataManager
+            return
+        }
+
+        guard let isLoggedIn = UserDefaults(suiteName: SimplenoteConstants.sharedGroupDomain)?.bool(forKey: .accountIsLoggedIn),
+              isLoggedIn else {
+            throw WidgetError.appConfigurationError
+        }
+
+        self.coreDataManager = coreDataManager
```
jleandroperez: There's a bit of duplication: `self.coreDataManager = coreDataManager` is implemented twice
One alternative could be:


```
 init(coreDataManager: CoreDataManager, isPreview: Bool = false) throws {
    self.coreDataManager = coreDataManager

    if isPreview {
        return
    }

    let authenticated = UserDefaults(suiteName: SimplenoteConstants.sharedGroupDomain)?.bool(forKey: .accountIsLoggedIn) ?? false

    guard authenticated else {
        throw WidgetError.appConfigurationError
        return
    }

```

SimplenoteWidgets/Tools/WidgetDataController.swift
```diff
@@ -0,0 +1,100 @@
...
+
+    /// Sort mode for widgets.  Fetched from main apps sort setting
+    ///
+    private var noteSortMode: SortMode {
+        guard let defaults = UserDefaults(suiteName: SimplenoteConstants.sharedGroupDomain) else {
+            return SortMode.alphabeticallyAscending
+        }
+
+        return SortMode(rawValue: defaults.integer(forKey: .listSortMode))
+            ?? SortMode.alphabeticallyAscending
```
jleandroperez: I'm starting to think we could maybe use a class called... perhaps WidgetUserDefaults (similar to Options.swift).  Or not.

jleandroperez:
I may have found a glitch @charliescheer !! Steps:

1. Fresh install Simplenote
2. Add the Note Widget
3. Tap over the Widget
4. Simplenote will launch
5. Log into your account and wait for Sync to happen
6. Go back to the springboard
7. Long press over the widget > Edit Widget
8. Choose any of your notes

As a result, the Widget will remain in "No content" / "Placeholder mode". Pressing over the widget will not open the associated Note.


https://user-images.githubusercontent.com/1195260/130161744-afe1a8d0-f408-44d7-b9ab-257e60f53e6b.mov

jleandroperez:
"Glitch" 2:

1. Fresh install Simplenote
2. **DO NOT** log into your account
3. Long press over the widget > Edit Widget
4. Press over the `Choose` button

As a result, you'll get the error attached in the image below. Do you think we could customize that?

SimplenoteWidgets/Models/Note+Widget.swift
charliescheer: This is serving two roles.  It is making sure that there is content in the note and that there is a title on the first line.  If we just check that there is a first line it could be \n which would make the note appear empty in the intents.  How about removing `private var first` and changing `var title` to:

```    var title: String {
        guard let firstLine = lines?.first,
              firstLine.count > 0 else {
            return Constants.defaultTitle
        }
        return firstLine
    }
```

That serves both roles in fewer lines/properties

SimplenoteWidgets/Tools/WidgetDataController.swift
charliescheer: So.... The problem here is SwiftUI previews.  SwiftUI previews have no access to UserDefaults, so by adding in this check to make sure a user is logged in, it breaks the SwiftUI previews.  The first guard to see if it is a preview is meant to exit with a setup WidgetDataController without checking if the user is logged in.  I this solution above  we would still setup the WidgetDataController if the user is logged out, but would throw an error.  I'm not sure that is what we want here, but this solution works

```
    init(coreDataManager: CoreDataManager, isPreview: Bool = false) throws {
        if !isPreview {
            guard let isLoggedIn = UserDefaults(suiteName: SimplenoteConstants.sharedGroupDomain)?.bool(forKey: .accountIsLoggedIn),
                  isLoggedIn else {
                throw WidgetError.appConfigurationError
            }
        }

        self.coreDataManager = coreDataManager
    }
```

The other solution is something I have been toying with, and that is removing SwiftUI previews from the widgets before it is merged.

SimplenoteWidgets/Tools/WidgetDataController.swift
charliescheer: I like it in it's own class.

Simplenote/Controllers/WidgetController.swift
```diff
@@ -0,0 +1,24 @@
...
+            if widgets.contains(where: { widget in
+                widget.configuration as? NoteWidgetIntent != nil
+            }) {
+                WidgetCenter.shared.reloadAllTimelines()
+            }
+        }
+    }
+
+    static func syncWidgetDefaults(authenticated: Bool) {
+        let widgetDefaults = WidgetDefaults.shared
```
jleandroperez, line 27: Beautiful!!

SimplenoteIntents/IntentHandler.swift
```diff
@@ -2,18 +2,62 @@ import Intents
...
+            self.coreDataManager = try CoreDataManager(StorageSettings().sharedStorageURL, for: .intents)
+        } catch {
+            fatalError()
+        }
+        super.init()
+    }
+
+
     override func handler(for intent: INIntent) -> Any {
+        NSLog("failed to return handler")
```
jleandroperez: We should prob drop this one (or format this a bit more, like `[IntentHandler] Failed to return handler` ?

P.s.: Does this imply the handler failed? since it's returning `self`

SimplenoteWidgets/Models/Note+Widget.swift
```diff
@@ -30,3 +30,42 @@ extension Note {
...
+        guard var lines = lines else {
+            // Note. Swift UI Text will crash if given String() so need to use this version of an empty string
+            return ""
+        }
+        lines.removeFirst()
+        return lines.joined(separator: .newline)
+    }
+
+    var limitedTitle: String {
+        String(title.prefix(50))
```
jleandroperez: Nipticky: Constant!

SimplenoteWidgets/Tools/WidgetDataController.swift
```diff
@@ -0,0 +1,85 @@
...
+        }
+        try performFetch()
+
+        return notesController.fetchedObjects
+    }
+
+    /// Returns note given a simperium key
+    ///
+    func note(forSimperiumKey key: String) -> Note? {
+        notesController.predicate = predicateForNotes()
```
jleandroperez: Let's chat over Slack, I believe we don't really need the whole ResultsController / NSFetchedResultsController machinery

SimplenoteWidgets/Tools/WidgetDataController.swift
jleandroperez: Gotcha!! thanks for changing that!!

SimplenoteIntents/IntentHandler.swift
```diff
@@ -2,18 +2,62 @@ import Intents
...
         return self
     }
 }
 
 extension IntentHandler: NoteWidgetIntentHandling {
     func provideNoteOptionsCollection(for intent: NoteWidgetIntent, with completion: @escaping (INObjectCollection<WidgetNote>?, Error?) -> Void) {
-        var notes: [WidgetNote] = []
+        NSLog("attempting to provied note options collection")
```
jleandroperez: Let's drop this log!

SimplenoteWidgets/Tools/WidgetDataController.swift
jleandroperez: Quick sketch, but the same machinery, without FRC, would look as follows:

```
        let request = NSFetchRequest<Note>(entityName: Note.entityName)
        request.predicate = predicate
        request.sortDescriptors = sortDescriptors
        request.fetchLimit = limit

        // This should be an ivar?
        let mainContext = coreDataManager.managedObjectContext!
        mainContext.fetch(request)
```

jleandroperez (commented):
Confirming that [**Bug 1** is solved!](https://github.com/Automattic/simplenote-ios/pull/1375#issuecomment-902348107)

SimplenoteWidgets/Tools/DemoContent.swift
charliescheer: I wonder if we should use a different text for this demo content then.  I had been using the text out of the zeplin designs just as something to put in there, but it might make sense to actually consider what this will say.

@SylvesterWilmott can you weigh in on what you would want as far as copy for placeholder text for the widgets?  There are 3 texts I can think of that may appear to users.
Title for the note widget
body for the note widget
8 note titles for the list widget

Currently all of these are using the zeplin text, which may not translate (Like I don't know how well Twelve Tone Serialism will read to users translated into another language?)  But we can continue to use these strings if you prefer

SimplenoteIntents/IntentHandler.swift
charliescheer: I think this got left in there cause I was doing some testing on the previous glitches and wasn't sure where the issue was yet.  Will drop it

SimplenoteWidgets/Views/NoteWidgetView.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    var body: some View {
+        GeometryReader { geometry in
+            ZStack {
+                VStack(alignment: .leading) {
+                    Text(entry.title)
+                        .widgetHeader(widgetFamily,
+                                      color: Color(for: colorScheme, light: .gray100, dark: .white))
+                    Text(entry.content)
+                        .subheadline(color: Color(for: colorScheme, light: .gray100, dark: .white))
```
jleandroperez: Suggestion: For reusability purposes, you would prob. want something like:

`Color+Widget.swift`

```
extension Color {

    static var bodyTextColor: Color {
        Color(for: colorScheme, light: .gray100, dark: .white)
    }
}
```

SimplenoteWidgets/Views/NewNoteWidgetView.swift
```diff
@@ -21,7 +21,7 @@ struct NewNoteWidgetView: View {
             }
             .padding(Constants.overallPadding)
         }
-        .background(Color(UIColor(studioColor: .spBlue50)))
+        .background(Color(studioColor: .spBlue50))
```
jleandroperez: (Same note as below). You would typically want something like:

```
extension Color {

   var newNoteBackgroundColor: Color {
      Color(studioColor: .spBlue50)
   }
}
```

SimplenoteWidgets/Tools/WidgetDataController.swift
```diff
@@ -0,0 +1,104 @@
...
+        fetchRequest.fetchLimit = limit
+        fetchRequest.sortDescriptors = [sortDescriptorForNotes()]
+        fetchRequest.predicate = predicateForNotes(filteredBy: filter)
+
+        return fetchRequest
+    }
+
+    // MARK: Fetching
+
+    private func performFetch<T: NSManagedObject>(from request: NSFetchRequest<T>) -> [T]? {
```
jleandroperez: ❤️

SimplenoteWidgets/Tools/WidgetDataController.swift
```diff
@@ -0,0 +1,104 @@
+import Foundation
+import SimplenoteFoundation
+import SimplenoteSearch
+import CoreData
+
+class WidgetDataController {
```
jleandroperez: Nipticky: Since this one manages access to `Note`, perhaps a more accurate name would be `WidgetNoteController`?  (or plural?)

SimplenoteWidgets/Providers/NoteWidgetProvider.swift
```diff
@@ -0,0 +1,88 @@
...
+        self.title = title
+        self.content = content
+        self.url = url
+    }
+
+    let date: Date
+    let title: String
+    let content: String
+    let url: URL
+}
```
jleandroperez, line 22: Tip: if you add the convenience initializer in an extension, you get to keep the "default" initializer(and save some lines):

```
struct NoteWidgetEntry: TimelineEntry {
    let date: Date
    let title: String
    let content: String
    let url: URL
}

extension NoteWidgetEntry {
    static let placeholder = NoteWidgetEntry(date: Date(),
                                             title: DemoContent.singleNoteTitle,
                                             content: DemoContent.singleNoteContent,
                                             url: DemoContent.demoURL)

    init(date: Date, note: Note) {
        self.init(date: date,
                  title: note.title,
                  content: note.body,
                  url: note.url)
    }
}
```

SimplenoteIntents/SimplenoteIntents.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.com.codality.NotationalFlow.Alpha</string>
```
jleandroperez: I think this should be `group.$(CFBundleIdentifier)` instead.

SimplenoteIntents/SimplenoteIntentsDebug.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.com.codality.NotationalFlow.Development</string>
```
jleandroperez: Same as above, I believe this should be `group.$(CFBundleIdentifier)`

SimplenoteIntents/SimplenoteIntentsDistribution Internal.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.com.codality.NotationalFlow.Internal</string>
```
jleandroperez:
```
group.$(CFBundleIdentifier)
```

SimplenoteIntents/SimplenoteIntentsRelease.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.com.codality.NotationalFlow</string>
```
jleandroperez:
```
group.$(CFBundleIdentifier)
```

SimplenoteIntents/SimplenoteIntents.entitlements
jleandroperez: cc @oguzkocer OR @mokagio OR @jkmassel  just in case!!

jleandroperez (approved):
✅ Code looks sharp
✅ Verified "Adding Widget after Authenticated" flow
✅ Verified "Adding Widget before Authenticated" flow
✅ Verified that the wiget's content gets refreshed
✅ Code is sharp

Few notes added. Only important ones are:

- Please verify with our Platform9 friends that the Entitlements are correct
- Please make sure the `localize.py` script is patched, to account for `SimplenoteWidget`. Otherwise Localizable Strings will **not be picked**

:shipit: when ready!!

SimplenoteWidgets/Providers/NoteWidgetProvider.swift
charliescheer: Neat!

SimplenoteIntents/SimplenoteIntents.entitlements
mokagio: Nice suggestion @jleandroperez.

The result at compile-time or (run-time 🤔 ?) would be the same as this syntax, because `group.$(CFBundleIndentifier)` resolves to the bundle identifier value of the target and build configuration combination.

Having said that, it would be useful to use this syntax because it would allow us to condense `.entitlements` files that are otherwise identical. E.g.:

https://github.com/Automattic/simplenote-ios/blob/08d15d7240a7e8d23544708969451bf6f6ff4815/SimplenoteIntents/SimplenoteIntentsRelease.entitlements#L1-L10
and 

https://github.com/Automattic/simplenote-ios/blob/08d15d7240a7e8d23544708969451bf6f6ff4815/SimplenoteIntents/SimplenoteIntentsDistribution%20Internal.entitlements#L1-L10

I think my suggestion is more appropriate for a dedicated PR, though.

mokagio: I opened a PR against this branch to address an entitlements issue I noticed while researching a good answer for the question above: https://github.com/Automattic/simplenote-ios/pull/1408

In the PR, I point out an issue with the Xcode UI, which I _think_ might be due to using the `CFBundleGroup` interpolation. Pending confirmation, that might be an argument _against_ going down that route and use explicit identifiers instead.

It would be a shame, though, as it's nice to have a single source of truth.

charliescheer:
I made some merging mistakes while preparing this to go back into the feature branch.  I am closing this PR in favor of #1413

PR #1378, opened 2021-08-02
Add bulk trash test

jostnes:
### Fix
This PR doesn't fix anything but rather adds an additional UI test. There is also an additional function that is used to dismiss alerts seen on certain accounts.

### Test
Test flow: Login -> Add multiple notes -> Select one note by long press -> Select all notes -> Tap trash icon on note list screen

### Review
@pachlava 

### Release

### DIFF
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -47,18 +47,17 @@ func toggleSwitchIfNeeded(_ switchElement: XCUIElement, _ value: String) {
     switchElement.tap()
 }
 
-func dismissVerifyEmailIfNeeded(using app: XCUIApplication) {
-    guard app.staticTexts["Review Your Account"].waitForExistence(timeout: 15) else { return }
-    app.buttons["icon cross"].tap()
+func dismissVerifyEmailIfNeeded() {
+    guard app.staticTexts[Text.alertReviewAccount].waitForExistence(timeout: 15) else { return }
+    app.buttons[UID.Button.crossIcon].tap()
 }
 
 func populateNoteList(title: String, numberToCreate: Int) {
     for i in 1...numberToCreate {
-        let noteTitle = title + String(i)
+        let noteTitle = title + "-" + String(i)
         let noteContent = "\n\nSome content here for note " + String(i)
 
         NoteList.addNoteTap()
-        NoteEditorAssert.editorShown()
         NoteEditor.clearAndEnterText(enteredValue: noteTitle + noteContent)
         NoteEditor.leaveEditor()
     }
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -47,6 +47,18 @@ func toggleSwitchIfNeeded(_ switchElement: XCUIElement, _ value: String) {
     switchElement.tap()
 }
 
+func populateNotesList(title: String, numberToCreate: Int) {
+    for i in 1...numberToCreate {
+        let noteTitle = title + String(i)
+        let noteContent = "\n\nSome content here for note " + String(i)
+
+        NoteList.addNoteTap()
+        NoteEditorAssert.editorShown()
+        NoteEditor.clearAndEnterText(enteredValue: noteTitle + noteContent)
+        NoteEditor.leaveEditor()
+    }
+}
+
 class Table {
 
     class func getAllCells() -> XCUIElementQuery {
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -47,7 +47,12 @@ func toggleSwitchIfNeeded(_ switchElement: XCUIElement, _ value: String) {
     switchElement.tap()
 }
 
-func populateNotesList(title: String, numberToCreate: Int) {
+func dismissVerifyEmailIfNeeded(using app: XCUIApplication) {
+    guard app.staticTexts["Review Your Account"].waitForExistence(timeout: 15) else { return }
+    app.buttons["icon cross"].tap()
+}
+
+func populateNoteList(title: String, numberToCreate: Int) {
     for i in 1...numberToCreate {
         let noteTitle = title + String(i)
         let noteContent = "\n\nSome content here for note " + String(i)
--- a/SimplenoteUITests/NoteList.swift
+++ b/SimplenoteUITests/NoteList.swift
@@ -129,6 +129,28 @@ class NoteList {
         guard searchCancelButton.exists else { return }
         searchCancelButton.tap()
     }
+
+    class func longPressNote(title: String) {
+        app.tables.cells[title].press(forDuration: 3)
+    }
+
+    class func selectNote() {
+        let selectButton = app.buttons[UID.Button.select]
+        guard selectButton.exists else { return }
+        selectButton.tap()
+    }
+
+    class func selectAll() {
+        let selectAllButton = app.buttons[UID.Button.selectAll]
+        guard selectAllButton.exists else { return }
+        selectAllButton.tap()
+    }
+
+    class func tapTrashButton() {
+        let trashIcon = app.buttons[UID.Button.trashIcon]
+        guard trashIcon.exists else { return }
+        trashIcon.tap()
+    }
 }
 
 class NoteListAssert {
@@ -273,5 +295,9 @@ class NoteListAssert {
 
         print(">>> Actual search string is '\(actualSearchString)'")
         XCTAssertEqual(actualSearchString, searchString)
+    }
+
+    class func isDeselectAllButtonDisplayed() {
+        XCTAssertTrue(app.buttons[UID.Button.deselectAll].waitForExistence(timeout: maxLoadTimeout))
     }
 }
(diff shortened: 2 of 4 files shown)


pachlava:
Thank you for creating the PR @jostnes! The new test does what's intended and passed on first attempt - this is the main part 👍 

I'll be adding some review notes - just ideas for improvements and minor nitpicks.

SimplenoteUITests/Generic.swift
```diff
@@ -47,6 +47,23 @@ func toggleSwitchIfNeeded(_ switchElement: XCUIElement, _ value: String) {
     switchElement.tap()
 }
 
+func dismissVerifyEmailIfNeeded(using app: XCUIApplication) {
```
pachlava: Since `app` is defined as a global variable in all the files in `SimplenoteUITests` (the approach in `SimplenoteScreenshots` is different), it's not really necessary to have the argument here:

```suggestion
func dismissVerifyEmailIfNeeded() {
```

SimplenoteUITests/Generic.swift
```diff
@@ -47,6 +47,23 @@ func toggleSwitchIfNeeded(_ switchElement: XCUIElement, _ value: String) {
...
+    guard app.staticTexts["Review Your Account"].waitForExistence(timeout: 15) else { return }
+    app.buttons["icon cross"].tap()
```
pachlava: In Simplenote UI tests, we try to have all objects identifiers stored in a single place - `UIDs.swift`. I think it's a good idea to move `icon cross` there - it will be easier to reuse.

SimplenoteUITests/Generic.swift
```diff
@@ -47,6 +47,23 @@ func toggleSwitchIfNeeded(_ switchElement: XCUIElement, _ value: String) {
...
+}
+
+func populateNoteList(title: String, numberToCreate: Int) {
+    for i in 1...numberToCreate {
+        let noteTitle = title + String(i)
```
pachlava: What do you think of adding a space here:

```suggestion
        let noteTitle = title + " " + String(i)
```

So that you will not have to include the space in the end of `title` argument?

SimplenoteUITests/Generic.swift
```diff
@@ -47,6 +47,23 @@ func toggleSwitchIfNeeded(_ switchElement: XCUIElement, _ value: String) {
...
+        let noteContent = "\n\nSome content here for note " + String(i)
+
+        NoteList.addNoteTap()
+        NoteEditorAssert.editorShown()
```
pachlava: Philosophical topic: what do you think of having assertions in non-asserting methods?

Currently, in Simplenote iOS, we have assertions really exposed - they are either in the test body, or in the methods that contain `Assert` is their name / class. This way, we know that we have a set of non-asserting methods, that do stuff, but do not care how correctly it goes. And the other set of asserting methods, that care only about how correctly something went.

Also, in that particular case, this method is expected to execute a loop of actions, and repeat the assertion `i` times - which means spending 5 * `i` seconds to assert something, that is already asserted in other tests of `Note Editor`.

^ This is only my opinion, and I can also see value is having as many assertions as possible. The main concern here is time: how many tests can we run on CI under 5 minutes (this is just an approximately acceptable time our developers are okay with, if we talk about running a subset of tests on every commit to PR - which happens now).

SimplenoteUITests/NoteList.swift
```diff
@@ -129,6 +129,28 @@ class NoteList {
         guard searchCancelButton.exists else { return }
         searchCancelButton.tap()
     }
+
+    class func longPressNote(title: String) {
+        app.tables.cells[title].press(forDuration: 3)
+    }
+
+    class func selectNote() {
```
pachlava: Method naming question: we can select a note from the context menu (what this method does), and we can also select an individual note when we are already in multiselect mode. What do you think of renaming this method so that it's obvious that it's about tapping `Select` in the currently opened context menu?

SimplenoteUITests/NoteList.swift
```diff
@@ -274,4 +296,8 @@ class NoteListAssert {
         print(">>> Actual search string is '\(actualSearchString)'")
         XCTAssertEqual(actualSearchString, searchString)
     }
+
+    class func isDeselectAllButtonDisplayed() {
```
pachlava: The current way of calling the asserting methods:

```swift
NoteListAssert.noteContentIsShownInSearch()
````

makes the calls look as a human-readble text, making`is` part not really necessary:

```suggestion
    class func deselectAllButtonDisplayed() {
```

or

```suggestion
    class func deselectAllButtonIsDisplayed() {
```

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
```diff
@@ -282,4 +288,22 @@ class SimplenoteUISmokeTestsNoteEditor: XCTestCase {
...
+        trackStep()
+        NoteList.longPressNote(title: noteTitle + "1")
+        NoteList.selectNote()
+        NoteList.selectAll()
+        NoteListAssert.isDeselectAllButtonDisplayed()
+
+        trackStep()
+        NoteList.tapTrashButton()
+        NoteListAssert.notesNumber(expectedNotesNumber: 0)
+    }
```
pachlava, line 314: What do you think of adding one more step, that will check that these two notes are now present in `Trash`? (you can use existing `Trash` tests as a guidance)

SimplenoteUITests/UIDs.swift
```diff
@@ -53,6 +53,10 @@ enum UID {
         static let cancel = "Cancel"
         static let dismissKeyboard = "Dismiss keyboard"
         static let deleteTagConfirmation = "Delete Tag"
+        static let select = "Select"
+        static let selectAll = "Select All"
+        static let deselectAll = "Deselect All"
+        static let trashIcon = "Trash Notes"
```
pachlava: Since we already have `itemTrash` and `trashNote` and `deleteNote` 🤦  buttons, what do you think of renaming this new button in this way:

```suggestion
        static let trashNotes = "Trash Notes"
```

SimplenoteUITests/NoteList.swift
```diff
@@ -129,6 +129,28 @@ class NoteList {
...
+        selectButton.tap()
+    }
+
+    class func selectAll() {
+        let selectAllButton = app.buttons[UID.Button.selectAll]
+        guard selectAllButton.exists else { return }
+        selectAllButton.tap()
+    }
+
+    class func tapTrashButton() {
```
pachlava:
```suggestion
    class func tapTrashNotesButton() {
```

to better reflect that this is about trashing multiple notes, and not about trashing a single note by swiping it left ad then tapping the `Trash` button.

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
```diff
@@ -16,6 +16,12 @@ class SimplenoteUISmokeTestsNoteEditor: XCTestCase {
         let _ = attemptLogOut()
         EmailLogin.open()
         EmailLogin.logIn()
+
+        // Extra check for certain emails
+        if testDataEmail.contains("trial") {
+            dismissVerifyEmailIfNeeded(using: app)
+        }
```
pachlava: A smart way of not waiting for the alert when we are not in the trial environment 👍

pachlava (changes requested):
While my notes are mainly nitpicky, I think it makes sense to address the one about adding one more step to the test (checking that deleted notes are now in `Trash`) and the ones about naming - to avoid possible confusions once this code goes live. After these are addressed, I think we can have this test added 👍 

Thank you for you work @jostnes !

SimplenoteUITests/Generic.swift
jostnes: Yeah, missed this. I'll move this as well as the `Review Your Account` text to that file

SimplenoteUITests/Generic.swift
jostnes: Yup can do that, I'll change it from a space to a `-` cos the title also needs to be called later in the test file `NoteList.longPressNote(title: noteTitle + "1")`

Having a `" 1"` would also look weird and as a bonus to changing it to a `-` means automatically testing special characters on note title.

SimplenoteUITests/Generic.swift
jostnes: Yup makes sense, I've used it sometimes on non-assert methods to make test fail faster, but usually prefer to have it separated as well. In this case, I do agree that it's not really needed/add much value being there. Will remove it.

SimplenoteUITests/NoteList.swift
jostnes: Yeah, guess this is more of a habit for me, I'm used to starting methods that assert with `is...` ,`are...`,`has...` so that just by looking at the first few letters we know what it is going to do. In this case since it's structured in a way that all are under the `...Assert` it makes sense to not use it, will update this.

pachlava:
Thanks for addressing all the remarks from me @jostnes !

There are two comments related to the newly added step:
1. We have a dedicated `TrashAssert.notesNumber(...)` that you can use instead of `NoteListAssert.notesNumber(...)`. They do the same thing, over the same object, but the error text for the case of fail is different.
2. Sice we know the names of deleted notes, we can additionally check that not only we have two notes is `Trash` after their trashing, but also check their names, to add more precision to the test.

jostnes:
Ah ok, didn't know there's a separate assert for that, I've updated it and also added the additional title check.

pachlava (approved):

PR #1379, opened 2021-08-04
Finish intents extension entitlements setup

mokagio:
This finishes @charliescheer's job from #1374.

It's been ages since I worked with entitlements, so I based everything off how the widgets extension was configured. 

@charliescheer FYI the steps I took are documented in the commit history. You should be able to replay them in case we'll add a new extension. I also checked on the Developer Portal to ensure the App Ids matched the groups, which was the case. 👍 

The result match, so I'm fairly confident everything is wired up correctly.

<table>
<tr>
	<td>Widgets</td>
	<td>Intents</tb>
<tr>
	<td>

image

</td>
	<td>

image

</td>
</table>

I archived it locally and it worked, although I'm not sure whether that hits App Store Connect or the dev portal to check if everything is fine. 🤔 

I started an installable build in CI to verify if everything works on that end, with the same caveat that it might not be testing the assignment. _Currently waiting for it to finish._

The final test would be attempting an upload to App Store Connect. 🤔 We could do that, too, maybe with an hotfix build number so the ASC upload succeeds but we don't compromise the build number sequence for the current release?


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -4569,6 +4577,7 @@
 				CLANG_ENABLE_OBJC_WEAK = YES;
 				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
+				CODE_SIGN_ENTITLEMENTS = SimplenoteIntents/SimplenoteIntentsDebug.entitlements;
```
mokagio, line 4580: These changes were done via the Xcode UI:

image

SimplenoteIntents/SimplenoteIntents.entitlements
```diff
@@ -4,7 +4,7 @@
 <dict>
 	<key>com.apple.security.application-groups</key>
 	<array>
-		<string>group.com.codality.NotationalFlow.Alpha</string>
+		<string>group.com.codality.NotationalFlow</string>
```
mokagio, line 7: This and the other changes to the `.entitlements` were only to keep them consistent with the widget's ones.

These are just file names at the end of day, they could have stayed the way they were, but consistency is useful to lower the cognitive load of navigating the project.

mokagio:
@jkmassel @AliSoftware: I asked your review so you can double check my work in configuring the entitlements, since, as I mentioned, I haven't done this in a while.

Also, what do you think of this? Worth doing or is the setup straightforward enough?

> The final test would be attempting an upload to App Store Connect. 🤔 We could do that, too, maybe with an hotfix build number so the ASC upload succeeds but we don't compromise the build number sequence for the current release?

charliescheer (approved):
Looks good to me.  Thanks!!

jkmassel:
>The final test would be attempting an upload to App Store Connect. 🤔 We could do that, too, maybe with an hotfix build number so the ASC upload succeeds but we don't compromise the build number sequence for the current release?

I would suggest always doing a throwaway upload just to be sure everything is fine 😄

jkmassel (approved):
Code LGTM, though with entitlements it's so hard to know 😅

PR #1381, opened 2021-08-04
Fixed a bug where the account deletion token was  created incorrectly

charliescheer:
### Fix
This PR is in response to feedback from  @eshurakov after the account deletion was merged.
I have:
1. fixed a bug where the account deletion token was created each time the app came from the background
2. removed guards for url request and url creations in the remotes

### Test
1. Attempt to do an account signup to confirm the remotes still work correctly
2. attempt to request deletion for that account
3. after requesting deletion, send the app into the background
4. complete the deletion
5. bring the app out of the background and confirm that the login screen appears

### Review

### DIFF
--- a/Simplenote/AccountRemote.swift
+++ b/Simplenote/AccountRemote.swift
@@ -8,7 +8,7 @@ class AccountRemote: Remote {
     /// Send verification request for specified email address
     ///
     func verify(email: String, completion: @escaping (_ result: Result<Data?, RemoteError>) -> Void) {
-        let request = verificationURLRequest(with: email)!
+        let request = verificationURLRequest(with: email)
 
         performDataTask(with: request, completion: completion)
     }
@@ -16,17 +16,14 @@ class AccountRemote: Remote {
     /// Send account deletion request for user
     ///
     func requestDelete(_ user: SPUser, completion: @escaping (_ result: Result<Data?, RemoteError>) -> Void) {
-        let request = deleteRequest(with: user)!
-
+        let request = deleteRequest(with: user)
         performDataTask(with: request, completion: completion)
     }
 
     // MARK: URL Requests
 
-    private func verificationURLRequest(with email: String) -> URLRequest? {
-        guard let base64EncodedEmail = email.data(using: .utf8)?.base64EncodedString() else {
-            return nil
-        }
+    private func verificationURLRequest(with email: String) -> URLRequest {
+        let base64EncodedEmail = email.data(using: .utf8)!.base64EncodedString()
         let verificationURL = URL(string: SimplenoteConstants.verificationURL)!
 
         var request = URLRequest(url: verificationURL.appendingPathComponent(base64EncodedEmail),
@@ -37,7 +34,7 @@ class AccountRemote: Remote {
         return request
     }
 
-    private func deleteRequest(with user: SPUser) -> URLRequest? {
+    private func deleteRequest(with user: SPUser) -> URLRequest {
         let url = URL(string: SimplenoteConstants.accountDeletionURL)!
 
         var request = URLRequest(url: url,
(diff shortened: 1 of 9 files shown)


Simplenote/AccountRemote.swift
```diff
@@ -8,32 +8,26 @@ class AccountRemote: Remote {
...
 
     // MARK: URL Requests
 
     private func verificationURLRequest(with email: String) -> URLRequest? {
-        guard let base64EncodedEmail = email.data(using: .utf8)?.base64EncodedString(),
-              let verificationURL = URL(string: SimplenoteConstants.verificationURL) else {
+        guard let base64EncodedEmail = email.data(using: .utf8)?.base64EncodedString() else {
             return nil
         }
+        let verificationURL = URL(string: SimplenoteConstants.verificationURL)!
```
eshurakov: If we force-unwrap the whole `URLRequest` later, we are better to force unwrap `base64EncodedEmail` here and change method signature to return non-optional request.

Simplenote/AccountRemote.swift
```diff
@@ -44,9 +38,7 @@ class AccountRemote: Remote {
     }
 
     private func deleteRequest(with user: SPUser) -> URLRequest? {
```
eshurakov: Same here, we don't need to return an optional request here, we can change it to be non-optional.

Simplenote/Classes/SignupRemote.swift
```diff
@@ -4,12 +4,9 @@ import Foundation
 //
 class SignupRemote: Remote {
     func signup(with email: String, completion: @escaping (_ result: Result<Data?, RemoteError>) -> Void) {
-        guard let request = request(with: email) else {
-            completion(.failure(RemoteError.urlRequestError))
-            return
-        }
+        let urlRequest = request(with: email)!
```
eshurakov: Let's force-unwrap URL and return non-optional request here. If we're gonna crash, we should crash as close to the initial source as possible.

Simplenote/Controllers/AccountDeletionController.swift
```diff
@@ -2,18 +2,24 @@ import Foundation
...
         return Date() > expirationDate
     }
 
     func requestAccountDeletion(_ user: SPUser, completion: @escaping (_ result: Result<Data?, RemoteError>) -> Void) {
-
-        AccountRemote().requestDelete(user, completion: completion)
+        accountDeletionRequestDate = Date()
+        AccountRemote().requestDelete(user) { [weak self] (result) in
+            if case .success = result {
+                self?.accountDeletionRequestDate = Date()
```
eshurakov: Curious why do we need to set `accountDeletionRequestDate` twice? 
I think it should be enough to do it only when we get a successful result?

Simplenote/Controllers/AccountDeletionController.swift
```diff
@@ -2,18 +2,24 @@ import Foundation
 
 @objc
 class AccountDeletionController: NSObject {
-    private let accountDeletionRequestDate = Date()
+    private var accountDeletionRequestDate: Date?
 
     var deletionTokenHasExpired: Bool {
-        guard let expirationDate = accountDeletionRequestDate.increased(byDays: 1) else {
+        guard let requestDate = accountDeletionRequestDate,
+              let expirationDate = requestDate.increased(byDays: 1) else {
```
eshurakov: Note: this can also be written like this without an extra var :-)
```
guard let expirationDate = accountDeletionRequestDate?.increased(byDays: 1) else {
```

eshurakov (commented):
Hey Charlie 👋 

Looks good! Left a few comments.

Simplenote/Controllers/AccountDeletionController.swift
charliescheer: Ha!  We don't.  I had it running when the user requested the delete at first, but changed it to if a success is received and must have forgotten to remove the first one

Simplenote/Controllers/AccountDeletionController.swift
charliescheer: ah, good call. Fixed

charliescheer:
@eshurakov I have fixed those suggestions, this is ready for another look

eshurakov (approved):
Looks good, thank you 👍

PR #1385, opened 2021-08-06
Undo via snackbar and copy internal link tests

jostnes:
### Fix
Adding two new tests to the test suite:
1. Undo trash via snackbar on note list
2. Copy internal link via context menu and pasting to another note

### Test
Test flow:
1. Login > Create 1 note > Trash note > Validate note list is empty > Undo trash > Validate note list not empty
2. Login > Create 1 note > Copy internal link via context menu > Create another note > Paste link into new note > Validate on note list that the new note has the correct title and internal link as content

### Review
@pachlava 

### Release

### DIFF
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -55,7 +55,7 @@ class NoteEditor {
     class func enterTitle(enteredValue: String) {
         let noteContentTextView = app.textViews.element
 
-        noteContentTextView.clearAndEnterText(text: enteredValue)
+        noteContentTextView.clearAndEnterText(text: enteredValue + "\n")
     }
 
     class func pasteNoteContent() {
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -52,6 +52,17 @@ class NoteEditor {
         }
     }
 
+    class func enterTitle(enteredValue: String) {
+        let noteContentTextView = app.textViews.element
+
+        noteContentTextView.clearAndEnterText(text: enteredValue)
+    }
+
+    class func pasteNoteContent() {
+        app.press(forDuration: 1.2)
+        app.menuItems["Paste"].tap()
+    }
+
     class func getEditorText() -> String {
         return app.textViews.element.value as! String
     }
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -60,7 +60,7 @@ class NoteEditor {
 
     class func pasteNoteContent() {
         app.press(forDuration: 1.2)
-        app.menuItems["Paste"].tap()
+        app.menuItems[UID.ContextMenuItem.paste].tap()
     }
 
     class func getEditorText() -> String {
--- a/SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
+++ b/SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
@@ -331,30 +331,27 @@ class SimplenoteUISmokeTestsNoteEditor: XCTestCase {
         NoteListAssert.noteExists(noteName: noteName + "-1")
     }
 
-    func testCopyAndPasteInterlink() throws {
+    func testCopyAndPasteInternalLink() throws {
         trackTest()
-        let toCopyNoteTitle = "Copy Internal Link Title"
-        let toPasteNoteTitle = "Paste Internal Link Title\n"
-        let note = NoteData(
-            name: "Paste Internal Link Title",
-            content: "[Copy Internal Link Title-1](simplenote://note/"
-        )
+        let originalNoteTitle = "Original Note Title"
+        let referringNoteTitle = "Referring Note Title"
+        let referringNoteData = NoteData(
+                name: referringNoteTitle,
+                content: "[\(originalNoteTitle)](simplenote://note/"
+            )
 
         trackStep()
-        populateNoteList(title: toCopyNoteTitle, numberToCreate: 1)
-
-        trackStep()
-        NoteList.longPressNote(title: toCopyNoteTitle + "-1")
+        NoteList.createNoteAndLeaveEditor(noteName: originalNoteTitle)
+        NoteList.longPressNote(title: originalNoteTitle)
         NoteList.copyInternalLinkFromContextMenu()
+        NoteListAssert.notesNumber(expectedNotesNumber: 1)
 
         trackStep()
         NoteList.addNoteTap()
-        NoteEditor.enterTitle(enteredValue: toPasteNoteTitle)
+        NoteEditor.enterTitle(enteredValue: referringNoteTitle)
         NoteEditor.pasteNoteContent()
         NoteEditor.leaveEditor()
-
-        trackStep()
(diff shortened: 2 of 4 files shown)


SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
```diff
@@ -312,4 +312,49 @@ class SimplenoteUISmokeTestsNoteEditor: XCTestCase {
...
+        NoteListAssert.noteExists(noteName: noteName + "-1")
+    }
+
+    func testCopyAndPasteInterlink() throws {
+        trackTest()
+        let toCopyNoteTitle = "Copy Internal Link Title"
+        let toPasteNoteTitle = "Paste Internal Link Title\n"
+        let note = NoteData(
+            name: "Paste Internal Link Title",
+            content: "[Copy Internal Link Title-1](simplenote://note/"
```
jostnes: a lot of strings here and not sure how to make this one cleaner, they're similar but not entirely the same.

SimplenoteUITests/UIDs.swift
```diff
@@ -77,6 +79,10 @@ enum UID {
     enum SearchField {
         static let search = "Search notes or tags"
     }
+
+    enum ContextMenuItem {
+        static let paste = "Paste"
+    }
```
jostnes, line 85: created new enum for context menu, figured it would be better to have it's own instead of storing it with buttons.

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
```diff
@@ -312,4 +312,49 @@ class SimplenoteUISmokeTestsNoteEditor: XCTestCase {
         TrashAssert.noteExists(noteName: noteTitle + "-1")
         TrashAssert.noteExists(noteName: noteTitle + "-2")
     }
+
+    func testUndoTrashFromSnackbar() throws {
+        trackTest()
+        let noteName = "Note Title"
+
+        trackStep()
+        populateNoteList(title: noteName, numberToCreate: 1)
```
pachlava: `populateNoteList` is good when we talk about multiple notes creation, but here we need to create only one.

By using `NoteList.createNoteAndLeaveEditor` instead of `populateNoteList` here, you would not need to append "-1" on lines 324 and 331.

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
```diff
@@ -312,4 +312,49 @@ class SimplenoteUISmokeTestsNoteEditor: XCTestCase {
...
+        trackTest()
+        let toCopyNoteTitle = "Copy Internal Link Title"
+        let toPasteNoteTitle = "Paste Internal Link Title\n"
+        let note = NoteData(
+            name: "Paste Internal Link Title",
+            content: "[Copy Internal Link Title-1](simplenote://note/"
+        )
+
+        trackStep()
+        populateNoteList(title: toCopyNoteTitle, numberToCreate: 1)
```
pachlava: It's not something official, or accepted as a rule, it's just the way it's done for all existing tests in this project: if the test has a step, the step ends with an assertion. For the sake of consistency, I'd either add an assertion for new note existence to this step, or would merge it with the next step.

Just for some context: as you can see from the [testing checklist](https://github.com/Automattic/simplenote-ios/blob/develop/TESTING-CHECKLIST.md) for Simplenote (it's the only testing document we have for Simplenote), there is no concept of steps, or expected results. A single test is just a single sentence, which is quite vague. 

This current structure for automated tests (multiple steps / expected results) is just a format that I was very used to because of previous experience, and it helped me to logically break down the tests. But it's not official, I'm just proposing you the change because of the consistency.

SimplenoteUITests/NoteEditor.swift
```diff
@@ -52,6 +52,17 @@ class NoteEditor {
         }
     }
 
+    class func enterTitle(enteredValue: String) {
+        let noteContentTextView = app.textViews.element
+
+        noteContentTextView.clearAndEnterText(text: enteredValue)
```
pachlava: This method does not add a new line after the first characters of note content (instead, you pass a string that already contains the `\n` in the end) so in its current state, it's no different from `NoteEditor.clearAndEnterText`.

In this case, adding this:

```suggestion
        noteContentTextView.clearAndEnterText(text: enteredValue + "\n")
```

Would add meaning to the method, and would also free you from the necessity of having the "\n" in the strings you define in test.

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
```diff
@@ -312,4 +312,49 @@ class SimplenoteUISmokeTestsNoteEditor: XCTestCase {
...
+    }
+
+    func testCopyAndPasteInterlink() throws {
+        trackTest()
+        let toCopyNoteTitle = "Copy Internal Link Title"
+        let toPasteNoteTitle = "Paste Internal Link Title\n"
+        let note = NoteData(
+            name: "Paste Internal Link Title",
+            content: "[Copy Internal Link Title-1](simplenote://note/"
+        )
```
pachlava: There are several ways to improve this situation a bit:

1. Keeping in mind that you can have a line break automatically added by `NoteEditor.enterTitle`, you no longer need to have "\n" at the end of `toPasteNoteTitle`
2. Since you create only one note via `populateNoteList`, you can use `NoteList.createNoteAndLeaveEditor` instead, and then you have no need to have "-1" in NoteData content.
3. You could adjust the naming a bit, so that the names reflect clearer what are they for.
4. Use [string interpolation](https://www.hackingwithswift.com/read/0/5/string-interpolation), since now you can refer both strings in the note data

In the end, it can look something like this:

```suggestion
        let originalNoteTitle = "Original Note Title"
        let referringNoteTitle = "Referring Note Title"
        let referringNoteData = NoteData(
            name: referringNoteTitle
            content: "[\(originalNoteTitle)](simplenote://note/"
        )
```

pachlava (changes requested):
Thanks for the other PR @jostnes ! 👍 

I left several notes, I think we can :shipit: after the one related to note input data improvements for `testCopyAndPasteInterlink` and the one related to "steps with no expected results" is adjusted.

SimplenoteUITests/NoteEditor.swift
jostnes: Good point, totally didn't see this one. I'll update this to include the new line so that it doesn't have to be passed.

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
jostnes: Yup make sense, would be better for tests to be consistent than having different ways for the same thing which would cause more confusion in the long run. I appreciate comments like this 👍 

In this case would be better to move it to the next step so I'll do that. Also this will be a two "step" test then which I think should be alright so it's:
1. First step - copy and validate
2. Second step - paste and validate

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
jostnes: Thanks for the link on string interpolation, I was actually looking for something similar to [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) on swift, now I know what it is 😄

pachlava (approved):

PR #1389, opened 2021-08-09
List Widget

charliescheer:
### Fix
This PR is the final step in the process of adding home screen widgets to Simplenote iOS.  In this PR I am adding the last widget, the list widget.  

(note this PR currently merges back into feature/note-widget cause a lot of the ground work for the list widget is in that PR)

The list widget has a few features to it that are good to check on this PR.
- Each note should open the note in the list when tapped
- In the top right is a new note button, this should start a new note when tapped.
- In the preview screen (when adding the simplenote widget) this should default to showing the user's all notes list
- when first added the widget should default to showing the users all notes list until configured

### Test
1. Build the app, and then go to the home screen and long press anywhere to get the edit mode to start
2. Find simplenote, select it, in the screen that appears swipe all the way to the right.  The last two widgets will be the list widget.  Make sure that it shows your all notes view
3. add the widget and notice that all notes is the default setting
4. press on the widget and select the tap menu, you will see a list of all of your tags
5. choose a tag and make sure the widget displays the top notes from that tag (they will have the same organization as your app)
6. Change to dark mode and make sure that it all looks right
7. Try loading some notes from the list
8. try creating a new note from the button
9. tap on the header to load the main notes list


SylvesterWilmott:
This looks great!

A couple of minor notes about the behaviour:
- If the user has set the widget to display a list of notes with a certain tag, we should automatically tag a new note created from the widget. This mirrors the behaviour of what we currently have in the app.
- Tapping the title of the widget brings you to the app (presumably because this is the default behaviour of widgets). Do you think we would be able to allow the title to also be tappable and open the app at the corresponding tag?

Design:
- Although it technically makes sense to show the note titles at 17pt, when the list widget appears alongside the note widget, the note titles in the list widget does appear quite large in comparison. Can we try sizing these to 15pts, the same as the note widget?
- The new note icon appears quite small, are we able to show this icon at its natural size ie. 24x24pt?

charliescheer:
@SylvesterWilmott Thanks for the feedback!  I was able to make all of those changes, tapping on the tag name will load the note list to that tag, the new note button adds the tag, and I have changed the sizes you requested.

This is ready to review again when you get a chance!

SylvesterWilmott:
Thanks, @charliescheer!

I noticed that on the large widget one of the keystrokes is thicker than the rest on a 3x density screen. It's always the same one that's thicker under the third item which seems to correspond with where the small widget cuts the list. Can we change this particular stroke to always be 1 pixel?

Screenshot 2021-08-13 at 09 20 43

charliescheer:
>I noticed that on the large widget one of the keystrokes is thicker than the rest on a 3x density screen. It's always the same one that's thicker under the third item which seems to correspond with where the small widget cuts the list. Can we change this particular stroke to always be 1 pixel?

That is truly bizarre @SylvesterWilmott I am not exactly sure what is going on there.  The divider is added as part of the row, which are repeated a certain number depending on the number of cells based on the widget size.  The large and medium sized widgets shouldn't know anything about each other.

Even weirder is this seems to only happen if the number of rows is divisible by 8.  Here are some bizarre examples:
Medium widget with 8 rows, here the thick line is after the 6th row:

Large widget with 16 rows, here it is also after the 6th row:

large widget with 7 rows, here there is no thicker line:


I'm honestly flummoxed on what could be causing this.  Easiest solution would be to increase or decrease the number of rows in the large widget.  Not ideal but it does seem to solve the issue.  @jleandroperez or @eshurakov  any ideas on what would cause a swift ui divider to be slightly thicker when created by a ForEach if it is a multiple of 8 times repeated?

charliescheer:
Also just noting a few additional changes I made.  The list widget was not updating when the app went into the background, that has been fixed.  Also I reformatted how the colors are working as to use the static UIColor variables directly in some cases.

charliescheer:
Okay. I chatted with some folks about what could be causing that single line to be slightly larger, and it was suggested that maybe there was an issue with a division remainders worth of pixels.  So I changed the number of pixels slightly and that got rid of the thicker line 🤷 
pushed a change that fixes the single thick line

Simplenote/Controllers/WidgetController.swift
```diff
@@ -8,8 +8,16 @@ struct WidgetController {
...
+            if widgets.contains(where: { (widget) -> Bool in
+                if widget.configuration as? NoteWidgetIntent != nil {
+                    return true
+                }
+
+                if widget.configuration as? ListWidgetIntent != nil {
+                    return true
+                }
+
+                return false
```
jleandroperez:
```
switch widget.configuration {
case is NoteWidgetIntent:
    return true

case is ListWidgetIntent:
    return true

default:
   return false
}
```

jleandroperez: 😎

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -107,6 +107,27 @@ extension SPAppDelegate {
...
+    @objc
+    func handleOpenTagList(url: NSURL) -> Bool {
+        guard let tag = url.internalTagKey else {
+            return false
+        }
+
+        let request = NSFetchRequest<Tag>()
+        request.entity = Tag.entity()
+        request.predicate = NSPredicate(format: "name = %@", tag)
+        let tagExists = (try? coreDataManager.managedObjectContext.count(for: request) > 0) ?? false
```
jleandroperez: We should extract this:

```
func tagExists(_ tag: String) -> Bool { }
```

SimplenoteWidgets/Providers/ListWidgetProvider.swift
```diff
@@ -0,0 +1,79 @@
...
+            ListWidgetNoteProxy(title: note.title, url: note.url)
+        }
+
+        completion(ListWidgetEntry(date: Date(), tag: WidgetConstants.allNotesIdentifier, noteProxys: proxies))
+    }
+
+    func getTimeline(for configuration: ListWidgetIntent, in context: Context, completion: @escaping (Timeline<ListWidgetEntry>) -> Void) {
+        // Confirm valid configuration
+        guard let tag = configuration.tag?.identifier,
+              let notes = try? widgetDataController()?.notes(withFilter: TagsFilter(from: tag), limit: 8) else {
```
jleandroperez: Magic numbers!

SimplenoteIntents/IntentHandler.swift
```diff
@@ -61,3 +51,32 @@ extension IntentHandler: NoteWidgetIntentHandling {
...
+
+        // Return collection to intents
+        let collection = tagNoteInObjectCollection(from: tags)
+        completion(collection, nil)
+    }
+
+    private func tagNoteInObjectCollection(from tags: [Tag]) -> INObjectCollection<WidgetTag> {
+        let allNotesWidgetTag: [WidgetTag] = [WidgetTag(identifier: WidgetConstants.allNotesIdentifier, display: Constants.allNotesDisplay)]
+        let fetchedWidgetTags = tags.map({ tag in
+            WidgetTag(identifier: tag.name ?? "Unamed Tag", display: tag.name ?? "Unamed Tag")
```
jleandroperez: Nipticky: Localized String in the placeholder. Plus: `Unnamed` or `Unamed` ? double N?

SimplenoteIntents/IntentHandler.swift
```diff
@@ -61,3 +51,32 @@ extension IntentHandler: NoteWidgetIntentHandling {
...
+        return INObjectCollection(items: allNotesWidgetTag + fetchedWidgetTags)
+    }
+
+    func defaultTag(for intent: ListWidgetIntent) -> WidgetTag? {
+        WidgetTag(identifier: WidgetConstants.allNotesIdentifier, display: Constants.allNotesDisplay)
+    }
+}
+
+private struct Constants {
+    static let allNotesDisplay = "All Notes"
```
jleandroperez: Nipticky: localized?

SimplenoteWidgets/Providers/ListWidgetProvider.swift
```diff
@@ -0,0 +1,79 @@
...
+            ListWidgetNoteProxy(title: note.title, url: note.url)
+        }
+
+        // Prepare timeline entry for every hour for the next 6 hours
+        // Create a new set of entries at the end of the 6 entries
+        let entries: [ListWidgetEntry] = WidgetConstants.rangeForSixEntries.compactMap { (index) in
+            guard let date = Date().increased(byHours: index) else {
+                return nil
+            }
+            return ListWidgetEntry(date: date, tag: tag, noteProxys: proxies)
```
jleandroperez: Beautiful!

SimplenoteWidgets/Tools/DemoContent.swift
```diff
@@ -4,6 +4,22 @@ struct DemoContent {
     static let singleNoteTitle = "Twelve Tone Serialism"
     static let singleNoteContent = "The twelve-tone technique is often closely related with the compositional style, serialism. Fundamentally, twelve-tone serialism is a compositional technique where all 12 notes of the chromatic scale are played with equal frequency throughout the piece without any emphasis on any one note. For this reason, twelve-tone serialism avoids being in any key. Arnold Schoenberg was arguably the most influential composers who embraced the twelve-tone technique. Schoenberg described the system as a “Method of composing with twelve tones which are related only with one another.”"
     static let singleNoteContentAlt = "- [] Check item one \n- [] check item two\n- [] check item three"
+    static let singleNoteURL = URL(string: SimplenoteConstants.simplenoteScheme + "://")!
 
     static let demoURL = URL(string: SimplenoteConstants.simplenoteScheme + "://")!
+
+    static let listTag = "Composition"
```
jleandroperez, line 11: Should this string be localized?

SimplenoteWidgets/Views/NoteListRow.swift
```diff
@@ -0,0 +1,23 @@
...
+            .font(.subheadline)
+            .lineLimit(1)
+            .frame(width: width, height: height, alignment: .leading)
+        Divider()
+    }
+}
+
+struct NoteListRow_Previews: PreviewProvider {
+    static var previews: some View {
+        NoteListRow(noteTitle: "Title for note", width: 300, height: 50)
```
jleandroperez: Nipticky: Constants?

jleandroperez (commented):
Few comments sir! Lookin' sharp!

Simplenote/Controllers/WidgetController.swift
charliescheer: nice!!

SimplenoteWidgets/Views/NoteListRow.swift
charliescheer: That is just for the preview, it has no real affect on the actual compiled code...  In WPiOS, they removed all of the preview code.  I was thinking we should do the same thing.  
On one hand if we come to work on the UI later we have to add in these sorts of previews.  But on the other hand, how often will be be coming back to redesign the widget UI???  WDYT?

SimplenoteWidgets/Tools/DemoContent.swift
charliescheer: Differing this the [conversation on the Note Widget PR](https://github.com/Automattic/simplenote-ios/pull/1375#discussion_r693405169) about what these strings should be.

SylvesterWilmott:
Thanks for the updates @charliescheer. The separator looks good now. All seem to be uniform.

I found a couple of small issues:
- When the widget shows "All Notes" and the "New Note" icon is tapped, the new note is tagged "All-Notes". Let's make sure that notes that are created in this scenario are not tagged with anything/tags are left blank.
- Can we remove the separator on the last item in the list? As far as I can tell, this is an issue only in the small list widget.

charliescheer:
@SylvesterWilmott thanks for finding those additional issues.  I have updated the list widget again and have made those changes

SylvesterWilmott:
Looks great @charliescheer

SylvesterWilmott (approved):

Simplenote/Classes/SimplenoteConstants.swift
```diff
@@ -18,6 +18,14 @@ class SimplenoteConstants: NSObject {
     ///
     static let simplenoteInterlinkHost = "note"
 
+    /// Simplenote: Tag list Host
+    ///
+    static let simplenoteInternalTagHost = "list"
+
+    /// Simplenote: Identifier for all notes list
+    ///
+    static let allNotesTagIdentifier = "All-Notes"
```
jleandroperez, line 27: I believe using the `All-Notes` as a keyword to map the `All Notes` filter is troublesome. Why:

- Nothing forbids you from adding a tag called `All-Notes`
- Okay, no reason (2). Just the above 😄 

Idea on how to address this noted a little bit below (we cooould try with a different data structure, defined via the intents editor, maybe)

Simplenote/Controllers/WidgetController.swift
```diff
@@ -8,8 +8,15 @@ struct WidgetController {
                 return
             }
 
-            if widgets.contains(where: { widget in
-                widget.configuration as? NoteWidgetIntent != nil
+            if widgets.contains(where: { (widget) -> Bool in
```
jleandroperez, line 11: (No need to change anything!!) `-> Bool` is not needed, since it's already defined in the method signature. Same for the parenthesis:

```
if widgets.contains(where { widget in
```

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -107,6 +107,28 @@ extension SPAppDelegate {
...
 
+    /// Opens the Note list displaying a tag associated with a given URL instance, when possible
+    ///
+    @objc
+    func handleOpenTagList(url: NSURL) -> Bool {
+        guard let tag = url.internalTagKey else {
+            return false
+        }
+
+        selectedTag = tagExists(tag) ? tag : nil
```
jleandroperez: (Note to self) we truly need to turn this `selectedTag` property into an enum. Blocked by the AppDelegate not being Swift, so can't do ATM!

SimplenoteIntents/IntentHandler.swift
```diff
@@ -15,30 +15,22 @@ class IntentHandler: INExtension {
...
-        NSLog("failed to return handler")
         return self
     }
 }
 
 extension IntentHandler: NoteWidgetIntentHandling {
     func provideNoteOptionsCollection(for intent: NoteWidgetIntent, with completion: @escaping (INObjectCollection<WidgetNote>?, Error?) -> Void) {
-        NSLog("attempting to provied note options collection")
-        guard let dataController = try? WidgetDataController(coreDataManager: coreDataManager) else {
+        guard let dataController = try? WidgetDataController(context: coreDataManager.managedObjectContext) else {
```
jleandroperez: I keep thinking about `WidgetDataController` (since `Data` makes you think about NSData right?).

Maybe `WidgetResultsController`... `WidgetDataSource` (brainstorming!) 🤔

SimplenoteIntents/IntentHandler.swift
```diff
@@ -51,13 +43,45 @@ extension IntentHandler: NoteWidgetIntentHandling {
...
-        guard let note = try? dataController.firstNote() else {
-            return nil
-        }
         return WidgetNote(identifier: note.simperiumKey, display: note.limitedTitle)
     }
 }
+
+extension IntentHandler: ListWidgetIntentHandling {
+    func provideTagOptionsCollection(for intent: ListWidgetIntent, with completion: @escaping (INObjectCollection<WidgetTag>?, Error?) -> Void) {
+        guard let dataController = try? WidgetDataController(context: coreDataManager.managedObjectContext) else {
```
jleandroperez: Question: is there any ("legal") scenario in which the `MOC` could be nil?

SimplenoteWidgets/Tools/WidgetDataController.swift
```diff
@@ -1,85 +1,111 @@
...
+
+enum TagsFilter {
+    case allNotes
+    case tag(String)
+}
+
+extension TagsFilter {
+    init(from tag: String) {
+        switch tag {
+        case SimplenoteConstants.allNotesTagIdentifier:
```
jleandroperez: (Noted above) **ideally** we should get rid of `allNotesTagIdentifier` and use the enum directly.

SimplenoteWidgets/Views/NewNoteButton.swift
```diff
@@ -0,0 +1,30 @@
...
+                .foregroundColor(foregroundColor)
+                .frame(side: size)
+        }
+        .background(backgroundColor)
+    }
+}
+
+struct NewNoteButton_Previews: PreviewProvider {
+    static var previews: some View {
+        NewNoteImage(size: 48, foregroundColor: .white, backgroundColor: .blue)
```
jleandroperez: Nipticky: Magic number 🔥

SimplenoteWidgets/Views/NoteListRow.swift
```diff
@@ -0,0 +1,31 @@
...
+            .lineLimit(1)
+            .frame(width: width, height: height, alignment: .leading)
+            .foregroundColor(.bodyTextColor)
+        Divider()
+            .opacity(lastRow ? Double.zero : Constants.fullOpacity)
+    }
+}
+
+private struct Constants {
+    static let fullOpacity = 1.0
```
jleandroperez, line 22: `UIKitConstants.alpha1_0` ?

Ahhh darn. This is SwiftUI. If there are any other occurrences, may be worth adding... god forgive me, `SwiftUIConstants` 🤦‍♂️

jleandroperez: Plus: Extremely annoying note (please forgive me!!!).

No need to change this, typically, in super long files, it's easier if you keep the constants at the very bottom. You just scroll down to the bottom, whenever you need to change anything. (Otherwise, you need to search!)

SimplenoteWidgets/Views/NoteListRow.swift
```diff
@@ -0,0 +1,31 @@
...
+}
+
+struct NoteListRow_Previews: PreviewProvider {
+    static var previews: some View {
+        NoteListRow(noteTitle: "Title for note", width: 300, height: 50, lastRow: false)
```
jleandroperez, line 28: Magic numbers / unlocalized string

SimplenoteWidgets/Views/NoteListTable.swift
```diff
@@ -0,0 +1,31 @@
+import SwiftUI
+
+struct NoteListTable: View {
+    let entry: ListWidgetEntry
+    let numberOfRows: Int
+    let geometry: GeometryProxy
+
+    var body: some View {
+        let width = geometry.size.width - Constants.sidePadding
+        let height = (geometry.size.height - Constants.headerSize) / CGFloat(numberOfRows)
```
jleandroperez: Just in case, we should probably round this value.

In UIKit, a screen location that has a floating value usually yields a blurry UI. This is the height, shouldn't bug, but just in case?

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -107,6 +107,28 @@ extension SPAppDelegate {
...
+        navigationController.popToRootViewController(animated: true)
+
+        return true
+    }
+
+    private func tagExists(_ tag: String) -> Bool {
+        let request = NSFetchRequest<Tag>(entityName: Tag.entityName)
+        request.predicate = NSPredicate(format: "\(\Tag.name) = %@", tag)
+        return (try? coreDataManager.managedObjectContext.count(for: request) > .zero) ?? false
+    }
```
jleandroperez: You can use `SPObjectManager.shared().tagExists()` here (this one is already implemented there)

SimplenoteWidgets/Extensions/Color+Widgets.swift
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    static var widgetBlueBackgroundColor: Color {
+        Color(studioColor: .spBlue50)
+    }
+
+    static var widgetTintColor: Color {
+        Color(light: .spBlue50, dark: .spBlue30)
+    }
+}
```
jleandroperez, line 19: Q: was this file added in the `note` widget PR?

SimplenoteWidgets/Providers/ListWidgetProvider.swift
```diff
@@ -0,0 +1,83 @@
...
+}
+
+struct ListWidgetProvider: IntentTimelineProvider {
+    typealias Intent = ListWidgetIntent
+    typealias Entry = ListWidgetEntry
+
+    let coreDataManager: CoreDataManager!
+
+    init() {
+        NSLog("Created a provider")
```
jleandroperez: 🔥

SimplenoteWidgets/Tools/DemoContent.swift
jleandroperez: IMO if it's gonna be presented of the user it shooould prooobably be localized. Maybe. Unless we use latin for all?

SimplenoteWidgets/Tools/WidgetConstants.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+struct WidgetConstants {
+    static let environmentXcodePreviewsKey = "XCODE_RUNNING_FOR_PREVIEWS"
```
jleandroperez: Suggestion:

Rather than running this in few spots:

```
let isPreview = ProcessInfo.processInfo.environment[WidgetConstants.environmentXcodePreviewsKey] != WidgetConstants.isPreviews
```

You can also add an extension on ProcessInfo, and just encapsulate the check in there. It'd save you some characters here and there, and it'd all be together

SimplenoteWidgets/Views/NoteListRow.swift
```diff
@@ -0,0 +1,31 @@
...
+struct NoteListRow: View {
+    var noteTitle: String
+    var width: CGFloat
+    var height: CGFloat
+    var lastRow: Bool
+
+    var body: some View {
+        Text(noteTitle)
+            .font(.subheadline)
+            .lineLimit(1)
```
jleandroperez: Nipticky: constant?

Simplenote/URL+Simplenote.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+
+extension URL {
+    static func internalUrl(for tag: String) -> URL {
```
jleandroperez: Suggestion: Either one of these two signatures would work super:

```
static func internalUrl(forTag Tag: String) -> {}
static func internalUrlForTag(_ tag: String) -> { }
```

Otherwise it's not clear what kind of URL will be produced. `URL.internalUrl(for: sometring)`

Simplenote/URL+Simplenote.swift
```diff
@@ -0,0 +1,29 @@
...
+        URL(string: SimplenoteConstants.simplenoteScheme +
+                "://" +
+                SimplenoteConstants.simplenoteInternalTagHost +
+                tagQuery(for: tag))!
```
jleandroperez: I've also noticed we've got `SimplenoteConstants.simplenoteScheme + "://" + SOMETHING"` in a ton of places. How about... maaayyybe:

```
extension String {
    static func simplenotePath(extension: String) -> String {
        SimplenoteConstants.simplenoteScheme + "://"
    }
}
```

SimplenoteIntents/IntentHandler.swift
```diff
@@ -51,13 +43,45 @@ extension IntentHandler: NoteWidgetIntentHandling {
...
+
+        // Return collection to intents
+        let collection = tagNoteInObjectCollection(from: tags)
+        completion(collection, nil)
+    }
+
+    private func tagNoteInObjectCollection(from tags: [Tag]) -> INObjectCollection<WidgetTag> {
+        let allNotesWidgetTag: [WidgetTag] = [WidgetTag(identifier: SimplenoteConstants.allNotesTagIdentifier, display: Constants.allNotesDisplay)]
+        let fetchedWidgetTags = tags.map({ tag in
+            WidgetTag(identifier: tag.name ?? Constants.unnamedTag, display: tag.name ?? Constants.unnamedTag)
```
jleandroperez: Nipticky: `tag.name ??...` is duplicated

SimplenoteIntents/IntentHandler.swift
```diff
@@ -51,13 +43,45 @@ extension IntentHandler: NoteWidgetIntentHandling {
...
+        })
+        return INObjectCollection(items: allNotesWidgetTag + fetchedWidgetTags)
```
jleandroperez: Potentially bad idea / untested: Since the `List.intentdefinitions` don't support Enums with Associated Values, you **could** try something like this:

```
struct ListFilter {
    let kind: ListFilterKind
    let payload: String?
}

enum ListFilterKind {
   case allNotes
   case tag
}
```

The point is to "emulate" an associated values. When `kind == .tag` then `payload` is the tag name`. Otherwise, all notes.

Notes:
- This would allow you get rid of `TagsFilter`, and have a single structure
- Yes, this is a workaround
- Didn't test it, so not 100% on spot if this would work, or if there's any other fields required.
- This would also fix the `"All-Notes"` reserved keyword problem (having a Tag with that name would break the scheme)

WDYT? (Just brainstorming, I'm not 100% sold on the idea)

jleandroperez (commented):
Looking good @charliescheer !! sending you few notes your way. Thank you sir!!

SimplenoteIntents/IntentHandler.swift
```diff
@@ -51,13 +43,45 @@ extension IntentHandler: NoteWidgetIntentHandling {
...
+            return
+        }
+
+        // Return collection to intents
+        let collection = tagNoteInObjectCollection(from: tags)
+        completion(collection, nil)
+    }
+
+    private func tagNoteInObjectCollection(from tags: [Tag]) -> INObjectCollection<WidgetTag> {
+        let allNotesWidgetTag: [WidgetTag] = [WidgetTag(identifier: SimplenoteConstants.allNotesTagIdentifier, display: Constants.allNotesDisplay)]
```
jleandroperez: Duplicated in L80. May be worth extracting, if possible, as a WidgetTag extension

SimplenoteIntents/IntentHandler.swift
```diff
@@ -51,13 +43,45 @@ extension IntentHandler: NoteWidgetIntentHandling {
     }
 
     func defaultNote(for intent: NoteWidgetIntent) -> WidgetNote? {
-        guard let dataController = try? WidgetDataController(coreDataManager: coreDataManager) else {
+        guard let note = try? WidgetDataController(context: coreDataManager.managedObjectContext).firstNote() else {
```
jleandroperez: Another Q: What's the `IntentHandler` lifecycle like? (does it get instantiated when the Widget goes live, then it gets killed?)

I've notice nearly all of the API(s) in here require a `WidgetDataController` instance. Maybe we should instantiate that in the `init`, and keep it around?

Simplenote/Classes/SimplenoteConstants.swift
charliescheer: I added an enum to the WidgetTag for all notes and tag which allows us to avoid this reserved key.

Simplenote/SPAppDelegate+Extensions.swift
charliescheer: someday

SimplenoteIntents/IntentHandler.swift
charliescheer: There shouldn't be....

SimplenoteWidgets/Views/NoteListRow.swift
charliescheer: Also an issue, the opacity in SwiftUI is using a double not a float.  I originally tried to use the UIKit constants... but nope note without changing the type

charliescheer: `This is SwiftUI. If there are any other occurrences, may be worth adding... god forgive me, SwiftUIConstants 🤦‍♂️`

We could do this.  I think it would make sense, but how much SwiftUI are we planning on using?  If we choose to move away from  IB and to SwiftUI then certainly we should make some global constants, but if it is just the widgets...  I don't know if it is worth it WDYT?

SimplenoteWidgets/Views/NoteListTable.swift
charliescheer: That could explain a problem I had with the UI earlier where one of the dividers was bigger than the others, but only in certain circumstances...
https://github.com/Automattic/simplenote-ios/pull/1389#issuecomment-899160237

SimplenoteWidgets/Tools/WidgetConstants.swift
charliescheer: Good idea!

Simplenote/URL+Simplenote.swift
charliescheer: 😄  made this change yesterday!

SimplenoteIntents/IntentHandler.swift
charliescheer: Fixed in an update yesterday

SimplenoteIntents/IntentHandler.swift
charliescheer: I had it setup that way originally, but that caused the issue where if someone added the widget before logging in, the data controller wasn't updated after they did login.  This cause the issues you highlighted here https://github.com/Automattic/simplenote-ios/pull/1375#issuecomment-902348107

From what I can tell the intent handler is created by springboard when the device is setup and is not released.  So if the data controller failed to setup because the user was not logged in, it would stay that way till the device was reset after logging in

SimplenoteWidgets/Extensions/Color+Widgets.swift
charliescheer: Yes.  I may have added it to both, which will create some issue when merging.  I will keep an eye out for that when I do fix the conflicts on this branch

Simplenote/Classes/SimplenoteConstants.swift
jleandroperez, line 27: 🔥 This one can go! (aaawesome!!)

Simplenote/Classes/String+Simplenote.swift
```diff
@@ -178,3 +178,17 @@ extension String {
...
+extension String {
+    static func simplenotePath(withHost host: String? = nil) -> String {
+        var path = SimplenoteConstants.simplenoteScheme + "://"
+
+        if let host = host {
+            path.append(host)
+            path.append("/")
+        }
+
+        return path
```
jleandroperez, line 192: The cool kids don't use `var` 😂 

```
let base = SimplenoteConstants.simplenoteScheme + "://"
guard let host = host else {
   return base
}

return base + host + "/"
```

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -107,6 +107,26 @@ extension SPAppDelegate {
...
+    func handleOpenTagList(url: NSURL) -> Bool {
+        guard url.isInternalTagURL else {
+            return false
+        }
+
+        if let tag = url.internalTagKey {
+            selectedTag = SPObjectManager.shared().tagExists(tag) ? tag : nil
+        } else {
+            selectedTag = nil
+        }
```
jleandroperez, line 122: You can simplify this API as follows:

```
    func handleOpenTagList(url: NSURL) -> Bool {
        guard let tag = url.internalTagKey else {
            return false
        }

        selectedTag = SPObjectManager.shared().tagExists(tag) ? tag : nil

        navigationController.presentedViewController?.dismiss(animated: true, completion: nil)
        navigationController.popToRootViewController(animated: true)

        return true
    }
```

Why:

- `internalTagKey` will return `nil` only if the URL is not a TagURL. So the second branch of the if (L121) should never run, granted that we've got a guard in L114
- No (2), just (1) 😛

SimplenoteWidgets/Extensions/ProcessInfo+Simplenote.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+extension ProcessInfo {
+    var environmentIsPreview: Bool {
```
jleandroperez, line 4: TY!

SimplenoteWidgets/Extensions/WidgetTag+Helpers.swift
```diff
@@ -0,0 +1,35 @@
...
+        switch kind {
+        case .allNotes, .unknown:
+            self.init(identifier: nil, display: Constants.allNotesDisplay, kind: .allNotes)
+        case .tag:
+            self.init(identifier: name, display: name ?? Constants.unnamedTag, kind: .tag)
+        }
+    }
+
+    static var allNotes: WidgetTag {
+        WidgetTag.init(kind: .allNotes)
```
jleandroperez, line 19:
```suggestion
        WidgetTag(kind: .allNotes)
```

SimplenoteWidgets/Providers/ListWidgetProvider.swift
```diff
@@ -0,0 +1,82 @@
...
+            return ListWidgetEntry(date: date, widgetTag: widgetTag, noteProxys: proxies)
+        }
+
+        completion(Timeline(entries: entries, policy: .atEnd)
+)
+    }
+
+    private func widgetResultsController() -> WidgetResultsController? {
+        return try? WidgetResultsController(context: coreDataManager.managedObjectContext,
+                                            isPreview: ProcessInfo.processInfo.environmentIsPreview)
```
jleandroperez: Question: 
- Should this be a property?
- Are we expecting this to actually ever fail?

SimplenoteWidgets/Providers/ListWidgetProvider.swift
```diff
@@ -0,0 +1,82 @@
...
+        // Create a new set of entries at the end of the 6 entries
+        let entries: [ListWidgetEntry] = WidgetConstants.rangeForSixEntries.compactMap { (index) in
+            guard let date = Date().increased(byHours: index) else {
+                return nil
+            }
+            return ListWidgetEntry(date: date, widgetTag: widgetTag, noteProxys: proxies)
+        }
+
+        completion(Timeline(entries: entries, policy: .atEnd)
+)
```
jleandroperez, line 76: Nipticky: this one belongs to L70 I think

SimplenoteWidgets/Views/NoteListRow.swift
jleandroperez: Over time, SwiftUI will probably take over. Not for now, though. Good to keep there! TY!

jleandroperez (commented):
Few notes!! testing as of now!! 

Lookin' sharp 😎

jleandroperez (approved):
✅ Verified logged out scenario works super
✅ Verified (all of the widgets) work great after logging in

Few pending comments sent a bit ago, plus: 
-`isPreview` can be dropped from WidgetResultsController
- `WidgetTag` initializer can be simplified (as discussed over slack!)
- And we need to handle the logout scenario, definitely good to do so in another PR

Other than that, please feel free to :shipit: when ready sir. Excellent work!!

PR #1391, opened 2021-08-10
Fixed an issue where tags aren't removed from notes when deleted

### LINKED ISSUE #1352: Tag removed from tags list still remains in the note
### Expected
If the tag is removed from tags list, it should also be removed inside notes

### Observed
Tag remains in the note after it was deleted in the sidebar:

https://user-images.githubusercontent.com/73365754/125072815-bd0e2f00-e0c3-11eb-80d9-08e43a05e141.MP4

### Reproduced
1. Open any note
2. Add a new tag into the note
3. Remove this tag in the sidebar
4. Open the note again ~> tag will still be there. It will also remain in the native macOS app.

-|-|-|-
iPhone|XR|14.6|4.39.0.0

charliescheer:
### Fix
This issue fixes #1352 

Currently there is an issue where tags are not removed from notes when they are deleted from the tag list in Simplenote iOS.

This PR fixes that issue so that if a tag is deleted in the list it is also removed from all notes (including deleted notes)

### Test
1. Create a few tags on a few different notes
2. delete one of those notes
3. go to the tags list and delete a tag
4. check the notes that had that tag and confirm they have been deleted
5. restore the deleted note and check the tag was removed from that note too

### Review

### DIFF
--- a/Simplenote/Classes/SPObjectManager+Simplenote.swift
+++ b/Simplenote/Classes/SPObjectManager+Simplenote.swift
@@ -31,10 +31,14 @@ extension SPObjectManager {
         }
 
         let request = NSFetchRequest<Note>(entityName: Note.entityName)
-        request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
-            .predicateForNotes(tag: tagName),
-            .predicateForNotes(deleted: includeDeleted)
-        ])
+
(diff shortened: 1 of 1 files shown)


eshurakov (approved):
Thank you for fixing it 👍 

:shipit:

PR #1395, opened 2021-08-11
Log out user account after account deletion success

### LINKED ISSUE #1390: Note list appears briefly before auth screen after account is deleted.
After recently adding in app account deletion, a bug was noticed in the MacOS version where the note list appears just for a moment before the auth screen loads.

This is caused in part by how simperium decides if it needs to verify accounts.  The solution that worked on mac os is to force the account to be logged out if simperium fails after an account deletion is requested

### Reproduced
1. create a new account in simplenote iOS
2. go to the settings menu -> Delete and follow the steps to delete the account
3. Once deleted you will be logged out and the auth screen will appear.
4. close the app completely (from the app switcher) and launch the app again.  For a moment the note list will appear before the auth screen reappears

### LINKED ISSUE #1393: Account deletion alert lacks two dots `.`
### Expected
The ends of sentences should be marked with dots. Also, [blueprints](https://simplenotep2.wordpress.com/2021/07/15/account-settings-delete-account-i3/).

### Observed
There are missing dots at the end of user email, and at the end of the very last sentence

photo_2021-08-10 15 55 35

### Reproduced
1. Log in
2. Tap `Delete Account` in Settings
3. Tap `Request Account Deletion` ~> you will see the dialog depicted above.

-|-|-|-
iPhone|XR|14.6|4.42.0.0

charliescheer:
### Fix
Per some discussions with in app account deletion for MacOS, we decided that it would be best to log out a user once the account deletion is complete, instead of simply displaying the onboarding.  We did this on macOS to deal with some windowing bugs, which were not apparent on iOS, but could lead to being able to navigate the app after an account was deleted making it seem like the deletion had failed.

By logging the user out the local data is removed and there is no way to navigate the app. 
### Test
1.  Create an account to delete
2. delete that account.
3. open simplenote iOS back up, notice that with in a moment or two the onboarding VC will appear.
4. send simplenote to the background and open it again
5. You will see the slide out menu for a short moment (this is in part because of the restoration state and I do not think is avoidable) but it can't be interacted with and the onboarding vc should appear quickly

### Review

### Release

### DIFF
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -286,7 +286,6 @@ extension SPAppDelegate: SimperiumDelegate {
         switch simperiumError {
         case .invalidToken:
             logOutIfAccountDeletionRequested()
-            break
         default:
             break
         }
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -278,7 +278,18 @@ extension SPAppDelegate: SimperiumDelegate {
 
     public func simperium(_ simperium: Simperium, didFailWithError error: Error) {
         SPTracker.refreshMetadataForAnonymousUser()
-        logOutIfAccountDeletionRequested()
+
+        guard let simperiumError = SPSimperiumErrors(rawValue: (error as NSError).code) else {
+            return
+        }
+
+        switch simperiumError {
+        case .invalidToken:
+            logOutIfAccountDeletionRequested()
+            break
+        default:
+            break
+        }
     }
 }
 
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -278,6 +278,18 @@ extension SPAppDelegate: SimperiumDelegate {
 
     public func simperium(_ simperium: Simperium, didFailWithError error: Error) {
         SPTracker.refreshMetadataForAnonymousUser()
+
+        guard let simperiumError = SPSimperiumErrors(rawValue: (error as NSError).code) else {
+            return
+        }
+
+        switch simperiumError {
+        case .invalidToken:
+            logOutIfAccountDeletionRequested()
+            break
+        default:
+            break
+        }
     }
 }
 
@@ -415,6 +427,15 @@ extension SPAppDelegate {
         }
 
         simperium.authenticateIfNecessary()
+    }
+
+    @objc
+    func logOutIfAccountDeletionRequested() {
+        guard accountDeletionController?.hasValidDeletionRequest == true else {
+            return
+        }
+
+        logoutAndReset(self)
     }
 }
 
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -278,18 +278,6 @@ extension SPAppDelegate: SimperiumDelegate {
 
     public func simperium(_ simperium: Simperium, didFailWithError error: Error) {
         SPTracker.refreshMetadataForAnonymousUser()
-
-        guard let simperiumError = SPSimperiumErrors(rawValue: (error as NSError).code) else {
-            return
-        }
-
-        switch simperiumError {
-        case .invalidToken:
-            logOutIfAccountDeletionRequested()
-            break
-        default:
-            break
-        }
     }
 }
 
@@ -427,15 +415,6 @@ extension SPAppDelegate {
         }
 
         simperium.authenticateIfNecessary()
-    }
-
-    @objc
-    func logOutIfAccountDeletionRequested() {
-        guard accountDeletionController?.hasValidDeletionRequest == true else {
-            return
-        }
-
-        logoutAndReset(self)
     }
 }
 
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -420,10 +420,7 @@ extension SPAppDelegate {
 
     @objc
     func logOutIfAccountDeletionRequested() {
-        guard let deletionController = accountDeletionController,
-              deletionController.hasValidDeletionRequest,
-              let user = simperium.user,
-              user.authenticated() else {
+        guard accountDeletionController?.hasValidDeletionRequest == true else {
             return
         }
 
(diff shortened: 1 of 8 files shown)


Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -278,6 +278,7 @@ extension SPAppDelegate: SimperiumDelegate {
 
     public func simperium(_ simperium: Simperium, didFailWithError error: Error) {
         SPTracker.refreshMetadataForAnonymousUser()
+        logOutIfAccountDeletionRequested()
```
eshurakov: We should only handle `SPSimperiumErrorsInvalidToken` error here as that's the only case we're interested in

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -408,16 +409,25 @@ extension SPAppDelegate {
...
-        if deletionController.deletionTokenHasExpired {
+        simperium.authenticateIfNecessary()
+    }
+
+    @objc
+    func logOutIfAccountDeletionRequested() {
+        guard let deletionController = accountDeletionController,
+              deletionController.hasValidDeletionRequest,
+              let user = simperium.user,
+              user.authenticated() else {
```
eshurakov: `user.authenticated()` will be false because we come here when user token is invalid, which means user is not authenticated. 
We don't need this condition here and can safely remove it.

Simplenote/SPConstants.h
```diff
@@ -23,6 +23,7 @@ extern NSString *const kAutomatticAnalyticLearnMoreURL;
 
 extern NSString *const kShareExtensionGroupName;
 extern NSString *const kSimplenoteWPServiceName;
+extern NSString *const SPUsername;
```
eshurakov: This is not used :-)

Simplenote/SPAppDelegate+Extensions.swift
eshurakov: After I removed `user.authenticated()` and tested the account deletion flow, I found another issue.

Calling `[self.simperium signOutAndRemoveLocalData:YES completion:^{}];` doesn't do anything because the first line there is checking if user is authenticated. It will return `false` in our case because the token is not valid and `singOut` won't do anything because of that. This means that all the data is still stored in the core data and we failed to remove it.

https://github.com/Simperium/simperium-ios/blob/develop/Simperium/Simperium.m#L589
```
- (void)signOutAndRemoveLocalData:(BOOL)remove completion:(SimperiumSignoutCompletion)completion {
    
    // Don't proceed, if the user isn't logged in
    if (!self.user.authenticated) {
        return;
    }

   ...
```

The issue here is that simperium checks for `self.user.authenticated` which can be false, but we will still have the user object and credentials stored in the keychain and on the next app launch Simperium will try to use the token from the keychain. The check is quite misleading.

I suggest that we change ` if (!self.user.authenticated)` to ` if (!self.user)` or even fully remove the line. This should allow us to fully clean the state even if the user is not authenticated.

@jleandroperez WDYT?

Simplenote/SPAppDelegate+Extensions.swift
jleandroperez: +1 to this one 😄

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -408,16 +409,25 @@ extension SPAppDelegate {
 //
 extension SPAppDelegate {
     @objc
-    func authenticateSimperiumIfNeeded() {
-        guard let deletionController = accountDeletionController else {
+    func authenticateSimperiumIfAccountDeletionRequested() {
+        guard let deletionController = accountDeletionController,
+              deletionController.hasValidDeletionRequest else {
```
jleandroperez, line 424: You can collapse this into a single check:

```
guard accountDeletionController?.hasValidDeletionRequest == true else { //...
```

Simplenote/SPAppDelegate+Extensions.swift
jleandroperez: Simperium [updated here](https://github.com/Simperium/simperium-ios/pull/622). TY Evgeny!! 

+1 yes let's just check if there's a valid deletion request pending.
Plus: we should point to the new Simperium branch (I'll make sure to release a new version once we know it's good!)

Simplenote/SPAppDelegate.m
```diff
@@ -325,6 +327,12 @@ - (void)logoutAndReset:(id)sender
     });
 }
 
+-(void)removeStoredCredentialsFromKeychain
+{
+    [self.simperium.authenticator reset];
```
jleandroperez: We should prob. also drop this invocation

jleandroperez (commented):
Thank you gentleman 😄

charliescheer:
Alright, tested the simperium changes, and that does work.  It created a new issue where the onboarding screen was loaded for a moment then it is dismissed, and then it is displayed again.  This happens because of simperium.authenticationDidFail schedules an onboarding VC to appear and logout calls dismissAllModals and then also calls authenticateIfNecessary which schedules a new onboarding VC.  I removed the call to dismiss  all modals to fix the issue.

Also I added a change that fixes #1393  (adding that in so it will auto close when this is merged)

To test these changes with the simperium changes in the SNiOS Pod file comment out 
`pod 'Simperium', '1.8.0'`

and add

`pod 'Simperium', :git => 'https://github.com/Simperium/simperium-ios.git', :branch => 'issue/drops-unneeded-logout-check'`

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -278,6 +278,18 @@ extension SPAppDelegate: SimperiumDelegate {
...
         SPTracker.refreshMetadataForAnonymousUser()
+
+        guard let simperiumError = SPSimperiumErrors(rawValue: (error as NSError).code) else {
+            return
+        }
+
+        switch simperiumError {
+        case .invalidToken:
+            logOutIfAccountDeletionRequested()
+            break
```
jleandroperez: No `break` needed here

Simplenote/SPAppDelegate.m
```diff
@@ -315,12 +315,11 @@ - (void)logoutAndReset:(id)sender
 			
 			// hide sidebar of notelist
             [self.sidebarViewController hideSidebarWithAnimation:NO];
-			
-			[self dismissAllModalsAnimated:YES completion:^{
```
jleandroperez, line 319: I believe `dismissAllModals` was introduced to dismiss the spinner (and we should probably check on iPad too).

How can I reproduce the loop-y behavior? (Already tried reverting the Simperium change, and dropping this). I think we should probably either keep the `dismissAllModals`, or analyze very closely all of the things that get dismissed by this call.

eshurakov (approved):
Worked great for me!

charliescheer:
I checked it out as well and it looks good to me @jleandroperez

jleandroperez (approved):
:shipit: sir!

PR #1397, opened 2021-08-16
Fixed a bug where non http/https urls crashed when opened from markdown

### LINKED ISSUE #725: Markdown Links to Shortcut URL Crashing App
### Expected
The user is adding shortcut URLs https://support.apple.com/guide/shortcuts/open-create-and-run-a-shortcut-apda283236d7/ios to their notes via markdown. They would expect these URLs to open in the app as any others

### Observed
When they try to open such a link the only thing that happens is that the app closes. The shortcut does not work as expected

(Image link: https://d.pr/v/UnxjTf )

-|-|-|-
DEVICE_MAKE|DEVICE_MODEL|IOS_VERSION|APP_VERSION
iPhone|8|13.4|4.16

#2822590-zen

charliescheer:
### Fix
This PR fixes #725 

Currently if you try to open an a URL scheme that is note simplenote:// http:// or https:// from markdown preview the app will crash because the urls can not be natively opened from web kit.

This change checks to see if the webkit can open a url and if it can't then it passes it along to safari to avoid the crash.

### Test
1. in a note with markdown enabled add this text `[shortcuts](shortcuts://)
2. open markdown mode
3. tap on the link.  You should be taken to the shortcuts app and simplenote should not crash

### DIFF
--- a/Simplenote/Classes/SPMarkdownPreviewViewController.m
+++ b/Simplenote/Classes/SPMarkdownPreviewViewController.m
@@ -130,8 +130,13 @@
         return;
     }
 
-    SFSafariViewController *sfvc = [[SFSafariViewController alloc] initWithURL:targetURL];
-    [self presentViewController:sfvc animated:YES completion:nil];
+
+    if ([WKWebView handlesURLScheme:targetURL.scheme]) {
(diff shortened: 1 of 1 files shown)


eshurakov (approved):
👍 🚢 :shipit:

PR #1398, opened 2021-08-17
Added error case at login if user email is not verified

charliescheer:
### Fix
This PR adds a new error case while logging in where a user's email address has not been verified they will receive an error alerting them to the fact that they can not login till their email is verified and then supplies them with a button to resend the verification email.

### Test
1. setup a tool like Charles proxy to get in the way of the login request response.
2. When the response returns, change the response value to 403
3. You should see an alert with two options, verify and okay.

Note this will require more testing when the backend changes are made to return the 403 response to an unverified email, currently that has not been added in, so it can not be fully tested.

### Review

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 4.43
 -----
 -   [Internal] added login alert for compromised password #1394
+- [Intenrnal] added error response case on login for unverified emails #1398
 
 4.42
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,7 +1,7 @@
 4.43
 -----
 -   [Internal] added login alert for compromised password #1394
-- [Intenrnal] added error response case on login for unverified emails #1398
+-   [Internal] added error response case on login for unverified emails #1398
 
 4.42
 -----
--- a/Simplenote/Classes/SPAuthError.swift
+++ b/Simplenote/Classes/SPAuthError.swift
@@ -25,7 +25,9 @@ extension SPAuthError {
         case 401:
             self = response == Constants.compromisedPassword ? .compromisedPassword : .loginBadCredentials
         case 403:
-            self = .unverifiedEmail
+            self = response == Constants.requiresVerification ?
+                .unverifiedEmail :
+                .unknown(statusCode: loginErrorCode, response: response, error: error)
         default:
             self = .unknown(statusCode: loginErrorCode, response: response, error: error)
         }
@@ -89,4 +91,5 @@ extension SPAuthError {
 
 private struct Constants {
     static let compromisedPassword = "compromised password"
+    static let requiresVerification = "requires verification"
 }
--- a/Simplenote/Classes/SPAuthError.swift
+++ b/Simplenote/Classes/SPAuthError.swift
@@ -22,12 +22,12 @@ extension SPAuthError {
     ///
     init(loginErrorCode: Int, response: String?, error: Error?) {
         switch loginErrorCode {
+        case 401 where response == Constants.compromisedPassword:
+            self = .compromisedPassword
         case 401:
-            self = response == Constants.compromisedPassword ? .compromisedPassword : .loginBadCredentials
-        case 403:
-            self = response == Constants.requiresVerification ?
-                .unverifiedEmail :
-                .unknown(statusCode: loginErrorCode, response: response, error: error)
+            self = .loginBadCredentials
+        case 403 where response == Constants.requiresVerification:
+            self = .unverifiedEmail
         default:
             self = .unknown(statusCode: loginErrorCode, response: response, error: error)
         }
--- a/Simplenote/Classes/SPAuthError.swift
+++ b/Simplenote/Classes/SPAuthError.swift
@@ -9,6 +9,7 @@ enum SPAuthError: Error {
     case signupUserAlreadyExists
     case network
     case compromisedPassword
+    case unverifiedEmail
     case unknown(statusCode: Int, response: String?, error: Error?)
 }
 
@@ -23,6 +24,8 @@ extension SPAuthError {
         switch loginErrorCode {
         case 401:
             self = response == Constants.compromisedPassword ? .compromisedPassword : .loginBadCredentials
+        case 403:
+            self = .unverifiedEmail
         default:
             self = .unknown(statusCode: loginErrorCode, response: response, error: error)
         }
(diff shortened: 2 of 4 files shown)


Simplenote/Classes/SPAuthError.swift
```diff
@@ -23,6 +24,8 @@ extension SPAuthError {
         switch loginErrorCode {
         case 401:
             self = response == Constants.compromisedPassword ? .compromisedPassword : .loginBadCredentials
+        case 403:
+            self = .unverifiedEmail
```
eshurakov: I believe we also wanted to send an error code in the body (`requires verification`) similarly to how we did it with compromised password.
Let's check with @nsakaimbo

Simplenote/Classes/SPAuthViewController.swift
```diff
@@ -443,6 +445,16 @@ private extension SPAuthViewController {
         present(alertController, animated: true, completion: nil)
     }
 
+    func presentUserUnverifiedError(error: SPAuthError) {
+        let alertController = UIAlertController(title: error.title, message: error.message, preferredStyle: .alert)
+        alertController.addCancelActionWithTitle(AuthenticationStrings.unverifiedCancelText)
+        alertController.addDefaultActionWithTitle(AuthenticationStrings.unverifiedActionText) { _ in
+            AccountRemote().verify(email: self.email) { _ in }
```
eshurakov: I think we should show an activity indicator (can use the one we have for account deletion) and handle errors similar to what we're doing in `AccountVerificationViewController`

RELEASE-NOTES.txt
```diff
@@ -1,6 +1,7 @@
 4.43
 -----
 -   [Internal] added login alert for compromised password #1394
+- [Intenrnal] added error response case on login for unverified emails #1398
```
jleandroperez:
```suggestion
-   [Internal] added error response case on login for unverified emails #1398
```

Simplenote/Classes/SPAuthError.swift
charliescheer: I chatted with @nsakaimbo during our one on one today and he agreed that we should also check for the requires verification body.  I have updated the changes to do that the same way we did in compromised password.

charliescheer:
Thanks for those suggestions @eshurakov I have updated the codes.  This is ready for another look

Simplenote/Classes/UIAlertController+Helpers.swift
```diff
@@ -26,4 +26,12 @@ extension UIAlertController {
 
         return action
     }
+
+    static func dismissableAlert(title: String,
+                                        message: String,
+                                        style: UIAlertController.Style = .alert) -> UIAlertController {
+        let alert = UIAlertController(title: title, message: message, preferredStyle: style)
+        alert.addCancelActionWithTitle("Okay")
```
eshurakov: Not localized string :-)

Simplenote/Classes/SPAuthViewController.swift
```diff
@@ -670,6 +697,12 @@ private enum AuthenticationStrings {
     static let loginActionText              = NSLocalizedString("Log In", comment: "Log In Action")
     static let compromisedAlertCancel       = NSLocalizedString("Not Yet", comment: "Cancel action for password alert")
     static let compromisedAlertReset        = NSLocalizedString("Change Password", comment: "Change password action")
+    static let unverifiedCancelText         = NSLocalizedString("Okay", comment: "Email unverified alert dismiss")
```
eshurakov: We never used "Okay" before, in all other alerts it says "OK"

eshurakov (approved):
Besides the fact that I find it strange to use alerts instead of reusing the UI that we already have for account verification and a few localization nitpicks, everything works as expected 👍

Simplenote/Classes/SPAuthError.swift
```diff
@@ -23,6 +24,10 @@ extension SPAuthError {
         switch loginErrorCode {
         case 401:
             self = response == Constants.compromisedPassword ? .compromisedPassword : .loginBadCredentials
+        case 403:
+            self = response == Constants.requiresVerification ?
+                .unverifiedEmail :
+                .unknown(statusCode: loginErrorCode, response: response, error: error)
```
jleandroperez: BTW you can also use `where`:

```
    init(loginErrorCode: Int, response: String?, error: Error?) {
        switch loginErrorCode {
        case 401:
            self = response == Constants.compromisedPassword ? .compromisedPassword : .loginBadCredentials
        case 403 where response == Constants.requiresVerification:
            self = .unverifiedEmail
        default:
            self = .unknown(statusCode: loginErrorCode, response: response, error: error)
        }
    }
```

Simplenote/Classes/SPAuthViewController.swift
```diff
@@ -443,6 +445,31 @@ private extension SPAuthViewController {
...
+
+            AccountRemote().verify(email: email) { result in
+                spinnerVC.dismiss(animated: false, completion: nil)
+                var alert: UIAlertController
+                switch result {
+                case .success:
+                    alert = UIAlertController.dismissableAlert(title: AuthenticationStrings.verificationSentTitle,
+                                                                   message: String(format: AuthenticationStrings.verificationSentTemplate, email))
+                case .failure:
+                    alert = UIAlertController.dismissableAlert(title: AuthenticationStrings.unverifriedErrorTitle,
```
jleandroperez: Nipticky: `unverifried` (fried!)

Simplenote/Classes/SPAuthViewController.swift
```diff
@@ -443,6 +445,31 @@ private extension SPAuthViewController {
...
+            let spinnerVC = SpinnerViewController()
+            self?.present(spinnerVC, animated: false, completion: nil)
+
+            AccountRemote().verify(email: email) { result in
+                spinnerVC.dismiss(animated: false, completion: nil)
+                var alert: UIAlertController
+                switch result {
+                case .success:
+                    alert = UIAlertController.dismissableAlert(title: AuthenticationStrings.verificationSentTitle,
+                                                                   message: String(format: AuthenticationStrings.verificationSentTemplate, email))
```
jleandroperez: Nipticky: alignment!

Simplenote/Classes/SPAuthViewController.swift
```diff
@@ -443,6 +445,31 @@ private extension SPAuthViewController {
...
+                case .failure:
+                    alert = UIAlertController.dismissableAlert(title: AuthenticationStrings.unverifriedErrorTitle,
+                                                                   message: AuthenticationStrings.unverifiedErrorMessage)
```
jleandroperez: Nipticky: Alignment!

jleandroperez (approved):
Couple nipticky notes. :shipit: when ready!!

Simplenote/Classes/SPAuthViewController.swift
charliescheer: I was having some crazy typo issues yesterday!

PR #1401, opened 2021-08-19
Adopt ScreenObject package in screenshots tests: Proof of Concept

mokagio:
I decided to link it only to the screenshots automation target because the work required to meaningfully integrate the framework in the UI tests target is more complex.

At this point in time, my goal is simply to get the framework setup and used in a leaf screen within the navigation tree (the passcode), just to serve as a proof of concept.

### Test

1. Checkout this branch
2. Either run the screenshot automation tests from Xcode, or via Fastlane with `bundle exec fastlane take_screenshots_from_app mode:dark devices:'iPhone 11 Pro Max' languages:en-US` and ensure they pass

### Review

### Release

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -4660,8 +4660,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/Automattic/ScreenObject";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 0.1.0;
+				branch = "always-compute-at-runtime";
+				kind = branch;
 			};
 		};
 		3F762E8D2677F19C0088CD45 /* XCRemoteSwiftPackageReference "XCUITestHelpers" */ = {
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -54,6 +54,7 @@
 		37FD30491FC4CFA2008D0B78 /* KeychainPasswordItem.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37FD30471FC4CFA2008D0B78 /* KeychainPasswordItem.swift */; };
 		3F1BB4A0242DB81E006D1A04 /* ScreenshotsCredentials.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F1BB49F242DB81E006D1A04 /* ScreenshotsCredentials.swift */; };
 		3F1BB4A3242DC162006D1A04 /* SnapshotHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3F1BB4A2242DC162006D1A04 /* SnapshotHelper.swift */; };
+		3F4BA07E26CDF295000619B1 /* ScreenObject in Frameworks */ = {isa = PBXBuildFile; productRef = 3F4BA07D26CDF295000619B1 /* ScreenObject */; };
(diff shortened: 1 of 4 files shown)


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -4652,6 +4660,14 @@
 /* End XCConfigurationList section */
 
 /* Begin XCRemoteSwiftPackageReference section */
+		3F4BA07C26CDF295000619B1 /* XCRemoteSwiftPackageReference "ScreenObject" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/Automattic/ScreenObject";
+			requirement = {
+				branch = "always-compute-at-runtime";
+				kind = branch;
```
mokagio, line 4668: Once https://github.com/Automattic/ScreenObject/pull/3 is merged and before merging this, I'll ship a new ScreenObject version, and point to it here.

SimplenoteScreenshots/SimplenoteScreenshots.swift
```diff
@@ -209,46 +208,19 @@ class SimplenoteScreenshots: XCTestCase {
...
+            try PasscodeScreen(app: app).type(passcode: 1234)
+        } catch {
+            XCTFail("Expected passcode screen to exist but it did not.")
         }
     }
 
-    func type(_ text: String, onFirstKeyboardOf app: XCUIApplication) {
-        text.forEach { app.keyboards.firstMatch.keys[String($0)].tap() }
-    }
-
```
mokagio, line 251: This was unused. Likely a leftover from the time the passcode screen used the system's numerical keyboard.

SimplenoteScreenshots/SimplenoteScreenshots.swift
```diff
@@ -263,7 +235,7 @@ extension XCUIElement {
 
     /// Workaround to type text into secure text fields due to the different behaviors across
     /// Simulators.
-    func typeSecureText(_ text: String, using app: XCUIApplication) -> Void {
+    func typeSecureText(_ text: String, using app: XCUIApplication) {
```
mokagio, line 238: `-> Void` is unnecessary because it's implicit in not specifying a return value.

mokagio (commented):
As I mentioned in the internal P2 Project Thread for this work, I don't plan to migrate the Simplenote automation to use ScreenObject immediately. It works fine the way it is. We may want to do some ad hoc migrations as we go, but we have bigger fishes to fry right now.

This PR was just a proof of concept to introduce the library and lower the barrier to entry when someone else will touch the tests.

SimplenoteScreenshots/PasscodeScreen.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    func type(passcode: Int) {
+        // This converts an Int into an [Int] of its digits
+        let digits = "\(passcode.magnitude)".compactMap(\.wholeNumberValue)
+
+        digits.forEach { digit in
+            let input = app.staticTexts["\(digit)"]
+            XCTAssertTrue(input.waitForExistence(timeout: 3))
+            // Both the custom UIButton and its UILabel match the "\(digit)" query. We need to pick
+            // one for the tap to work.
```
pachlava, line 35: When I tried dumping the object tree by printing app.debugDescription, I saw this (this is the example for the first row with 1-3 buttons):

Screenshot 2021-08-26 at 14 04 47

It looks like a button is represented with a StaticText the size of a button, with another StaticText inside, the size of the actual digit, and tapping both should work.

~~Because of this, I'm not sure if I fully understand the `We need to pick one for the tap to work` part 🤔~~

UPD: I understood `We need to pick one for the tap to work` as `We need to pick the one which can be tapped` instead of `We need any of both, as both can be tapped` 🤦

pachlava (approved):
Tested by running SimplenoteScreenshots, and it worked! It looks tidier this way 👍 , thanks!

PR #1410, opened 2021-08-25
[Tooling] Fix AppStore metadata scripts

AliSoftware:
Fix the tooling that handles metadata translation in order to properly handle the updates requested in pd8BQf-60-p2

 - Fix the `download_metadata.swift` script to properly set the subtitle's GlotPress translation to `subtitle.txt` instead of `promo_text.txt`
 - Update the Swift script and Fastfile to make the app name (aka title) be part of the list of translated metadata – given that we now include keywords in the app name too
 - Re-run the scripts to update the files, including the `.pot` so that title key gets imported in GP

### Known subtitle length issue

The current translations for the subtitle are exceeding the 30 chars limit. I'll post a P2 soon in +i18n to ask to fix them in GlotPress, alongside asking for the `title` translation.

Note that we initially talked in Slack about temporarily reverting the subtitle to use the English copy (which is obv <= 30 chars) for all locales until all the translations for it are fixed. But I figured since I'm hoping the copy will be fixed in GlotPress soon (🤞 ) after my upcoming P2, they should hopefully all be back to <30 by the time we run `finalize_release` then `fastlane deliver` to update the metadata in ASC with the latest GlotPress copies.

If that's not the case and some translations for the subtitle (and/or title) exceed their character limit, we'll need to manually replace the `subtitle.txt` file for each language _after_ running the `finalize_release` anyway, since that lane re-downloads the data from GlotPress and overwrites the metadata `.txt` files by then


fastlane/download_metadata.swift
```diff
@@ -2,11 +2,15 @@
...
+let glotPressTitleKey = "app_store_title"
 let glotPressSubtitleKey = "app_store_subtitle"
 let glotPressWhatsNewKey = "v4.43-whats-new"
 let glotPressDescriptionKey = "app_store_desc"
 let glotPressKeywordsKey = "app_store_keywords"
-let baseFolder = "./metadata"
+
+let scriptURL = URL(fileURLWithPath: CommandLine.arguments[0], relativeTo: URL(fileURLWithPath: FileManager.default.currentDirectoryPath))
+let fastlaneDir = scriptURL.deletingLastPathComponent()
+let baseFolder = fastlaneDir.appendingPathComponent("metadata").path
```
AliSoftware, line 13: This is to ensure that the path for `baseFolder` is computed relative to the script's path and not to the current directory.

Made that change because I kept accidentally running `fastlane/download_metadata.swift` from the root repo during my tests… and had the script create the `metadata/` folder at the repo root instead of inside the `fastlane/` subfolder.
This is not an issue when this script is run by our release-toolkit and its `ios_update_metadata` action because [we explicitly `cd fastlane` before invoking the script in this context](https://github.com/wordpress-mobile/release-toolkit/blob/develop/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb#L60), but it's still sometimes useful to run the script standalone/ourselves which is why I thought this to be useful

jkmassel (approved):
Nice – thanks for taking care of this!!

AliSoftware:
Follow-up PR so that the new key gets imported to GlotPress: #1411

PR #1412, opened 2021-08-26
Feature list widget

charliescheer:
### Fix
This PR is the same as #1389 While preparing to merge in the note widget and list widget I made some merging mistakes that were not easily unravelled, so it seemed best to step back a few commits and build the PRs/branches and to the merge again.  This PR contains all the same commits from #1389 See the original PR for the conversation and comments.

### Fix
This PR is the final step in the process of adding home screen widgets to Simplenote iOS.  In this PR I am adding the last widget, the list widget.  

(note this PR currently merges back into feature/note-widget cause a lot of the ground work for the list widget is in that PR)

The list widget has a few features to it that are good to check on this PR.
- Each note should open the note in the list when tapped
- In the top right is a new note button, this should start a new note when tapped.
- In the preview screen (when adding the simplenote widget) this should default to showing the user's all notes list
- when first added the widget should default to showing the users all notes list until configured

### Test
1. Build the app, and then go to the home screen and long press anywhere to get the edit mode to start
2. Find simplenote, select it, in the screen that appears swipe all the way to the right.  The last two widgets will be the list widget.  Make sure that it shows your all notes view
3. add the widget and notice that all notes is the default setting
4. press on the widget and select the tap menu, you will see a list of all of your tags
5. choose a tag and make sure the widget displays the top notes from that tag (they will have the same organization as your app)
6. Change to dark mode and make sure that it all looks right
7. Try loading some notes from the list
8. try creating a new note from the button
9. tap on the header to load the main notes list


charliescheer:
Realized we didn't need these new branches.  Closing

PR #1414, opened 2021-08-26
Merge release/4.43 into develop - AppStore keywords update

AliSoftware:
See request in pd8BQf-9O-p2

We need this to land in `develop` ASAP so it gets picked and imported by GlotPress soon after, such that translators start translating the new keywords and get the translations in before the 4.43 release.


mokagio (approved):
👍 the list matches the one in the P2 :shipit:

PR #1417, opened 2021-08-27
Fixes a bug where the select all button state might be off

### LINKED ISSUE #1377: Select all label displayed when there’s only 1 note and it’s selected
### Expected
When all notes are selected, label should read "Deselect All"

### Observed
When there is only 1 note on note list and it is selected, the top left label reads Select All instead of Deselect All

### Reproduced
1. Login to Simplenote
2. Go to notes list and add 1 note (if there are many remove others leaving just one)
3. Select the note using long press
4. See that label is Select All > Tap it
5. Note is deselected

Screen recording:

https://user-images.githubusercontent.com/17252150/127877753-ef72e776-726e-49d1-975a-80f10d82c5aa.mp4

-|-|-|-
iPhone|X (Simulator)|14.5|4.41

charliescheer:
### Fix
This PR fixes #1377 

Currently if you start the edit mode on the note list, and only one note is in the list currently the Select all button will continue to be set as Select all even though all of the notes in the list are selected.  

This PR adds a refresh to the start of select mode to update the state of the select all button after the cell has been selected.

### Test
1. Select a tag that only has one note in it
2. long press on the note to see the contextual menu and then tap on Select
3. The edit mode should appear, check the select all/deselect all button in the top left.  It should say "Deselect All"
4. try the button to make sure it behaves as expected

### Review

### Release
> `RELEASE-NOTES.txt` was updated in f17001 with:
> 
> > - Fixed an issue with the select all button when only one note is in the list #1417

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,6 @@
 4.44
 -----
-
+-   Fixed an issue with the select all button when only one note is in the list #1417
 
 4.43
 -----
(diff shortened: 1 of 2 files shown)


eshurakov (approved):
Looks good!

:shipit:

PR #1419, opened 2021-08-28
Adding UI tests for History feature

tiagomar:
I have added a few UI tests for the History feature.

### Tests
1. Run SimplenoteUITestsHistory.swift file from Xcode and make sure they all pass.

### Review

### Release

### DIFF
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -63,9 +63,9 @@ func populateNoteList(title: String, numberToCreate: Int) {
     }
 }
 
-func getRandonId() -> String.SubSequence {
+func getRandomId() -> String {
     let id = NSDate().timeIntervalSince1970.description.suffix(6)
-    return id
+    return String(id)
 }
 
 class Table {
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -63,6 +63,11 @@ func populateNoteList(title: String, numberToCreate: Int) {
     }
 }
 
+func getRandonId() -> String.SubSequence {
+    let id = NSDate().timeIntervalSince1970.description.suffix(6)
+    return id
+}
+
 class Table {
 
     class func getAllCells() -> XCUIElementQuery {
--- a/SimplenoteUITests/History.swift
+++ b/SimplenoteUITests/History.swift
@@ -18,19 +18,23 @@ class History {
 
 class HistoryAssert {
     class func historyShown() {
-        let dismissHistoryButton = app.buttons[UID.Button.dismissHistory]
-        let restoreNoteButton = app.buttons[UID.Button.restoreNote]
-
-        XCTAssertTrue(dismissHistoryButton.waitForExistence(timeout: minLoadTimeout), UID.Button.dismissHistory + buttonNotFound)
-        XCTAssertTrue(restoreNoteButton.waitForExistence(timeout: minLoadTimeout), UID.Button.restoreNote + buttonNotFound)
+        HistoryAssert.historyExistence(shouldExist: true, errorMessage: buttonNotFound)
     }
 
     class func historyDismissed() {
+        HistoryAssert.historyExistence(shouldExist: false, errorMessage: buttonNotAbsent)
+    }
+
+    class func historyExistence(shouldExist: Bool, errorMessage: String) {
         let dismissHistoryButton = app.buttons[UID.Button.dismissHistory]
         let restoreNoteButton = app.buttons[UID.Button.restoreNote]
 
-        XCTAssertFalse(dismissHistoryButton.exists, UID.Button.dismissHistory + buttonNotAbsent)
-        XCTAssertFalse(restoreNoteButton.exists, UID.Button.restoreNote + buttonNotAbsent)
+        XCTAssertEqual(dismissHistoryButton.waitForExistence(timeout: minLoadTimeout),
+                       shouldExist,
+                       UID.Button.dismissHistory + errorMessage)
+        XCTAssertEqual(restoreNoteButton.waitForExistence(timeout: minLoadTimeout),
+                       shouldExist,
+                       UID.Button.restoreNote + errorMessage)
     }
 
     class func restoreButtonIsDisabled() {
--- a/SimplenoteUITests/History.swift
+++ b/SimplenoteUITests/History.swift
@@ -1,0 +1,45 @@
+import XCTest
+
+class History {
+
+    class func close() {
+        guard app.staticTexts[UID.Button.restoreNote].waitForExistence(timeout: minLoadTimeout) else { return }
+        app.buttons[UID.Button.dismissHistory].tap()
+    }
+
+    class func setSliderPosition(position: CGFloat) {
+        app.sliders.element.adjust(toNormalizedSliderPosition: position)
+    }
+
+    class func restoreNote() {
+        app.buttons[UID.Button.restoreNote].tap()
+    }
+}
+
+class HistoryAssert {
+    class func historyShown() {
+        let closeHistoryButton = app.buttons[UID.Button.dismissHistory]
+        let restoreNoteButton = app.buttons[UID.Button.restoreNote]
+
+        XCTAssertTrue(closeHistoryButton.waitForExistence(timeout: minLoadTimeout), UID.Button.dismissHistory + buttonNotFound)
+        XCTAssertTrue(restoreNoteButton.waitForExistence(timeout: minLoadTimeout), UID.Button.restoreNote + buttonNotFound)
+    }
+
+    class func historyDismissed() {
+        let closeHistoryButton = app.buttons[UID.Button.dismissHistory]
+
+        XCTAssertFalse(closeHistoryButton.exists, UID.Button.dismissHistory + buttonNotFound)
+    }
+
+    class func restoreButtonIsDisabled() {
+        let restoreNoteButton = app.buttons[UID.Button.restoreNote]
+
+        XCTAssertFalse(restoreNoteButton.isEnabled)
+    }
+
+    class func restoreButtonIsEnabled() {
+        let restoreNoteButton = app.buttons[UID.Button.restoreNote]
+
+        XCTAssertTrue(restoreNoteButton.isEnabled)
+    }
+}
--- a/SimplenoteUITests/History.swift
+++ b/SimplenoteUITests/History.swift
@@ -18,17 +18,19 @@ class History {
 
 class HistoryAssert {
     class func historyShown() {
-        let closeHistoryButton = app.buttons[UID.Button.dismissHistory]
+        let dismissHistoryButton = app.buttons[UID.Button.dismissHistory]
         let restoreNoteButton = app.buttons[UID.Button.restoreNote]
 
-        XCTAssertTrue(closeHistoryButton.waitForExistence(timeout: minLoadTimeout), UID.Button.dismissHistory + buttonNotFound)
+        XCTAssertTrue(dismissHistoryButton.waitForExistence(timeout: minLoadTimeout), UID.Button.dismissHistory + buttonNotFound)
         XCTAssertTrue(restoreNoteButton.waitForExistence(timeout: minLoadTimeout), UID.Button.restoreNote + buttonNotFound)
     }
 
     class func historyDismissed() {
-        let closeHistoryButton = app.buttons[UID.Button.dismissHistory]
+        let dismissHistoryButton = app.buttons[UID.Button.dismissHistory]
+        let restoreNoteButton = app.buttons[UID.Button.restoreNote]
 
-        XCTAssertFalse(closeHistoryButton.exists, UID.Button.dismissHistory + buttonNotFound)
+        XCTAssertFalse(dismissHistoryButton.exists, UID.Button.dismissHistory + buttonNotAbsent)
+        XCTAssertFalse(restoreNoteButton.exists, UID.Button.restoreNote + buttonNotAbsent)
     }
 
     class func restoreButtonIsDisabled() {
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -58,10 +58,10 @@ class NoteEditor {
         noteContentTextView.clearAndEnterText(text: enteredValue + "\n")
     }
 
-    class func waitBeforeEditingText(delayInSeconds: TimeInterval, newText: String) {
+    class func waitBeforeEditingText(delayInSeconds: UInt32, newText: String) {
         let noteContentTextView = app.textViews.element
 
-        Thread.sleep(forTimeInterval: delayInSeconds)
+        sleep(delayInSeconds)
 
         noteContentTextView.clearAndEnterText(text: newText)
     }
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -186,9 +186,10 @@ class NoteEditor {
     }
 
     class func openHistory() {
-        app.navigationBars[UID.NavBar.allNotes].buttons[UID.Button.noteEditorMenu].tap()
-        guard app.tables.staticTexts[UID.Text.noteEditorOptionsHistory].exists else {return}
-        app.tables.staticTexts[UID.Text.noteEditorOptionsHistory ].tap()
+        app.buttons[UID.Button.noteEditorMenu].tap()
+        let historyButton = app.tables.staticTexts[UID.Text.noteEditorOptionsHistory]
+        guard historyButton.waitForExistence(timeout: minLoadTimeout) else {return}
+        historyButton.tap()
     }
 
 }
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -58,6 +58,14 @@ class NoteEditor {
         noteContentTextView.clearAndEnterText(text: enteredValue + "\n")
     }
 
+    class func waitBeforeEditingText(delayInSeconds: TimeInterval, newText: String) {
+        let noteContentTextView = app.textViews.element
+
+        Thread.sleep(forTimeInterval: delayInSeconds)
+
+        noteContentTextView.clearAndEnterText(text: newText)
+    }
+
     class func pasteNoteContent() {
         app.press(forDuration: 1.2)
         app.menuItems[UID.ContextMenuItem.paste].tap()
@@ -175,6 +183,12 @@ class NoteEditor {
         let matches = matchingTextViews.count
         print(">>> Found \(matches) TextView(s) with '\(label)' label")
         return matches
+    }
+
+    class func openHistory() {
+        app.navigationBars[UID.NavBar.allNotes].buttons[UID.Button.noteEditorMenu].tap()
+        guard app.tables.staticTexts[UID.Text.nodeEditorOptionsHistory].exists else {return}
+        app.tables.staticTexts[UID.Text.nodeEditorOptionsHistory ].tap()
     }
 
 }
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -187,8 +187,8 @@ class NoteEditor {
 
     class func openHistory() {
         app.navigationBars[UID.NavBar.allNotes].buttons[UID.Button.noteEditorMenu].tap()
-        guard app.tables.staticTexts[UID.Text.nodeEditorOptionsHistory].exists else {return}
-        app.tables.staticTexts[UID.Text.nodeEditorOptionsHistory ].tap()
+        guard app.tables.staticTexts[UID.Text.noteEditorOptionsHistory].exists else {return}
+        app.tables.staticTexts[UID.Text.noteEditorOptionsHistory ].tap()
     }
 
 }
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -188,7 +188,7 @@ class NoteEditor {
     class func openHistory() {
         app.buttons[UID.Button.noteEditorMenu].tap()
         let historyButton = app.tables.staticTexts[UID.Text.noteEditorOptionsHistory]
-        guard historyButton.waitForExistence(timeout: minLoadTimeout) else {return}
+        guard historyButton.waitForExistence(timeout: minLoadTimeout) else { return }
         historyButton.tap()
     }
 
--- a/SimplenoteUITests/SimplenoteUITestsHistory.swift
+++ b/SimplenoteUITests/SimplenoteUITestsHistory.swift
@@ -1,0 +1,198 @@
+import XCTest
+
+class SimplenoteUISmokeTestsHistory: XCTestCase {
+    override class func setUp() {
+        app.launch()
+        getToAllNotes()
+        let _ = attemptLogOut()
+        EmailLogin.open()
+        EmailLogin.logIn()
+    }
+
+    override func setUpWithError() throws {
+        getToAllNotes()
+        NoteList.trashAllNotes()
+        Trash.empty()
+        NoteList.openAllNotes()
+    }
+
+    override class func tearDown() {
+        NoteList.trashAllNotes()
+        Trash.empty()
+    }
+
+    func testHistoryCanBeDismissed() throws {
+        trackTest()
+        let timestamp = NSDate().timeIntervalSince1970.description
+        let noteText = "HistoryTest " + timestamp.suffix(6)
+
+        trackStep()
+        NoteList.addNoteTap()
+        NoteEditorAssert.editorShown()
+
+        trackStep()
+        NoteEditor.clearAndEnterText(enteredValue: noteText)
+        NoteEditor.openHistory()
+        HistoryAssert.historyShown()
+
+        trackStep()
+        History.close()
+        HistoryAssert.historyDismissed()
+
+        NoteEditor.leaveEditor()
+        NoteList.trashNote(noteName: noteText)
+    }
+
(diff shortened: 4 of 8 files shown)


SimplenoteUITests/UIDs.swift
```diff
@@ -65,6 +66,7 @@ enum UID {
     enum Text {
         static let noteEditorPreview = "Preview"
         static let noteEditorOptionsMarkdown = "Markdown"
+        static let nodeEditorOptionsHistory = "History"
```
pachlava: Just a typo:

```suggestion
        static let noteEditorOptionsHistory = "History"
```

SimplenoteUITests/Generic.swift
```diff
@@ -63,6 +63,11 @@ func populateNoteList(title: String, numberToCreate: Int) {
     }
 }
 
+func getRandonId() -> String.SubSequence {
+    let id = NSDate().timeIntervalSince1970.description.suffix(6)
+    return id
+}
```
pachlava: Two points here:
1. A typo in function name
2. Is there a reason to have the return value as `String.Subsequence` and not `String`, taking into account that the return value is used as a string? (please let me know if there's a context I might be missing here)

```suggestion
func getRandomId() -> String {
    let id = NSDate().timeIntervalSince1970.description.suffix(6)
    return String(id)
}
```

SimplenoteUITests/NoteEditor.swift
```diff
@@ -177,6 +185,12 @@ class NoteEditor {
         return matches
     }
 
+    class func openHistory() {
+        app.navigationBars[UID.NavBar.allNotes].buttons[UID.Button.noteEditorMenu].tap()
```
pachlava: I know this is the same in older code, so it's not technically incorrect. This is just something that we have to keep in mind when writing new code (and once we have time to refactor the old code). 

The navigation bar title depends on the current selection in note list. For example, if we open `All Notes`, the navbar title will be `All notes`. If we create a tag for a note, and open all notes tagged with a certain tag, the navbar title will be this tag, but not `All Notes`.

Because of this, hardcoding the navbar title in the button locator limits us to the usage of this method to only `All Notes`. This can be avoided just by removing the navbar part:

```suggestion
        app.buttons[UID.Button.noteEditorMenu].tap()
```

SimplenoteUITests/NoteEditor.swift
```diff
@@ -177,6 +185,12 @@ class NoteEditor {
...
+        guard app.tables.staticTexts[UID.Text.nodeEditorOptionsHistory].exists else {return}
+        app.tables.staticTexts[UID.Text.nodeEditorOptionsHistory ].tap()
```
pachlava: Just a tricky question: why there's a guard statement for `History` button, but not for the `Menu` button? 😉 

Also, if we use the control more than once in a method, I see sense in having it as a constant:

```suggestion
        let historyButton = app.tables.staticTexts[UID.Text.nodeEditorOptionsHistory]
        guard historyButton.exists else { return }
        historyButton.tap()
```

SimplenoteUITests/History.swift
```diff
@@ -0,0 +1,45 @@
...
+        let restoreNoteButton = app.buttons[UID.Button.restoreNote]
+
+        XCTAssertTrue(closeHistoryButton.waitForExistence(timeout: minLoadTimeout), UID.Button.dismissHistory + buttonNotFound)
+        XCTAssertTrue(restoreNoteButton.waitForExistence(timeout: minLoadTimeout), UID.Button.restoreNote + buttonNotFound)
+    }
+
+    class func historyDismissed() {
+        let closeHistoryButton = app.buttons[UID.Button.dismissHistory]
+
+        XCTAssertFalse(closeHistoryButton.exists, UID.Button.dismissHistory + buttonNotFound)
```
pachlava: `buttonNotFound` message is used here, while something like `buttonNotAbsent` would be correct.

Also, since we are checking the both `Dismiss` and `Restore` buttons presence in the `HistoryAssert.historyShown()` method, why not do the same here?

```suggestion
        let closeHistoryButton = app.buttons[UID.Button.dismissHistory]
        let restoreNoteButton = app.buttons[UID.Button.restoreNote]

        XCTAssertFalse(closeHistoryButton.exists, UID.Button.dismissHistory + buttonNotAbsent)
        XCTAssertFalse(restoreNoteButton.exists, UID.Button.restoreNote + buttonNotAbsent)
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        let editedNoteText = notePrefix + getRandonId()
+
+        trackStep()
+        NoteList.addNoteTap()
+        NoteEditorAssert.editorShown()
+
+        trackStep()
+        NoteEditor.clearAndEnterText(enteredValue: initialNoteText)
+        NoteEditor.waitBeforeEditingText(delayInSeconds: 3, newText: editedNoteText)
+
```
pachlava: Extra line break:
```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
```
pachlava: Extra line break:
```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        trackStep()
+        getToAllNotes()
```
pachlava: `getToAllNotes()` is used in setUp/tearDown methods, since we don't always know where the app might be. It will firstly try to close the WebView which might be shown in note preview mode, then leave the preview, and then leave the editor.

However, in this test, we do know that the way to get to all note is this one:

```suggestion
        NoteEditor.leaveEditor()
```

And this will save us some time by not trying to tap controls that we know do not exist anyway.

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        NoteList.trashAllNotes()
+        Trash.empty()
+        NoteList.openAllNotes()
+    }
+
+    override class func tearDown() {
+        NoteList.trashAllNotes()
+        Trash.empty()
+    }
```
pachlava: This instance of teardown will be called only once at the end of the whole suit. 

Currently, all the tests in this file have a sort of tearDown at their end, which is usually closing History, leaving the editor, and trashing the note. 

But, at the same time (minus the history close part), the same is done in `setUpWithError`. Because of this, I would see value in having the following as the only step of `tearDown()` (note how you can change the teaDown declaration to have it executed at the end of _each_ test):

```suggestion
    override func tearDownWithError() throws {
        History.close()
    }
```

By doing so, you can remove the 'tearDown' part in all tests.

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        NoteEditor.clearAndEnterText(enteredValue: noteText)
+        NoteEditor.openHistory()
+        HistoryAssert.historyShown()
+
+        trackStep()
+        History.close()
+        HistoryAssert.historyDismissed()
+
+        NoteEditor.leaveEditor()
+        NoteList.trashNote(noteName: noteText)
```
pachlava: See my comment for `tearDown()` method.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        NoteEditor.openHistory()
+        HistoryAssert.historyShown()
+
+        trackStep()
+        HistoryAssert.restoreButtonIsDisabled()
+
+        trackStep()
+        History.close()
+        NoteEditor.leaveEditor()
+        NoteList.trashNote(noteName: noteText)
```
pachlava: See my comment for `tearDown()` method.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        HistoryAssert.historyShown()
+
+        trackStep()
+        History.setSliderPosition(position: 0.0)
+        HistoryAssert.restoreButtonIsEnabled()
+
+        trackStep()
+        History.close()
+        NoteEditor.leaveEditor()
+        NoteList.trashNote(noteName: editedNoteText)
```
pachlava: See my comment for `tearDown()` method.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+
+        trackStep()
+        History.setSliderPosition(position: 0.0)
+        History.restoreNote()
+        NoteEditorAssert.textViewWithExactValueShownOnce(value: initialNoteText)
+
+        trackStep()
+        History.close()
+        NoteEditor.leaveEditor()
+        NoteList.trashNote(noteName: initialNoteText)
```
pachlava: See my comment for `tearDown()` method.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        NoteEditorAssert.textViewWithExactValueShownOnce(value: editedNoteText)
+
+        trackStep()
+        History.setSliderPosition(position: 0.0)
+        NoteEditorAssert.textViewWithExactValueShownOnce(value: initialNoteText)
+
+        trackStep()
+        History.close()
+        NoteEditor.leaveEditor()
+        NoteList.trashNote(noteName: initialNoteText)
```
pachlava: See my comment for `tearDown()` method.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        HistoryAssert.historyShown()
+
+        trackStep()
+        History.setSliderPosition(position: 0.0)
+        History.close()
+        NoteEditorAssert.textViewWithExactValueShownOnce(value: editedNoteText)
+
+        trackStep()
+        NoteEditor.leaveEditor()
+        NoteList.trashNote(noteName: editedNoteText)
```
pachlava: See my comment for `tearDown()` method.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+
+        trackStep()
+        History.setSliderPosition(position: 0.0)
+        History.restoreNote()
+        NoteEditorAssert.textViewWithExactValueShownOnce(value: initialNoteText)
+
+        trackStep()
+        History.close()
+        NoteEditor.leaveEditor()
+        NoteList.trashNote(noteName: initialNoteText)
```
pachlava: See my comment for `tearDown()` method.

```suggestion
```

SimplenoteUITests/History.swift
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    class func restoreNote() {
+        app.buttons[UID.Button.restoreNote].tap()
+    }
+}
+
+class HistoryAssert {
+    class func historyShown() {
+        let closeHistoryButton = app.buttons[UID.Button.dismissHistory]
```
pachlava: Nitpick (and it's only my opinion): having the button named after the string identifier makes the code easier on the eye:

```suggestion
        let dismissHistoryButton = app.buttons[UID.Button.dismissHistory]
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+
+    func testHistoryCanBeDismissed() throws {
+        trackTest()
+        let noteText = notePrefix + getRandonId()
+
+        trackStep()
+        NoteList.addNoteTap()
+        NoteEditorAssert.editorShown()
+
+        trackStep()
```
pachlava: In this test and all the others:

This assertion is quite slow because it checks for several controls presence (8 sec if I remember correctly), and it's already vastly used in `NoteEditor` tests, dedicated to note editor. 

Because of this, I think we will not lose much if we remove it here, and merge steps 1 and 2, since:
- It's already tested in the dedicated pack of tests.
- This file's main focus is History

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        NoteEditor.clearAndEnterText(enteredValue: noteText)
+        NoteEditor.openHistory()
+        HistoryAssert.historyShown()
+
+        trackStep()
+        HistoryAssert.restoreButtonIsDisabled()
+
```
pachlava: It's not something official, or accepted as a rule, it's just the way it's done for all existing tests in this project: every step contains some `action`, followed by an `assertion`.

In this case, there's an assertion in a separate step, but with no action:

```suggestion
        trackStep()
        NoteEditor.clearAndEnterText(enteredValue: noteText)
        NoteEditor.openHistory()
        HistoryAssert.historyShown()
        HistoryAssert.restoreButtonIsDisabled()

```

Similar to this, in some other tests, there are cases of steps with actions, but no assertions.

Just for some context: as you can see from the [testing checklist](https://github.com/Automattic/simplenote-ios/blob/develop/TESTING-CHECKLIST.md) for Simplenote (the only testing document we have for Simplenote), there is no concept of steps, or expected results. A single test is just a single sentence, which is quite vague.

This current structure for automated tests (multiple steps / expected results) is just a format that I was very used to because of previous experience, and it helped me to logically break down the tests. But it's not official, I'm just proposing you the change because of the consistency.

The decision is up to you 👍

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        trackStep()
+        NoteList.addNoteTap()
+        NoteEditorAssert.editorShown()
+
+        trackStep()
+        NoteEditor.clearAndEnterText(enteredValue: initialNoteText)
+        NoteEditor.waitBeforeEditingText(delayInSeconds: 3, newText: editedNoteText)
+
+        trackStep()
```
pachlava: Same sentiment as in my `steps/actions/assertions` comment.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        NoteList.addNoteTap()
+        NoteEditorAssert.editorShown()
+
+        trackStep()
+        NoteEditor.clearAndEnterText(enteredValue: initialNoteText)
+        NoteEditor.waitBeforeEditingText(delayInSeconds: 3, newText: editedNoteText)
+
+
+        trackStep()
```
pachlava: Same sentiment as in my `steps/actions/assertions` comment.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        NoteList.addNoteTap()
+        NoteEditorAssert.editorShown()
+
+        trackStep()
+        NoteEditor.clearAndEnterText(enteredValue: initialNoteText)
+        NoteEditor.waitBeforeEditingText(delayInSeconds: 3, newText: editedNoteText)
+
+        trackStep()
```
pachlava: Same sentiment as in my `steps/actions/assertions` comment.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        NoteList.addNoteTap()
+        NoteEditorAssert.editorShown()
+
+        trackStep()
+        NoteEditor.clearAndEnterText(enteredValue: initialNoteText)
+        NoteEditor.waitBeforeEditingText(delayInSeconds: 3, newText: editedNoteText)
+
+
+        trackStep()
```
pachlava: Same sentiment as in my `steps/actions/assertions` comment.

```suggestion
```

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,231 @@
...
+        NoteEditor.clearAndEnterText(enteredValue: initialNoteText)
+        NoteEditor.waitBeforeEditingText(delayInSeconds: 3, newText: editedNoteText)
+
+        trackStep()
+        getToAllNotes()
+        NoteList.trashNote(noteName: editedNoteText)
+
+        trackStep()
```
pachlava, line 157: Same sentiment as in my `steps/actions/assertions` comment.

```suggestion
```

pachlava (changes requested):
Thanks for this PR, @tiagomar 🥳 

I executed the tests on iPhone 12 Simulator, and they all worked correctly, which is the main part 👍 

I left _a lot_ of comments to the code - but don't get demoralized because of this - they are mainly related to test steps structure, and potential clean-up of the final part in every test. There are also a few about typos / naming and so on, which is nice to have addressed before the merge. Nothing critical. 🔥

pachlava:
An additional note: since all new tests get added to the UI testing schemes automatically, your new tests were added as selected for execution to both `SimplenoteUITests` and `SimplenoteUITests_Subset` schemes.

We keep `SimplenoteUITests_Subset` short (fast) on purpose - because this set is executed on our CI on every commit to any PR. There's a sort of agreement that the Subset should not take longer than 5 min, and we already are hitting this limit. A full UI test can be also executed on every commit, but it's optional (selectable by user).

Could you please exclude 6 tests only from `SimplenoteUITests_Subset` scheme, but feel free to keep one test that is most representative in your opinion 👍. All you tests will still be active in `SimplenoteUITests` scheme.

This change is the one that will result in changes in `SimplenoteUITests_Subset.xcscheme`:

https://user-images.githubusercontent.com/73365754/131338383-7b188f82-f235-4e2e-acf6-3be924a2e59a.mov

SimplenoteUITests/Generic.swift
tiagomar: Yeah, returning String was my first attempt but I've got the `Cannot convert return expression of type 'String.SubSequence' (aka 'Substring') to return type 'String'` message from Xcode and just set the return type to `String.SubSequence` instead of just converting variable. Returning `String` is definitely better.

SimplenoteUITests/SimplenoteUITestsHistory.swift
tiagomar: I thought about using the `tearDown()` running after each test to have them more isolated, however part of what I wanted to do in `tearDown()` was to delete the note created during the test but not impacting a possible parallel execution (which might be out of the scope in this project), that's why I started using ids for each note. As I couldn't find a way to pass the name of the note to be deleted in `tearDown()` I moved to having it running once after the whole class is executed and getting rid of all the remaining notes.

I like the idea of not having tearDown code within the test method, what I miss in your suggestion is actually deleting the note created during the test. I can think in two solutions:
Assumption: parallelism is not a reality
1. We can go with your suggestion but keeping an after class `tearDown()` that would actually delete all the notes.
2. Add to your suggestion extra steps to delete all the notes after each tests.

I don't really like the option 2 because deleting all notes after each test feels too costly since parallelism is already out of scope.

What are your thoughts on this?

SimplenoteUITests/NoteEditor.swift
tiagomar: Ok, let's see if I got it right. =P
The `Menu` button is in the navigation bar, so we 'know' it is there while the `History` button is not visible yet and there's some animation in place. If `History` button does not exist the guard statement will stop the `openHistory()` flow without trying to tap the button and the test will gracefully fail the `historyShown()` assertion. If we tried to tap a button that does not exist it would just break the test execution.

Looking at it now I that using `historyButton.waitForExistence(minLoadTimeout)` would be more interesting if dealing with some slow devices/simulators. Does it make sense?

SimplenoteUITests/SimplenoteUITestsHistory.swift
pachlava: The notes are already getting deleted as a part of `setUpWithError()`, which is executed before each test, this is why deleting them additionally at the end of the test or inside a tearDown means spending extra time 🤔 

Parallel execution under the same account is not the case: we have a dedicated solution for that on our CI, which manages a number of tests accounts and makes sure no account is used in parallel. While there is still a possibility for someone to run these tests locally using the test account which is hardcoded in the workspace, the chances and impact are so low, that we decided to neglect it, since this account is never used on CI.

The approach with using unique IDs for notes titles was one of the options we considered when we were solving the test accounts race condition, but a lot of existing tests rely on checking the total number of notes in `Trash` / `All Notes` etc, besides checking for the note content. Using this approach instead of having multiple accounts would mean loosing this possibility, as the number of notes cannot be predicted in this case (and this is the other reason why all notes are deleted at the beginning of each test).

In your case, using IDs in titles is a robust approach to check that `History` really reverts the revision 👍 . You are right about the fact that deleting all notes at the beginning of each test (instead of doing this only once at the beginning or end of the suite) is a bit of overkill, and you are free to adopt a different approach in the tests you've added. However, doing so for the other existing tests will cause fails, since they are written assuming that there are zero notes at the beginning of each test (except for `Search` tests).

The content for setUp/tearDown differs between the files with tests, and actually any form of tearDown is absent in the most files. You are free to have it defined the way you consider the best 🙂

SimplenoteUITests/NoteEditor.swift
pachlava: I think there is no _right_ answer, since wrapping everything in guards is a questionable time investment, and you can see that it's quite sporadic in the existing code. I was just curious about your opinion 🙂 

> Looking at it now I that using historyButton.waitForExistence(minLoadTimeout) would be more interesting if dealing with some slow devices/simulators. Does it make sense?

We can never be extra careful, so why not 🙂 In practice, the animation of note Menu is fast enough to not cause any issues even on notoriously slow CI runners (we know this because there's a very similar `toggleMarkdownState()` method that does not use waits and did not result in flakiness).

SimplenoteUITests/SimplenoteUITestsHistory.swift
tiagomar: You're right, it makes total sense. I was kind of ignoring the `tearDown()` and `setup()` overlap.  I'll apply the changes. =)

SimplenoteUITests/SimplenoteUITestsHistory.swift
```diff
@@ -0,0 +1,176 @@
...
+        NoteEditor.openHistory()
+        HistoryAssert.historyShown()
+
+        trackStep()
+        History.setSliderPosition(position: 0.0)
+        History.restoreNote()
+        NoteEditorAssert.textViewWithExactValueShownOnce(value: initialNoteText)
+    }
+
```
pachlava: An extra line break here

```suggestion
```

SimplenoteUITests/History.swift
```diff
@@ -0,0 +1,47 @@
...
+        XCTAssertTrue(restoreNoteButton.waitForExistence(timeout: minLoadTimeout), UID.Button.restoreNote + buttonNotFound)
+    }
+
+    class func historyDismissed() {
+        let dismissHistoryButton = app.buttons[UID.Button.dismissHistory]
+        let restoreNoteButton = app.buttons[UID.Button.restoreNote]
+
+        XCTAssertFalse(dismissHistoryButton.exists, UID.Button.dismissHistory + buttonNotAbsent)
+        XCTAssertFalse(restoreNoteButton.exists, UID.Button.restoreNote + buttonNotAbsent)
+    }
```
pachlava: (This is something really extra, feel free to ignore it.)

Since both methods are quite similar, we can apply the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle here, and do something like:

```swift
    class func historyShown() {
        HistoryAssert.historyExistence(shouldExist: true, errorMessage: buttonNotFound)
    }

    class func historyDismissed() {
        HistoryAssert.historyExistence(shouldExist: false, errorMessage: buttonNotAbsent)
    }

    class func historyExistence(shouldExist: Bool, errorMessage: String) {
        let dismissHistoryButton = app.buttons[UID.Button.dismissHistory]
        let restoreNoteButton = app.buttons[UID.Button.restoreNote]

        XCTAssertEqual(dismissHistoryButton.waitForExistence(timeout: minLoadTimeout),
                       shouldExist,
                       UID.Button.dismissHistory + errorMessage)
        XCTAssertEqual(restoreNoteButton.waitForExistence(timeout: minLoadTimeout),
                       shouldExist,
                       UID.Button.restoreNote + errorMessage)
    }
```

pachlava (approved):
Thanks for addressing my comments, and for the thoughtful approach @tiagomar !

I left a few really minor notes. The tests can go live 🚀 

Thank you 👍

SimplenoteUITests/NoteEditor.swift
```diff
@@ -177,6 +185,13 @@ class NoteEditor {
         return matches
     }
 
+    class func openHistory() {
+        app.buttons[UID.Button.noteEditorMenu].tap()
+        let historyButton = app.tables.staticTexts[UID.Text.noteEditorOptionsHistory]
+        guard historyButton.waitForExistence(timeout: minLoadTimeout) else {return}
```
pachlava: Just in order to keep it along with the existing styling in the workspace:

```suggestion
        guard historyButton.waitForExistence(timeout: minLoadTimeout) else { return }
```

SimplenoteUITests/History.swift
tiagomar: This is great, I'm changing it.

PR #1422, opened 2021-08-31
Updated check for verification required body response

charliescheer:
### Fix
This is a quick PR to update SPAuthError key for verification required.

Previously the expected response that had been added to the error was not correct.  It was `requires verification` which is not the correct key.  I have updated it to `verification required` 

(also fixed a type I found in one of the other error descriptions)

### Test
1. Setup some sort of proxy system like Charles to interrupt responses from the auth endpoint
2. modify the response to be a 403 error with a body of `verification required` 
3. You should see an alert with instructions for resending a verification

### Review

### Release


eshurakov (approved):
I think we should target `release/4.43` to jump on release train, otherwise we will have a version of the app live that listens to the wrong error message.

PR #1432, opened 2021-09-13
readme updates

charliescheer:
### Fix
This PR updates the OSS instructions for Simplenote iOS to be a bit clearer and help with some potential errors that may happen

I have updated:
- The swiftLint instructions
- The initial build instructions
- some common errors with setting up and how to fix them

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,10 @@ You can also open the project by double clicking on Simplenote.xcworkspace file,
 
 Once you have opened Simpleonte iOS in Xcode, depending on your setup, you may need to make a few changes before you can build the app.  In Xcode hit `Command + B` and see if you get any errors.  
 
+## Known Xcode Issues
+
+There are some common errors that can happen when first getting Simplenote iOS setup.  Check these instructions to see how to fix those.
+
 If you see `The server SSH fingerprint failed to verify` before you can build Simplenote you will need to mark the app as trusted.  To do this, tap on the warning and hit Trust
 
 If the build fails with an `Authentication failed because the credentials were missing` error most likely you need to update the Swift Packages before being able to build the app. We use Swift Package Manager for some internal dependencies which can be found on Github.  To fetch these packages, connect Xcode to Github by going to Xcode > Preferences > Accounts, then enter your Github account details.  To be able to fetch these dependencies Xcode will need to be connected to a Github account via [SSH](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)
--- a/README.md
+++ b/README.md
@@ -50,9 +50,9 @@ You can also open the project by double clicking on Simplenote.xcworkspace file,
 
 Once you have opened Simpleonte iOS in Xcode, depending on your setup, you may need to make a few changes before you can build the app.  In Xcode hit `Command + B` and see if you get any errors.  
 
-If you see (go do the setup again and get the error) before you can build Simplenote you will need to mark the app as trusted.  To do this, tap on the warning and hit Trust
+If you see `The server SSH fingerprint failed to verify` before you can build Simplenote you will need to mark the app as trusted.  To do this, tap on the warning and hit Trust
 
-If the build fails with an "Account Creation Failed" error and Xcode asks for your GitHub credentials most likely you need to update the Swift Packages before being able to build the app. We use Swift Package Manager for some internal dependencies which can be found on Github.  To fetch these packages, connect Xcode to Github by going to Xcode > Preferences > Accounts, then enter your Github account details.  To be able to fetch these dependencies Xcode will need to be connected to a Github account via [SSH](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,9 +30,7 @@ This is the recommended way to include SwiftLint in your workflow, as it catches
 
 Alternately, a SwiftLint scheme is exposed within the project; Xcode will show a warning if you don't have SwiftLint installed.
 
-Finally, you can also run SwiftLint manually from the command line with:
-
-`rake lint`
+SwiftLint is connected to Simplenote iOS, so lint errors will appear as warnings in Xcode after you build the project.
```
jleandroperez: How about something in the lines of:

```
SwiftLint is integrated directly into the Xcode project, so lint errors appear as warnings after you build the project
```

SwiftLint connected to Simplenote iOS feels like it's doing something with the actual service maybe?

README.md
```diff
@@ -50,24 +48,33 @@ This will ensure any dependencies are ready before launching Xcode.
 
 You can also open the project by double clicking on Simplenote.xcworkspace file, or launching Xcode and choose `File` > `Open` and browse to `Simplenote.xcworkspace`.
 
+Once you have opened Simpleonte iOS in Xcode, depending on your setup, you may need to make a few changes before you can build the app.  In Xcode hit `Command + B` and see if you get any errors.  
+
+If you see `The server SSH fingerprint failed to verify` before you can build Simplenote you will need to mark the app as trusted.  To do this, tap on the warning and hit Trust
```
jleandroperez, line 57: Where would this error show up? In the device itself?

README.md
```diff
@@ -50,24 +48,33 @@ This will ensure any dependencies are ready before launching Xcode.
...
+
+If the build fails with an `Authentication failed because the credentials were missing` error most likely you need to update the Swift Packages before being able to build the app. We use Swift Package Manager for some internal dependencies which can be found on Github.  To fetch these packages, connect Xcode to Github by going to Xcode > Preferences > Accounts, then enter your Github account details.  To be able to fetch these dependencies Xcode will need to be connected to a Github account via [SSH](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)
```
jleandroperez, line 59: Is this error related to Xcode not able to download the GH dependencies?

README.md
```diff
@@ -50,24 +48,33 @@ This will ensure any dependencies are ready before launching Xcode.
...
 ## Setup Credentials
 
-Simplenote is powered by the [Simperium Sync'ing protocol](https://www.simperium.com). We distribute **testing credentials** that help us authenticate your application, and verify that the API calls being made are valid. Once the Simperium account is created, you can register an app and access the APP ID and necessary API keys.
+Simplenote is powered by the [Simperium Sync'ing protocol](https://www.simperium.com). To be able to connect a development build of Simplenote iOS to the Simperium syncing service you will first need to setup app credentials.
+We distribute **testing credentials** that help us authenticate your application, and verify that the API calls being made are valid. 
 
 **⚠️ Please note → We're not accepting any new Simperium accounts at this time.**
+First you will need to setup an account with [Simperium.com](https://www.simperium.com) 
+
+After creating your account, login and click on `Add an App`, give the app a name, and you will be taken to your test apps details page.  From there you need to copy the `App ID` and the `Default API Key`
```
jleandroperez: I've reworded this section in the macOS app ([PR here](https://github.com/Automattic/simplenote-macos/pull/1035/files)).

I'd suggest dropping these two sections, since signup in simperium.com is not allowed (and we're not sure if it'll ever be, back again). WDYT?

jleandroperez (commented):
Sending few questions! WDYT?

README.md
charliescheer: It shows up in Xcode if you have not verified the project

charliescheer: And then stops you from building the app or downloading the SPM packages

README.md
charliescheer: Both of these errors are related to fetching the GH dependencies

README.md
charliescheer: I had thought possibly the same thing.  But I think we should say `something` cause if a non a8c OSS found simplenote and wanted to contribute there would be no details on the readme about why they can't login.

README.md
charliescheer: Sounds good, have made that change

README.md
jleandroperez: Can we confirm this still happens with Xcode 13? (since that was an Xcode glitch AFAIK?).

If not, we could probably drop it!

jleandroperez (approved):
:shipit:

charliescheer:
>Can we confirm this still happens with Xcode 13? (since that was an Xcode glitch AFAIK?).

I will give it a shot and see if the same thing happens

charliescheer:
>Can we confirm this still happens with Xcode 13? (since that was an Xcode glitch AFAIK?).
If not, we could probably drop it!

I tested setting up a new repo on a new account with Xcode 13 and both of these errors do appear still.  I suspect we wouldn't see them with a new repo because we already have our GitHub accounts connected to Xcode and have already marked a8c as a verified SSH fingerprint.  But a new contributor may see these things.  

I think that the warnings that appear are vague enough and could happen with any new install of SNiOS, I think we should keep these instructions in. WDYT?

jleandroperez (approved):
Sounds good @charliescheer !! Only note I've got is, if possible, I'd just keep that under a section called `Known Xcode Issues` (or something like that), since it's an Xcode problem, and not necessarily part of the project's setup..

:shipit: at will sir!

PR #1408, opened 2021-08-25
Update Intents "Distribution AppStore" configuration to match other extensions

mokagio:
### Fix
While checking what discussed [here](https://github.com/Automattic/simplenote-ios/pull/1375#discussion_r693999625), I noticed that the entitlements for the "Distribution AppStore" build configuration was pointing to the Alpha build. That was inconsistent with the other apps, but most importantly incorrect: "Distribution AppStore" is the build configuration we use to App Store builds and needs to point to a valid app group (the Alpha one exists only in the Enterprise account).

### Test
In Xcode, verify the "Signing & Capabilities > App Groups" are consistent between extensions

**Share**

Screen Shot 2021-08-25 at 3 10 49 pm

**Widgets**

Screen Shot 2021-08-25 at 3 18 23 pm

**Intents**

Screen Shot 2021-08-25 at 3 11 06 pm

### Review
Only one developer required to review these changes, but anyone can perform the review.

### Release
These changes do not require release notes.

PR #1436, opened 2021-09-21
Rename "Supporting Files" to "Supporting-Files"

mokagio:
### Fix

[`e6a2a16` (#1435)](https://github.com/Automattic/simplenote-ios/pull/1435/commits/e6a2a168a33c72be2050dffc9fcc65144655eac5) deleted the entire `en.lproj/Localizable.strings` file because the `find` command in `localize.py` returned a path with a space in it, `./Simplenote/Supporting Files/main.m`. When parsing that space,
`genstring` failed:

```
genstrings: error: failed to read input file ./Simplenote/Supporting
```

I haven't dug deeper, but I'm guessing at some point there's a generic "stage all changes in the folder" command for Git, which would have resulted in the file being removed.

That never happened before because the "Supporting Files" folder was added as part of the recent https://github.com/Automattic/simplenote-ios/pull/1408.

This PR removes the spaces from the directory name to avoid this kind of issue with `find` and other scripts. I like this approach because, as the issue we just encountered prooves, _spaces in paths are dangerous_! On the other hand, "Supporting Files" (with a space) is a standard name in Xcode projects, so I wonder if this change makes the project setup too "odd". 🤔 

## To test

[Compare](https://github.com/Automattic/simplenote-ios/compare/develop...release/4.45-fix-string-by-removing-spaces) the changes generated with this fix with `develop` and notice the `en.lproj/Localizable.strings` diff.


mokagio:
Closing in favor of the implementation from #1437.

PR #1437, opened 2021-09-21
Update `find` usage in `localize.py` script to account for spaces in paths

mokagio:
### Fix

[`e6a2a16` (#1435)](https://github.com/Automattic/simplenote-ios/pull/1435/commits/e6a2a168a33c72be2050dffc9fcc65144655eac5) deleted the entire `en.lproj/Localizable.strings` file because the `find` command in `localize.py` returned a path with a space in it, `./Simplenote/Supporting Files/main.m`. When parsing that space,
`genstring` failed:

```
genstrings: error: failed to read input file ./Simplenote/Supporting
```

I haven't dug deeper, but I'm guessing at some point there's a generic "stage all changes in the folder" command for Git, which would have resulted in the file being removed.

That never happened before because the "Supporting Files" folder was added as part of the recent https://github.com/Automattic/simplenote-ios/pull/1408.

This PR fixes the issue by updating the command in `localize.py` to be more robust against spaces and other paths oddities. This approach is cleaner than https://github.com/Automattic/simplenote-ios/pull/1436, but doesn't fix the issue at the source. Now, one could argue that scripts should handle spaces, so this is actually a good fix, also because "Supporting Files" (with the space) is a standard name in an Xcode project.

## To test

[Compare](https://github.com/Automattic/simplenote-ios/compare/develop...release/4.45-fix-string-in-find-cmd) the changes generated with this fix with `develop` and notice the `en.lproj/Localizable.strings` diff

### DIFF
--- a/Scripts/localize.py
+++ b/Scripts/localize.py
@@ -122,7 +122,7 @@ def localize(path):
     new = original + '.new'
 
     # Using with `-print0` and `xargs -0` to account for spaces in the paths `find` might return
-    genstrings_cmd = 'find ./Simplenote ./SimplenoteShare ./SimplenoteWidgets -name "*.m" -o -name "*.swift" -print0 | xargs -0 genstrings -q -o "%s"'
+    genstrings_cmd = 'find ./Simplenote ./SimplenoteShare ./SimplenoteWidgets \( -name "*.m" -o -name "*.swift" \) -print0 | xargs -0 genstrings -q -o "%s"'
 
     if os.path.isfile(original):
         os.rename(original, old)
--- a/Scripts/localize.py
+++ b/Scripts/localize.py
@@ -121,7 +121,8 @@ def localize(path):
     old = original + '.old'
     new = original + '.new'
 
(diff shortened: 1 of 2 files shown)


AliSoftware:
@mokagio I took some time to dig into this and trying to understand why this solution generated such a big diff in the strings, with so many deleted/missing.

After some fiddling I found out that while without the `print0` the list of files were as expected (and the same as for the other altermative fix in the other PR), with `print0` it only included the `.swift` files, not the `.m` ones. That's when I realized that the `-print0` has higher priority than `-o`, so the whole expression was interpreted as "if .m do nothing, or if .swift then -print0".

As soon as I fixed it using parentheses (to apply the `-print0` to the whole expression). the diff on the generated `.lproj` came back to be the same as the one on the other PR (you can check using the "Compare" link in this PR description.

So in conclusion, now that this issue about missing strings is fixed, I think this PR is a better choice than the other one at fixing the source issue.

AliSoftware (approved):

mokagio:
Thank you for taking the time to work on this @AliSoftware 🙌

PR #1441, opened 2021-09-21
Updated text for Compromised password alert

charliescheer:
### Fix
This is a quick PR to clarify the text that appears when a user's password has been found to be compromised.  This text should be clearer and imply more urgency around changing that password

internal ref: p1632202617005500-slack-C01KRRK0MLJ

### Test
1. setup a tool like charles proxy to interrupt web auth traffic and set a breakpoint on the auth endpoint
2. attempt to log into the app with a bad username/password combe
2. edit the response from the auth endpoint to 401 with a body of "compromised password"
3. the warning will appear

### Review

### Release

### DIFF
--- a/Simplenote/Classes/SPAuthError.swift
+++ b/Simplenote/Classes/SPAuthError.swift
@@ -80,7 +80,7 @@ extension SPAuthError {
         case .network:
             return NSLocalizedString("The network could not be reached.", comment: "Error when the network is inaccessible")
         case .compromisedPassword:
-            return NSLocalizedString("It is recommended that you change your password immediately. To protect your data, you'll need to update your password before being able to log in again.", comment: "error for compromised password")
+            return NSLocalizedString("This password has appeared in a data breach, which puts your account at high risk of compromise. To protect your data, you'll need to update your password before being able to log in again.", comment: "error for compromised password")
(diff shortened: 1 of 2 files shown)


eshurakov (approved):
Looks good, thank you! 
:shipit:

ghost (approved):

SylvesterWilmott:
Thanks @charliescheer 

To clarify, the text here should be appended to the existing text in the dialog. So in it's entirety the string should be:

`This password has appeared in a data breach, which puts your account at high risk of compromise. It is recommended that you change your password immediately. To protect your data, you'll need to update your password before being able to log in again.`

And instead of `Dismiss` for the button label, let's use `Cancel`.

SylvesterWilmott:
Now that I see the whole string written out, we could probably shorten it to:

`This password has appeared in a data breach, which puts your account at high risk of compromise. To protect your data, you'll need to update your password before being able to log in again.`

But I'll let you make the final call on that one.

charliescheer:
Thanks @SylvesterWilmott I have updated the copy to these new suggestions



If that looks better I will go ahead and merge this

SylvesterWilmott:
:shipit:

PR #1443, opened 2021-09-22
Update some localized string definitions to correct the English rendering

mokagio:
After the localization tooling improvement from #1437, I noticed a few localized values changed in a way that would have resulted in an incorrect representation for the English locale. E.g.:

Screen Shot 2021-09-22 at 11 13 58 am

I believe this is an issue we always had, but that was revealed by the change in how `genstrings` run.

I updated the strings definition in the source to the best of my understanding. The resulting `en.lproj/Localizable.strings` might not be perfect from a copy writing point of view, but at least has human-readable English words and sentences. 😅

I think we should merge this initial version I'm proposing to unblock the 4.45 code freeze, and proceed with copy writing updates where needed against `develop` to ship with the next version.

### Test
Nothing to test, just take a look at the `.strings` diff.

### Review

### Release

### DIFF
--- a/Simplenote/Classes/NotesListFilter.swift
+++ b/Simplenote/Classes/NotesListFilter.swift
@@ -44,7 +44,7 @@ extension NotesListFilter {
         case .everything:
             return NSLocalizedString("All Notes", comment: "Title: No filters applied")
         case .deleted:
-            return NSLocalizedString("Trash-noun", comment: "Title: Trash Tag is selected")
+            return NSLocalizedString("Trash", comment: "Title: Trash Tag is selected")
         case .untagged:
             return NSLocalizedString("Untagged", comment: "Title: Untagged Notes are onscreen")
         case .tag(let name):
--- a/Simplenote/Classes/SPAddCollaboratorsViewController.m
+++ b/Simplenote/Classes/SPAddCollaboratorsViewController.m
@@ -145,7 +145,11 @@
 - (NSString *)tableView:(UITableView *)tableView titleForFooterInSection:(NSInteger)section
 {
     if ([tableView isEqual:primaryTableView]) {
-        return self.dataSource.count > 0 ? nil : NSLocalizedString(@"collaborators-description", nil);
+        if (self.dataSource.count > 0) {
+            return nil;
+        } else {
+            return NSLocalizedString(@"Add an email address to share this note with someone. Then you can both make changes to it.", @"Description text for the screen that allows users to share notes with other users");
+        }
     }
 
     return [super tableView:tableView titleForFooterInSection:section];
--- a/Simplenote/Classes/SPTagEntryField.m
+++ b/Simplenote/Classes/SPTagEntryField.m
@@ -29,7 +29,7 @@ CGFloat const TagEntryFieldPadding = 40;
 
 
         self.accessibilityLabel = NSLocalizedString(@"Add tag", @"Label on button to add a new tag to a note");
-        self.accessibilityHint = NSLocalizedString(@"tag-add-accessibility-hint", @"Accessibility hint for adding a tag to a note");
+        self.accessibilityHint = NSLocalizedString(@"Add a tag to the current note", @"Accessibility hint for adding a tag to a note");
 
         [self sizeField];
 
--- a/Simplenote/Classes/SPTagEntryField.m
+++ b/Simplenote/Classes/SPTagEntryField.m
@@ -22,12 +22,11 @@ CGFloat const TagEntryFieldPadding = 40;
         self.textColor = [UIColor simplenoteTagViewTextColor];
         self.placeholdTextColor = [UIColor simplenoteTagViewPlaceholderColor];
         self.textAlignment = NSTextAlignmentNatural;
-        self.placeholder = NSLocalizedString(@"Add a tag...", nil);
+        self.placeholder = NSLocalizedString(@"Tag...", @"Placeholder text in textfield when adding a new tag to a note");
         self.returnKeyType = UIReturnKeyNext;
         self.autocorrectionType = UITextAutocorrectionTypeNo;
         self.autocapitalizationType = UITextAutocapitalizationTypeNone;
 
-        [self setPlaceholder:NSLocalizedString(@"Add a tag...", @"Placeholder test in textfield when adding a new tag to a note")];
 
         self.accessibilityLabel = NSLocalizedString(@"Add tag", @"Label on button to add a new tag to a note");
         self.accessibilityHint = NSLocalizedString(@"tag-add-accessibility-hint", @"Accessibility hint for adding a tag to a note");
@@ -43,10 +42,10 @@ CGFloat const TagEntryFieldPadding = 40;
 - (void)setText:(NSString *)text
 {
     [super setText:text];
-    
+
     // size field appropriately
     [self sizeField];
-    
+
     if ([self.tagDelegate respondsToSelector:@selector(tagEntryFieldDidChange:)]) {
         [self.tagDelegate tagEntryFieldDidChange:self];
     }
@@ -55,12 +54,12 @@ CGFloat const TagEntryFieldPadding = 40;
 - (void)sizeField
 {
     [self sizeToFit];
-    
+
     CGRect frame = self.frame;
-    
+
     frame.size.width += 2 * TagEntryFieldPadding;
     frame.size.height = self.frame.size.height;
-    
+
     self.frame = frame;
 }
 
--- a/Simplenote/TagListViewController.swift
+++ b/Simplenote/TagListViewController.swift
@@ -860,7 +860,7 @@ private struct Localization {
     static let done = NSLocalizedString("Done", comment: "Done editing tags")
 
     static let allNotes = NSLocalizedString("All Notes", comment: "All Notes filter button")
-    static let trash = NSLocalizedString("Trash-noun", comment: "Trash filter button")
+    static let trash = NSLocalizedString("Trash", comment: "Trash filter button")
     static let settings = NSLocalizedString("Settings", comment: "Settings button")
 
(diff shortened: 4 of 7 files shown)


Simplenote/Classes/SPAddCollaboratorsViewController.m
```diff
@@ -145,7 +145,11 @@ - (NSString *)tableView:(UITableView *)tableView titleForHeaderInSection:(NSInte
 - (NSString *)tableView:(UITableView *)tableView titleForFooterInSection:(NSInteger)section
 {
     if ([tableView isEqual:primaryTableView]) {
-        return self.dataSource.count > 0 ? nil : NSLocalizedString(@"collaborators-description", nil);
+        if (self.dataSource.count > 0) {
+            return nil;
+        } else {
+            return NSLocalizedString(@"Add an email address to share this note with someone. Then you can both make changes to it.", @"Description text for the screen that allows users to share notes with other users");
+        }
```
mokagio, line 152: I removed the ternary operator because the line was already quite long.

Simplenote/Classes/NotesListFilter.swift
```diff
@@ -44,7 +44,7 @@ extension NotesListFilter {
         case .everything:
             return NSLocalizedString("All Notes", comment: "Title: No filters applied")
         case .deleted:
-            return NSLocalizedString("Trash-noun", comment: "Title: Trash Tag is selected")
+            return NSLocalizedString("Trash", comment: "Title: Trash Tag is selected")
```
mokagio, line 47: I picked this and all the other values from the translations that were set in the `.strings` file itself before the tooling fix. 

You can verify them from [this compare view](https://github.com/Automattic/simplenote-ios/compare/develop...3025190a3fe031a5bc2c34d5fe94b4e782fef021#diff-b008231bd522825760f4af589e37ded40359d336760494c92cfe707ef0417429).

Simplenote/Classes/SPTagEntryField.m
```diff
@@ -22,15 +22,14 @@ - (instancetype)init
         self.textColor = [UIColor simplenoteTagViewTextColor];
         self.placeholdTextColor = [UIColor simplenoteTagViewPlaceholderColor];
         self.textAlignment = NSTextAlignmentNatural;
-        self.placeholder = NSLocalizedString(@"Add a tag...", nil);
+        self.placeholder = NSLocalizedString(@"Tag...", @"Placeholder text in textfield when adding a new tag to a note");
         self.returnKeyType = UIReturnKeyNext;
         self.autocorrectionType = UITextAutocorrectionTypeNo;
         self.autocapitalizationType = UITextAutocapitalizationTypeNone;
 
-        [self setPlaceholder:NSLocalizedString(@"Add a tag...", @"Placeholder test in textfield when adding a new tag to a note")];
```
mokagio, line 30: I took the time to remove this redundant double assignment, and to fix the "test" vs "text" typo 💅

For this change in particular, I took extra time to verify the version we get at runtime here (left) matches what we have currently on the App Store (right) 

<table>
<tr>
	<td>
Simulator Screen Shot - iPod touch (7th generation) - 2021-09-22 at 12 16 08
</td>
	<td>
IMG_E59A16748543-1

</td>
</table>

AliSoftware: I appreciate that you took the time to check this 👍 

~But I'm still wondering if we shouldn't go with "Add a tag…" for this string anyway, since that's what was in the original code and thus feels like this was the intent all along?~ [EDIT: Maybe not, as "Tag..." is what was in the `.strings` file even from the beginning, see below 🤔 ]

Note that we should use `…` (ellipsis character) and not `...` (3 dots) in this text. See [this comment](https://github.com/Automattic/simplenote-ios/pull/1443/files#r713803804) for the GH commit-able suggestion.

---

To be honest, I am not really sure how we ended up with `"Tag..."` copy at runtime given this `NSLocalizedString` key. This feels very odd.

I mean, the reason for it is [because `Simplenote/en.lproj/Localizable.strings` contains `"Add a tag..." = "Tag...";`](https://github.com/Automattic/simplenote-ios/blob/develop/Simplenote/en.lproj/Localizable.strings#L78) – and the translation in all other locales also match "Tag" not "Add a tag".

But this shouldn't be possible given our tooling is supposed to overwrite this `en.lproj` during the `genstrings` call, right? Yet, if I `git blame` the `en.lproj/Localizable.strings` this line goes all the way back from when the `localize.py` script was first introduced in https://github.com/Automattic/simplenote-ios/pull/535/files#diff-b008231bd522825760f4af589e37ded40359d336760494c92cfe707ef0417429R54 … so it seems it was always that way, while with the process and scripts we have this is never supposed to happen for the base locale, right? So… I'm a bit puzzled there.

Simplenote/TagListViewController.swift
```diff
@@ -860,7 +860,7 @@ private struct Localization {
     static let done = NSLocalizedString("Done", comment: "Done editing tags")
 
     static let allNotes = NSLocalizedString("All Notes", comment: "All Notes filter button")
-    static let trash = NSLocalizedString("Trash-noun", comment: "Trash filter button")
+    static let trash = NSLocalizedString("Trash", comment: "Trash filter button")
```
AliSoftware, line 863:
```suggestion
    static let trash = NSLocalizedString("Trash", comment: "Trash filter button (noun)")
```

There might be an issue with this one, as it seems like `NSLocalizedString("Trash", "Trash filter button")` is also used elsewhere in the code in a different context, at least based on the fact that `genstrings` [aggregates not only the comment for this one, but for another as well ("Title: Trash Tag is selected") in the generated `.strings` file](https://github.com/Automattic/simplenote-ios/pull/1443/files#diff-b008231bd522825760f4af589e37ded40359d336760494c92cfe707ef0417429R822-R823).

So we're using the same English term for a noun here and a title elsewhere, while in different locales the term for those two contexts might be different.

I don't have a solution for this though 🤷‍♂️  (that's a long-known issue of our process and way we handle localization in iOS with English copy as the key… which is something I wish we could change, especially to avoid this kind of issues, but it would be quite a big change and it's probably too late at this point anyway). The only hack I can think of to distinguish them is to add a dummy special character (or even just a space, assuming they're not trimmed by GlotPress?) in the key/text, and remove it at runtime… but this would feel quite dirty…

AliSoftware (approved):
Approving to keep things moving and because I don't want to hold up the release and translation process for too long, but I still think those two points are worth checking and looking into (either in this PR if we can, or in a separate one if we don't want to hold things up)

Simplenote/Classes/SPTagEntryField.m
```diff
@@ -22,15 +22,14 @@ - (instancetype)init
         self.textColor = [UIColor simplenoteTagViewTextColor];
         self.placeholdTextColor = [UIColor simplenoteTagViewPlaceholderColor];
         self.textAlignment = NSTextAlignmentNatural;
-        self.placeholder = NSLocalizedString(@"Add a tag...", nil);
+        self.placeholder = NSLocalizedString(@"Tag...", @"Placeholder text in textfield when adding a new tag to a note");
```
AliSoftware, line 25: If you want to keep "Tag", but fix the use of proper ellipsis character:
```suggestion
        self.placeholder = NSLocalizedString(@"Tag…", @"Placeholder text in textfield when adding a new tag to a note");
```

If you want to go with [my rationale](https://github.com/Automattic/simplenote-ios/pull/1443/files#r713792664) (since [current `develop` uses "Add a tag" in both `self.placeholder` and `[self setPlaceholder:]` calls](https://github.com/Automattic/simplenote-ios/blob/develop/Simplenote/Classes/SPTagEntryField.m#L25-L30) anyway)

```suggestion
        self.placeholder = NSLocalizedString(@"Add a tag…", @"Placeholder text in textfield when adding a new tag to a note");
```

Simplenote/TagListViewController.swift
AliSoftware: EDIT: btw for now I think it's ok to go with this (even if this means 2 different contexts for the same string) for now.
I'm hoping to address that kind of edge case with some refactoring of our tooling when I get to tackle our Localization Project (ref: paaHJt-sd-p2) soon.

_(TL;DR the likely idea being to separate the source copies in `Base.lproj` and have `en.lproj` be an actual locale in GlotPress, so that we can have a key (in `NSLocalizedString` code and in `Base.lproj` extracted by `genstrings`) that is slightly different from the English copy when it is really needed. At least that's the idea that I'm likely to propose and write a RFC about when I get to that)_

oguzkocer:
Since @AliSoftware gave a 👍, I'll proceed with merging this PR so that translation process can start sooner rather than later. @mokagio Please feel free to open a follow up PR if you'd like to make any more changes. Thanks again for working on this fix and thank you @AliSoftware for your review! 🙇‍♂️

Simplenote/Classes/SPTagEntryField.m
mokagio: > So… I'm a bit puzzled there.

So am I. I went through the same blame exercise while working on this.

I had a theory that could have been due to the fact that there were two instances of an `NSLocalizedString` using "Add a tag..." and that `genstrings` sort of didn't want to touch the original value then. I tested it by removing the duplication, but the value in the generated `.strings` file did not change. See 82f3e90c65e5f9fa90e483b4259cfe628e8b723c.

I think there's diminishing returns in looking into this at this point. But I think we should update the command in the other apps so they all use `xargs`, just in case there are other issues like this hiding from us in those codebases.

AliSoftware: This python script is gonna be moved in Ruby in our toolkit very soon anyway as I plan to work on our "localization toolchain improvements" project soon anyway (it's my next priority after my support rotation and after finishing my current project) and this migration will be part of it.

PR #1448, opened 2021-09-24
Use ellipsis (`…`) instead of three dots in "Tag…" localized string

mokagio:
This is a followup on the discussion from #1443. Hat tip @AliSoftware for the [suggestion](https://github.com/Automattic/simplenote-ios/pull/1443#discussion_r713792664).

### Test
Nothing to test, when we'll run the code freeze automation for the next version, this new value will be used in the generated `.strings` file.

### Review
Only one developer are required to review these changes, but anyone can perform the review.

### Release


Simplenote/Classes/SPTagEntryField.m
```diff
@@ -22,7 +22,7 @@ - (instancetype)init
         self.textColor = [UIColor simplenoteTagViewTextColor];
         self.placeholdTextColor = [UIColor simplenoteTagViewPlaceholderColor];
         self.textAlignment = NSTextAlignmentNatural;
-        self.placeholder = NSLocalizedString(@"Tag...", @"Placeholder text in textfield when adding a new tag to a note");
+        self.placeholder = NSLocalizedString(@"Tag…", @"Placeholder text in textfield when adding a new tag to a note");
```
mokagio, line 25: Alternatively, we could revert back to the text that was in the code originally:

```suggestion
        self.placeholder = NSLocalizedString(@"Add a tag…", @"Placeholder text in textfield when adding a new tag to a note");
```

I updated the code to "Tag" in the PR against the release branch just so we'd keep consistent behavior with the App Store build. The release branch is not the place where to make non-bugfix changes.

My rational for keeping "Tag" would be that if someone updated the `.strings` file they must have wanted the copy to be that. From a UX perspective, I think both "Add a tag…" and "Tag…" get to the point of what tapping that bit of UI would do. But I don't have a strong opinion on this.

@SylvesterWilmott what's your two cents?

mokagio: I enabled auto-merge on this PR. If you are reviewing and reading this and feel that "Add a tag…" would be a better option, feel free to apply the suggestion above and approve 👍

AliSoftware (commented):
I'm 👍 with the change but will wait for Sylvester to come in on "Tag…" vs "Add a tag…" before approve-and-auto-merge the PR.

Simplenote/Classes/SPTagEntryField.m
SylvesterWilmott: Thanks @mokagio 

`Tag...` looks good to me!

pachlava:
Hey @mokagio ! Could I please ask you to change `...` to `…` in the similar [UID](https://github.com/Automattic/simplenote-ios/blob/9bce25ed187e45b4a1121d4d84317af7de27728a/SimplenoteUITests/UIDs.swift#L78) for UI tests too? So that we will not need to create a separate PR for this. After this, the `Search` tests from `Full Test` should start passing.

mokagio:
Thanks @pachlava. I should have stuck around to see what the tests did on GitHub, but hey... this is why we run them in CI in the first place 💪 

I addressed the request, and CI validates that the tests pass ✅ . Mind leaving an approval so auto-merge can get this in?

Thanks!

pachlava (approved):
Thanks a lot @mokagio for helping me with the UI part 🙂 

I had no chance to run the optional Full UI test, because CircleCI is in a partial outage at the moment.

Giving an approval, since I know that with this change the tests will pass (tested locally).

PR #1450, opened 2021-09-27
Dropping iOS 12 Support

jleandroperez:
### Description
In this PR we're bumping up the deployment target to iOS 13, and fixing a ton of warnings along the way.

cc @charliescheer (Thanks in advance!!)

### Details
Noteworthy Changes:

- We're switching the `Widget Intents` not to codegen in the main app's target, since that was causing +180 duplication warnings. Details in the `WidgetController` class.
- I've patched up the `copy-secrets.sh` script, to prevent Copying whenever both files match. Xcode was failing (A LOT) because it was picking up changes in `SPCredentials.swift`

### Test
Please run a smoke test over the app, and verify everything looks as expected:

- [ ] Authentication spinner
- [ ] Logout spinner
- [ ] Shake animation (when Pinlock is incorrect)
- [ ] NavigationBar Backgrounds

Please also make sure to test the app in iOS 13 / 14 / 15!!

### Release
`RELEASE-NOTES.txt` was updated in 9be96977 with:

> Drops iOS 12 support #1450


Simplenote/Classes/UIView+Animations.swift
```diff
@@ -142,9 +142,10 @@ extension UIView {
         let originalTransform = transform
 
         transform = leftTranslation
-        UIView.animate(withDuration: 0.07, delay: 0.0, options: [.autoreverse, .repeat]) {
-            UIView.setAnimationRepeatCount(5)
-            self.transform = rightTranslation
+        UIView.animate(withDuration: 0.07, delay: 0.0, options: []) {
```
charliescheer, line 145: Should we replace these numbers with some sort of constant?  Replace the `0.0` s with `.zero` and either add some local constants to the file for the others, or add them to `UIKitConstants`

Simplenote/Controllers/WidgetController.swift
```diff
@@ -8,15 +10,15 @@ struct WidgetController {
...
-                }
+            if widgets.contains(where: { widget in
+                /// Note:
+                /// We *used to* check if `widget.configuration` was of the `NoteWidgetIntent` / `ListWidgetIntent`.
+                /// This caused (a million) Swift compiler errors in Xcode 13.
+                /// In order to remediate this, we're disabling `codegen` in the `intentdefinition` file, and simply checking if there's
+                /// an Intent set up there.
+                /// The only scenario in which we wouldn't wanna reload the Timelines is the `New Note` widget (which does not display any dynamic content).
+                ///
+                return widget.configuration != nil
```
charliescheer, line 21: Nice!  Much simpler than before!

Simplenote/Supporting Files/Simplenote-Info.plist
```diff
@@ -75,6 +75,7 @@
 	<key>NSUserActivityTypes</key>
 	<array>
 		<string>ListWidgetIntent</string>
+		<string>NoteWidgetIntent</string>
 		<string>SPNoteWidgetIntent</string>
```
charliescheer:
```suggestion

```

I'm pretty sure that SPNoteWidgetIntent is an earlier version of NoteWidgetIntent.  With the code generated intents files, by default the system gives the class a name and adds the app prefix.  `NoteWidgetIntent` should be the custom class that replaces `SPNoteWidgetIntent` so there is no prefix.  

We should be able to drop this `SPNoteWidgetIntent` here

charliescheer (commented):
Thanks for making all of these changes @jleandroperez it looks good.  I added a few comments of things that I saw.

charliescheer:
@jleandroperez checked out this PR for 13/14/15 and everything seemed fine to me.

Simplenote/Classes/UIView+Animations.swift
jleandroperez: We should *always* avoid magic numbers (inline constants).However, we should probably cap the kind of changes we're introducing in this PR (which is already super large!!)

charliescheer (approved):

jleandroperez:
Thank you Charliee!!

PR #1451, opened 2021-09-28
Crash when publishing deleted note

### LINKED ISSUE #1305: Crash when publishing fully deleted note
This issue is a 2-in-1, let me know if I should separate it.

### Expected
App should not crash

### Observed
It crashes when the `Actions` screen is open for a note, while the note is deleted forever on the other device. This allows the user to tap 'Publish', which will cause the crash:

https://user-images.githubusercontent.com/73365754/120666898-a3127880-c495-11eb-82ef-f9c50c7ad528.MP4

### Reproduced
1. iOS: Open any note and open options/actions screen for it
3. Other device: trash the note. This enables **bug1:** since the note is now trashed, but the Collab and Publish options are still enabled, it's possible to start a collab/publish the trashed note. (Corresponding menu items become disabled on Android in similar case).
4. Other device: delete the note forever. 
5. iOS: `The note was deleted on the other device` message is shown. Confirm it ~> **bug2:** Options screen is still shown. Tapping 'Publish' will cause a crash. **This would not be possible if Note List was opened automatically after message confirmation.**

-|-|-|-
iPhone|XR|14.4.2|4.37.0.0

charliescheer:
### Fix
This PR fixes #1305 

This PR adds a bit of logic when a note is updated to check if that note is the currently selected note in the note editor and if the change is `delete`  if so, the editor will dismiss any modals it is displaying, show an alert that the current note has been deleted, and return to the note list.

These changes fix two issues listed in #1305:
1. if a note is sent to the trash on another device while open on iOS the note can still be accessed and even though it has moved to the trash
2. If that note is then published the app crashes.

By having iOS exit out of a note if it has been deleted, the note can not be published avoiding the crash issue

@jleandroperez I did not make any logic changes to avoid the crash, just made it so that iOS will exit the note settings and the note editor if this happens. Do you think we should approach the crash as well?  Or is making the settings in accessible enough?

### Test
1. open a note on iOS and go to the note settings.  Leave the modal open
2. trash that note from a different device
3. return to iOS and see the modal has been dismissed and an alert telling you the note has been deleted is displayed.

### Review

### Release
> `RELEASE-NOTES.txt` was updated in efd1f1 with:
> 
> > Fixed an issue where a note could be edited after being trashed #1451


Simplenote/SPAppDelegate.m
```diff
@@ -355,6 +355,10 @@ - (void)bucket:(SPBucket *)bucket didChangeObjectForKey:(NSString *)key forChang
             {
                 if ([key isEqualToString:self.noteEditorViewController.note.simperiumKey]) {
                     [self.noteEditorViewController didReceiveNewContent];
+
+                    if ([memberNames.firstObject isEqual:@"deleted"]) {
+                        [self dismissTrashedNote];
+                    }
```
jleandroperez, line 362: There's already code, in place, to handle the `Note deleted from Trash` scenario. (And ends up running `didDeleteCurrentNote` as well).

How about if we, instead, ensure any presented VC is dismissed? (that way the Options VC should just go away).

WDYT?


```
// Class: SPNoteEditorViewController

- (void)dismissEditor:(id)sender
{
    // ...
    [self.presentedViewController dismissViewControllerAnimated:YES completion:nil];
    // ... 
}
```

charliescheer:
Thanks for those suggestions @jleandroperez 

I have updated this and it can be reviewed again when ever.

jleandroperez (approved):
:shipit:

PR #1453, opened 2021-09-28
Fixed a crash for ios 12 and 13 caused by widgets when app enters bg

charliescheer:
### Fix
Currently  SNiOS 4.45 on iOS 12 or 13, when the app enters the background it will crash.

This is caused by the app delegate trying to call libraries that are not available before iOS14. 

This PR fixes that issue

### Test
1. on a device or simulator running iOS 13 or iOS 12 open simplenote and login
2. enter the background.  The app should not crash

### Review

### DIFF
--- a/Simplenote/SPAppDelegate.m
+++ b/Simplenote/SPAppDelegate.m
@@ -168,7 +168,9 @@
 - (void)applicationDidBecomeActive:(UIApplication *)application
 {
     [SPTracker trackApplicationOpened];
-    [self syncWidgetDefaults];
+    if (@available(iOS 14.0, *)) {
+        [self syncWidgetDefaults];
+    }
 }
 
 - (void)applicationDidEnterBackground:(UIApplication *)application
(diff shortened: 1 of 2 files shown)


Simplenote/SPAppDelegate.m
```diff
@@ -182,8 +184,11 @@ - (void)applicationDidEnterBackground:(UIApplication *)application
 
     [self showPasscodeLockIfNecessary];
     [self cleanupScrollPositionCache];
-    [self syncWidgetDefaults];
-    [self resetWidgetTimelines];
+
+    if (@available(iOS 14.0, *)) {
+        [self syncWidgetDefaults];
+        [self resetWidgetTimelines];
+    }
```
jleandroperez: Suggestion: Rather than safeguarding every spot that calls `syncWidget` / `resetWidget`, you can just safeguard at the API level:

```
    @objc
    func resetWidgetTimelines() {
        guard #available(iOS 14.0, *) else {
            return
        }

        WidgetController.resetWidgetTimelines()
    }
```

Or even better, you can just add that snippet in `resetWidgetTimelines`. That shields you from any accidental bug like this, in the future.

jleandroperez (approved):
Please rebase against `release/4.45`, and Platform will take care of bringing the fix back into Develop.

Thank you Charlie, nice catch!!

jleandroperez (approved):
:shipit: sir!!

charliescheer:
Thanks for those suggestions @jleandroperez I have made those changes and tested that the crash is still fixed.  I am gonna go ahead and merge this.

pinging @oguzkocer for visibility.  This should be all set for release 4.45.

PR #1455, opened 2021-10-01
Added 429 too many login attempts error

charliescheer:
### Fix
This PR adds a new login error case for a response 429 too many attempts

Internal ref: paFIJd-ou#comment-1094-p2

### Test
1. setup some sort of proxy to interrupt incoming web traffic for the auth endpoint
2. change the response to 429
3. you should see an error for too many attempts

### Review


jleandroperez (approved):
:shipit:

PR #1495, opened 2022-11-08
Add repo automation to register new devices

mokagio:
~~We may or may not want to merge this PR, given that we have new internal automation for devs to self-serve when adding new devices to the Developer Portal.~~`*` Still, I just used this automation and thought it useful to track the changes I made.

`*` – We _should_ merge this PR because it's handy to have the manual automation, just in case. If that wasn't true, then by we should go remove that from all other projects, since it's theoretically no longer needed. Clearly, right now, the cost of keeping that code in is lower than the cost of having to manually add devices in case something goes in other parts of our tooling.

_I won't lie, though, getting this in also removes noise from my https://github.com/pulls list._

---

Note: The UI tests are failing because the kvdb.io URL we use to lookup test accounts is 404ing. That is obviously unrelated to the kind of changes this PR does.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -815,6 +815,51 @@ end
       UI.user_error!("This lane should be run only from CI")
     end
   end
+
+  desc 'Registers a Device in the developer console'
+  lane :register_new_device do |options|
+    device_name = UI.input('Device Name: ') if options[:device_name].nil?
+    device_id = UI.input('Device ID: ') if options[:device_id].nil?
+    all_bundle_ids = simplenote_app_identifiers
+
+    UI.message "Registering #{device_name} with ID #{device_id} and registering it with any provisioning profiles associated with these bundle identifiers:"
+    all_bundle_ids.each do |identifier|
+      UI.message "\t#{identifier}"
+    end
+
+    team_id = get_required_env('EXT_EXPORT_TEAM_ID')
+
+    # Register the user's device
+    register_device(
+      name: device_name,
+      udid: device_id,
+      team_id: team_id,
+      api_key_path: APP_STORE_CONNECT_KEY_PATH
+    )
+
+    # We're about to use `add_development_certificates_to_provisioning_profiles` and `add_all_devices_to_provisioning_profiles`.
+    # These actions use Developer Portal APIs that don't yet support authentication via API key (-.-').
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -20,8 +20,14 @@ before_all do
 
   # Check that the env files exist
   unless is_ci || File.file?(USER_ENV_FILE_PATH)
-    UI.user_error!("~/.simplenoteios-env.default not found: Please copy env/user.env-example to #{USER_ENV_FILE_PATH} and fill in the values")
+    error_message = <<~ERROR
+    ~/.simplenoteios-env.default not found. Please copy env/user.env-example to #{USER_ENV_FILE_PATH} and fill in the values") using the following command:
+
+    \tcp #{PROJECT_ROOT_FOLDER}/fastlane/env/user.env-example #{USER_ENV_FILE_PATH}
+    ERROR
+    UI.user_error!(error_message)
```
mokagio, line 30: This looks like this when printed:

image

fastlane/Fastfile
```diff
@@ -770,6 +776,51 @@ end
...
+    add_development_certificates_to_provisioning_profiles(
+      team_id: team_id,
+      app_identifier: all_bundle_ids
+    )
+
+    # Add all devices to the provisioning profiles
+    add_all_devices_to_provisioning_profiles(
+      team_id: team_id,
+      app_identifier: all_bundle_ids
+    )
```
mokagio, line 864: Worth noting that, when I run the automation, these two printed warnings about not finding profiles matching the given identifiers:

image

This is actually the correct behavior, because Simplenote uses a different bundle identifier with an additional `Development` reverse-DNS component for the development builds and is what made me realize I had to change the code as per 737aa165

fastlane/Fastfile
```diff
@@ -770,6 +776,51 @@ end
       UI.user_error!("This lane should be run only from CI")
     end
   end
+
+  desc 'Registers a Device in the developer console'
+  lane :register_new_device do |options|
```
mokagio, line 820: The following logic has been tried and tested dozens of times across our other apps.

AliSoftware (approved):
I haven't run the lane to test that it works, but since I see you did use it, I guess that's enough of a test 🙂 

Code looks good, and the lane is mostly similar to what we do with other apps (with the slight tweak about the additional Development profile), so :shipit:

PR #1497, opened 2022-11-11
Homogenize indentation in `Podfile`

mokagio:
This was done by opening the file in Vim and using the indent command (`=G`)

### Test
Nothing to test. If CI doesn't error on the `Podfile.lock`, we're good to go.

### Review

### Release


spencertransier (approved):
Thanks @mokagio! Is there a standard set of indenting rules that we use across the apps?

mokagio:
@spencertransier good question. There have been efforts in the past to set some standard, e.g. via [EditorConfig](https://editorconfig.org/) (see also paaHJt-1U5-p2) but I don't think we have an explicit repos with what our preferred configurations should be.

Also, RuboCop should know about consistent indentation and I wonder why it didn't tell us here.

Probably not worth investigating though. This is a super minor thing and I expect this sort of "issue", or "stones in the shoe" like @jkmassel calls them to ease out as we organically improve our static analysis coverage and CI integration.

PR #1504, opened 2022-11-16
[UI Tests] Adjust UI tests to support iOS 16 & switch to iPhone SE

pachlava:
### Fix
The last time Simplenote UI tests were actively executed was in the times of iOS 14 and CircleCI. Now, Buildkite uses iOS 16.

#### During the local execution with iPhone 14, a number of tests were failing due to differences between iOSes:

1. Note Editor tests involving links were failing. Previously, a link was placed inside a TextView. Now, the usable link is a child element of another link object, which has zeroes as starting coordinates. This was addressed by e9dca25c3570b26529da4e28c288072e79d9ce39

Screenshot 2022-11-16 at 17 37 59

2. As of iOS 14, the undo command resulted only in last entered character removal. Now, it undoes the whole entered word. Addressed by 05557e7afe19f934073931e0787a03ce50425a8b
3. Search tests failed because the test failed to delete existing tags on a preconditions stage. This is because the delete confirmation was a child of `Sheet`, and now it's a child of `ScrollView`. Addressed by a4b2f59fefa01a7810be4f13550320fbacbb97ab.

#### After those from above got fixed, some tests still failed on CI, but never failed locally.

These fails were really random, like a failure to open `Trash` in one of 50 tries, which is a sign of instability related to performance, so I replaced `iPhone 14` with `iPhone SE` in 77571e83e13a792ee155f712a75911d65c97a898, since the latter is a bit faster (the `All Tests` suite tests approximately a minute faster)

### Test
- Make sure the UI tests tasks are green on Buildkite. (Unrelated FYI: the `ui-test-full` is expected to run only after approval, while `ui-test-subset` should run on every commit. This setting was lost after migration to Buildkite, and now both run on every commit. This should be addressed separately).
- Optionally, run the tests locally using `iPhone SE`.

### Release

### DIFF
--- a/SimplenoteUITests/AssertGeneric.swift
+++ b/SimplenoteUITests/AssertGeneric.swift
@@ -35,8 +35,7 @@ let numberOfNotesInAllNotesNotExpected = "Notes Number" + inAllNotesEnding + not
     numberOfNotesInTrashNotExpected = "Notes Number" + inTrashEnding + notExpectedEnding,
     numberOfTagsSuggestionsNotExpected = "Tags search suggestions " + inNoteListEnding + notExpectedEnding
 
-let linkContainerNotFoundInEditor = "\" link container" + notFoundEnding + inEditorEnding,
-    linkNotFoundInEditor = "\" link" + notFoundEnding + inEditorEnding,
+let linkNotFoundInEditor = "\" link" + notFoundEnding + inEditorEnding,
     linkNotFoundInPreview = "\" link" + notFoundEnding + inNotePreviewEnding
 
 let textNotFoundInEditor = "\" text" + notFoundEnding + inEditorEnding,
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -230,6 +230,19 @@ class NoteEditorAssert {
         XCTAssertEqual(matches, 0)
     }
 
+    class func oneOfTheStringsIsShownInTextView(strings: [String]) {
+        var matchFound = false
+
+        for string in strings {
+            if NoteEditor.getTextViewsWithExactValueCount(value: string) > 0 {
+                matchFound = true
+                break
+            }
+        }
+
+        XCTAssertTrue(matchFound)
+    }
+
     class func textViewWithExactLabelShownOnce(label: String) {
         let matches = NoteEditor.getTextViewsWithExactLabelCount(label: label)
         XCTAssertEqual(matches, 1)
--- a/SimplenoteUITests/NoteEditor.swift
+++ b/SimplenoteUITests/NoteEditor.swift
@@ -132,9 +132,19 @@ class NoteEditor {
         toggleMarkdownState()
     }
 
-    class func pressLink(containerText: String, linkifiedText: String) {
+
+    class func getLink(byText linkText: String) -> XCUIElement {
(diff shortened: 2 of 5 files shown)


SimplenoteUITests/AssertGeneric.swift
```diff
@@ -35,8 +35,7 @@ let numberOfNotesInAllNotesNotExpected = "Notes Number" + inAllNotesEnding + not
     numberOfNotesInTrashNotExpected = "Notes Number" + inTrashEnding + notExpectedEnding,
     numberOfTagsSuggestionsNotExpected = "Tags search suggestions " + inNoteListEnding + notExpectedEnding
 
-let linkContainerNotFoundInEditor = "\" link container" + notFoundEnding + inEditorEnding,
```
pachlava, line 38: Links are no longer children of TextViews, so this is not needed.

SimplenoteUITests/NoteEditor.swift
```diff
@@ -89,7 +89,6 @@ class NoteEditor {
 
     class func undo() {
         app.textViews.element.tap(withNumberOfTaps: 2, numberOfTouches: 3)
-        app.otherElements["UIUndoInteractiveHUD"].children(matching: .other).element(boundBy: 1).children(matching: .other).element(boundBy: 1).tap()
```
pachlava, line 92: Double tap with three fingers now does the undo, no need to additionally tap `Undo` in the context menu.

spencertransier (approved):
These updates look great @pachlava! Thank you for jumping on those failures. 🙌

> I replaced iPhone 14 with iPhone SE in 77571e8, since the latter is a bit faster (the All Tests suite tests approximately a minute faster)

That's really interesting! I didn't know there could be a performance difference between simulators.

tiagomar (commented):
👋 @pachlava !
I did run all the tests locally from `trunk`, saw a few tests failing, moved to this PR's branch and saw them all passing, except for one, `testUndoUndoesTheLastEdit()`, which hadn't failed initially. However, I see it passing in CI. 🤷 

I checked the comment in the code (below) and it looks like we experience different things.
> // Depending on iOS version, undoing might result in last character only "removal"
        // or the whole entered string removal. As of iOS 16.0, former is true.

These are my results from running locally with iPhone SE on Xcode 14.1:
| iOS | Text after test | Test result | 
|-|-|-|
| 14.4 | "ABC" | ❌ |
| 16.0 | "ABC" | ❌ |
| 16.1 | "ABC" | ❌ |

If we can assure it's not a bug, we could 1) further investigate it; 2) accept both "" and "ABC" or 3) wait for it to become a problem since it's working in CI.

pachlava:
Thanks for the review @spencertransier!

> That's really interesting! I didn't know there could be a performance difference between simulators.

For some reason there is, even though I can't really imagine why, and iOS / Xcode version also sometimes play a big difference, especially for key input events.

pachlava:
Thanks for the additional local research @tiagomar! 👋 

Funny enough, using undo today resulted only in last character removal for me too... maybe it's a seasonal thing.

> If we can assure it's not a bug, we could 1) further investigate it; 2) accept both "" and "ABC" or 3) wait for it to become a problem since it's working in CI.

I decided to go with `2` (bf2cd4d4b5914075759c49b02747fbfcb8f34687), since this was a test I updated around 2-3 times already.

Thank you!

PR #1505, opened 2022-11-17
Merge 4.48 strings into trunk

spencertransier:
This PR includes: 
- Adding a lane to update the `Localizable.strings` file with new strings (it's necessary to get the strings translated for 4.48 or else I would have added this in a separate PR). SNiOS was using the deprecated `ios_localize_project` from a much earlier version of the Release Toolkit which is why it was necessary to update it. 
- The updated `Localizable.strings` file
- The updated release notes for 4.48

The failing UI tests are a known issue that [Sergiy is working on](https://github.com/Automattic/simplenote-ios/pull/1504) and are unrelated to the changes in this PR.


fastlane/Fastfile
```diff
@@ -919,6 +919,26 @@ end
...
+########################################################################
+
+desc 'Updates the main `Localizable.strings` file — that will be imported by GlotPress'
+lane :generate_strings_file_for_glotpress do
+  en_lproj_path = File.join(PROJECT_ROOT_FOLDER, 'Simplenote', 'en.lproj')
+  ios_generate_strings_file_from_code(
+    paths: ['Simplenote/', 'SimplenoteIntents/', 'SimplenoteShare/',
+            'SimplenoteWidgets/', 'Pods/Simperium/Simperium/',
+            'Pods/Simperium/Simperium-iOS'],
+    output_dir: en_lproj_path)
```
mokagio, line 933: Have you looked into adopting the two-step setup from WordPress and WooCommerce and including the `AppLocalizedString` custom routine?

https://github.com/wordpress-mobile/WordPress-iOS/blob/b624f735b505546096249d0514672c39b08a47cc/fastlane/lanes/localization.rb#L129-L144

I don't think it's necessary at this time, though, in the context of what this release wants to achieve. Something to keep on the radar for potential future work, though.

mokagio (approved):

fastlane/Fastfile
spencertransier: @mokagio I looked at it briefly while preparing this PR, but decided to keep the changes here as simple as possible since it's a release branch. 

I will take a deeper look soon so that we can have parity with SN and the other apps. Thank you!

AliSoftware: Note that, as App Extensions have their own `Bundle`, different from the `Bundle.main` one, that means that `NSLocalizedString` calls from within the code used in those targets… will search for the strings in the `.strings` files of those App Extensions bundles, and not in the `.strings` files of the main app bundle.

This is why we're using our `AppLocalizedString` function for those in WP/WC apps, so that those App Extensions targets will look at the `.strings` in the main app bundle instead of defaulting to searching for them in their own.

All this means that, if you call `ios_generate_strings_file_from_code` with paths like `'SimplenoteIntents/', 'SimplenoteShare/', 'SimplenoteWidgets'`, this will ultimately put those targets' strings in the main `Localizable.strings` files from the main target and main bundle… which is good because that means they will be sent to GlotPress, so that's a good first step… but then those translations will never be used by the code and shown in the UI of those Intents/Share extension/Widgets, as the code for those still uses `NSLocalizedString` and not `AppLocalizedString`.

> I will take a deeper look soon so that we can have parity with SN and the other apps. Thank you!

Yep, so the next steps should be:

 - Add the custom `AppLocalizedString.swift` file and function in the codebase
 - Add the script build phases to the Xcode project to lint those `AppLocalizedString` usages and to ensure that no Swift/ObjC file that is included in at least one app extension is still using `NSLocalizedString`. See [`LintAppLocalizedStringUsage.{sh,swift}` scripts here](https://github.com/wordpress-mobile/WordPress-iOS/tree/trunk/Scripts/BuildPhases)
 - Fix this script's violations, by making sure that all the source code that is added in an app extension target uses that `AppLocalizedString` function instead of `NSLocalizedString`.
 - Add `routines: ['AppLocalizedString'],` parameter to your call of `ios_generate_strings_file_from_code` in the `Fastfile`, and add `'**/AppLocalizedString.swift'` to its `exclude:` parameter (to not risk false-positive-parsing the helper function itself)
 - _(I think that's all, but I might be forgetting something?)_

Until you do that in your future follow-up PR, those app extension targets (Widgets, Share, …) won't show up translated in the UI to the end user 😉

PR #1518, opened 2023-03-10
Add background color to code block in dark mode preview

### LINKED ISSUE #1372: Dark Mode: code block is not dark in note preview
When using dark mode, the preview note option in the web app shows a nice dark color scheme for code blocks. The same does not happen when previewing the note from the iOS app instead.

### Expected
To see code blocks in a dark color scheme like we do in the web app:

Captura de Tela 2021-07-30 às 12 03 09

### Observed
Code blocks are not changed at all:

Captura de Tela 2021-07-30 às 12 03 09

### Reproduced
1. Add a code block to a note
2. Change the theme to dark
3. Choose to preview the note

#### Internal reference: p1616480324000200-slack-C0133BTEDLJ

mrpolanco:
### Fix
This PR fixes [#1372](https://github.com/Automattic/simplenote-ios/issues/1372)

When using dark mode, the preview note option on the web app shows a dark color scheme for code blocks. The same does not happen when previewing the note from the iOS app. This fix adds a background color to dark mode that matches the same color as found on the web app (#002b36).

Simplenote-web-app

### Test
1. Change the theme to Dark
2. Add code block to note
3. Choose to preview the note

Simplenote-code-block

### Observation

This addresses the code block background color not matching the web app (#002b36).

To add a color scheme similar to that of the web app, it may be worth adding [syntax highlighting](https://www.markdownguide.org/extended-syntax/#syntax-highlighting) to the app's roadmap, if not already done so.

### DIFF
--- a/Simplenote/Resources/markdown-dark.css
+++ b/Simplenote/Resources/markdown-dark.css
@@ -20,11 +20,11 @@ html {
 }
 
 .note-detail-markdown pre {
-    background: #f6f7f8;
+    background: #002b36;
 }
 
 .note-detail-markdown pre code {
-    color: #616870;
+    color: #84878a;
 }
 
 .note-detail-markdown table tr:nth-child(2n) {


jleandroperez (approved):
Thank you @mrpolanco, nice fix!!

:shipit:

PR #1527, opened 2023-04-26
Restore `Localizable.strings` file which was accidentally emptied

mokagio:
In 6058f24edf150a573ba977e95a6296b55b2c531b, I converted the file from UTF-16 to UTF-8. Or, rather, I thought I did.

The command I used was:

```
iconv \
  --from-code=utf-16le \
  --to-code=utf-8 \
  ./Simplenote/en.lproj/Localizable.strings > ./Simplenote/en.lproj/Localizable.strings
```

It seems like updating the file in place resulted in it being emptied.

This time, I reverted 6058f24edf150a573ba977e95a6296b55b2c531b and repeated the `iconv` conversion using a temporary file in between.

### Test
Because the issue has to do with encoding, I couldn't find a way to `diff` one file against the other. I even got ChatGPT to write me a script to read each file by line and compare them, but it, too, run into issues.

So, I opted for a low accuracy eye-based diffing. I sampled the two files and they looked the same:

image

### Review

### Release

### DIFF
--- a/Simplenote/en.lproj/Localizable.strings
+++ b/Simplenote/en.lproj/Localizable.strings
@@ -1,0 +1,1029 @@
+﻿/* Monthly Subscription Option. Please preserve the special marker! */
+"%@ per Month" = "%@ per Month";
+
+/* Yearly Subscription Option. Please preserve the special marker! */
+"%@ per Year" = "%@ per Year";
+
+/* Number of found search results */
+"%d Result" = "%d Result";
+
+/* Number of found search results */
+"%d Results" = "%d Results";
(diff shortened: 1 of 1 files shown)


mokagio:
The diff offers no opportunity for review, given it replaces an empty file with a full one and for the encoding diff complexities discussed in the PR review. The sooner we merge this PR the better, because it's currently blocking the import of the strings to GlotPress due to the file being empty on `trunk`, internal ref p1682436251706159-slack-C02KLTL3MKM. As such, I'll admin merge

Issue #1536, opened 2023-11-01
UI tests can't run on CI


crazytonyli:
The UI tests CI step is failing because [fastlane fails to find the test accounts](https://github.com/Automattic/simplenote-ios/blob/8d061a19283ea8d113203a5b80c92e7c8753a214/fastlane/Fastfile#L1064) which are stored in kvdb.io.

pachlava:
@crazytonyli 👋 

The free variant of buckets on kvdb.io lasts for 3 months IINM. To restore the bucket, it's enough to run one command from terminal: p1667903243007039/1658172902.518279-slack-C02H2SVA1L5

I just did that, and the testing PR #1537 [started executing the tests](https://buildkite.com/automattic/simplenote-ios/builds/598#018b8f2b-5567-4f24-8bf3-4e7642418ef5/443-466). All of them failed, however 😆, which is a good sign, because most probably the reason is the same for all. I'll have a look at the reason later this week, but I think the current issue can be closed.

**P.S.** Given the repo activity level, I don't think it makes sense to pay for the bucket. When the repo was active, the selection of test accounts was really needed, to avoid race scenarios, because the tests hit the real API. Now, I think we can either stick to running this command every 3 months or even get back to using one test account. WDYT?

crazytonyli:
@pachlava Thanks for looking into this issue.

Do you think we can put the `UI_TESTS_ACCOUNTS_JSON_URL` content into the pipeline's secret file? So that we don't need to go through kvdb.io?

pachlava:
@crazytonyli, I think it's a matter of scope / visibility. This should be an environment variable which can be read/written by different builds simultaneously. AFAIK, this was hard to achieve when CircleCI was in use. Possibly with BK this is easier?

charliescheer:
I brought this issue up recently cause we are cleaning up the SNiOS repo right now, and the CI failing got noticed.  @spencertransier got the CI running again.  Gonna close this

PR #1540, opened 2023-12-27
Migrate from Intel to Apple Silicon CI

spencertransier:
This PR migrates Simplenote iOS from Intel to Apple Silicon CI.

The "Build and Test" step succeeded, which builds the app and runs the unit tests: https://buildkite.com/automattic/simplenote-ios-apple-silicon/builds/3. The "UI Test (Full)" step fails due to a known issue with the test accounts that are used in the UI tests: https://github.com/Automattic/simplenote-ios/issues/1536#issuecomment-1790309542

I ran the UI tests locally on my Apple Silicon Mac and 37 out of 43 passed. The six that failed don't appear to be due to any Intel vs. Apple Silicon difference, so I don't think they should be a blocker to this migration. 

Accompanying `buildkite-ci` PR to update the Terraform config: https://github.com/Automattic/buildkite-ci/pull/338


AliSoftware (approved):
I'm happily surprised that there were no changes needed in the codebase to migrate SN from Xcode 14.3 to Xcode 15.1, especially given the app is not really maintained anymore those days! 🎉

AliSoftware:
PS: CI is red because the config was apparently reverted to unclustered in Buildkite (probably via a `terraform apply` from an unrelated buildkite-ci PR) and the last build was thus run on Intel infra, where `xcode-15.1` does not exist. So this is not surprising.

@spencertransier  Once you apply https://github.com/Automattic/buildkite-ci/pull/338 (which I just approved) to [deploy the changes](https://buildkite.com/automattic/buildkite-ci/builds/2070#018cadde-7c3d-46c3-ac5c-f47f7d8b800e), I'd suggest you to rebuild the last commit to validate that the changes were taken into account
 - I think you can just do this using the "Rebuild" button in Buildkite… but that being said, I'm not 100% sure that this won't reuse the settings of the build it is rebuilding (instead of re-evaluating the updated pipeline settings, like the cluster to use)
 - So, alternatively you might have to push an empty commit (`git commit --allow-empty -m"Empty commit to retrigger CI" && git push`) instead.

PR #1541, opened 2024-01-10
Update Simplenote iOS to fix build errors

### LINKED ISSUE #1545: Update Simplenote iOS to fix build errors
Simplenote iOS can currently build 😱 

There is currently a provisioning issue and a rake script issue.  Both should be fixed so the project can build

- [ ] Fix provisioning issue
- [ ] Fix `rake dependencies` so it runs with out failing.

charliescheer:
### Fix
Currently Simplenote iOS can't be built for device.  

There are two problems:
1. the rake file isn't compatible to the specified version of ruby, so you can't update the dependencies
2. The provisioning profile is wrong.

This PR Fixes both problems.

While I was in there I fixed a few warnings that have built up over time

### Test
Confirm you can build the app to device.  Confirm you can log into an account.


.gitignore
```diff
@@ -57,3 +57,5 @@ default.profraw
 # Everything without a .enc extension is ignored
 .configure-files/*
 !.configure-files/*.enc
+
+Gemfile.lock
```
charliescheer: Wasn't sure if we needed to or wanted to have the gem file/gemlock file in git.  Seemed odd to me to track it, but happy to undo this change if people disagree

Rakefile
```diff
@@ -98,7 +98,7 @@ namespace :dependencies do
           # Try first using a binary release
           zipfile = "#{tmpdir}/swiftlint-#{SWIFTLINT_VERSION}.zip"
           sh "curl --fail --location -o #{zipfile} https://github.com/realm/SwiftLint/releases/download/#{SWIFTLINT_VERSION}/portable_swiftlint.zip || true"
-          if File.exists?(zipfile)
+          if File.exist?(zipfile)
```
charliescheer, line 101: Fixes the ruby version error when running `rake dependencies`

Simplenote/Supporting Files/Simplenote.entitlements
```diff
@@ -13,11 +13,6 @@
 	<key>keychain-access-groups</key>
 	<array>
 		<string>$(AppIdentifierPrefix)$(CFBundleIdentifier)</string>
-		<string>4ESDVWK654.com.codality.NotationalFlow</string>
```
charliescheer, line 16: fixes the provisioning issue

.gitignore
spencertransier: @charliescheer Yep, we track `Gemfile.lock` in all the apps so the dependency versions stay consistent.

charliescheer: Okay, I removed the gem lock file from the gitignore

jleandroperez (approved):
✅ Verified the app can be built
✅ Verified I can run on the device!!

:shipit:

charliescheer:
I'm gonna merge this even though the UI tests are failing.  Will fix in an upcoming PR

PR #1547, opened 2024-01-11
GHA Workflow for validating issues

iangmaia:
This PR reimplements our [current Issues check done in Peril](https://github.com/Automattic/peril-settings/blob/master/org/issue/label.ts) with a GHA workflow. It uses the reusable workflow implemented on https://github.com/Automattic/dangermattic/pull/31.

## How to test
This is a bit hard to test as GHA Issue events are only triggered for workflows already on the main branch, and a complete test can be done only once this PR is merged.

I've tested it with a fork on my account. Feel free to create issues and play around with the labels on my forked project: https://github.com/iangmaia/woocommerce-android/

### DIFF
--- a/.github/workflows/validate-issues.yml
+++ b/.github/workflows/validate-issues.yml
@@ -1,0 +1,18 @@
+name: 📝 Validate Issues
+
+on:
+  issues:
+    types: [opened, labeled, unlabeled]
+
+jobs:
+  check-labels-on-issues:
+    uses: Automattic/dangermattic/.github/workflows/reusable-check-labels-on-issues.yml@iangmaia/gha-labels-on-issues-check
+    with:
+      label-format-list: '[
+        "^\[.+\]",
+        "^[[:alnum:]]"
+      ]'
+      label-error-message: '🚫 Please add a type label (e.g. **[Type] Enhancement**) and a feature label (e.g. **Stats**) to this issue.'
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-issues.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 jobs:
   check-labels-on-issues:
-    uses: Automattic/dangermattic/.github/workflows/reusable-check-labels-on-issues.yml@iangmaia/gha-labels-on-issues-check
+    uses: Automattic/dangermattic/.github/workflows/reusable-check-labels-on-issues.yml@trunk
```
mokagio: How do you feel about pinning to a version?

Or... maybe using `trunk` is cool because it will auto-update? 

Then again, auto-updates can backfire if something was to change.

🤔 🤔 🤔

iangmaia: Yep, it will auto-update for now. My plan is to keep it in trunk until we ship a first stable version of Dangermattic after all the API changes we've been doing, then we can move to a `@v1` pinning instead.

mokagio: Ah, right. I didn't realize Dangermattic wasn't v1 yet. I should have checked...

mokagio (approved):
:shipit:

PR #1548, opened 2024-01-11
Update snios minimum os version to iOS 15

### LINKED ISSUE #1542: Update minimum supported iOS version to iOS 15
Simplenote iOS has been static for a little while.  It would be good to update Simplenote's minimum supported version to something a bit more modern.

This is gonna require a couple of tasks
 - [ ] Update the minimum version
 - [ ] Fix any depreciation warnings caused by the minimum version change
 - [ ] Remove any version specific code for older versions of iOS
 - [ ] Update any public documentation that states the minimum version

### LINKED ISSUE #1544: Update Xcode project settings
There are a few warning in Simplenote iOS right now where Xcode wants us to update the project settings to recommended settings.  This should be a fairly easy task, except that when you do this update for cocoa pods the app stops building.... 

So:
- [ ] Diagnose what updated setting is breaking the project
- [ ] Update the settings
- [ ] Fix any errors introduced after updating settings

charliescheer:
### Fix
We are going to update the minimum supported version of iOS for Simplenote to iOS 15.

This PR sets those minimums and then fixes all the warnings that cropped up after making those changes.

### Test
This changes a bunch of things, so a general app smoke test would be a good idea.  Somethings that definitely need to be confirmed:

- [x] All three Simplenote widgets look right on iOS 17
- [x] viewing markdown preview still works
- [x] sharing from other apps to simplenote still works
- [x] sharing from the notes app Bear using Text Bundle as the export type works

### Review
> Only one developer  are required to review these changes, but anyone can perform the review.

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -3993,7 +3993,7 @@
 				);
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				INFOPLIST_FILE = SimplenoteScreenshots/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 13.2;
+				IPHONEOS_DEPLOYMENT_TARGET = 15.0;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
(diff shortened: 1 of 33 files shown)


charliescheer:
Converting back to draft cause I realized the @available version checking wasn't removed yet

charliescheer:
Okay, this is finally ready now

jleandroperez (approved):
✅ All test scenarios verified
✅ Changes look beautiful

Nice work sir!!!

:shipit:

PR #1549, opened 2024-01-16
Replaced automattic tracks cocoa pod with automattic track SPM

### LINKED ISSUE #1543: Update use of Automattic Tracks iOS to SPM
Simplenote iOS is currently using version tracks version 0.13.  This works fine but it is really really old and the sentry version included in that version of tracks is causing an error in the project.  We should update tracks to the most recent version, but this is gonna cause issues cause we are importing tracks from cocoa pods and the latest version available on cocoa pods is 2, which is also very old.  So we should move tracks to SPM so we can get the most recent version of tracks.

This is gonna require a few tasks:
- [x] Add Automattic tracks iOS to SPM
- [x] Remove Automattic tracks iOS from cocoa pods
- [x] This will break stuff, so fix the usage of tracks inside the project to call to the SPM package (fix the errors)

charliescheer:
### Fix
Simplenote iOS is currently using tracks version 0.13, which is a little old...  Tracks stopped supporting cocoa pods on version 2 as well. So it seemed like a good idea to update to using the SPM version of tracks and get onto the newest release.  In this PR I have done that.

Relies on #1545 
### Test
1. pull the branch and run `rake dependencies` on the repo to update the pods
2. make sure the app can build
3. Trigger a trackable event (like opening the sidebar or making a note)
4. Go to tracks live, wait a while, and then confirm your event comes through.  Be aware that you can't filter SN events by user id, so you will have to find it manually

### Review

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -25,8 +25,6 @@ abstract_target 'Automattic' do
 
     # Automattic
     #
-    pod 'Automattic-Tracks-iOS', '~> 0.13'
-    #		pod 'Automattic-Tracks-iOS', :git => 'https://github.com/Automattic/Automattic-Tracks-iOS.git', :branch => 'add/support-for-tracking-crashes'
     pod 'Simperium', '1.9.0'
     pod 'WordPress-Ratings-iOS', '0.0.2'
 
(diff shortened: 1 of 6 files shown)


jleandroperez (approved):
I **LOVE** this PR!!!

:shipit: sir!!

charliescheer:
Merging this even though the UI Tests are failing.  Will look into those in a future PR

PR #1550, opened 2024-01-17
Fix UI tests

charliescheer:
### Fix
The UI Tests are failing currently so the CI is unhappy.  We can't have that, so I have gone ahead and fixed the failures.

Note that I am currently getting different failures locally as compared to what is happening on the CI.  What this PR fixes is the UI Tests and how they run locally.  If the CI fails this PR I will phone an infrastructure friend to take a look and see what if we can figure out why.

### Test
1.  Select the UI test target and run all the tests.  
Confirm they all pass.

### Review

### DIFF
--- a/SimplenoteUITests/Generic.swift
+++ b/SimplenoteUITests/Generic.swift
@@ -154,8 +154,10 @@ class WebView {
 
     class func tapDone() {
         let doneButton = app.buttons[UID.Button.done]
-        guard doneButton.exists else { return }
-        doneButton.tap()
+        guard doneButton.exists else {
+            return
+        }
+        doneButton.tapCenterCoordinates(in: app)
     }
 }
 
@@ -189,5 +191,13 @@ class Alert {
         guard confirmationButton.exists else { return }
 
         confirmationButton.tap()
+    }
+}
+
+extension XCUIElement {
+    func tapCenterCoordinates(in app: XCUIApplication) {
+        let coordinates = app.coordinate(withNormalizedOffset: .zero)
+            .withOffset(CGVector(dx: frame.midX, dy: frame.midY))
+        coordinates.tap()
     }
 }
--- a/SimplenoteUITests/NoteList.swift
+++ b/SimplenoteUITests/NoteList.swift
@@ -32,11 +32,11 @@ class NoteList {
     }
 
     class func getNoteCellHeight(_ noteName: String) -> CGFloat {
-        return NoteList.getNoteCell(noteName).frame.height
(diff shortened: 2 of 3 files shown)


SimplenoteUITests/Generic.swift
```diff
@@ -154,8 +154,10 @@ class WebView {
 
     class func tapDone() {
         let doneButton = app.buttons[UID.Button.done]
-        guard doneButton.exists else { return }
-        doneButton.tap()
+        guard doneButton.exists else {
+            return
+        }
+        doneButton.tapCenterCoordinates(in: app)
```
charliescheer, line 160: For some reason the webkit done button couldn't be tapped on by the test runner.  So getting the center point and tapping the coords instead

SimplenoteUITests/NoteList.swift
```diff
@@ -32,11 +32,11 @@ class NoteList {
     }
 
     class func getNoteCellHeight(_ noteName: String) -> CGFloat {
-        return NoteList.getNoteCell(noteName).frame.height
+        return NoteList.getNoteCell(noteName).frame.height.rounded()
     }
 
     class func openAllNotes() {
-        guard !NoteList.isAllNotesListOpen() else { return }
+        guard !isAllNotesListOpen() else { return }
```
charliescheer, line 39: Linter warning

SimplenoteUITests/NoteList.swift
```diff
@@ -32,11 +32,11 @@ class NoteList {
     }
 
     class func getNoteCellHeight(_ noteName: String) -> CGFloat {
-        return NoteList.getNoteCell(noteName).frame.height
+        return NoteList.getNoteCell(noteName).frame.height.rounded()
```
charliescheer, line 35: This was returning `80.99999999999997` consistently.  But we are looking for `81`.  close enough...

SimplenoteUITests/SimplenoteUITestsNoteEditor.swift
```diff
@@ -114,7 +114,7 @@ class SimplenoteUISmokeTestsNoteEditor: XCTestCase {
 
         trackStep()
         NoteEditor.swipeToPreview()
-        PreviewAssert.staticTextWithExactValueShownOnce(value: noteNameInitial)
+        PreviewAssert.staticTextWithExactLabelShownOnce(label: noteNameInitial)
```
charliescheer, line 117: When the checkbox was added the value was not discoverable but the label was.

jleandroperez (approved):
Niiiiiceee 🚀 !!!

PR #1555, opened 2024-03-26
Charlie/drop sustainer

charliescheer:
### Fix
We are dropping the Sustainer plan from Simplenote.  This PR removes the Sustainer banner in the tags list, the sustainer banner in the settings view, and the alert controller that guides users to upgrading.  While looking at this we decided that the store and header banner tools may be useful in the future so I have preserved these tools and updated the naming of `SustainerView` to `BannerView` to make it more usable in the future.

### Test
1. Launch Simplenote iOS and confirm that in the tags list you do not see the sustainer banner.  
2. Open the settings view and confirm there is no sustainer banner in the settings view
3. Make sure you can not access the alert controller that permits purchasing sustainer
4. Make sure both the tags and settings views respond to the device orientation changes.

### Review

### Release


Simplenote/BannerView.swift
```diff
@@ -55,17 +49,19 @@ class SustainerView: UIView {
...
 }
 
 // MARK: - Private API(s)
 //
-private extension SustainerView {
+private extension BannerView {
 
-    func refreshInterface() {
-        let style = isActiveSustainer ? Style.sustainer : Style.notSubscriber
+    func refreshInterface(with style: Style? = nil) {
```
charliescheer, line 61: Modified this function so that we can define a style for the BannerView instead of using the predefined sustainer Styles that exist in the file

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/BannerView.swift
jleandroperez: If this API doesn't do anything whenever `Style` is nil, perhaps we should drop the optional param?

jleandroperez (approved):
Nice work!!

:shipit:

PR #1557, opened 2024-03-27
Make sure that updating the subscriptions doesnt remove sustainer status

charliescheer:
### Fix
Currently if the subscription status of a users account changes in store kit then we will update and sync those values through Simperium.  As we drop the sustainer plan, we don't want to remove these values so we can still identify the previous sustainers.

So in this PR I have made it so that we only update the Simperium values if the subscription status is active.  Since the UI has been removed from the rest of the app this won't effect how the existing sustainers see the app and no new sustainers will be able to sign up.  This should make the subscription level values sticky for all users accounts.

### Test
Nothing really to test here, just verify the code makes sense

### Review

### Release

### DIFF
--- a/Simplenote/StoreManager.swift
+++ b/Simplenote/StoreManager.swift
@@ -17,7 +17,6 @@ enum StoreError: Error {
 
 // MARK: - StoreManager
 //
-@available(iOS 15, *)
 class StoreManager {
 
     // MARK: - Static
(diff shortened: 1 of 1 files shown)


Simplenote/StoreManager.swift
```diff
@@ -17,7 +17,6 @@ enum StoreError: Error {
 
 // MARK: - StoreManager
 //
-@available(iOS 15, *)
```
charliescheer, line 20: min version is 15 now.  So we don't need these.

jleandroperez (approved):
Thank you Charlie!!

:shipit:

PR #1558, opened 2024-04-09
Persist was sustainer value in preferences

### LINKED ISSUE #1442: iOS Analytics is not defaulting to opt out for new accounts
### Expected
On Simplenote, currently we should not be syncing the privacy settings around analytics to Simperium, and it should be defaulting to opt out for a new user.

### Observed
The analytics setting for SNiOS is currently being fetched from Simperium and not being stored locally.  This process is causing the analytics to get set to enabled without the user being asked.  This is different from our other platforms which are opt out by default and store the setting locally.

### Reproduced
1. Create a new account
2. check the analytics setting by going to Menu > privacy (you should see it enabled)
3. create an account on another client and check the privacy (you should see it disabled)

###References
iOS privacy setting:
https://github.com/Automattic/simplenote-ios/blob/bfa530c86e3cbd4fe89c01666de9f39351e83221/Simplenote/SPPrivacyViewController.swift#L28-L35

MacOS privacy setting:
https://github.com/Automattic/simplenote-macos/blob/57e013379603f7ea7e811709e290f3f5ac56ca5e/Simplenote/PrivacyViewController.swift#L46-L48

charliescheer:
### Fix
We currently check to see if a user is an active sustainer if they have a subscription level that is set to sustainer.  Since we are going to drop that value, in the future we may want to be able to identify which users were once sustainers.  We have added a value for this in the user preferences, but need to be able to store and parse that information on the iOS client.

In this PR I have done that.  I have also created a migration so that existing clients will be able to pickup and store the value into their user preferences.

As a bonus I noticed that a fresh install of the app will override some existing in the preferences bucket with the app install defaults.  This isn't ideal as it clears the `was_sustainer` value and more importantly it clears `analytics_enabled`. I have removed the defaults for these values so when a new install happens and we fetch the preferences for the first time the local preferences are populated with the server values.

As a bonus I think this fixes #1442 will need to do some testing on that to confirm

### Test
1. on trunk, log into an account that has `was_sustainer` marked as true (contact me directly if you need steps to set that)
2. just to confirm, also go to Settings > Privacy and disable Share Analytics.
3. Stop running, checkout this branch.  In Xcode filter the logs for `Preferences`
4. run this branch.  After the app launches:

- [x] confirm you see `Migrating Simperium Preferences object to include was_sustainer value` in the logs
- [x] confirm you see simperium has pulled values for preferences, you are looking for something like

```
Simperium client ios-26bdebdd58cc46b79f002e1fe840a65a received change (Preferences) ios-26bdebdd58cc46b79f002e1fe840a65a [M] : {
    ccids =     (
        e59671b66ce24824aa8158e517b7de9b
    );
    clientid = "ios-26bdebdd58cc46b79f002e1fe840a65a";
    cv = 6615aa48d9b5bb0a654fb674;
    ev = 16;
    id = "preferences-key";
    o = M;
    sv = 15;
    v =     {
    "was_sustainer" : {
      "o" : "+",
      "v" : 1
    },
        "analytics_enabled" =         {
            o = r;
            v = 0;
        };
    };
}
```

- [x] Go to settings > Privacy and confirm that analytics are disabled

### Review
> Only one developer if required to review these changes, but anyone can perform the review.

### DIFF
--- a/Simplenote/Classes/SPSettingsViewController+Extensions.swift
+++ b/Simplenote/Classes/SPSettingsViewController+Extensions.swift
@@ -45,6 +45,11 @@ extension SPSettingsViewController {
     }
 
     @objc
+    var wasSustainer: Bool {
+        SPAppDelegate.shared().simperium.preferencesObject().was_sustainer == true
+    }
+
+    @objc
     func sustainerSwitchDidChangeValue(sender: UISwitch) {
         let isOn = sender.isOn
 
--- a/Simplenote/Classes/SPSettingsViewController+Extensions.swift
+++ b/Simplenote/Classes/SPSettingsViewController+Extensions.swift
@@ -40,13 +40,9 @@ fileprivate extension SPSettingsViewController {
 
 extension SPSettingsViewController {
     @objc
-    var isActiveSustainer: Bool {
-        SPAppDelegate.shared().simperium.preferencesObject().isActiveSubscriber
-    }
-
-    @objc
-    var wasSustainer: Bool {
-        SPAppDelegate.shared().simperium.preferencesObject().was_sustainer == true
+    var showSustainerSwitch: Bool {
+        let preferences = SPAppDelegate.shared().simperium.preferencesObject()
+        return preferences.isActiveSubscriber || preferences.wasSustainer
     }
 
     @objc
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -507,13 +507,13 @@ extension SPAppDelegate {
 
 // MARK: - Sustainer migration
 extension SPAppDelegate {
-    @objc 
+    @objc
     func migrateSimperiumPreferencesIfNeeded() {
         guard UserDefaults.standard.bool(forKey: .hasMigratedSustainerPreferences) == false else {
             return
         }
 
-        guard Options.shared.firstLaunch == false else {
+        guard isFirstLaunch() == false else {
             UserDefaults.standard.set(true, forKey: .hasMigratedSustainerPreferences)
             return
         }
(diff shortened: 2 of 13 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -44,6 +44,11 @@ extension SPSettingsViewController {
         SPAppDelegate.shared().simperium.preferencesObject().isActiveSubscriber
     }
 
+    @objc
+    var wasSustainer: Bool {
+        SPAppDelegate.shared().simperium.preferencesObject().was_sustainer == true
+    }
```
jleandroperez, line 46: Nit: we can probably combine `isActiveSustainer` and `wasSustainer` into a single property?

```
var showsSustainerSwitch: Bool {
   let preferences = SPAppDelegate.shared().simperium.preferencesObject()
   return preferences.isActiveSubscriber || preferences.was_sustainer
}
```

?

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -504,3 +504,26 @@ extension SPAppDelegate {
...
+            return
+        }
+
+        NSLog("Migrating Simperium Preferences object to include was_sustainer value")
+        UserDefaults.standard.removeObject(forKey: Simperium.preferencesLastChangedSignatureKey)
+        let prefs = simperium.preferencesObject()
+        prefs.ghostData = ""
+        simperium.saveWithoutSyncing()
+
+        UserDefaults.standard.set(true, forKey: .hasMigratedSustainerPreferences)
```
jleandroperez, line 527: You can prob. move this into a `defer` clause, and remove duplication between L517 and L527

jleandroperez (approved):
Nice work Charlieee!!!

:shipit:

PR #1559, opened 2024-04-10
Updated rake file to fix bundle install --path depreciation

charliescheer:
### Fix
While working on SNiOS I noticed that bundle has depreciated using `--path` when invoking commands and suggests setting a config path value instead.

```
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local path 'vendor/bundle'`, and stop using this flag
```

In this PR I have updated the rake file to fix this depreciation

### Test
1.  The bundle commands only run and show anything if it has to do work, so in terminal go to the SNiOS folder and type `bundle exec gem uninstall nokogiri 1.15.4`. Hit yes a couple of times to remove this gem
2. type `rake dependencies`
3. confirm that bundler runs without any depreciation notices

### Review

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,7 @@ require 'English'
 XCODE_WORKSPACE = 'Simplenote.xcworkspace'
 XCODE_SCHEME = 'Simplenote'
 XCODE_CONFIGURATION = 'Debug'
+LOCAL_PATH = "vendor/bundle"
 
 require 'fileutils'
 require 'tmpdir'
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -4,6 +4,7 @@ require 'English'
 XCODE_WORKSPACE = 'Simplenote.xcworkspace'
 XCODE_SCHEME = 'Simplenote'
 XCODE_CONFIGURATION = 'Debug'
+LOCAL_PATH = "vendor/bundle"
```
dangermattic, line 7: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/simplenote-ios/blob/97ee2b721bba1d6abdd082b67db8ccc7e826004d/Rakefile#L7"></span>Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.</td>
    </tr>
  </tbody>
</table>

jleandroperez (approved):
Working as advertised!!

:shipit:

PR #1560, opened 2024-04-12
Fixed an issue where note tableview cells highlighted before selection

charliescheer:
### Fix
So... right now if you are in multiselect on the notes list and you tap on a note, the highlighting for the cell changes on press down events, but the selection changes on press up.  Because you can scroll after pressing down this can create a weird experience where it seems like you selected a cell (but you haven't) then you start swiping and the selection disappears (but it wasn't ever selected) This experience is a little bizarre.

In this PR I have updated the note list cells to only change their highlighting if the cell is selected.  `setHighlighting` is called on press down and on `setSelected` so we can piggy back off both of those calls to only update the high lighting if the cell is selected.

### Test
1. Open up Simplenote and long press on a note.  Choose select to enter multi select
2. on a cell that is not selected, press down on it but don't lift your finger.  Confirm that the cell appearance doesn't change. 
3. Lift your finger and confirm that the appearance changes to being selected.

### Review

### Release
> `RELEASE-NOTES.txt` was updated in c46161 with:
> 
> > Fixed issue where note cell appearance changes on press down instead of when the cell is selected

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,6 @@
 4.51
 -----
-
+- Fixed issue where note cell appearance changes on press down instead of when the cell is selected
 
 4.50
 -----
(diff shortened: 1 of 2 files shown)


jleandroperez (approved):
Works exactly as advertised!!

:shipit:

PR #1561, opened 2024-04-13
Fastlane `match`: Migrate Code Signing from Google Cloud to S3

spencertransier:
### Summary

This PR: 
- Migrates the code signing certificates and provisioning storage from Google Cloud to S3
- Uses the DRY'd code signing lane setup from WooCommerce iOS and WordPress iOS to better organize the code signing lanes
- Updates Fastlane from `2.216.0` to `2.220.0`
- Deletes the `Appfile` because the `team_id` is already read in the lanes from environment variables
- Deletes the `Matchfile` and moves that info into the appropriate lanes within the `Fastfile`
- Removes unneeded values from the example user `env` file

### Test
1. In your local `~/.simplenoteios-env.default` env file, add values for `MATCH_S3_ACCESS_KEY`, `MATCH_S3_SECRET_ACCESS_KEY`, and `MATCH_PASSWORD` from Mobile Secrets
2. Run `bundle exec fastlane update_certs_and_profiles`
3. Verify that it completes successfully
4. Run `bundle exec fastlane update_certs_and_profiles readonly:false`
5. Verify that it completes successfully (everything is up to date, so no changes will be made)

### Review
You can view the Installable build that uses the S3 `match` storage here in Buildkite: https://buildkite.com/automattic/simplenote-ios/builds/771#018ee8a0-ff7e-4505-87ee-53f09f0be1c7/481-503

### Release


fastlane/Fastfile
```diff
@@ -919,55 +919,103 @@ platform :ios do
...
+  # NOTE: It might be neccessary to add `force: true` alongside `readonly: true` in order to regenerate some provisioning profiles.
+  # If this turns out to be a hard requirement, we should consider updating the method with logic to toggle the two setting based on whether we're fetching or renewing.
+
+  # Fail early if secrets not available (via ENV.fetch with not default).
+  # Otherwise, Fastlane will prompt to type them.
+  access_key = ENV.fetch('MATCH_S3_ACCESS_KEY')
+  secret_access_key = ENV.fetch('MATCH_S3_SECRET_ACCESS_KEY')
+  # This ENV var is required but match has no input parameter for it, likely for security
+  password_key = 'MATCH_PASSWORD'
+  raise "Missing #{password_key} in the environment" unless ENV.key?(password_key)
```
AliSoftware: Mmmh what if we don't pass the password via `MATCH_PASSWORD` (like we do on CI via the secret being injected via `.mobile-secrets`) but instead want to rely on the password being in our keychain (like when we run the lane locally)?

fastlane/Fastfile
```diff
@@ -919,55 +919,103 @@ platform :ios do
...
+lane :update_certs_and_profiles do |options|
+  alpha_code_signing(options)
+  internal_code_signing(options)
+  appstore_code_signing(options)
 end
 
-########################################################################
-# Fastlane match code signing
-########################################################################
+def update_code_signing_enterprise(readonly:, app_identifiers:)
```
AliSoftware: Suggestion: move this helper function next to the similar `def update_code_signing` helper down below

AliSoftware (commented):
Haven't tested it yet (nor looked  at the S3 bucket for the new files, nor checked mobile-secrets for the `MATCH_PASSWORD` to be provided to SNiOS's CI, nor checked Tumblr-iOS won't break while failing to decode SNiOS files with the password…) so I'll make another pass later, but just wanted to drop initial feedback in the meantime.

AliSoftware:
🎗️  Completely unrelated to this PR so obviously not to be solved in here, but dropping a note that we should probably fix the Prototype Build PR comments (especially now that the Google service we used to use for generating the QRCode image has been retired). Simply using our dedicated release-toolkit action for generating the prototype build PR comment text (instead of hardcoding it like now) should be enough to fix that issue.

fastlane/Fastfile
spencertransier: Good point. I removed the extra check here so that there won't be problems if `MATCH_PASSWORD` is provided locally as an env var. https://github.com/Automattic/simplenote-ios/commit/4a23e75e25f2e52d44e6789a90a4df48b64d8685

fastlane/Fastfile
spencertransier: 👍  Moved in this commit: https://github.com/Automattic/simplenote-ios/commit/664a20c4c44905e1c82b7fbc654675deed9046ed

spencertransier:
> 🎗️ Completely unrelated to this PR so obviously not to be solved in here, but dropping a note that we should probably fix the Prototype Build PR comments (especially now that the Google service we used to use for generating the QRCode image has been retired). Simply using our dedicated release-toolkit action for generating the prototype build PR comment text (instead of hardcoding it like now) should be enough to fix that issue.

Agreed, that's one of the many SNiOS things I'd like to get to soon 😃

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Gemfile.lock
```diff
@@ -400,7 +406,7 @@ PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
   arm64-darwin-23
-  ruby
+  x86_64-linux
```
AliSoftware, line 409: Ran `bundle lock --add-platform x86_64-linux` as instructed by the [CI error message on previous commit](https://github.com/Automattic/simplenote-ios/actions/runs/8714042403/job/23903563796#step:3:40)

AliSoftware (approved):
- [x] Checked that files are present in the S3 bucket
- [x] Ran the lanes mentioned in testing instructions, worked as expected. I didn't add `MATCH_PASSWORD` to my `~/.simplenoteios-env.default` and instead relied on it being in my keychain, and it still worked.

I also see that today's Tumblr alpha and beta didn't have issues decoding the files in the S3 bucket when running match (like it did previously) ✅ 

PS: I pushed a https://github.com/Automattic/simplenote-ios/pull/1561/commits/19fae80545d6def74aca1db364dece9ed957c949 to fix the `danger` CI failure while at it.

PR #1562, opened 2024-04-15
Option to disable spotlight search

charliescheer:
### Fix
Spotlight indexing and search was added to Simplenote iOS some time ago, but while implementing spotlight in Mac we realized that maybe that should be an optional setting. 

So I have added a setting to the app preferences to enable or disable spotlight search.  It will be off by default

NOTE: it seems that on iOS spotlight maintains a history of some results and interacted with spotlight indexes.  So even after deleting all of the index there may still be some ghost data.  This is not the same for Mac.  So I have added an alert here after deleting the indexes to inform the users that they may still see some results.

### Test
1.  Erase and reset sim.  Clean install of Simplenote.  Log into an account with notes
2. go to the Home Screen and open spotlight.  Type in some text from a note and confirm that no Simplenote notes appear
3. Go to settings > Index Notes in Spotlight and enable the switch.  Give it a minute to index all the notes
4. Return to the Home Screen and open spotlight.  Type in some text from a note and confirm that it appears in the results
5. Return to the settings screen in Simplenote and disable the Index Note in Spotlight button.  You will see an alert that says some history may still appear in spotlight (my testing says it is any search result you tapped on before)
6. Return to spotlight search and search for text from another note, confirm it does not appear

### Review

### Release
> `RELEASE-NOTES.txt` was updated in aff30a with:
> 
> > Added ability to enable and disable indexing notes in Spotlight


Simplenote/Classes/CSSearchable+Helpers.swift
```diff
@@ -32,8 +32,29 @@ extension CSSearchableItem {
...
+            return
+        }
+
+        context.perform {
+            if let deleted = context.fetchObjects(forEntityName: "Note", with: NSPredicate(format: "deleted == YES")) as? [Note] {
+                CSSearchableIndex.default().deleteSearchableNotes(deleted)
+            }
+
+            if let notes = context.fetchObjects(forEntityName: "Note", with: NSPredicate(format: "deleted == NO")) as? [Note] {
+                CSSearchableIndex.default().indexSearchableNotes(notes)
```
jleandroperez: I think you can replace `CSSearchableIndex.default()` with `self` ? (same note a couple lines above!!)

jleandroperez (approved):
Nice work @charliescheer !!!

:shipit:

PR #1563, opened 2024-04-16
Charlie/fix provisioning and settings

charliescheer:
### Fix
There are a few issues right now with SNiOS building and Xcode settings.

This PR fixes those issues.

### Test
nothing to test, just confirm you can build the app to device and that there are no warnings

### Review

### Release


jleandroperez (approved):
✅ Verified that the app builds + installs correctly, without warnings

:shipit:

PR #1564, opened 2024-04-16
New note widget lock screen

### LINKED ISSUE #1444: Using `New Note` via widget will create and save an empty note if user was stopped by passcode
### Expected
Since this is a use case which wasn't possible before, it's hard to come up with an expected result. I can think of two options:

- New empty notes should go directly to trash if user did not get to note editor
- Note should be created once user sees the editor, not once they click `New Note` button

### Observed
If the device has a passcode activated, and the user taps `New Note` (no matter if via `New Note` or `Note List` widget), but does not enter the passcode and closes it, the note will still be created and saved into `All notes`:

https://user-images.githubusercontent.com/73365754/134309027-e26849df-677e-41ee-bad6-f8a43273adb4.MP4

### Reproduced
1. Make sure you have either a `New Note` or `Note List` (with `All Notes` selected for display) widget added
2. Activate passcode in the app
3. Kill the app and tap `New Note` button ~> you will see passcode
4. Close the passcode ~> new empty note will be placed in `All Notes`

-|-|-|-
iPhone|XR|14.7.1|4.45.0.0

charliescheer:
### Fix
So this one is a little odd... but, if you are using the New Note Widget and you have passcode enabled, it is possible to tap on the widget, and then not enter your passcode before leaving SN.  If you follow those steps then a note is created and the editor is presented even though you never actually unlocked the app.  

We shouldn't be making changes to the app's content if the app has not been unlocked.  So in this PR I have deferred the creation of new notes from the new note widget until after the user has unlocked the app.  As a bonus, during testing I realized the same problem would have occurred using shortcuts, and this fixes that too.

fixes #1444 

### Test
1. In simplenote, go to settings > Passcode and add a passcode (this fix works for biometrics as well, but it is easier to test with just a passcode
2. make sure Simplenote is on the note list
3. Go to the iOS Home Screen and add the New Note Widget for Simplenote.
4. Tap on it, the app will launch to the Lock Screen.  Enter your passcode and confirm that you are in the editor in a new note.  Dismiss this note
5. Put Simplenote into the background, tap on the new note widget again.  When the passcode screen appears, don't enter your passcode, instead swipe up to put Simplenote in the background
6. Now using the app switcher select Simplenote.  This will put you back on the Lock Screen.
7. Enter your password and confirm that you do NOT end up in the editor with a new note.

### Review

### Release
> `RELEASE-NOTES.txt` was updated in 0dba22 with:
> 
> > - Fixed an issue where using the new note widget with the lock screen could create empty notes while the app was locked

### DIFF
--- a/Simplenote/Classes/PinLock/PinLockVerifyController.swift
+++ b/Simplenote/Classes/PinLock/PinLockVerifyController.swift
@@ -18,6 +18,8 @@ final class PinLockVerifyController: PinLockBaseController, PinLockController {
     private let application: ApplicationStateProvider
     private weak var delegate: PinLockVerifyControllerDelegate?
 
+    private var onSuccesBlocks = [(() -> Void)]()
+
     init(pinLockManager: SPPinLockManager = .shared,
          application: ApplicationStateProvider = UIApplication.shared,
          delegate: PinLockVerifyControllerDelegate) {
@@ -37,6 +39,7 @@ final class PinLockVerifyController: PinLockBaseController, PinLockController {
         }
 
         delegate?.pinLockVerifyControllerDidComplete(self)
+        runOnSuccessBlocks()
     }
 
     func handleCancellation() {
@@ -49,6 +52,21 @@ final class PinLockVerifyController: PinLockBaseController, PinLockController {
 
     func applicationDidBecomeActive() {
         verifyBiometry()
+    }
+
+    private func runOnSuccessBlocks() {
+        onSuccesBlocks.forEach({
+            $0()
+        })
+        removeSuccesBlocks()
+    }
+
+    func addOnSuccesBlock(block: @escaping () -> Void) {
+        onSuccesBlocks.append(block)
+    }
+
+    func removeSuccesBlocks() {
+        onSuccesBlocks.removeAll()
     }
 }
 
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">This PR is still a Draft: some checks will be skipped.</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/Classes/PinLock/PinLockVerifyController.swift
```diff
@@ -50,6 +53,21 @@ final class PinLockVerifyController: PinLockBaseController, PinLockController {
     func applicationDidBecomeActive() {
         verifyBiometry()
     }
+
+    private func runOnSuccessBlocks() {
+        onSuccesBlocks.forEach({
+            $0()
+        })
+        removeSuccesBlocks()
```
jleandroperez, line 61: 😎

jleandroperez (approved):
✅ Fix confirmed, no more new empty notes without unlocking!!

Nice changes!!

:shipit:

PR #1578, opened 2024-05-03
Shortcuts update part 1: remove existing shortcuts

### LINKED ISSUE #1567: Remove existing shortcuts
As part of the project to improve the iOS shortcut support for Simpleonte, we first need to remove the existing shortcuts.  

When we remove them, we want to make sure that we do not remove the Home Screen actions which are currently powered by the same user activity actions.

ref: #1566

charliescheer:
### Fix
We are reimplementing the current shortcuts using Siri kit intents.  This will open up a lot more interesting possibilities for shortcuts, it will also add the intents into shortcuts automatically, (where our current system is using donation to add the intents which means they may not appear in shortcuts) This is going to happen in several steps and the first one is to drop the existing NSUserActivity donation so it no longer is contributed to shortcuts.

### Test
Run this PR and then go to the shortcuts app -> Create a new shortcut - > Add action -> Find Simplenote and confirm you don't have new note or open note (you will still see publish to Wordpress, that comes from elsewhere and I don't think needs to be replaced)

### Review

### DIFF
--- a/Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,13 +1,12 @@
 {
-  "originHash" : "4c803bfc0eef41e640a11de86de8bd9cb6ec711e343dadfe7d14d904ae2a7283",
   "pins" : [
     {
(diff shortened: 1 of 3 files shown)


jleandroperez (approved):
:shipit:

PR #1579, opened 2024-05-03
Shortcuts update part 2: refactor widget intents

charliescheer:
### Fix
Step two for implementing siri kit intents is to refactor the widget intents.  Currently the only intents on Simplenote iOS are for the widgets, so the intents extension handler is the class that handles all intents.  That is gonna get messy, so I am breaking out the widget intent handlers into their own classes and using the IntentHandler class to martial the intents to the correct handler.

### Test
Confirm that you can still add and use all three Simplenote iOS widgets.

### Review


jleandroperez (approved):
✅ Verified the three widgets are working!!

:shipit:

PR #1580, opened 2024-05-03
Shortcuts update part 3: Added open new note intent and handler

charliescheer:
### Fix
First shortcut to reimplement with siri kit is open new note intent.  This shortcut simply opens Simplenote and then creates a new note and opens that in the editor.  It doesn't take any parameters or have any actions, so is fairly simple.

### Test
1. Launch the app
2. go to the Shortcut app, create a new shortcut, add action, find Simplenote and select Open New Note.
3. Run the shortcut and confirm the app opens and creates a new note

### Review


jleandroperez (approved):
✅ Verified that the new Open new Note intent works as expected

:shipit:

PR #1582, opened 2024-05-08
Add Append note Intent

charliescheer:
### Fix
So continuing on with the shortcut improvements.  One of the requested shortcuts is to be able to append to an existing note.  Makes sense that you would want to be able to use shortcuts to easily add to one of your existing notes.  So in this PR I have done just that.  

This PR adds a new intent Append Note.  With Append note you can specify an existing note and content that you want to add to it.  You can also choose to select the note and the content at the time which makes this shortcut a pretty versatile way to add content to existing Simplenote notes from outside of the app.

### Test
1. launch Simplenote with this branch and log into an account (you may need to do this on device to make the new entitlements work)
2. background Simplenote and go to the shortcuts app.  Add a new shortcut, add action and look for Simplenote.  Select Append note
3. Tap on the content button and choose "ask every time" and then tap on the note and select a note to use
4. Run the shortcut, add some content to your note.
5. Open Simplenote again and confirm that your note has been updated with the new content.

A few things to consider while testing this:
- make sure that tags stay the same
- make sure that system tags stay the same
- make sure a published note continues to be published

### Review

### Release

(When the last of the shortcuts are added then I will update the release notes)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

SimplenoteIntents/Intent Handlers/AppendNoteIntentHandler.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    func handle(intent: AppendNoteIntent) async -> AppendNoteIntentResponse {
+        guard let identifier = intent.note?.identifier,
+              let content = intent.content,
+              let note = coreDataWrapper.resultsController()?.note(forSimperiumKey: identifier),
+              let token = KeychainManager.extensionToken else {
+            return AppendNoteIntentResponse(code: .failure, userActivity: nil)
+        }
+
+        note.content? += "\n\n\(content)"
```
jleandroperez: You may wanna do something like:

```
let oldContent = note.content ?? ""
note.content = oldContent + "\n\n\(content)"
```

(Or so?) Otherwise whenever the content is currently nil, the new text would get lost

jleandroperez (commented):
Leaving a note, already discussed over Slack!.

I've found a glitch while testing this PR:

1. Add a new `Append note` Shortcut
2. Invoke the shortcut multiple times
3. Launch Simplenote

As a result, **only** the last text you've appended will be visible in the note.

SimplenoteShare/Simperium/Uploader.swift
```diff
@@ -61,10 +61,46 @@ extension Uploader: URLSessionTaskDelegate {
     }
 }
 
+class Downloader: NSObject {
```
jleandroperez: Maaaybe we should either extract this into its own file, OR just rename as `SimperiumREST.swift` ?

SimplenoteIntents/Intent Handlers/AppendNoteIntentHandler.swift
```diff
@@ -35,7 +35,11 @@ class AppendNoteIntentHandler: NSObject, AppendNoteIntentHandling {
             return AppendNoteIntentResponse(code: .failure, userActivity: nil)
         }
 
-        note.content? += "\n\n\(content)"
+        guard let existingContent = try? await Downloader(simperiumToken: token).getNoteContent(for: identifier) else {
+            return AppendNoteIntentResponse(code: .failure, userActivity: nil)
+        }
+
+        note.content = existingContent + "\n\n\(content)"
```
jleandroperez: I wonder if we'd rather using `+ \n` (rather than a double newline. Dunno?

jleandroperez (approved):
✅ Verified you can, as of now, append text to an Entry
✅ Verified that multiple Append OPs don't overwrite each other

Nice work Charlie!!

:shipit:

SimplenoteShare/Simperium/Uploader.swift
charliescheer: moved it to it's own file

SimplenoteIntents/Intent Handlers/AppendNoteIntentHandler.swift
charliescheer: done

Issue #1569, opened 2024-04-17
Add additional shortcuts


charliescheer:
in #1487 we received a request to extend the shortcut support for the iOS app to include a few extra features.  Requested:

- Create note with provided data
- Append to specific note
- List notes
- Find notes

We should explore the possibility of each of these options and implement them where possible.

ref: #1566

dangermattic:
Thanks for reporting! 👍

PR #1583, opened 2024-05-09
Add create note with content intent

charliescheer:
### Fix
Okay so if we are talking about shortcuts, one I think would be really useful is to be able to paste or write content into a new note and send that over to Simplenote so you can reference it later.  So this PR adds that shortcut intent.

With the new "Create New Note" intent you can create a new note with any given content and push it to Simplenote.  This shortcut does not require you to launch the app, it is pushed to the server directly from the shortcut and is available on all connected devices.

moves forward #1569 
relies on #1582 

### Test
1. Launch branch on device (the entitlements need to load on device before they work, so this may be required to access the api token) 
2. Background the app and go to shortcuts
3. Add a shortcut, add an action, find Simplenote, and select "create new note with content" 
4. Run the shortcut, add some content.
5. Confirm that the new note is created and available next time you launch Simplenote 

### Review

 (Will add release notes when I add the last shortcut)


jleandroperez (approved):
✅ Verified that newly created notes sync, and show up in the main app

:shipit:

PR #1584, opened 2024-05-09
Add find note by content intent

### LINKED ISSUE #1487: Better iOS Shortcuts support
It would be nice to have additional options in iOS Shortcuts on top of _Add New Note_ and open specific note.

### What
The Shortcuts app should contains at least the following commands:
- Create note with provided data
- Append to specific note
- List notes
- Find notes

### Why
Simplenote is not really usable with Shortcuts right now. It would be nice to have better integration there.

### How
As you can see the existing Notes shortcuts

### LINKED ISSUE #1569: Add additional shortcuts

charliescheer:
### Fix
So another thing that might be useful for shortcuts is to be able to find a note by its content and not its title.  This intent itself doesn't do anything except for search notes and present them as an option, but if you connect it to Append Note then you can find a note by it's content, and add content to it in one shortcut.  pretty cool I think.  So that is what I have done in this PR.

As a bonus, While I was at it I realized that you might also want to open a note that you found based on its content.  So I have updated the Open Note Intent to accept a note from other shortcuts.

This is the last planned shortcut to be added, so this: 
fixes #1569 
fixes #1487 

### Test
okay, so testing this requires a bit of understanding of shortcuts.

### find and append to note
1. build app to device to ensure the entitlements are handled correctly
2. background the app, open shortcuts, add a shortcut and tap on add action.  Find Simplenote and add Find Note With Content
3. tap on content and choose "Ask each time" and hit done in the drawer
4. at the bottom of the page you should see Simplenote in a hidden modal, slide that up and select Append to note
5. tap on content and hit Ask each time
6. now run the shortcut. 
7. type in some text you know is in one of your notes and then hit done. (note if you have two notes with the same content you will be asked to choose which note you want.  Make sure to test this too)
8. then enter in some content in the next view and hit done

Open Simplenote and confirm your text was added to your note

### find and open note
1. background the app again, open shortcuts, add a shortcut and tap on add action.  Find Simplenote and add Find Note With Content
2. tap on content and choose "Ask each time" and hit done in the drawer
3. at the bottom of the page you should see Simplenote in a hidden modal, slide that up and select Open Note
4. now run the shortcut. 

confirm that the note you searched for is opened in Simplenote

### Review

### Release
> `RELEASE-NOTES.txt` was updated in fc201c with:
> 
> > Extended support for iOS Shortcuts

### DIFF
--- a/SimplenoteIntents/ResolutionResults/IntentNote+Helpers.swift
+++ b/SimplenoteIntents/ResolutionResults/IntentNote+Helpers.swift
@@ -21,6 +21,31 @@ extension IntentNoteResolutionResult {
 
         return IntentNoteResolutionResult.success(with: intentNote)
     }
+
+    static func resolveIntentNote(for content: String, in coreDataWrapper: ExtensionCoreDataWrapper) -> IntentNoteResolutionResult {
+        guard let notes = coreDataWrapper.resultsController()?.notes() else {
+            // TODO: Better error
+            return IntentNoteResolutionResult.unsupported()
+        }
+        let filteredNotes = notes.filter({ $0.content?.contains(content) == true })
+        let intentNotes = filteredNotes.map({ IntentNote(identifier: $0.simperiumKey, display: $0.title) })
+
+        guard intentNotes.isEmpty == false else {
+            // TODO: Better error
+            return IntentNoteResolutionResult.unsupported()
+        }
+
+        guard intentNotes.count == 1 else {
+            return IntentNoteResolutionResult.disambiguation(with: intentNotes)
+        }
+
+        //This shouldn't happen but we to unwrap the existing note we need to return something
+        guard let matchingNote = intentNotes.first else {
+            return IntentNoteResolutionResult.unsupported()
+        }
+
+        return IntentNoteResolutionResult.success(with: matchingNote)
+    }
 }
 
 extension IntentNote {
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,6 @@
 4.52
 -----
-
+- Extended support for iOS Shortcuts
 
 4.51
 -----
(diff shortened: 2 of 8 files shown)


SimplenoteIntents/ResolutionResults/IntentNote+Helpers.swift
```diff
@@ -21,6 +21,31 @@ extension IntentNoteResolutionResult {
 
         return IntentNoteResolutionResult.success(with: intentNote)
     }
+
+    static func resolveIntentNote(for content: String, in coreDataWrapper: ExtensionCoreDataWrapper) -> IntentNoteResolutionResult {
+        guard let notes = coreDataWrapper.resultsController()?.notes() else {
+            // TODO: Better error
+            return IntentNoteResolutionResult.unsupported()
+        }
+        let filteredNotes = notes.filter({ $0.content?.contains(content) == true })
```
jleandroperez, line 30: Ideaaaally this happens at the Core Data level.

I haven't seen massively big accounts, but this will load all the things in memory, and run the lookup there.

jleandroperez (approved):
Woah chained Intents, love it!!

:shipit:

SimplenoteIntents/ResolutionResults/IntentNote+Helpers.swift
jleandroperez: @charliescheer ^ Good to disregard though, I don't think this should be a problem in Simplenote

charliescheer: I'm gonna disregard it for now I think.  If we get any complaints then we can optimize

PR #1585, opened 2024-05-09
Shortcuts: Copy note text intent

charliescheer:
### Fix
Okay. so in #1584 I said that was the last shortcut intent I was adding, but then I thought of two more that were too good to pass up.

So first, copy note text.  This intent will fetch a selected note or given note and then copy the full text content and return it from the shortcut.  This will allow users to connect the content in their notes to any other shortcut they might want to use it with.  Pretty neat right?

This intent on it's own doesn't do much, but combined with search intents and output intents it has a lot of potential

### Test
Okay so easiest way to test this one.  
1. build the app, background the app, go to shortcuts, add a shortcut and then tap on add action
2. Find Simplenote and select "Copy Note Content" 
3. Select a note (we will test accepting a note from another shortcut in the next PR so we can skip that for now)
4. then at the bottom tap on the x to dismiss the Simplenote modal hiding at the bottom
5. in the search field that appears type in "clipboard" and then select the copy to clipboard shortcut.  This will connect the output of the "Copy Note Content" shortcut to the copy to clipboard shortcut.
6. go to another app where you can paste text and attempt to past the text from the note you just copied.

### Review
> Only one developer and one designer is to review these changes, but anyone can perform the review.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is larger than 500 lines of changes. Please consider splitting it into smaller PRs for easier and faster reviews.</td>
    </tr>
  </tbody>
</table>

</p>

jleandroperez (approved):
Works as expected!!

:shipit:

PR #1586, opened 2024-05-09
Shortcuts: List notes by tag intent

### LINKED ISSUE #1569: Add additional shortcuts

charliescheer:
### Fix
Okay, I mean it this time, this is the last PR for adding new shortcuts, so this really fixes #1569 

The last Idea I realized might be useful is being able to find your notes via tag.  On its own this shortcut doesn't do much, it shows you your tags and then lets you select one note from the notes associated with that tag, then returns that from the shortcut.  However combined with other actions like append to note or copy note content this can be really useful.

### Test
1. Build the app, background the app, go to shortcuts, add a new shortcut, add a new action
2. Select Find Note With Tag.  Choose to select Ask Each Time as your tag choice
3. pull up the drawer at the bottom for other simple note actions and select Copy Note Content
4. Close the drawer, in the search bar type "clipboard". Select the copy to clipboard option
5. Run the shortcut. You will be prompted to choose a tag.  Select one with multiple notes in it
6. You will be asked to choose a note, select one
7. Go to another app where you can paste content and confirm the note content was copied.

Note: if you select a tag with just 1 note in it the content of that note will be copied without confirming

### Review


SimplenoteIntents/ResolutionResults/IntentNote+Helpers.swift
```diff
@@ -24,27 +24,35 @@ extension IntentNoteResolutionResult {
 
     static func resolveIntentNote(for content: String, in coreDataWrapper: ExtensionCoreDataWrapper) -> IntentNoteResolutionResult {
         guard let notes = coreDataWrapper.resultsController()?.notes() else {
-            // TODO: Better error
```
jleandroperez, line 27: No, no better error 😆

jleandroperez (approved):
Love the way these Shortcuts can be chained up!!

:shipit:

PR #1588, opened 2024-05-10
Updated note list view controller top content inset

charliescheer:
### Fix
There is a small gap at the top of the note list right now that doesn't look correct.
image

This PR updates the content inserts on the list view so that the gap isn't there.

### Test
1. Open up Simplenote and go to the notes list.  Confirm there isn't a weird gap at the top of the list above the first note.
2. scroll and make sure the notes still look correct

Try on a few devices and orientations to guard against any regressions

### Review

### Release

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController+Extensions.swift
+++ b/Simplenote/Classes/SPNoteListViewController+Extensions.swift
@@ -157,8 +157,8 @@ extension SPNoteListViewController {
     ///
     @objc
     func refreshTableViewTopInsets() {
-        tableView.contentInset.top = searchBarStackView.frame.height / 2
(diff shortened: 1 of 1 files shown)


roundhill (approved):
LGTM, tested on iOS 16.4 as well and on iPad.  Not sure if this exactly matches up how it used to look, but there is a little gap when selecting the top note between the note and the search bar:

Screenshot 2024-05-10 at 2 52 04 PM

roundhill:
Follow-up, the screenshot on the homepage I think has a bit more space in-between the search bar and the top note:

Screenshot 2024-05-10 at 2 54 52 PM

This PR:
Screenshot 2024-05-10 at 2 56 16 PM

PR #1592, opened 2024-05-14
Fix installable build

crazytonyli:
Set `MATCH_CERTIFICATE_ID` env var so that fastlane match can install the right certificate for us. See the code comments for details.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1017,7 +1017,7 @@ private_lane :alpha_code_signing do |options|
   # fastlane match can install the right certificate for us.
   #
   # [1]: https://github.com/fastlane/fastlane/blob/2.220.0/match/lib/match/runner.rb#L255-L258
-  ENV['MATCH_CERTIFICATE_ID'] = 'S8CMX4U6QW'
+  # ENV['MATCH_CERTIFICATE_ID'] = 'S8CMX4U6QW'
 
   update_code_signing_enterprise(
     app_identifiers: simplenote_app_identifiers(root_bundle_id: "#{APP_STORE_BUNDLE_IDENTIFIER}.Alpha"),
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -630,6 +630,7 @@ platform :ios do
   #####################################################################################
   desc 'Builds and uploads an installable build'
   lane :build_and_upload_installable_build do
+    ENV['MATCH_CERTIFICATE_ID'] = 'S8CMX4U6QW'
     alpha_code_signing
 
     # Get the current build version, and update it if needed
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -630,7 +630,6 @@ platform :ios do
   #####################################################################################
   desc 'Builds and uploads an installable build'
   lane :build_and_upload_installable_build do
-    ENV['MATCH_CERTIFICATE_ID'] = 'S8CMX4U6QW'
     alpha_code_signing
 
     # Get the current build version, and update it if needed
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/Fastfile
```diff
@@ -1007,6 +1007,18 @@ end
 # @option [Boolean] readonly (default: true) Whether to only fetch existing certificates and profiles, without generating new ones.
 #
 private_lane :alpha_code_signing do |options|
+  # IMPORTANT: We need to update this value when the app's provision profiles are renewed.
+  #
+  # At the moment, we have two certificates for in-house distribution in the fastlane
+  # match repo. `fastlane match` will only install one of them (See [1]), which may not be
```
AliSoftware, line 1013: Is there a reason for us having two certificates in the match repo currently? Maybe as we go through transitioning to a soon-to-be-expired one to a new further-expiration-date one?

I think I remember an issue or even a PR in fastlane exactly about that though (making fastlane always use the latest cert instead of just the first one) 🤔

Which, if that's the root cause for this issue in this repo and what this workaround aims to solve here, would be a better solution long term, and if so, worth taking note of them in that comment so that we can adopt the better solution once fixed in fastlane itself in the future.

AliSoftware (commented):
🎗️ Not related to this PR so could be addressed separately (just thought of it since this PR is about Prototype Builds) but would be nice to use our `release-toolkit`'s `prototype_build_details_comment` action to generate the PR comment (which is formatted nicer and solves the issue with QR code image)

fastlane/Fastfile
AliSoftware: > I think I remember an issue or even a PR in fastlane exactly about that though

[This is the one I was thinking about](https://github.com/fastlane/fastlane/pull/21809)

crazytonyli: > Is there a reason for us having two certificates in the match repo currently? Maybe as we go through transitioning to a soon-to-be-expired one to a new further-expiration-date one?

I haven't been following certificates renewal closing, so I'm not sure why we have two.

> making fastlane always use the latest cert instead of just the first one

I don't think that's a proper fix. That's no difference than picking first or last one. The correct fix should be installing the certificate that's used by the provisioning profiles. Or, just install all certificates and let Xcode figure out the rest.

AliSoftware: > The correct fix should be installing the certificate that's used by the provisioning profiles. Or, just install all certificates and let Xcode figure out the rest.

Good point. You might want to leave a comment in the _fastlane_ PR I mention above about this btw, as if it gets merged and shipped then the next time we update fastlane that might break our setups if they start picking the latest? 🤔

AliSoftware: And also in the short term if there's no good reason for us to have two distinct certificates in the Enterprise account, I think we should remove the one that is only used my Simplenote profiles to instead only keep the cert used by all other apps and profiles, even if that means you'll need to refresh the Simplenote profiles to use the other certificate.

That way that will avoid confusion in the future (about "why do we have 2 certificates? Which one is used by what? Which one should we renew?") and ease maintenance of our Enterprise account (including freeing up a slot for helping doing certificate rotation when the main cert will be about to expire, as iirc we can only have a maximum of two distribution certificates per account at a time?)

And finally doing such a cleanup in our account will also avoid you needing that fix (and us having to update the fix when certificates are renewed)

crazytonyli: I thought about that, but didn't do it because `S8CMX4U6QW` is newer than the other certificate (UW99FF5HN9): S8CMX4U6QW expires at 2027-01-07 17:22:57 UTC and UW99FF5HN9 expires at 2025-02-22 08:40:35 UTC.

It feels strange to revoke a newer certificate and keep an older one. But I guess that doesn't really
matter in how we use them in practice. I can look into renewing profiles.

crazytonyli: The in-house distribution provisioning profiles are renewed and [installable build step now passes](https://buildkite.com/automattic/simplenote-ios/builds/861#018f83c4-81da-4044-8c17-95e4b1cbc79a). I'll close this PR.

PR #1593, opened 2024-05-14
Dont use tags on shortcut

### LINKED ISSUE #1424: Update quick actions to not tag a new note
### What
As discussed in this comment https://github.com/Automattic/simplenote-ios/pull/1416#issuecomment-909135307 currently there is an inconsistency in how new notes from different quick actions.

We should update the new note quick action to not tag a note.

### Why
There is no visual information that hints that the new note will be tag

### How
Current flow:

Screen%20Capture%20on%202021-09-01%20at%2011-58-26

The note should not be tagged

charliescheer:
### Fix
Currently we have a few different ways to create and open new notes from outside of Simplenote:
- The new note widget
- The long press Home Screen shortcut
- The Open New Note shortcut

If you select one of those options and inside the app you have a tag selected then the new note that is created is tagged.  This is probably fine ish... but since you can't see what the tag is when the shortcuts are run it is a little odd if you weren't expecting it.  We shouldn't add the tag when making a new note.  So here we are not doing that.

fixes #1424 

### Test
1.  In Simplenote select a tag from the sidebar
2. background the app and long press on the Simplenote icon, select new note.  When the editor appears confirm that the new note does not have a tag
3. Go to the shortcut app, create a new shortcut, add action, select Simplenote and add an Open New Note action.  Run the action and confirm that the new note that is opened is not tagged
4. Go to the desktop and long press on the Home Screen, when the home screen edit appears tap on the plus, search the widgets for Simplenote, and add the new note widget.  Tap on that widget, confirm that no tag is added to the note that is created.

### Review

### Release

### DIFF
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -165,6 +165,11 @@ extension SPAppDelegate {
         }
     }
 
+    @objc
+    func presentNewNoteEditor(animated: Bool = false) {
+        presentNewNoteEditor(useSelectedTag: true, animated: animated)
+    }
+
     var verifyController: PinLockVerifyController? {
         (pinLockWindow?.rootViewController as? PinLockViewController)?.controller as? PinLockVerifyController
     }
--- a/Simplenote/SPAppDelegate+Extensions.swift
+++ b/Simplenote/SPAppDelegate+Extensions.swift
@@ -151,13 +151,17 @@ extension SPAppDelegate {
     /// Opens editor with a new note
     ///
     @objc
-    func presentNewNoteEditor(animated: Bool = false) {
-        if isPresentingPasscodeLock && SPPinLockManager.shared.isEnabled {
-            verifyController?.addOnSuccesBlock {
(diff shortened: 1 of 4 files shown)


Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -204,6 +213,16 @@ extension SPAppDelegate {
     func setupNoticeController() {
         NoticeController.shared.setupNoticeController()
     }
+
+    func performActionAfterUnlock(action: @escaping () -> Void) {
+        if isPresentingPasscodeLock && SPPinLockManager.shared.isEnabled {
+            verifyController?.addOnSuccesBlock {
```
jleandroperez, line 219: I know it's code just moved around... but should we unwrap first?

Otherwise this whole path will be a NO-OP, and the block will be lost?

jleandroperez (approved):
✅ Verified all test scenarios!!

:shipit:

Simplenote/SPAppDelegate+Extensions.swift
charliescheer: discussed, and decided that we are gonna disregard this one.  Thanks 🫡

PR #1594, opened 2024-05-15
Preserve tableView selection when app entering background

charliescheer:
### Fix
Currently if you have notes selected in the notes table view and you put Simplenote into the background, the selection is lost.  Not an ideal situation, I would expect that you could background and then return to Simplenote and the state of the app should be the same, so the selection would be preserved.  So in this PR I have fixed that.

### Test
1. Open up the note list, long press on a note and choose Select. 
2. Pick a few notes
3. put the app in the background
4. Go back into Simplenote

- [ ] confirm that your notes are still selected

### Review

### Release
> `RELEASE-NOTES.txt` was updated in ea5992 with:
> 
> > Fixed issue where notes selection was lost when app backgrounded

### DIFF
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -529,8 +529,6 @@
     [self refreshTitle];
     [self refreshSearchBar];
 
-    BOOL isTrashOnScreen = self.isDeletedFilterActive;
-
     [self refreshEmptyTrashState];
     
     [self displayPlaceholdersIfNeeded];
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -197,6 +197,7 @@
 
     // App Background
     [nc addObserver:self selector:@selector(appWillEnterBackground) name:UIApplicationDidEnterBackgroundNotification object:nil];
+    [nc addObserver:self selector:@selector(appWillEnterForeground) name:UIApplicationWillEnterForegroundNotification object:nil];
 }
 
 - (void)condensedPreferenceWasUpdated:(id)sender
@@ -220,6 +221,10 @@
 
 - (void)appWillEnterBackground {
     self.selectedNotesEnteringBackground = self.tableView.indexPathsForSelectedRows;
+}
+
+- (void)appWillEnterForeground {
+    [self restoreSelectedRowsAfterBackgrounding];
 }
 
 - (void)refreshStyle {
--- a/Simplenote/Classes/SPNoteListViewController.m
+++ b/Simplenote/Classes/SPNoteListViewController.m
@@ -63,6 +63,8 @@
         [self refreshStyle];
         [self update];
         self.mustScrollToFirstRow = YES;
+
+        self.selectedNotesEnteringBackground = [NSArray new];
     }
     
     return self;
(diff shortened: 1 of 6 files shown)


Simplenote/SPAppDelegate.m
```diff
@@ -41,6 +41,7 @@
 @interface SPAppDelegate ()
 
 @property (weak,   nonatomic) SPModalActivityIndicator      *signOutActivityIndicator;
+@property (nonatomic) NSArray *selectedNotesEnteringBackground;
```
jleandroperez: @charliescheer do you think we could encapsulate this in the NoteList itself?

(So that the AppDelegate doesn't keep track of IndexPath(s))

Simplenote/SPAppDelegate.m
```diff
@@ -189,12 +190,19 @@ - (void)applicationDidEnterBackground:(UIApplication *)application
     [self cleanupScrollPositionCache];
     [self syncWidgetDefaults];
     [self resetWidgetTimelines];
+
+    self.selectedNotesEnteringBackground = self.noteListViewController.tableView.indexPathsForSelectedRows;
```
jleandroperez: Same note here, perhaps we could move this into the Note List itself?

(The appDelegate could relay the event, or the Note List could listen to the `didEnterBackground` notification)

jleandroperez (commented):
Hey there @charliescheer !! Changes look good, buuut sending you a couple annoying comments 😆 

LMK what you think!!

Simplenote/SPAppDelegate.m
charliescheer: Ooff that would have been so much easier if the whole thing was in swift.  Yes I have made that change. 👍

Simplenote/SPAppDelegate.m
charliescheer: done

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -520,8 +529,6 @@ - (void)update
     [self refreshTitle];
     [self refreshSearchBar];
 
-    BOOL isTrashOnScreen = self.isDeletedFilterActive;
```
charliescheer, line 523: This was a warning in the project that I probably caused recently, but I was tired of it staring at me.  So tossing that in here to shut it up.

Simplenote/URL+Simplenote.swift
```diff
@@ -29,7 +29,7 @@ extension URL {
     }
 
     static func newNoteWidgetURL() -> URL {
-        guard var components = URLComponents.simplenoteURLComponents(with: Constants.newNotePath) else {
+        guard let components = URLComponents.simplenoteURLComponents(with: Constants.newNotePath) else {
```
charliescheer, line 32: another warning that was poking me in the brain.

Simplenote/Classes/SPNoteListViewController.m
```diff
@@ -192,6 +194,9 @@ - (void)startListeningToNotifications {
 
     // Themes
     [nc addObserver:self selector:@selector(themeDidChange) name:SPSimplenoteThemeChangedNotification object:nil];
+
+    // App Background
+    [nc addObserver:self selector:@selector(appWillEnterBackground) name:UIApplicationDidEnterBackgroundNotification object:nil];
```
jleandroperez, line 199: Nit: `UIApplicationWillEnterForegroundNotification` and you can remove the AppDelegate from the equation!

jleandroperez (approved):
✅ Verified that the selected notes aren't getting lost

:shipit:

Simplenote/Classes/SPNoteListViewController.m
charliescheer: yeah, good call.  Done

PR #1595, opened 2024-05-15
Fixed an issue can activate the faceid switch without a passcode

charliescheer:
### Fix
Currently there is an issue where you can activate the FaceID/Biometry lock switch for Simplenote without entering a passcode.  We do not permit this, and the biometry doesn't actually activate, but default setting changes and the switch appears like it is active.  This PR fixes that.

Steps to repro on trunk:
1. with PIN code and biometry disabled go to the settings view
2. enable the faceID switch without putting in a passcode
3. Swipe up and close out of the app without dealing with the modal.
4. Relaunch the app and go back to settings, you will see that the biometry switch is active, but it shouldn't be.

### Test
1. with PIN code and biometry disabled go to the settings view
2. enable the faceID switch without putting in a passcode
3. Swipe up and close out of the app without dealing with the modal.
4. Relaunch the app and go back to settings, you will see that the biometry switch is not active

Confirm you can set a passcode and activate biometry.

### Review

### Release
> `RELEASE-NOTES.txt` was updated in b2f88d with:
> 
> > Fixed an issue can activate the faceid switch without a passcode

### DIFF
--- a/Simplenote/Classes/SPSettingsViewController+Extensions.swift
+++ b/Simplenote/Classes/SPSettingsViewController+Extensions.swift
@@ -87,6 +87,7 @@ extension SPSettingsViewController: PinLockSetupControllerDelegate {
     func pinLockSetupControllerDidComplete(_ controller: PinLockSetupController) {
         SPTracker.trackSettingsPinlockEnabled(true)
         dismissPresentedViewController()
+        SPPinLockManager.shared.shouldUseBiometry = biometrySwitch.isOn
     }
 
     func pinLockSetupControllerDidCancel(_ controller: PinLockSetupController) {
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">PR requires at least one label.</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/Classes/SPSettingsViewController+Extensions.swift
```diff
@@ -87,6 +87,7 @@ extension SPSettingsViewController: PinLockSetupControllerDelegate {
     func pinLockSetupControllerDidComplete(_ controller: PinLockSetupController) {
         SPTracker.trackSettingsPinlockEnabled(true)
         dismissPresentedViewController()
+        SPPinLockManager.shared.shouldUseBiometry = biometrySwitch.isOn
```
jleandroperez: I think you don't really need to check on `biometrySwitch.isOn`, it'll always be `true`, if you got into this spot?

If so, perhaps, for readability, we could set it to `= true`?

jleandroperez (approved):
:shipit:

PR #1597, opened 2024-06-06
Fixing Crash

jleandroperez:
### Fix
In this PR we're fixing a crash, triggered by a force unwrap of a nil property.

### Test
1. Launch Simplenote
2. Send the app to Background
3. Bring the app back into Foreground

- [x] Verify the app doesn't crash

### Review
Fixed a crash that occurred after switching to / from background


charliescheer (approved):
oops... 👍  this works thanks for fixing it

jleandroperez:
Thank youuu

roundhill:
Yay!

PR #1600, opened 2024-06-13
Charlie/shortcut data loss failsafe

charliescheer:
### Fix
With the new append shortcuts there is the possibility that your shortcut fails and if that happens the content you sent to the shortcut will be lost.  There isn't a lot we can do to prevent that unfortunately, but we can attempt to save the content locally and then restore it on the next launch.  So in this PR I have added a content recovery mechanism incase a shortcut fails.  That way the user doesn't have to restore the written content if a shortcut fails.

### Test
1. Build and run simplenote ios
2. Open up the shortcuts app and create an append to note or a create note with content shortcut
3. Disable the internet connections
4. run the shortcut
6. launch simplenote.

 - [ ] Confirm that a new note is created with the content from your shortcut

### Review

### DIFF
--- a/Simplenote/ContentRecoveryManager.swift
+++ b/Simplenote/ContentRecoveryManager.swift
@@ -56,20 +56,19 @@ public class ContentRecoveryManager {
 
     private func prepareNoteContent(at url: URL, in context: NSManagedObjectContext) -> String? {
         guard let data = fileManager.contents(atPath: url.path),
-              let content = String(data: data, encoding: .utf8) else {
+              let recoveredContent = String(data: data, encoding: .utf8) else {
             return nil
         }
-        return content
-    }
 
-    private func identifier(from url: URL) -> String? {
-        let fileName = url.deletingPathExtension().lastPathComponent
-        let components = fileName.components(separatedBy: "-")
-        return components.last
+        var content = Constants.recoveredContentHeader
+        content += "\n\n"
+        content += recoveredContent
+        return content
     }
  }
 
 private struct Constants {
     static let recoveredContent = "recoveredContent"
     static let richTextKey = "richText"
+    static let recoveredContentHeader = NSLocalizedString("Recovered Note Cotent - ", comment: "Header to put on any files that need to be recovered")
 }
--- a/Simplenote/ContentRecoveryManager.swift
+++ b/Simplenote/ContentRecoveryManager.swift
@@ -1,0 +1,75 @@
+import Foundation
+import UniformTypeIdentifiers
+import CoreData
+
+public class ContentRecoveryManager {
+    private let fileManager: FileManager
+
+    public init(fileManager: FileManager = .default) {
+        self.fileManager = fileManager
+    }
+
+    private func createRecoveryDirIfNeeded() {
+        guard !fileManager.directoryExistsAtURL(fileManager.recoveryDirectoryURL) else {
+            return
+        }
+
+        try? fileManager.createDirectory(at: fileManager.recoveryDirectoryURL, withIntermediateDirectories: true)
+    }
+
+    // MARK: Archive
+    //
+    public func archiveContent(_ content: String) {
+        createRecoveryDirIfNeeded()
+
+        guard let data = content.data(using: .utf8) else {
+            return
+        }
+        try? data.write(to: url(for: UUID().uuidString))
+    }
+
+    private func url(for identifier: String) -> URL {
+        let formattedID = identifier.replacingOccurrences(of: "/", with: "-")
+        let fileName = "\(Constants.recoveredContent)-\(formattedID)"
+        return fileManager.recoveryDirectoryURL.appendingPathComponent(fileName, conformingTo: UTType.json)
+    }
+
+    // MARK: Restore
+    //
+    public func prepareRecoveredNoteContentIfNeeded(in context: NSManagedObjectContext) async -> [String] {
+        createRecoveryDirIfNeeded()
+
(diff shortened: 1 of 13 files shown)


Simplenote/ContentRecoveryManager.swift
```diff
@@ -0,0 +1,74 @@
...
+    public init(fileManager: FileManager = .default) {
+        self.fileManager = fileManager
+    }
+
+    private func createRecoveryDirIfNeeded() {
+        guard !fileManager.directoryExistsAtURL(fileManager.recoveryDirectoryURL) else {
+            return
+        }
+
+        try? fileManager.createDirectory(at: fileManager.recoveryDirectoryURL, withIntermediateDirectories: true)
```
jleandroperez: This could probably live within a FileManager extension, so that you don't need to invoke it from `archiveContent` or `prepareRecoveredNoteContentIfNeeded`

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -557,3 +557,29 @@ extension SPAppDelegate {
...
+    private func insertNote(with content: String) {
+        guard let note = simperium.notesBucket.insertNewObject() as? Note else {
+            return
+        }
+        note.modificationDate = Date()
+        note.creationDate = Date()
+        note.content = content
+        note.markdown = UserDefaults.standard.bool(forKey: "kMarkdownPreferencesKey")
+        simperium.save()
+    }
```
jleandroperez, line 567: This should probably be part of `ContentRecoveryManager`

Simplenote/ContentRecoveryManager.swift
```diff
@@ -0,0 +1,74 @@
...
+            guard let content =  prepareNoteContent(at: url, in: context) else {
+                return nil
+            }
+
+            try? fileManager.removeItem(at: url)
+            return content
+        }
+    }
+
+    private func prepareNoteContent(at url: URL, in context: NSManagedObjectContext) -> String? {
```
jleandroperez: The context here isn't being used

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -557,3 +557,29 @@ extension SPAppDelegate {
...
     }
 }
+
+// MARK: - Content Recovery
+//
+extension SPAppDelegate {
+    @objc
+    func attemptContentRecoveryIfNeeded() {
+        let context = NSManagedObjectContext(concurrencyType: .privateQueueConcurrencyType)
+        context.persistentStoreCoordinator = coreDataManager.persistentStoreCoordinator
```
jleandroperez: (Same note as above), the context here isn't being used

jleandroperez (commented):
Found a glitch!

1. Enable airplane mode
2. Trigger the `Create new Note` shortcut. Add something
3. Switch back to Simplenote

As a result, the content won't be restored. This is caused by the hook spot (should prob. be applicationDidBecomeActive)

**Also:** Can we adjust the copy of the alert you get? (maybe indicate that it was preserved somehow?)

SimplenoteIntents/Intent Handlers/AppendNoteIntentHandler.swift
```diff
@@ -38,11 +30,25 @@ class AppendNoteIntentHandler: NSObject, AppendNoteIntentHandling {
         guard let existingContent = try? await Downloader(simperiumToken: token).getNoteContent(for: identifier) else {
             return AppendNoteIntentResponse(code: .failure, userActivity: nil)
         }
```
jleandroperez: This snippet is, now, repeated, and the compiler prints a warning

Simplenote/SPAppDelegate+Extensions.swift
```diff
@@ -557,3 +557,29 @@ extension SPAppDelegate {
...
+
+        Task {
+            let restoredContent = await ContentRecoveryManager().prepareRecoveredNoteContentIfNeeded(in: context)
+            restoredContent.forEach({ insertNote(with: $0) })
```
jleandroperez: Just noticed in macOS, this is a Core Data threading violation. The Task is on a BG queue, and the Simperium API isn't threadsafe

Simplenote/RecoveryUnarchiver.swift
```diff
@@ -0,0 +1,53 @@
...
+
+public class RecoveryUnarchiver {
+    private let fileManager: FileManager
+
+    public init(fileManager: FileManager = .default) {
+        self.fileManager = fileManager
+    }
+
+    var simperium: Simperium {
+        SPAppDelegate.shared().simperium
```
jleandroperez: Same note as in mac, since the AppDelegate will use this, and has direct access to Simperium, this class can get the reference via initializer

SimplenoteIntents/Intent Handlers/AppendNoteIntentHandler.swift
```diff
@@ -35,14 +27,25 @@ class AppendNoteIntentHandler: NSObject, AppendNoteIntentHandling {
...
-        uploader.send(note)
 
-        return AppendNoteIntentResponse(code: .success, userActivity: nil)
+        do {
+            _ = try await uploader.send(note)
+            return AppendNoteIntentResponse(code: .success, userActivity: nil)
+        } catch {
+            return handleFailure(with: error, content: content)
+        }
+    }
```
jleandroperez, line 39: Same note as in macOS, not sure you *need* two catch statements!

jleandroperez (approved):
**Beautiful PR**, thank you for building this one too!!

:shipit:

PR #1616, opened 2024-07-15
Refactored passkey remote to handle all json conversions for passkeys

charliescheer:
### Fix
In my last PR, I refactored passkey authorization so that the PasskeyRemote only received and returned `Codable` objects and did any json conversions its self.  This was a nice improvement because you couldn't then feed just any old data into the PasskeyRemote, it needed to be structured a certain way.  Well, I noticed that for the passkey registration process we the remote was still accepting and returning Data instead of the `Codable` objects.   So in this PR I have updated that.

### Test
There shouldn't be any actual changes to how passkeys operate here.  So just smoke test that you can register and authenticate using Passkeys.

### Review

### Release


jleandroperez (approved):
Nice!!

:shipit:

PR #1631, opened 2024-07-31
Validate deleting tag before completing

### LINKED ISSUE #1312: Extra tag is deleted when initial tag was removed remotely
### Expected
App should delete only the tag that was selected for trashing, and for which the delete confirmation is shown.

### Observed
Maybe caused by using indices when deleting a tag... 🤷 

When a tag was selected for removal (confirmation dialog is shown), and this tag was removed on the other device, confirming the message will delete another additional tag (the original tag is removed via Electron app approx at second 10):

https://user-images.githubusercontent.com/73365754/120800083-373d1800-c548-11eb-93d4-36eb5a9508b8.MP4

### Reproduced
1. Have at least two tags
2. Select on tag for removal ~> confirmation dialog is shown
3. Delete this tag via other device ~> this tag is no longer in the list on iOS
4. Confirm the message on iOS ~> another tag will be removed

-|-|-|-
iPhone|XR|14.4.2|4.37.0.0

charliescheer:
### Fix
There is currently a bug when deleting tags on iOS.  If you have multiple tags and then..:
1. Show the tag list > Tap on edit > Tap on the trash button next to a tag.  A ui alert appears asking you to confirm that you want to delete that tag.  Don't confirm or dismiss the alert
2. On another device delete that same tag and let that sync across devices
3. Confirm deleting on the first device.  This will cause a different tag to be deleted than the one you originally selected.

This is a bit of a weird use case, but in a situation where syncing was in progress you could easily delete the wrong tag.  So we shouldn't permit this.  

In this PR I have added a validation method before tags are deleted that confirm the tag that was originally selected is still the same tag that found at the original indexPath.  If it is not then the same then show an alert and don't delete the tag.

fixes #1312 

### Test
1. Show the tag list > Tap on edit > Tap on the trash button next to a tag.  A ui alert appears asking you to confirm that you want to delete that tag.  Don't confirm or dismiss the alert
2. On another device delete that same tag and let that sync across devices
3. Confirm deleting on the first device.

- [x] Confirm that you see an alert and no additional tag is deleted
- [x] Try deleting a tag without deleting it on another device, confirm you can still delete

### Review

### Release

### DIFF
--- a/Simplenote/TagListViewController.swift
+++ b/Simplenote/TagListViewController.swift
@@ -501,7 +501,7 @@ extension TagListViewController: TagListViewCellDelegate {
                                                 preferredStyle: .actionSheet)
 
         alertController.addDestructiveActionWithTitle(Localization.TagDeletionConfirmation.confirmationButton) { (_) in
-            guard self.shouldRemove(tag, at: indexPath) else {
+            guard self.verifyTagIsAtIndexPath(tag, at: indexPath) else {
                 self.present(UIAlertController.dismissableAlert(
                     title: Localization.tagDeleteFailedTitle,
                     message: Localization.tagDeleteFailedMessage), animated: true)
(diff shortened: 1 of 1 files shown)


Simplenote/TagListViewController.swift
```diff
@@ -519,6 +526,17 @@ extension TagListViewController: TagListViewCellDelegate {
 
         present(alertController, animated: true, completion: nil)
     }
+
+    private func shouldRemove(_ tagToRemove: Tag, at indexPath: IndexPath) -> Bool {
```
jleandroperez: Perhaps, for clarity, this could be named...

- `func verifyTagIsAtIndexPath(tag: Tag, indexPath: IndexPath) -> Bool`

?

Thinking about adding a bit of extra clarity!

jleandroperez (approved):
Minor comment, no blockers!! Thanks for fixing this Charlie!

:shipit:

PR #1633, opened 2024-08-01
Prepare app to build on Xcode 16

charliescheer:
### Fix
Xcode 16 is coming!  So time to check and make sure that we can still build Simplenote on the newest version of Xcode.  I did a quick check and there is an issue in the newest version of Xcode where SN can't build because of one of its package dependencies.

In this PR I have updated Automattic tracks to get the newest version of Sentry-cocoa which fixes the build error.

### Test
1. download the latest Xcode beta
2. Build and run SN
3. smoke test the app

### Review

### Release


jleandroperez (approved):
:shipit:

PR #1638, opened 2024-08-13
Updates Translations

jleandroperez:
### Fix
In this PR we're updating the Localizations. For the record:

1. I've edited `fastlane/Fastfile` and commented out L205-212 and L179
2. Executed `bundle exec fastlane finalize_release`

@mokagio Would it be possible to confirm if doing this was required, or it's just part of the regular release process?
I **think** it wasn't really needed?

Thanks in advance!!

### Test
- [ ] Smoke test Simplenote and make sure everything looks good!

PR #1635, opened 2024-08-08
Split building for ASC from uploading to TestFlight and run as separate CI steps

mokagio:
### Fix
This will:

- Ensure better retry-ability. If the TestFlight upload fails because of HTTP errors, we can retry that without rebuilding the whole `ipa`
- Better CI resource management. The TestFlight upload does not need macOS, so it could run on a cheaper Linux box. This is not ready yet, but at some point we'll an agent optimized for running this kind of tasks.

### Test

I tested this by triggering a release build on this branch:

image

image

image

### Review
Only one developer or infrastructure engineer required to review these changes, but anyone can perform the review.

### Release

These changes do not require release notes.

PR #1639, opened 2024-08-15
Fix broken localization automation in release branch

mokagio:
I was checking the diff for #1638 after @roundhill pointed out that it looked like the Simplenote iOS build did not pick up new localizations.

I discovered that was indeed the fact and it was due to an old tooling migration that never had the chance to be properly tested IRL due to Simplenote being in maintenance mode.

I was surprised that I didn't get a failure, however, when working on #1635 recently. But then I realized that the _build and upload_ step does not include updating the codebase with localization etc. That's the job of `new_beta` and `finalize_release`. I didn't run either of those lanes at the time because I wasn't working at that level.

I decided not to commit the metadata changes that would result from running the new automation and leave that as something that will happen during a new beta or release finalization, as it should normally be.

### Fix
The fix really is more of a completion of the tooling upgrade, actually implementing the "modern" localization workflows.

### Test

I tested these locally and you can do the same by running:

- `bundle exec fastlane download_localized_strings_from_glotpress` ... Nothing to commit (unless new localizations changed on GlotPress)
- `bundle exec fastlane download_localized_metadata_from_glotpress` ... Same (or similar depending on GlotPress' state) to https://github.com/Automattic/simplenote-ios/compare/mokagio/fix-download-localizations...mokagio/test-new-localization-automation

### Review


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">This PR has the <code>Releases</code> label: some checks will be skipped.</td>
    </tr>
  </tbody>
</table>

</p>

.gitignore
```diff
@@ -48,6 +48,8 @@ fastlane/report.xml
 fastlane/README.md
 fastlane/screenshots/
 fastlane/test_output/
+# Ignore Fastlane's default/release_notes.txt because the automation copies it from the Simplenote/Resources/release_notes.txt file in
+fastlane/metadata/default/release_notes.txt
```
mokagio, line 52: What do you think of this?

My rationale is to have less noisy diffs and honor the (current) single source of truth that is the file in `Simplenote/Resources/`

fastlane/Fastfile
```diff
@@ -88,38 +95,6 @@ platform :ios do
...
-
-      english_comma = ','
-      arabic_comma = '،'
-      keywords = keywords.gsub(arabic_comma, english_comma) if File.basename(locale_dir) == 'ar-SA'
-
-      keywords = keywords.split(english_comma)[0...-1].join(english_comma) until keywords.length <= app_store_connect_keywords_length_limit
-
-      File.write(keywords_path, keywords)
-    end
-  end
```
mokagio, line 121: Moved to the localization lanes file

fastlane/lanes/localization.rb
```diff
@@ -8,7 +8,6 @@
   'cy' => 'cy',             # Welsh
   'de' => 'de',             # German
   'el' => 'el',             # Greek
-  'en' => 'en',             # English
```
mokagio, line 11: Does not exist in GlotPress...

fastlane/lanes/localization.rb
```diff
@@ -48,4 +69,130 @@
...
+
+      keywords = keywords.split(english_comma)[0...-1].join(english_comma) until keywords.length <= app_store_connect_keywords_length_limit
+
+      File.write(keywords_path, keywords)
+
+      git_commit(
+        path: keywords_path,
+        message: "Trim #{locale_code} keywords to be less than #{app_store_connect_keywords_length_limit} characters",
+        allow_nothing_to_commit: false
+      )
```
mokagio, line 164: I thought it clearer to individually commit each modified file, rather than in a batch or as part of the update-after-download commit. I think it valuable to know when this happen in detail, because it should help us fix the problem at the root.

.gitignore
AliSoftware: I think we should instead make `fastlane/metadata/default/release_notes.txt` be the new source of truth, and get rid of `Simplenote/Resources/release_notes.txt`.

This will simplify the setup drastically, reducing the number of intermediate files for release notes and the confusion it has always led to.

This is also something we've always wanted to simplify; and is an approach @iangmaia and I have already adopted in Tumblr. So I think that's the right move forward 🙂

The only other things to remember to do when migrating to that solution is to:
 - Update the release scenario accordingly to mention the new location where to update the release notes during code freeze
 - Make the `complete_code_freeze` deletes previous translations from `fastlane/metadata/{any-lang-other-than-default}/release_notes.txt` in parallel of updating the `.pot` with the new English copy from `fastlane/metadata/default/release_notes.txt`, so that we don't risk using old release notes from previous version if some locales haven't translated the new ones in time by the time we pull them before submission

mokagio: I think that's the right way forward, too. See also the `FIXME` in the download lane below.

However, I admit that if you hadn't left this comment, implementing it would have been low on my priority list...

I'm a bit unsure about making the switch in the context of this PR, though. I'd feel better about making it happen in one go and test it "end-to-end" on the next code freeze.

mokagio:
I'm going to admin-merge this to move on with shipping a new beta.

@AliSoftware I'll try to followup with the `release_notes.txt` source of truth update within the week...

PR #1644, opened 2024-08-30
Add method for milestone moving 

mokagio:
Updates the code freeze automation to move PRs targeting the milestone for which the code freeze just started to the next milestone. Part of the internal project paaHJt-6Zx-p2

The implementation is based on what other repos already do, example: [WooCommerce Android](https://github.com/woocommerce/woocommerce-android/blob/trunk/fastlane/Fastfile#L181-L212), but I had a stab at moving it to a dedicated method. The idea being to make it easier to extract to `release-toolkit` or an helper file in `lib/`, or even just make it simpler to copy-paste.

### Test
I haven't tested this change. I will do that as part of the upcoming code freeze, see also p2XJRt-3QQ-p2

### Review
Only one infra engineer required to review these changes, but anyone can perform the review.

### Release
These changes do not require release notes

### DIFF
--- a/fastlane/lanes/release.rb
+++ b/fastlane/lanes/release.rb
@@ -58,9 +58,10 @@ platform :ios do
       from_branch: DEFAULT_BRANCH,
       to_branch: computed_release_branch_name
     )
-    set_milestone_frozen_marker(
-      repository: GITHUB_REPO,
-      milestone: new_version
+
+    freeze_milestone_and_move_assigned_prs_to_next_milestone(
+      milestone_to_freeze: new_version,
+      next_milestone: release_version_next
     )
 
     trigger_beta_build(branch_to_build: computed_release_branch_name)
--- a/fastlane/lanes/release.rb
+++ b/fastlane/lanes/release.rb
@@ -253,7 +253,7 @@ def freeze_milestone_and_move_assigned_prs_to_next_milestone(
 
   UI.message("Moved the following PRs to milestone #{next_milestone}: #{moved_prs.join(', ')}")
 
-  next unless is_ci
+  return unless is_ci
 
   moved_prs_info = if moved_prs.empty?
                      "No open PRs were targeting `#{milestone_to_freeze}` at the time of code-freeze."
--- a/fastlane/lanes/release.rb
+++ b/fastlane/lanes/release.rb
@@ -218,6 +218,49 @@ rescue StandardError => e
   UI.user_error!(error_message)
 end
 
+def freeze_milestone_and_move_assigned_prs_to_next_milestone(
+  milestone_to_freeze:,
+  next_milestone:,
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/simplenote-ios/milestone/68">4.53</a>. This milestone is due in less than 4 days.<br />
Please make sure to get it merged by then or assign it to a milestone with a later deadline.</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/lanes/release.rb
```diff
@@ -218,6 +219,49 @@ def create_release_backmerge_pr(version_to_merge:, next_version:)
...
+      from_milestone: milestone_to_freeze,
+      to_milestone: next_milestone,
+      comment: "Version `#{milestone_to_freeze}` has entered code-freeze. The milestone of this PR has been updated to `#{next_milestone}`."
+    )
+
+    # Add ❄️ marker to milestone title to indicate we entered code-freeze
+    set_milestone_frozen_marker(
+      repository: github_repository,
+      milestone: milestone_to_freeze
+    )
```
mokagio, line 247: I considered providing two methods with dedicated `begin rescue end` but decided to ask your thoughts on this first.

It might be overkill given the two steps always run together. 

One alternative might be to wrap each call in its own `begin rescue end` block, so that we can give a clear error message when one or the other fails.

AliSoftware: Actually one more strong reason to keep those 2 together is that they have to be run in that specific order and not the other way around, otherwise that would silently fail.

The reason for that is that if you rename the milestone first (from "X.Y" to "X.Y ❄️"), the search API we use for `update_assigned_milestone` to search for opened PRs targeting said milestone will return an empty list and fail to find any.
From my prior testing when I noticed this back in the Tumblr repo, I think that might be because GitHub will take a second to update its internal database and search index to account for the new name, as if you retry the same search query a couple seconds later it will find the PRs again (which is what makes me think about a database index update taking some time). I didn't dig into this too much back then so maybe the real reason is something else, but in any case I know that doing the search query to `update_assigned_milestone` first and only then rename the milestone via `set_milestone_frozen_marker` avoids the problem altogether.

(All that will definitively be worth a code comment to be left in the implementation of the action once we move this all to `release-toolkit` at some point…)

AliSoftware (approved):

fastlane/lanes/release.rb
mokagio: 👍 added a note here in the meantime 9705d7ac

PR #1646, opened 2024-09-02
Run code freeze in CI

mokagio:
What it say in the title.

I will use my admin privileges to merge on `trunk` in order to test the process. The code is all an adaptation from other projects that have already used it in production for a while.

I'd be thankful for a posthumous review and will address any feedback coming that way, @Automattic/apps-infrastructure . Thanks!


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">This PR has the <code>Releases</code> label: some checks will be skipped.</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/lanes/ci_releases.rb
```diff
@@ -0,0 +1,26 @@
...
+    buildkite_trigger_build(
+      buildkite_organization: BUILDKITE_ORGANIZATION,
+      buildkite_pipeline: BUILDKITE_PIPELINE,
+      branch: release_branch_name(release_version: release_version),
+      pipeline_file: File.join(PIPELINES_ROOT, 'complete-code-freeze.yml'),
+      message: "Complete code freeze for #{release_version}",
+      environment: { RELEASE_VERSION: release_version }
+    )
+  end
+end
```
mokagio, line 26: I decided to name this `ci_releases.rb` instead of `release_management_in_ci.rb` like other projects because it makes tab completion easier. 

Otherwise, with `re...` one gets hits for `relaese.rb` and `release_management_in_ci.rb`

fastlane/Fastfile
```diff
@@ -50,9 +50,10 @@ require_relative 'lib/env_manager'
 
 # Important: These need to be imported after all the constants have be defined because they access them.
```
spencertransier:
```suggestion
# Important: These need to be imported after all the constants have been defined because they access them.
```

mokagio: Thanks!

.buildkite/release-pipelines/start-code-freeze.yml
```diff
@@ -0,0 +1,17 @@
...
+        queue: android
+    command: |
+      echo '--- :git: Configure Git for release management'
+      .buildkite/commands/configure-git-for-release-management.sh
+
+      echo '--- :ruby: Setup Ruby tools'
+      install_gems
+
+      echo '--- :shipit: Start code freeze'
+      bundle exec fastlane code_freeze skip_confirm:true
```
spencertransier, line 18: We recently started disallowing retries from Buildkite ([example from DOApple](https://github.com/bloom/DayOne-Apple/pull/24201)) so that the release driver manages the build from Releases V2 instead: 

(I _think_ the indenting should work as expected, but with yaml, 🤷‍♂️)

```suggestion
      bundle exec fastlane code_freeze skip_confirm:true
    retry:
      manual:
        # If those jobs fail, one should always prefer re-triggering a new build from ReleaseV2 rather than retrying the individual job from Buildkite
        allowed: false
```

.buildkite/release-pipelines/complete-code-freeze.yml
```diff
@@ -0,0 +1,17 @@
...
+
+# The code freeze completion needs to run on macOS because it uses genstrings under the hood.
+agents:
+    queue: mac
+
+steps:
+  - label: Complete Code Freeze
+    key: complete_code_freeze
+    plugins: [$CI_TOOLKIT_PLUGIN]
+    command: .buildkite/commands/complete-code-freeze.sh $RELEASE_VERSION
```
spencertransier: See https://github.com/Automattic/simplenote-ios/pull/1646/files#r1742990234 for more context about this suggestion:

```suggestion
    command: .buildkite/commands/complete-code-freeze.sh $RELEASE_VERSION
    retry:
      manual:
        # If those jobs fail, one should always prefer re-triggering a new build from ReleaseV2 rather than retrying the individual job from Buildkite
        allowed: false
```

mokagio: 3ad44174

.buildkite/release-pipelines/start-code-freeze.yml
mokagio: 3ad44174

.buildkite/commands/complete-code-freeze.sh
```diff
@@ -0,0 +1,26 @@
...
+.buildkite/commands/configure-git-for-release-management.sh
+
+echo '--- :git: Checkout release branch'
+.buildkite/commands/checkout-release-branch.sh "$RELEASE_NUMBER"
+
+echo '--- :ruby: Setup Ruby tools'
+install_gems
+
+echo "--- :swift: Set up Swift Packages"
+install_swiftpm_dependencies
```
spencertransier: Is SPM needed for this pipeline? I didn't think `generate_strings_file_for_glotpress` or `update_appstore_strings` would need the packages to be present or else I would think Cocoapods would be required as well.

mokagio: Ah, you are right. Must have come across from a repo that either needed it or erroneously called it.

I'll need to chase that, too...

fastlane/lanes/release.rb
```diff
@@ -56,12 +54,13 @@
...
 
-    generate_strings_file_for_glotpress
-
     UI.important('Pushing changes to remote, configuring the release on GitHub, and triggering the beta build...')
     UI.user_error!("Terminating as requested. Don't forget to run the remainder of this automation manually.") unless skip_confirm || UI.confirm('Do you want to continue?')
 
-    push_to_git_remote(tags: false)
+    push_to_git_remote(
+      tags: false,
+      set_upstream: is_ci == false # only set upstream when running locally, useless in transient CI builds
```
spencertransier, line 62: For my learning, why do we only want to set the upstream when running locally?

.buildkite/commands/complete-code-freeze.sh
spencertransier: I see that WordPress iOS has it in the `complete-code-freeze.sh`, but it looks like it is needed there for the library localizations?

fastlane/lanes/release.rb
mokagio: Otherwise, we'll find ourselves on a `release/x.y` branch that does not point to `origin` and we won't be able to push pending configuring the upstream ourselves. It's just a convenience.

.buildkite/release-pipelines/start-code-freeze.yml
```diff
@@ -0,0 +1,23 @@
...
+# Variables used in this pipeline are defined in `shared-pipeline-vars`, which is `source`'d before calling `buildkite-agent pipeline upload`
+
+steps:
+  - label: Start Code Freeze
+    plugins:
+      - $CI_TOOLKIT_PLUGIN
+    # The first client to implement releases in CI was Android so the automation works in that queue.
+    # We might want to move it to a leaner one in the future.
+    agents:
+        queue: android
```
mokagio, line 13: After looking at @iangmaia's recent work adding a release-publishing pipeline to one of our apps that is not yet open source, it occurred to me that starting the code freeze could run on the `tumblr-metal` agent to leverage the more secure read vs read-write agent, since clearly it does not need macOS

However, I'm not 100% sure of the what's exactly required and want to get testing the code freeze in CI. I'll leave that as something to investigate next.

mokagio: 🤔 I might have to switch sooner 

image

https://buildkite.com/automattic/simplenote-ios/builds/1087#0191bb69-078a-43fe-9053-0291750baf22

PR #1647, opened 2024-09-02
Make Fastlane the single source of truth for release notes

mokagio:
Following up on what discussed at in https://github.com/Automattic/simplenote-ios/pull/1639#discussion_r1718251107. Also see internal ref pdnsEh-1Qv-p2.

Removes `Simplenote/Resources/release_notes.txt` and uses `fastlane/metadata/default/release_notes.txt` as the _single_ source of truth for release notes to send to localization.

### Implementation considerations

**Ignore `fastlane/metadata/default/release_notes.txt`? (No)** I briefly considered not tracking the file under Git because we generate it from `RELEASE-NOTES.txt`. I discarded that option because `RELEASE-NOTES.txt` contains "raw" release notes, or, let's say, release notes formatted for an audience of developers. The release notes in `fastlane/metadata/default/release_notes.txt` on the other hand, should be formatted for users to read.

**Ignore `fastlane/metadata/**/release_notes.txt`? (No)** When implementing the code to deleted the previous localized `release_notes.txt` entries and commit the result I also wondered whether we wouldn't be better served by Git-ignoring those files and downloading them from GlotPress when it's time to update the metadata on ASC. I decided against that for two reasons:

- When creating hotfixes, GlotPress might have already been updated with the new notes and no longer have the ones we need. Having the localizations in the repo ensures we can easily reuse the release notes for the version which the hotfix targets
- It's handy to have the localizations show in the diff to "smoke test" them. There have been various occasions in the past where release managers noticed issues in the release notes this way.

**Other pieces to update?**

- We don't need to update the GlotPress upload cron job because it reads the `.pot` file, which has not moved. 
- I decided not to update the script that generates the release checklist for this app in favor of deleting it because of the migration to our internal GUI tool.
- Our internal GUI tool will need to be updated to 1) use the new path in the user facing instructions; 2) use the new path in the setting that the automation to generate the announcement post uses (I'll need @hannahtinkler guidance to find it)

### Test
I created a test branch on top of this one, hacked `code_freeze` to run the automation and verified the results. See XXX

### Review
Only one infra engineer required to review these changes, but anyone can perform the review.

### Release


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/simplenote-ios/milestone/68">4.53</a>. The due date for this milestone has already passed.<br />
Please assign it to a milestone with a later deadline or check whether the release for this milestone has already been finished.</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/Fastfile
```diff
@@ -35,17 +35,16 @@ SECRETS_ROOT = File.join(Dir.home, '.configure', 'simplenote-ios', 'secrets')
...
-
 GLOTPRESS_BASE_URL = 'https://translate.wordpress.com/projects'
 # Notice the trailing / is required.
 # Without it, GlotPress will redirect to the version with /
 GLOTPRESS_APP_STRINGS_PROJECT_URL = "#{GLOTPRESS_BASE_URL}/simplenote/ios/".freeze
 GLOTPRESS_STORE_METADATA_PROJECT_URL = "#{GLOTPRESS_APP_STRINGS_PROJECT_URL}release-notes/".freeze
 
+APP_RESOURCES_DIR = File.join(PROJECT_ROOT_FOLDER, 'Simplenote', 'Resources')
 STORE_METADATA_FOLDER = File.join(PROJECT_ROOT_FOLDER, 'fastlane', 'metadata')
+DEFAULT_METADATA_FOLDER = File.join(STORE_METADATA_FOLDER, 'default')
```
mokagio: I'm not sure about this name... Maybe `DEFAULT_LOCALE_METADATA_FOLDER` or `STORE_METADATA_DEFAULT_LOCALE_FOLDER` would be clearer?

fastlane/lanes/release.rb
```diff
@@ -39,12 +39,22 @@
...
+    changelog_path = File.join(PROJECT_ROOT_FOLDER, 'RELEASE-NOTES.txt')
     extract_release_notes_for_version(
       version: new_version,
-      release_notes_file_path: File.join(PROJECT_ROOT_FOLDER, 'RELEASE-NOTES.txt'),
-      extracted_notes_file_path: File.join(PROJECT_ROOT_FOLDER, 'Simplenote', 'Resources', 'release_notes.txt')
+      release_notes_file_path: changelog_path,
+      extracted_notes_file_path: RELEASE_NOTES_SOURCE_PATH
+    )
+    # Add a new section to the changelog for the version _after_ the one we are code freezing
+    ios_update_release_notes(
```
mokagio, line 54: I had a quick look at the source for the only difference from its `android_update_release_notes` is the _internal_ logic they use to compute the version value to add at the top of the changelog file.

Given our move to the `Versioning` module, I think it would be appropriate to consolidate the two action into a single one, which receives either a precomputed next version (not to be confused with current `new_version` which is the value of the version we are code freezing and that the actions bump internally) or a `Versioning` object(s) to use for the computation.

I didn't go down that path in the interest of keeping momentum with the changes in this PR.

fastlane/Fastfile
```diff
@@ -73,30 +72,10 @@ end
 ########################################################################
 ENV['PROJECT_NAME'] = 'Simplenote'
 ENV['PROJECT_ROOT_FOLDER'] = PROJECT_ROOT_FOLDER
-ENV['APP_STORE_STRINGS_FILE_NAME'] = 'AppStoreStrings.pot'
```
mokagio, line 76: No longer needed (:phew:)

https://github.com/wordpress-mobile/release-toolkit/blob/b6eb4278f96779c47888e1b57af830e9258c083f/MIGRATION.md#L34

fastlane/Fastfile
```diff
@@ -73,30 +72,10 @@ end
...
-      app_store_keywords: File.join(source_metadata_folder, 'keywords.txt')
-    }
-
-    ios_update_metadata_source(
-      po_file_path: File.join(APP_RESOURCES_DIR, 'AppStoreStrings.pot'),
-      source_files: files,
-      release_version: version
-    )
-  end
-
```
mokagio, line 99: Moved to the dedicated `localization.rb`

AliSoftware:
> Ignore fastlane/metadata/**/release_notes.txt? (No)

A 3rd reason I like for not git-ignoring those is simple so that they are tracked in git. So we can see what translations were used in an old commit and have the history.

fastlane/Fastfile
AliSoftware: I like `STORE_METADATA_DEFAULT_LOCALE_FOLDER` better indeed.

fastlane/lanes/localization.rb
```diff
@@ -90,12 +106,6 @@
   end
 
   lane :download_localized_metadata_from_glotpress do
-    # FIXME: We should make the `fastlane/metadata/default/release_notes.txt` path be the source of truth for the original copies in the future.
-    # (will require changes in the `update_appstore_strings` lane, the Release Scenario, the MC tool to generate the announcement post…)
-    #
-    # In the meantime, just copy the file to the right place for `deliver` to find, for the `default` pseudo-locale which is used as fallback
-    FileUtils.cp(RELEASE_NOTES_SOURCE_PATH, File.join(STORE_METADATA_FOLDER, 'default', 'release_notes.txt'))
-
```
AliSoftware, line 98: 🎉 🧹

fastlane/lanes/release.rb
```diff
@@ -39,12 +39,22 @@
 
     new_version = release_version_current
 
+    # Delete all release notes metadata, including the source of truth.
+    # We'll generate a new source of truth next, and the localized versions will be re-downloaded once translated on GlotPress.
+    # It's important we delete them, otherwise we risk using old release notes for locales that won't get translated in time for the release finalization.
```
AliSoftware, line 44: 👍

fastlane/lanes/release.rb
```diff
@@ -280,3 +290,15 @@ def report_milestone_error(error_title:)
 
   buildkite_annotate(style: 'warning', context: 'error-with-milestone', message: error_message) if is_ci
 end
+
+def delete_all_metadata_release_notes(store_metadata_folder: STORE_METADATA_FOLDER)
+  pattern = File.join(store_metadata_folder, '**', 'release_notes.txt')
+  Dir[pattern].each { |path| File.delete(path) }
+  git_add(path: pattern, shell_escape: true)
```
AliSoftware: I'm always confused with that `shell_escape` parameter of `git_add` action. Doesn't that mean that it will shell-escape the pattern, and thus replace `**` with `\*\*` when calling the `git add` command… which will then interpret those star characters as literals (since they'd be escaped) and not as a pattern?

fastlane/lanes/release.rb
```diff
@@ -280,3 +290,15 @@ def report_milestone_error(error_title:)
...
+  git_commit(
+    path: pattern,
+    message: 'Delete release notes source and localization before code freeze',
+    # Even if no locale was translated in the previous cycle, default/relaese_notes.txt should always be present, and therefore deleted at this stage.
+    allow_nothing_to_commit: false
+  )
+end
```
AliSoftware: wdyt of this instead, to avoid using the pattern itself in the git commands, and only use them for the `glob` instead?

```suggestion
def delete_all_metadata_release_notes(store_metadata_folder: STORE_METADATA_FOLDER)
  files = Dir.glob(File.join(store_metadata_folder, '**', 'release_notes.txt'))
  files.each { |path| File.delete(path) }
  git_add(path: files)
  git_commit(
    path: files,
    message: 'Delete release notes source and localization before code freeze',
    # Even if no locale was translated in the previous cycle, default/relaese_notes.txt should always be present, and therefore deleted at this stage.
    allow_nothing_to_commit: false
  )
end
```

AliSoftware (approved):
Just left a code change suggestion (and disabled auto-merge to git you the opportunity to apply it if you like it), otherwise LGTM 👍

fastlane/lanes/release.rb
mokagio: Sounds good!

fastlane/lanes/release.rb
```diff
@@ -280,3 +290,15 @@ def report_milestone_error(error_title:)
...
   buildkite_annotate(style: 'warning', context: 'error-with-milestone', message: error_message) if is_ci
 end
+
+def delete_all_metadata_release_notes(store_metadata_folder: STORE_METADATA_FOLDER)
+  files = Dir.glob(File.join(store_metadata_folder, '**', 'release_notes.txt'))
+  files.each { |path| File.delete(path) }
+  git_add(path: files)
+  git_commit(
+    path: files,
+    message: 'Delete release notes source and localization before code freeze',
```
AliSoftware, line 300: nitpick for a future PR:

```suggestion
    message: 'Delete release notes of previous versions before code freeze'
```

fastlane/lanes/release.rb
```diff
@@ -280,3 +290,15 @@ def report_milestone_error(error_title:)
...
+    # Even if no locale was translated in the previous cycle, default/relaese_notes.txt should always be present, and therefore deleted at this stage.
```
AliSoftware, line 301:
```suggestion
    # Even if no locale was translated in the previous cycle, default/release_notes.txt should always be present, and therefore deleted at this stage.
```

mokagio:
Thanks for the suggestions @AliSoftware. Addressed via 439bc6c9

PR #1655, opened 2024-09-11
Add CI pipeline to generate a new beta

mokagio:
### Fix
Adds the CI pipeline required to trigger the new beta deployment from CI and updates the `new_beta_release` lane to also create the backmerge PR.

To get there, I ended up updating a lot more code, including:

- Upgrading to `release-toolkit` 12.0
- Consolidating to `create_release_backmerge_pull_request`
- Remove some unused lanes
- Move more lanes to dedicated files and adopt named block parameters

See also paaHJt-6Zx-p2 paaHJt-6ps-p2

### Test
This will be tested properly only once landed on the release branch.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is assigned to the milestone <a href="https://github.com/Automattic/simplenote-ios/milestone/69">4.54</a>. This milestone is due in less than 4 days.<br />
Please make sure to get it merged by then or assign it to a milestone with a later deadline.</td>
    </tr>
  </tbody>
</table>

</p>

fastlane/Fastfile
```diff
@@ -77,206 +77,20 @@ ENV['PUBLIC_CONFIG_FILE'] = VERSION_FILE_PATH
...
-  #
-  # Example:
-  # bundle exec fastlane finalize_hotfix_release skip_confirm:true
-  #####################################################################################
-  desc 'Performs the final checks and triggers a release build for the hotfix in the current branch'
-  lane :finalize_hotfix_release do |options|
-    ios_finalize_prechecks(options)
-    version = ios_get_app_version(public_version_xcconfig_file: VERSION_FILE_PATH)
-    trigger_release_build(branch_to_build: "release/#{version}")
-  end
```
mokagio, line 117: 👆  `new_hotfix_release` and `finalize_hotfix_release` have been moved to `release.rb`.

👇 the other `build_and_upload_...` lanes below were unused and have been removed.

fastlane/Fastfile
```diff
@@ -77,206 +77,20 @@ ENV['PUBLIC_CONFIG_FILE'] = VERSION_FILE_PATH
...
     # Skip screenshots by default. The naming is "with" to make it clear that
     # callers need to opt-in to adding screenshots. The naming of the deliver
     # (upload_to_app_store) parameter, on the other hand, uses the skip verb.
-    with_screenshots = options.fetch(:with_screenshots, false)
     skip_screenshots = !with_screenshots
 
     upload_to_app_store(
       app_identifier: APP_STORE_BUNDLE_IDENTIFIER,
-      app_version: ios_get_app_version(public_version_xcconfig_file: VERSION_FILE_PATH),
+      app_version: release_version_current,
```
mokagio, line 92: This was required because of the `release-toolkit` 12.0 upgrade.

In hindsight, the upgrade was not required because `create_release_backmerge_pull_request` was already available in version 11, but it ought to have been done sooner or later, so might as well do it now.

fastlane/lanes/build.rb
```diff
@@ -61,7 +61,7 @@
   next unless create_release
 
   version = beta_release ? build_code_current : release_version_current
-  create_release(
+  create_github_release(
```
mokagio, line 64: Another 12.0 upgrade requirement.

fastlane/lanes/ci_releases.rb
```diff
@@ -23,4 +23,15 @@
...
+  lane :trigger_new_beta_in_ci do |release_version: release_version_current|
+    buildkite_trigger_build(
+      buildkite_organization: BUILDKITE_ORGANIZATION,
+      buildkite_pipeline: BUILDKITE_PIPELINE,
+      branch: release_branch_name(release_version: release_version),
+      pipeline_file: File.join(PIPELINES_ROOT, 'new-beta-release.yml'),
+      message: "New beta for #{release_version}",
+      environment: { RELEASE_VERSION: release_version }
+    )
+  end
```
mokagio: So... this lane is not strictly necessary. After all, we use our internal release management app to trigger release builds.

I added it here because I plan to use it to test the new beta workflow once this PR lands as I don't want to reset the scenario for 4.53 in the release management app.

One could argue that it's useful to have these here as a fallback if the app is offline or something, but I think it might be overkill. In fact, maybe is best if I delete the whole file and use the Buildkite GUI to trigger this during testing.

What do you think?

WordPress has these lanes too, but that's because they were added before Rv2 was available. If we decide to remove the lanes from here, we shall delete them from there, too.

fastlane/lanes/release.rb
```diff
@@ -221,6 +224,77 @@
     end
   end
 
+  desc 'Creates a new hotfix branch for the given version:x.y.z. The branch will be cut from the tag x.y of the previous release'
+  lane :new_hotfix_release do |version:, skip_confirm: false, skip_prechecks: false|
+    ensure_git_status_clean unless skip_prechecks
+
+    # Parse the provided version into an AppVersion object
```
mokagio:
```suggestion
```

fastlane/lanes/release.rb
```diff
@@ -221,6 +224,76 @@
...
+    hotfix_version = release_version_current
+
+    UI.important("Will triggrer hotfix build for version #{hotfix_version}")
+    UI.user_error!("Terminating as requested. Don't forget to run the remainder of this automation manually.") unless skip_confirm || UI.confirm('Do you want to continue?')
+
+    trigger_release_build(branch_to_build: release_branch_name(release_version: hotfix_version))
+
+    create_backmerge_prs
+
+    # Close hotfix milestone
```
mokagio:
```suggestion
```

fastlane/lanes/release.rb
```diff
@@ -262,18 +332,40 @@ def trigger_buildkite_release_build(branch:, beta:)
   buildkite_annotate(style: 'info', context: 'trigger-release-build', message: message)
 end
 
-def create_release_backmerge_pr(version_to_merge:, next_version:)
+def create_backmerge_pr!
+  pr_urls = create_backmerge_prs
+
+  return pr_urls unless pr_urls > 1
```
iangmaia: I haven't tested, but don't you need to compare the size?
```suggestion
  return pr_urls unless pr_urls.length > 1
```

fastlane/lanes/release.rb
```diff
@@ -262,18 +332,40 @@ def trigger_buildkite_release_build(branch:, beta:)
   buildkite_annotate(style: 'info', context: 'trigger-release-build', message: message)
 end
 
-def create_release_backmerge_pr(version_to_merge:, next_version:)
+def create_backmerge_pr!
```
iangmaia, line 335: Given an error is thrown in an alternative flow (isn't guaranteed to always happen), does it still make sense to add a `!` to the method name?

fastlane/lanes/release.rb
```diff
@@ -262,18 +332,40 @@ def trigger_buildkite_release_build(branch:, beta:)
...
+    Unexpectedly opened more than one backmerge pull request. URLs:
+    #{pr_urls.map { |url| "- #{url}" }.join("\n")}
+  ERROR
+  buildkite_annotate(style: 'error', context: 'error-creating-backmerge', message: backmerge_error_message) if is_ci
+  UI.user_error!(backmerge_error_message)
+end
+
+# Notice the plural in the name.
+# The action this method calls may create multiple backmerge PRs, depending on how many releas branches with version greater than the source are in the remote.
+def create_backmerge_prs
```
iangmaia, line 350: Interesting approach. I wonder if this is really safe to assume and, if so, if this approach should be ported to other apps.

fastlane/lanes/release.rb
```diff
@@ -262,18 +332,40 @@ def trigger_buildkite_release_build(branch:, beta:)
...
 rescue StandardError => e
   error_message = <<-MESSAGE
-    Error creating backmerge pull request: #{e.message}
-    If this is not the first time you are running the release task, the backmerge PR for the version `#{version_to_merge}` might have already been previously created.
-    Please close any previous backmerge PR for `#{version_to_merge}`, delete the previous merge branch, then run the release task again.
+    Error creating backmerge pull request:
+
+    ```
+    #{e.message}
+    ```
```
iangmaia:
```suggestion
    #{e.message}
```

One thing I realize happened is that the ``` does something strange, where the entire message (above and below) becomes part of this block. For now in new apps I'm working on (WPiOS/WPAndroid), I've just removed it and let only the line breaks, but perhaps we can think of a better formatting.

fastlane/lanes/ci_releases.rb
iangmaia: Hmm I'm leaning towards deleting it.
I myself was confused for some time looking at the WPiOS automation, as we even mention it in the task in Rv2:
```
Run <code>bundle exec fastlane trigger_code_freeze_in_ci</code>.
```

While this lane is never really called, unless someone would do it directly (which [almost?] never happens these days).

fastlane/lanes/release.rb
AliSoftware: @iangmaia [the `create_release_backmerge_pull_request` action from `release-toolkit` is already the one taking care of creating multiple PRs if there are multiple `release/*` branches with a greater version](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/create_release_backmerge_pull_request_action.rb#L61-L73). So no need to port anything to other repos, it's already handled as long as you use the toolkit action.

fastlane/lanes/release.rb
```diff
@@ -262,18 +332,40 @@ def trigger_buildkite_release_build(branch:, beta:)
...
+  backmerge_error_message = UI.user_error! <<~ERROR
+    Unexpectedly opened more than one backmerge pull request. URLs:
+    #{pr_urls.map { |url| "- #{url}" }.join("\n")}
+  ERROR
+  buildkite_annotate(style: 'error', context: 'error-creating-backmerge', message: backmerge_error_message) if is_ci
+  UI.user_error!(backmerge_error_message)
+end
+
+# Notice the plural in the name.
+# The action this method calls may create multiple backmerge PRs, depending on how many releas branches with version greater than the source are in the remote.
```
AliSoftware:
```suggestion
# The action this method calls may create multiple backmerge PRs, depending on how many release branches with version greater than the source are in the remote.
```

mokagio: Thanks @AliSoftware . I don't know what it is with the word "release" but I seem to have a high typo ratio for it 🙃

fastlane/lanes/release.rb
mokagio: Woops! 😳 

Thanks @iangmaia

mokagio: For the record: I updated @iangmaia original suggestion from `size` to `length` and applied it from GitHub to maintain attribution.

Rationale: [`size` is an alias for `length`](https://ruby-doc.org/3.3.5/Array.html#method-i-size) and as far as I've seen, `length` is the most commonly used.

fastlane/lanes/ci_releases.rb
mokagio: Done via 24fd4a6471928a84e70af3d1db58f3d8f725b484. See also https://github.com/wordpress-mobile/WordPress-iOS/pull/23589

fastlane/lanes/release.rb
mokagio: Removing the code fencing sounds good to me. Hopefully we'll see little of this error, but if it happens often and users found it poorly rendered, we can address then.

fastlane/lanes/release.rb
iangmaia: Yeah I know, but I meant **throwing an error when there are multiple (unexpected) PRs**, which is what @mokagio is doing differently here.

mokagio:
I'm going to admin-merge this because I feel like the various comments have been addressed and there was nothing that was a "request changes". 

In particular, the backmerge chages will be useful in the upcoming 4.54 code freeze. As always, I'll be happy to follow up to address anything I might have missed.

mokagio:
😓 because I used an integration branch in #1659, the `release/4.53` branch was still around and that's where this PR landed 🤦‍♂️ 

I'll cherry-pick the commits from here into a new branch against `trunk` and admin-merge that...

PR #1656, opened 2024-09-11
Add fall back login with username and password option to login

### LINKED ISSUE #1654: Add label to log in with email with option to use password log in
Beneath the Log in with email button, let's add this text:

`We'll email you a code to log in, or you can log in manually.`

Link `log in manually` with a button to the username/password form.

charliescheer:
### Fix
We have received some feedback about the new login flow that some people thought we had dropped the login with password option.  We have decided to add a button on the login with email page where users can choose to login with email and password instead.  This surfaces this option for users who wants it and doesn't require receiving the login email to before you can enter your password.  This also helps in circumstances where GAE might be down and login with email isn't possible.

| Login | Login with pass |
|--------|--------|
|  |  | 

### Test
1. Clean install Simplenote
2. On the onboarding page, tap on Sign up.  Make sure that you see the username field and the terms and conditions info
3. Go back, tap on Log In.

- [ ] confirm that you see the option `We'll email you a code to log in, or you can log in manually` 
- [ ] Tap on that and confirm you are shown a login page with username and password fields
- [ ] confirm that you can login from that page if you enter both pieces of info

Smoke test the magic links flow and make sure it still looks as expected

### Review


jleandroperez (commented):
@charliescheer Just one minor note!!

Nice work BTW!!!

### Header
1. Launch Simplenote iOS
2. Press on `Log In`
3. Press on `log in manually`

As a result, you'll get the header `Log in with email failed (...)`.
We should probably display no header here!

charliescheer:
@jleandroperez Good call.  Dropped the header

jleandroperez (approved):
Looks great!!

:shipit:

PR #1686, opened 2024-11-19
[Tooling] Update CI image from `xcode-16.1-rc.1` to `xcode-16.1-v4`

spencertransier:
## What it does
This updates the CI image from `xcode-16.1-rc.1` to `xcode-16.1.v4`. The original 16.1 image only had 92GB of space. The v4 image was upgraded to use 120GB instead. 

It also updates the CI Toolkit from `3.4.2` to `3.7.1`. 

### How to test
Good to go as long as CI is green.


.buildkite/shared-pipeline-vars
```diff
@@ -5,7 +5,8 @@
 
 # The ~> modifier is not currently used, but we check for it just in case
 XCODE_VERSION=$(sed -E -n 's/^(~> )?(.*)/xcode-\2/p' .xcode-version)
-export CI_TOOLKIT_PLUGIN_VERSION="3.4.2"
+export CI_TOOLKIT_PLUGIN_VERSION="3.7.1"
 
-export IMAGE_ID="$XCODE_VERSION"
+# Note: `-v4` suffix was added to use xcode-16.1-v4 image; remember to remove that suffix during the next Xcode update
+export IMAGE_ID="$XCODE_VERSION-v4"
```
AliSoftware: 💄  Nitpick: in most other repos we insert the `xcode-` prefix during the `export IMAGE_ID=` instead of via the `sed`, i.e.:

```bash
XCODE_VERSION=$(sed -E 's/^~> ?//' .xcode-version)
…
export IMAGE_ID="xcode-$XCODE_VERSION-v4"
```

This not only makes the `sed` command simpler (and thus more readable), but it also better matches the variable names. Indeed, with that approach `XCODE_VERSION` contains just the Xcode version itself, without the `xcode-` prefix. And it's only when assigning the `IMAGE_ID` value that we add the `xcode-` prefix because that's how we name our VM images.

AliSoftware (approved):
Left a nitpick to adjust consistency in how we do it in other repos, but approving to unblock

.buildkite/shared-pipeline-vars
spencertransier: Updated in https://github.com/Automattic/simplenote-ios/commit/0ecc10dd1c8e68ad46b7f8419f4ac20ab19b1e74

PR #1690, opened 2024-12-21
TextKit 2 - Implement basic TK2 class and graceful fallback to TK1

charliescheer:
### Fix
This PR is the first step in implementing Textkit 2 into Simplenote.  Why do we need this?  Well primarily it is to keep up with the next paradigms in apple development.  Also with the addition of Textkit 2 we should get all of the new Apple intelligence features for free.  Lastly, by removing the reliance on glyphs and glyph ranges in TextKit 2, we should be able to fix many of the out of range crashes that we sometimes see with multi bit character languages.

Here I have taken the first steps.  I have:
- created the TextKit 2 layout properties if the device can run them
- layout the text with the same appearance as text in TextKit 1

NOTE: this PR is broken for iOS 16 and older.  That is fixed in #1691 

### Test
1.  Launch the app, open a note
- [ ] Confirm that the text is styled correctly for SN
- [ ] Confirm that if you tap into the text you are able to become first responder
- [ ] Confirm that you can enter text
- [ ] Confirm that after you tap into the text you see this warning in the console `UITextView 0x10500fe00 is switching to TextKit 1 compatibility mode because its layoutManager was accessed` but the text is still styled

NOTE: When the editor becomes first responder the text will lose formatting because it has fallen back to text kit 1.  

### Review
> Only one developer and required to review these changes, but anyone can perform the review.

### Release
These changes will require an update, but not yet


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/SPFallbackTextContainer.swift
```diff
@@ -0,0 +1,19 @@
...
+            guard let layoutManager,
+                  let newManagerStorage = layoutManager.textStorage,
+                  let textView,
+                  textView.interactiveTextStorage.layoutManagers.isEmpty else {
+                return
+            }
+
+            textView.interactiveTextStorage.addLayoutManager(layoutManager)
+            textView.interactiveTextStorage.setAttributedString(newManagerStorage)
+            textView.text = newManagerStorage.string
```
jleandroperez: Can we move this so that it's explicit during the initialization flow, rather than a lazy thing?

Simplenote/SPFallbackTextContainer.swift
```diff
@@ -0,0 +1,19 @@
+class SPFallbackTextContainer: NSTextContainer {
```
jleandroperez: Annoying comment: Can we add either an `UIKit` import or something, so that the class declaration isn't the very first one in the file?

jleandroperez (changes requested):
@charliescheer I spotted a couple issues (while running on an iOS 18 device):

### Bug: No Text + Crasher
1. Add a new Note with a single line
2. Go back to the Notes List
3. Open the note again

As a result, the text is not visible. If you attempt to type anything, Simplenote crashes.

https://github.com/user-attachments/assets/07f2bafb-3d16-4ebd-983a-d99922e8ac3e

Screenshot 2024-12-23 at 12 20 51 PM

### Bug: No Style
1. Add a new Note
2. Write more than one line

As a result, the H1 style is never applied.

https://github.com/user-attachments/assets/d8945ee2-31a6-4965-a892-a8a435706b51

charliescheer:
So the problem with the crashing was how the fallback was being setup.  It could mean that there was no text when changes were applied, so the ranges would always be out of bounds.

Discussed in slack and decided instead to approach this in a feature branch rather than building a fallback.  In theory if we are not using TK1 properties then we don't need to fallback.  We don't use them too much already, so if we patch over those for iOS 17 and higher then we should be fine to stay in TK2 always.  Updating the PR to reflect that

charliescheer:
As for the editing of the header in a new note, this PR is just setting up the TK2 properties and displaying existing notes, it doesn't yet handle editing. That will come later.

Simplenote/SPTextView.swift
```diff
@@ -0,0 +1,94 @@
...
+
+        if #available(iOS 16.0, *) {
+            let textLayoutManager = NSTextLayoutManager()
+            let contentStorage = NSTextContentStorage()
+            contentStorage.delegate = self
+            contentStorage.addTextLayoutManager(textLayoutManager)
+            textLayoutManager.textContainer = container
+
+        } else {
+            layoutManager.addTextContainer(container)
```
jleandroperez, line 26: Quick note: `layoutManager` may not be init'ed by the time it's called here

Simplenote/SPTextView.swift
```diff
@@ -0,0 +1,94 @@
...
+
+        return container
+    }
+}
+
+// MARK: NSTextContentStorageDelegate
+//
+extension SPTextView: NSTextContentStorageDelegate {
+    public func textContentStorage(_ textContentStorage: NSTextContentStorage, textParagraphWith range: NSRange) -> NSTextParagraph? {
+        guard let originalText = textContentStorage.textStorage?.attributedSubstring(from: range) as? NSMutableAttributedString else {
```
jleandroperez, line 38: Missing `.mutabelCopy()` here

jleandroperez (approved):
Discussed over Slack, the `mutableCopy()` change is coming up in a follow up.

Please feel free to :shipit: once the `NSLayoutManager` initialization is fixed (fallback flow!).

Nice work @charliescheer !!!

PR #1691, opened 2024-12-21
Textkit 2: Fixed gesture recognizer so tapping into text does not fallback to text kit 1

charliescheer:
### Fix
In my previous PR with TextKit 2, #1690 , I added the basic setup for SPTextView to use Text Kit 2 if available, and I created a mechanism to fallback to text kit 1 without losing all formatting.  However in that PR if you did anything other than view a note you were immediately dropped back into Text Kit 1.  

In this PR I have updated gesture recognizer to not use `layoutManager`  so the view does not drop back to text kit 1.  This should allow you to tap into, interact with, edit, and save text while still in text kit 2.

Next up: Search

### Test
1. Select a note
2. Tap on any part of the text and start typing in it.  Do some basic editing of the note

- [x] confirm that you do not see `is switching to TextKit 1 compatibility mode because its layoutManager was accessed` in the console
- [x] Also smoke test iOS 16 or older to make sure the old processes still work

### Review

### Release


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/SPTextView.swift
```diff
@@ -8,11 +8,13 @@
...
 extension SPTextView {
     @objc
-    func setupTextContainer(with textStorage: SPInteractiveTextStorage) -> NSTextContainer {
+    func setupTextContainer() -> NSTextContainer {
         let container = NSTextContainer(size: .zero)
         container.widthTracksTextView = true
         container.heightTracksTextView = true
 
+        let textStorage = SPInteractiveTextStorage()
+        interactiveTextStorage = textStorage
```
jleandroperez: I'm not sure I'm fond of this change: the API name implies you're setting up a `TextContainer`, but it's got the side effect of also initializing the `interactive Text Storage`.

Do you think it may be better if we moved this into a `NSTextContainer` extension, and keep the Storage initialization elsewhere?

If the `SPTextView`'s **init** was Swift based, the NSTextContainer extension approach would build correctly (otherwise, calling `setupTextContainer` wouldn't work, because it's being called before all properties get initialized)

jleandroperez (approved):
No blockers, just one comment on the way the Storage is init'ed.

Feel free to :shipit: when ready!

Simplenote/SPTextView.swift
charliescheer: 👍

PR #1693, opened 2025-01-20
Tweak code signing lanes naming and make them all public

mokagio:
This was done in the context of regenerating soon expiring certificates, see pdnsEh-1ZG-p2 and https://github.com/wordpress-mobile/WordPress-iOS/pull/23991

Renamed the existing App Store and Enterprise code signing lanes to start with `update_certs_and_profiles` like the lane that calls them both, so the relationship between the three lane is clear in code completion. Also made them public so that they can be called individually without having to hack the `Fastfile`.

I also removed the unused lane for internal code signing. I will follow up this with a proposal to remove all the Internal related code and build configuration. I thought this had been discussed already but I couldn't find the conversation. The latest Internal build was shipped in 2021...

image


fastlane/Fastfile
```diff
@@ -419,45 +419,32 @@ end
...
-    readonly: options.fetch(:readonly, true)
-  )
-end
-
-# Downloads all the required certificates and profiles (using `match`) for the internal builds in the enterprise account.
-# Optionally, it can create any new necessary certificates or profiles.
-#
-# @option [Boolean] readonly (default: true) Whether to only fetch existing certificates and profiles, without generating new ones.
-#
-private_lane :internal_code_signing do |options|
```
iangmaia, line 445: > The latest Internal build was shipped in 2021

👌

iangmaia (approved):
👍

PR #1712, opened 2025-06-11
Provision SwiftLint via SwiftPM instead of CocoaPods

mokagio:
See https://linear.app/a8c/issue/AINFRA-449

### Test
Checkout locally and run `bundle install && bundle exec pod install && bundle exec rake lint`

You should see something like, possibly with additional logs such as removing SwiftLint from your pods.

image

> [!NOTE]  
> The UI tests failure is a long standing issue and unrelated with the changes in this PR.

### Review
> Only one developer required to review these changes, but anyone can perform the review.

### Release

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -3306,7 +3306,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "[ $CI ] && exit 0\n./Pods/SwiftLint/swiftlint\n";
+			shellScript = "[ $CI ] && exit 0\nrake lint\n";
 		};
 		B553F5A11D10474600F7E397 /* Delete Frameworks folder (temporary fix for CocoaPods) */ = {
 			isa = PBXShellScriptBuildPhase;
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ default.profraw
 # Everything without a .enc extension is ignored
 .configure-files/*
 !.configure-files/*.enc
+
+# BuildTools
+BuildTools/.build
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,7 +1,8 @@
-swiftlint_version: 0.54.0
+swiftlint_version: 0.58.2
 
 # Project configuration
 excluded:
+  - DerivedData
   - Pods
   - fastlane
   - vendor
@@ -49,8 +50,8 @@ only_rules:
 
 # Rules configuration
 
-control_statement: 
-  severity: error 
+control_statement:
+  severity: error
 
 custom_rules:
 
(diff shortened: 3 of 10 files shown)


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -3306,7 +3306,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "[ $CI ] && exit 0\n./Pods/SwiftLint/swiftlint\n";
+			shellScript = "[ $CI ] && exit 0\nrake lint\n";
```
AliSoftware, line 3309: I was wary of if running `rake` from an Xcode build phase would work as expected, given that:
 - When Xcode runs shell scripts it runs them in a separate shell that might not have the same env / settings as your Terminal.app's shell (e.g. `$PATH` and where `rake` is installed)
 - That shell script runs in `/bin/sh`, while macOS users are using `zsh`, so all their customization and setup of their shell is in `.zshrc` and not `.bashrc`. So again, any custom `$PATH` or `rbenv` and `bundle`-related config etc might be set up properly in `.zshrc` but not in `.bashrc` for the `/bin/sh` shell to pick up, which could lead to this `rake lint` command to fail
 - Notice how you run `rake lint`, not `bundle exec rake lint`. Which means one might accidentally run this command with a different version of `rake` than the one in the `Gemfile.lock` bundle of this repo. Probably not a big deal (the most important thing is for `swiftlint` to be using the right version, and `rake` doesn't have that many updates anyway, and if it does, nothing likely breaking anything) but still inconsistent.

---
It still worked when I tested building the project with Xcode on my Mac:
image

But I think that might be a happy coincidence in how my Mac is set up, and that might not be the case for every developer? I wonder if switching to `shellPath = /bin/zsh;` and using `bundle exec rake lint` wouldn't be better (maybe with a need to `source ~/.zshrc` explicitly first, in case Xcode running this shell in non-interactive mode would prevent that from loading… not sure, always confused with the rc files and when they're loaded wrt to interactive vs non-interactive shells etc)

AliSoftware (approved):
Approving to unblock and because it worked on my machine, but left a note about running `rake lint` from the Xcode shell script build phase and its potential issues depending on developers' setup

Simplenote.xcodeproj/project.pbxproj
mokagio: I remember looking into this with Jeremy a while back and coming to the conclusion that calling `rake` from Xcode works because of a number of happy coincidences in how Apple set up Ruby in macOS.

In the long run, consistently with our goal of removing Ruby from day-to-day DX, I think we'd be better off using `make` which we know would be consistent without depending on Ruby setup.

mokagio: Moved this discussion internally to paaHJt-8yZ-p2

PR #1713, opened 2025-06-12
Fetch ZIPFoundation with SwiftPM instead of CocoaPods

mokagio:
### Fix
This is part of our effort to move away from CocoaPods, which is now in maintenance mode, in favor of Swift Package Manager. See https://linear.app/a8c/issue/AINFRA-448.

### Test
I made sure the app built locally.

### Review

### Release

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -59,6 +59,7 @@
 		3FA323292D63E5A90017AE66 /* XCUITestHelpers in Frameworks */ = {isa = PBXBuildFile; productRef = 3FA323282D63E5A90017AE66 /* XCUITestHelpers */; };
 		3FA3232B2D63E6F50017AE66 /* XCUITestHelpers in Frameworks */ = {isa = PBXBuildFile; productRef = 3FA3232A2D63E6F50017AE66 /* XCUITestHelpers */; };
 		3FA6013C242C5AAE0068FC52 /* SimplenoteScreenshots.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3FA6013B242C5AAE0068FC52 /* SimplenoteScreenshots.swift */; };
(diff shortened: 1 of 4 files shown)


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -6180,6 +6145,14 @@
 				minimumVersion = 0.3.0;
 			};
 		};
+		3FA8817C2DFAADD200E1F4A3 /* XCRemoteSwiftPackageReference "ZIPFoundation" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/weichsel/ZIPFoundation";
+			requirement = {
+				kind = exactVersion;
+				version = 0.9.15;
```
mokagio, line 6153: I used exact to match the version resolved in `Podfile.lock` and avoid introducing changes because of fetching a newer version (at the time of writing, the latest is 0.9.19).

We can always follow up with a dedicated PR to update.

AliSoftware:
The [UI Tests CI job is failing](https://buildkite.com/automattic/simplenote-ios/builds/1287/steps/canvas?sid=019762e3-c6c8-4a96-af2a-d28909ce55d7#019762e3-c70e-4f5b-a62c-8e43236c9203/347-380
) because it seems it doesn't find the JSON [used to handle test accounts used for those UI Tests](https://github.com/Automattic/simplenote-ios/blob/trunk/fastlane/Fastfile#L582-L593)

This seems unrelated to the changes in this PR and more likely related to the URL we used to store those not being valid anymore (hence the body of the request being just the text `Not Found` instead of the expected JSON, leading to JSON parsing error)

The value of `UI_TESTS_ACCOUNTS_JSON_URL` env var—as provided via `.mobile-secrets`—is a URL to `https://kvdb.io/<someid>/…` so I guess maybe our subscription to that service expired or something? (though I haven't found credentials for such service in Secret Store, so unable to verify)

AliSoftware (approved):

PR #1716, opened 2025-06-13
Fetch Simperium via SwiftPM and remove CocoaPods

mokagio:
This is the final part of the CocoaPods removal work. See https://linear.app/a8c/issue/AINFRA-446.

The changes in https://github.com/Simperium/simperium-ios/pull/624 made the Simperium Apple client available via SwiftPM, and this PR switches from the CocoaPods version to it.

While at it, I also removed CocoaPods entirely. In hindsight, I should have saved the removal for a dedicated PR. But the footprint is small enough that I guess we can manage here, too.

> [!NOTE]  
> UI tests failing because of https://github.com/Automattic/simplenote-ios/issue/1717

### Test
I logged into my account from the Simulator and verified that the notes syncing — which is what Simperium is in charge of — still works.

https://github.com/user-attachments/assets/bd13438c-e426-4d69-92eb-deecf6036704

### Release


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">This PR is still a Draft: some checks will be skipped.</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/Classes/SPNoteEditorViewController+Extensions.swift
```diff
@@ -391,7 +392,7 @@ extension SPNoteEditorViewController {
     func presentInformationController(for note: Note, from barButtonItem: UIBarButtonItem) {
         let informationViewController = NoteInformationViewController(note: note)
 
-        let presentAsPopover = UIDevice.sp_isPad() && traitCollection.horizontalSizeClass == .regular
+        let presentAsPopover = UIDevice.isPad && traitCollection.horizontalSizeClass == .regular
```
mokagio, line 394: When building Simperium via CocaoPods, all headers are accessible. When building with SwiftPM, only those explicitly configured as public in the Xcode project and imported in the umbrella header are. 

This resulted in `sp_isPad()` no longer being available on my first Simperium-via-SwiftPM go. I could have made it public, but I noticed we already have an implementation in the app code base, so I just used that.

README.md
```diff
@@ -13,11 +13,6 @@ We use a few tools to help with development. To install or update the required d
 
 `rake dependencies`
 
-#### CocoaPods
-
-Simplenote for iOS uses [CocoaPods](http://cocoapods.org/) to manage third party libraries.
-Third party libraries and resources managed by CocoaPods will be installed by the `rake dependencies` command above.
-
```
AliSoftware, line 20: Maybe instead of deleting this completely, replace it with `#### SwiftPM` section mentioning that's what it uses now?

Rakefile
```diff
@@ -198,35 +175,19 @@ end
 
 desc 'Open the project in Xcode'
 task xcode: [:dependencies] do
-  sh "open #{XCODE_WORKSPACE}"
+  sh "open #{XCODE_PROJECT}"
```
AliSoftware, line 178: I wonder if using `xed` (command shipped with Xcode Command Line Tools to open a project in Xcode Editor) rather than `open` wouldn't be more idiomatic there? 🤷

AliSoftware (approved):
Apart from docs nits, LGTM.

I also ran the `pod deintegrate` to validate that there was not trace left of CocoaPods on the project, and confirmed it didn't generate any diff[^1].

[^1]: well to be fair it did generate one, but only about formatting the line 1337 of the `project.pbxproj` from a one-liner into multiline without semantic change)

Rakefile
mokagio: To be honest, I think we should get rid of this task.

Maybe it's just me, but I always just use `xed` from the terminal. Never this.

I'd feel differently if we were code generating the Xcode project with Tuist or similar, because we could make the task kill any open Xcode, regenerate, and reopen Xcode—that's what Beeper does. But here, especially now that we don't have CocoaPods in the `dependencies` task, it seems unnecessary.

AliSoftware: Agreed

README.md
mokagio: Given SwiftPM is now the standard, or at least the recommended way to manage dependencies, would it be fair to expect folks to assume that's what's in use?

It's also apparent once one opens the project:

image

AliSoftware: Fair enough.

And as long as the README for simperium itself indicates how to integrate it via SwiftPM (which arguably is also the now-standard and could be seen as not useful to document, but I still think it's useful to mention, including given it's known to not be super active/rarely updated so one could still wonder if it has been updated to SPM, and because it's integrated as a binary xcframework not source code and one has to provide the GH release path not just the repo and the checksum etc, right?). Once we have than on simperium then I agree no need to mention SPM specifically in SN client repos like here.

mokagio: > as long as the README for simperium itself indicates how to integrate it via SwiftPM

Oh, for sure. I haven't had a chance to address your comments on Simplenote macOS and Simperium but I agree that for the library the SwiftPM + XCFramework setup deserves to be documented.

mokagio: Begun adding docs for SwiftPM in Simperium via https://github.com/Simperium/simperium-ios/pull/624/commits/ba866785ac84b086153963cf5f3cb8ffaab297e9 and https://github.com/Simperium/simperium-ios/pull/624/commits/e3a3a1beab9e81aa610a8195cffc2df384abee80

Issue #1717, opened 2025-06-15
UI Tests Failing


mokagio:
This is a reference issue to explain why the UI tests fail and why we are not fixing them at this time.

## Why they fail

As @AliSoftware recently [observed](https://github.com/Automattic/simplenote-ios/pull/1713#issuecomment-2965885552):

> The [UI Tests CI job is failing](https://buildkite.com/automattic/simplenote-ios/builds/1287/steps/canvas?sid=019762e3-c6c8-4a96-af2a-d28909ce55d7#019762e3-c70e-4f5b-a62c-8e43236c9203/347-380
) because it seems it doesn't find the JSON [used to handle test accounts used for those UI Tests](https://github.com/Automattic/simplenote-ios/blob/trunk/fastlane/Fastfile#L582-L593)
>
> This seems unrelated to the changes in this PR and more likely related to the URL we used to store those not being valid anymore (hence the body of the request being just the text `Not Found` instead of the expected JSON, leading to JSON parsing error)
>
> The value of `UI_TESTS_ACCOUNTS_JSON_URL` env var—as provided via `.mobile-secrets`—is a URL to `https://kvdb.io/<someid>/…` so I guess maybe our subscription to that service expired or something? (though I haven't found credentials for such service in Secret Store, so unable to verify)

## Why we are not fixing them

At this time, we have very little bandwidth for Simplenote work. One would argue that having automated tests working is even more valuable then, but in the wider scheme of things, fixing these tests that only run occasionally is less valuable that helping other teams move forward.

Moreover, fixing is unlikely to be a matter of loosening a timeout or updating an API. This adds to the difficulty of finding enough time for it.

We hope things will change at some point and we'll have time for a deeper dive to address the issue.

dangermattic:
Thanks for reporting! 👍

twstokes:
This doesn't really fix the UI Tests which are quite broken, but it does address some of the mystery – see this Slack thread: p1667903243007039-slack-C02H2SVA1L5

I was able to make the `curl` call mentioned there (without authentication) and UI tests got past that failure [in this PR](https://github.com/Automattic/simplenote-ios/pull/1766). It still ultimately failed, however.

PR #1722, opened 2025-10-30
[Tooling] Use ReleasesV2 version as the source of truth during Code Freeze

iangmaia:
See AINFRA-1478

## Description

This PR enhances the version handling added in the previous iteration to assume the version received by the code freeze lane as the source of truth, ignoring potential mismatches between the ReleasesV2 version and the project's computed version.

## Testing

You can make sure the calculated release branch, current version and the next version are correctly shown in the "Continue" prompt. Just **remember to answer `N` when asked to continue**, to cancel the actual code freeze.

To avoid unexpected results, comment out the lines doing `ensure_git_status_clean` and `Fastlane::Helper::GitHelper.checkout_and_pull(DEFAULT_BRANCH)` at the beginning of the lane.

You can run the `start_code_freeze` lane, with and without the version parameters:
- Run `bundle exec fastlane start_code_freeze` 
- Run `bundle exec fastlane start_code_freeze version:30.6`

This will be fully tested in the next release cycle during code freeze.

### DIFF
--- a/fastlane/lanes/release.rb
+++ b/fastlane/lanes/release.rb
@@ -19,15 +19,17 @@ platform :ios do
     new_version = version || computed_version
     computed_release_branch_name = release_branch_name(release_version: new_version)
 
-    # Warn if provided version differs from computed version
+    # Fail if provided version differs from computed version
     if version && version != computed_version
-      warning_message = <<~WARNING
-        ⚠️ Version mismatch: The explicitly-provided version was '#{version}' while new computed version would have been '#{computed_version}'.
-        If this is unexpected, you might want to investigate the discrepency.
-        Continuing with the explicitly-provided verison '#{version}'.
-      WARNING
-      UI.important(warning_message)
-      buildkite_annotate(style: 'warning', context: 'start-code-freeze-version-mismatch', message: warning_message) if is_ci
+      error_message = <<~ERROR
+        ❌ Version mismatch detected!
+
+        The explicitly-provided version from the release tool is '#{version}' but the computed version from the codebase is '#{computed_version}'.
+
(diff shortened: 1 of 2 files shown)


fastlane/lanes/release.rb
```diff
@@ -425,3 +416,14 @@ def release_is_hotfix?
...
+def ensure_branch_does_not_exist!(branch_name)
+  return unless Fastlane::Helper::GitHelper.branch_exists_on_remote?(branch_name: branch_name)
+
+  error_message = "The branch `#{branch_name}` already exists. Please check first if there is an existing Pull Request that needs to be merged or closed first, " \
+                  'or delete the branch to then run again the release task.'
+
+  buildkite_annotate(style: 'error', context: 'error-checking-branch', message: error_message) if is_ci
+
+  UI.user_error!(error_message)
+end
```
mokagio, line 429: I'm sure this has been thought already, but just in case: This method, reused across many of our projects, could be extracted in `release-toolkit`. In a followup, of course!

mokagio (approved):
Tested locally. Default reads from file:

image

(The darker color of the `version` parameter is because it's my prompt suggesting it from history for tab completion)

Passing parameter overrides:

image


---


It goes without saying given the content of this PR, but we can ignore (_sigh_) the failing UI tests. More at https://github.com/Automattic/simplenote-ios/issues/1717

fastlane/lanes/release.rb
iangmaia: Yeah, totally. There are several helpers that are repeated across repos, they easily emerge when we do this kind of effort to standardize behavior. At the same time... this is very simple and kinda specific (e.g. Buildkite), and maybe we could even have a separate library for this kind of helper 🤔 anyway just thinking out loud.

PR #1724, opened 2025-11-20
[Tooling] Improve `new_hotfix_release` lane to work when there's no release tag

iangmaia:
Related to AINFRA-1518

## Description

This PR aims to add more robustness and require less manual intervention to the hotfix creation lane `new_hotfix_release`.

We have observed that sometimes it is possible that the release has been finalized but a release tag isn't present yet for some reason (likely due to the fact that the release hasn't been published yet), so release managers will get an error and require manual intervention.

This PR changes the code so that, if no release tag is present, we start the hotfix branch based on the corresponding release branch.

## Testing instructions

These are some test scenarios that can be simulated:
1. There is a release tag that can be used to create the hotfix branch
2. There's no tag, but there's still a release branch that can be used to create the hotfix branch
3. There's no tag or branch (error)
4. The provided version doesn't have a patch component like `30.6` instead of `30.6.1` (error)

To test it, you can create test branches / tags (⚠️ delete them later or, even better, use [this technique](https://github.com/bloom/DayOne-Apple/pull/27670#issuecomment-3553138802) of pointing your working copy’s `origin` remote to a local bare repo instead ⚠️ ) and run `bundle exec fastlane new_hotfix_release skip_confirm:false version:"$VERSION"` and then answer `no` after the prompt so that the lane execution stops and check the printed values.

For example:
1. Create and push release branch `release/30.6` (which doesn't exist and has no corresponding tag for that version)
2. Run `bundle exec fastlane new_hotfix_release skip_confirm:false version:30.6.1`
3. It should use the `release/30.6` branch instead of trying to use a tag. It should print:
```
ℹ️  Tag '30.6' not found on the remote. Using release branch 'release/30.6' as the base for hotfix instead.
Hotfix release:
- New hotfix version: 30.6.1
- New build code: 30.6.1.0
- Branching from release/30.6

Do you want to continue? (y/n)
```

### DIFF
--- a/fastlane/lanes/release.rb
+++ b/fastlane/lanes/release.rb
@@ -271,9 +271,12 @@ platform :ios do
     # Fetch the base ref to ensure it's available locally
     sh('git', 'fetch', 'origin', base_ref_for_hotfix)
 
+    hotfix_branch = release_branch_name(release_version: version)
+    ensure_branch_does_not_exist!(hotfix_branch)
+
     UI.message("Creating hotfix branch from '#{base_ref_for_hotfix}'...")
     Fastlane::Helper::GitHelper.create_branch(
-      release_branch_name(release_version: version),
+      hotfix_branch,
       from: base_ref_for_hotfix
     )
     UI.success("Done! New hotfix branch is: '#{git_branch}'")
--- a/fastlane/lanes/release.rb
+++ b/fastlane/lanes/release.rb
@@ -240,18 +240,20 @@ platform :ios do
     ensure_git_status_clean unless skip_prechecks
 
     parsed_version = VERSION_FORMATTER.parse(version)
+    # Validate that this is a hotfix version (must have a patch component > 0)
+    UI.user_error!("Invalid hotfix version '#{version}'. Must include a patch number.") unless parsed_version.patch.to_i.positive?
     build_code_hotfix = BUILD_CODE_FORMATTER.build_code(version: parsed_version)
     previous_version = VERSION_FORMATTER.release_version(VERSION_CALCULATOR.previous_patch_version(version: parsed_version))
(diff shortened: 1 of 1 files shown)


mokagio (approved):
Approving to unblock, but @iangmaia notice I pushed two commits to address issues I found when testing locally. Please have a look at them (left details in the commit message) and if they look good go ahead and merge.

iangmaia:
@AliSoftware FYI I've implemented the last change we discussed about ensuring the hotfix branch doesn't exist on https://github.com/Automattic/simplenote-ios/pull/1724/commits/96e4e41e0d320348a5e873e681af7800025f7f69.

fastlane/lanes/release.rb
```diff
@@ -240,32 +240,53 @@
...
+      - New build code: #{build_code_hotfix}
+      - Branching from #{base_ref_for_hotfix}
     MESSAGE
+    UI.important(message)
+
     UI.user_error!("Terminating as requested. Don't forget to run the remainder of this automation manually.") unless skip_confirm || UI.confirm('Do you want to continue?')
 
-    UI.user_error!("Version #{version} already exists! Abort!") if git_tag_exists(tag: version)
-    UI.user_error!("No tag found for version #{previous_version}. A hotfix branch cannot be created.") unless git_tag_exists(tag: previous_version)
+    UI.user_error!("Version '#{version}' already exists on the remote! Abort!") if git_tag_exists(tag: version, remote: true)
```
AliSoftware, line 269: Would be probably better moved earlier in the lane, e.g. after line 244?

AliSoftware (approved):
Tested all 4 scenarios (invalid hotfix version number, neither tag nor release branch, no tag but a release branch, tag present) and all worked as expected.

Left a nitpick (that can be addressed in separate PR). Approving to unblock.

PR #1777, opened 2026-05-20
Enable SwiftLint rule: return_value_from_void_function

mokagio:
Part of the SwiftLint rules rollout campaign — one rule per PR.

The `return_value_from_void_function` rule flags returning a value (including an expression) from a `Void`-returning function — usually a copy-paste artefact that obscures the intent.

## Changes

- Adds the rule to `.swiftlint.yml` (alphabetically placed).
- Autofix split a `return XCTFail(...)` into two statements in `SimplenoteUITests/NoteList.swift` (1 occurrence in 1 file).

## Testing

- `bundle exec rake lint` passes locally.
- `xcodebuild build-for-testing` against the `SimplenoteUITests` scheme succeeds.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -82,6 +82,9 @@ only_rules:
   # Type annotations are redundant when the type can be inferred.
   - redundant_type_annotation
 
+  # Returning a value from a function that has Void return type.
+  - return_value_from_void_function
+
   # Prefer `someBool.toggle()` over `someBool = !someBool`.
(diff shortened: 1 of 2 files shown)


mokagio:
@crazytonyli we decided to not adopt this in WordPress iOS https://github.com/wordpress-mobile/WordPress-iOS/pull/25572

What do you think of giving it a shot here?

crazytonyli (approved):

PR #1786, opened 2026-06-16
Add bundler block to Dependabot config and route Ruby reviews

mokagio:
Updates Dependabot to check Ruby dependencies daily so we can more easily address security updates. See [AINFRA-2437](https://linear.app/a8c/issue/AINFRA-2437).

Also routes review of the Apps Infra surface (dependency manifests, CI config, toolchain pins) to @Automattic/apps-infra-tooling via CODEOWNERS — GitHub retired the dependabot.yml `reviewers` key, and CODEOWNERS is its designated replacement. The routing is path-based, so it applies to any PR touching those files, not just Dependabot's; deliberate, since no source-scoped alternative exists.

---

Posted by Claude Code, Opus 4.8 (1M context) on behalf of @mokagio.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,20 @@
+# Ruby dependency surface
+Gemfile*              @Automattic/apps-infra-tooling
+.bundle/              @Automattic/apps-infra-tooling
+.rubocop*.yml         @Automattic/apps-infra-tooling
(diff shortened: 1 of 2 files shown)


iangmaia (approved):
Approving to unblock.
I wonder why copilot keeps complaining about `patterns` 😅

PR #1826, opened 2026-09-02
Prioritise release jobs ahead of feature builds

mokagio:
Adds `priority: 1` to the release pipelines so release jobs jump the queue ahead of feature builds, and the release train stalls less.

Closes [AINFRA-3045](https://linear.app/a8c/issue/AINFRA-3045).

<details>
<summary>AI-generated details.</summary>

## Description

The Releases V2 tasks run on the same pipeline and queue as every PR build, so a release manager waits behind queued feature work. The release build files already had `priority`; `release-pipelines/*` never did.

Set pipeline-wide rather than per step so that steps added later inherit it.

Part of [AINFRA-3038](https://linear.app/a8c/issue/AINFRA-3038); audit in [AINFRA-3032](https://linear.app/a8c/issue/AINFRA-3032).

## Testing instructions

CI here does not exercise these files — the upload step only uploads `pipeline.yml`. Locally, all 6 parse and every step resolves to priority `1`.

Real confirmation is the next release: the job should show **Priority** `1` in Buildkite, against `0` for a PR build.

</details>

### DIFF
--- a/.buildkite/release-pipelines/complete-code-freeze.yml
+++ b/.buildkite/release-pipelines/complete-code-freeze.yml
@@ -10,6 +10,8 @@ env:
 agents:
     queue: mac
 
+priority: 1
+
 steps:
   - label: Complete Code Freeze
     key: complete_code_freeze
--- a/.buildkite/release-pipelines/finalize-release.yml
+++ b/.buildkite/release-pipelines/finalize-release.yml
@@ -9,6 +9,8 @@ agents:
 env:
   IMAGE_ID: $IMAGE_ID
 
+priority: 1
+
 steps:
   - label: Finalize Release
     plugins: [$CI_TOOLKIT_PLUGIN]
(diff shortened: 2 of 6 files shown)


AliSoftware (changes requested):
~Don't listen to Copilot.~
~[`priority` is a job/step-level attribute](https://buildkite.com/docs/pipelines/configure/workflows/job-priority), and not valid at pipeline level. So we should NOT merge this as is as it would break.~

AliSoftware (approved):
Ok, nevermind, [TIL that this was actually supported](https://buildkite.com/docs/pipelines/configure/workflows/job-priority#prioritizing-whole-builds).

It was apparently [added to the documentation back in Aug 2024](https://github.com/buildkite/docs/pull/2923), and [to the JSON schema back in Dec 2025](https://github.com/buildkite/pipeline-schema/pull/152), and doesn't seem to depend on a `buildkite-agent` version—as it is instead an attribute managed by the control node (on the buildkite servers' side) assigning jobs to agents—so should be supported transparently regardless of what `buildkite-agent` version our CI agents are using.

